PR #5, opened 2018-03-31
ios build broken

shageman:
Hi all,

I would love to see deltachat build on ios. However, there are a number of issues currently preventing things from working. With this pull request I updated to the latest delta-chat-core and fixed a broken method reference.

I get a bunch of errors that will take me longer to understand. Any guidance would be much appreciated.

screen shot 2018-03-31 at 8 38 13 am
 
Thanks,
Stephan


r10s:
without knowing ios, this looks good to me :)

PR #16, opened 2018-11-15
Upload images

humanlikedisaster:
Updated some components (MessageKit and added SnapKit)

Implemented uploading images (beta)


Podfile.lock
```diff
@@ -1,20 +1,24 @@
...
+  - MessageKit (1.0.0)
   - openssl-ios-bitcode (1.0.210)
+  - SnapKit (4.0.1)
 
 DEPENDENCIES:
-  - MessageKit
-  - openssl-ios-bitcode
+  - MessageKit (= 1.0.0)
+  - openssl-ios-bitcode (= 1.0.210)
+  - SnapKit (~> 4.0.0)
```
hpk42: what's the license of SnapKit?

jonasreinsch: https://github.com/SnapKit/SnapKit/blob/develop/LICENSE

hpk42: k, that's MIT, fine.

jonasreinsch:
@humanlikedisaster thank you, the code looks good at a first glance, will build & test prob. this evening!

Could you perhaps tell me a little bit why you addded the SnapKit dependency (its not used in your code)? Is this just something you like to work with in the future (am just a little wary about adding 3rd party dependencies if not needed 100%)?

humanlikedisaster:
> @humanlikedisaster thank you, the code looks good at a first glance, will build & test prob. this evening!

Thanks!

> Could you perhaps tell me a little bit why you addded the SnapKit dependency (its not used in your code)? Is this just something you like to work with in the future (am just a little wary about adding 3rd party dependencies if not needed 100%)?

Yep, I saw you not using storyboards/xibs at all, and SnapKit helps to create contraint layout in code without pain in the ass. So it will help me to create complex layout without storyboards.

humanlikedisaster:
> > Yep, I saw you not using storyboards/xibs at all, and SnapKit helps to create contraint layout in code without pain in the ass. So it will help me to create complex layout without storyboards.
> 
> ok got it - would it be possible perhaps to leave that out for now and discuss further if the need arises? IMHO since the addition of layout anchors Autolayout became a lot less painful - not sure if you compare SnapKit to layout anchors or to the 'traditional' way of setting up constraints, which was indeed cumbersome. @hpk42 and @r10s also favour leaving out 3rd party deps whenever possible...

Okay, we can get rid from it right now. 

But what is problem with reusing open source components? In futher development it will be good idea to reuse someone work (for UI for example or animation tweaks), not create own from scratch. Is it license issue?

jonasreinsch:
> Okay, we can get rid from it right now.
> 
> But what is problem with reusing open source components? In futher development it will be good idea to reuse someone work (for UI for example or animation tweaks), not create own from scratch. Is it license issue?

No, no licence problem and also no general problem with reusing open source components (we already have some dependencies like MessageKit). I would just like a little bit more (concrete) evidence that it simplifies things enough to justify inclusion (it's just the general tradeoff: each 3rd party component also has to be upgraded etc).

hpk42:
On Thu, Nov 15, 2018 at 02:42 -0800, Jonas Reinsch wrote:
> > Okay, we can get rid from it right now.
> > 
> > But what is problem with reusing open source components? In futher development it will be good idea to reuse someone work (for UI for example or animation tweaks), not create own from scratch. Is it license issue?
> 
> No, no licence problem and also no general problem with reusing open source components (we already have some dependencies like MessageKit). I would just like a little bit more (concrete) evidence that it simplifies things enough to justify inclusion (it's just the general tradeoff: each 3rd party component also has to be upgraded etc).

what i don't know is also what the practises are around libraries with Swift/IOS.
In JS-land frameworks come and go on a yearly basis it seems.
In Python frameworks and libs are usually longer maintained and supported. 
Is SnapKit a well-known longer-term maintained thing?

humanlikedisaster:
> On Thu, Nov 15, 2018 at 02:42 -0800, Jonas Reinsch wrote: > Okay, we can get rid from it right now. > > But what is problem with reusing open source components? In futher development it will be good idea to reuse someone work (for UI for example or animation tweaks), not create own from scratch. Is it license issue? No, no licence problem and also no general problem with reusing open source components (we already have some dependencies like MessageKit). I would just like a little bit more (concrete) evidence that it simplifies things enough to justify inclusion (it's just the general tradeoff: each 3rd party component also has to be upgraded etc).
> what i don't know is also what the practises are around libraries with Swift/IOS. In JS-land frameworks come and go on a yearly basis it seems. In Python frameworks and libs are usually longer maintained and supported. Is SnapKit a well-known longer-term maintained thing?

Yep, it's swift version of Masonry https://github.com/SnapKit/Masonry, so I've used it in cases when I need some layout instrument.

But I will get rid from Masonry in few hours.

r10s:
great - looking forward to see this in action :)

PR #30, opened 2018-12-25
[wip] Various improvements

dignifiedquire:
- feat: make libetpan a git submodule
- feat: make deltachat-core a git submodule
- feat: add `sentbox` and `mvbox` threads
- fix: load images sent from the right file location
- feat: update messagekit to 2.0

Depends on #29

PR #31, opened 2018-12-26
Remove chat and chat list remove animation

humanlikedisaster:
Now you can remove chat (just swipe to left).
I've added Differ pod to make deletion/insertion animated (it's MIT).


dignifiedquire:
As far as I can tell this does not actually delete something in the backend, this needs to call `dc_delete_chat`, and we probably also want archiving using `dc_archive_chat`. 

Also we probably want to use sth like https://github.com/SwipeCellKit/SwipeCellKit for the different actions on a chat.

humanlikedisaster:
@dignifiedquire 
No, it's deletes chat. Look here
https://github.com/deltachat/deltachat-ios/pull/31/commits/b3520fa2d8dc293526d511c260db742b406fa127

Wrapper and ChatListController.

We can use it if it will have archive option in future. For current implementation there is no need for this.

dignifiedquire:
you are right, I missed this sorry

jonasreinsch:
@humanlikedisaster could you try to resolve the conflicts (sorry for the long time it did take me to merge https://github.com/deltachat/deltachat-ios/pull/30 )

r10s:
I think this pr is a bit stale, closing this for now.

PR #36, opened 2019-01-08
chore: better linting and formatting

dignifiedquire:
- add swiftlint
- fix swiftlint warnings
- add swiftformat config
- run swiftformat on build
- run swiftlint on build


dignifiedquire:
we can also use 4 spaces if you prefer, it is now just a one line change away :)

jonasreinsch:
> we can also use 4 spaces if you prefer, it is now just a one line change away :)

haha, no :-) I just wanted to fullfill the cliché that devs always love to bicker about coding guidelines, so I couldn't resist to comment. Besides, you are doing the real work here now, so as long as you don't set it to 0 spaces it's all ok with me!

Issue #38, opened 2019-01-13
Can't see or even receive incoming messages


csb0730:
Beta 6, Installation at IOS 12.1, provider _web.de_
- All settings checked: Seemed to be ok
- Can't see any message which is incoming. Neither to an existing chat nor any other messages.

Android version has the logging facility where you can see the internal log by "settings->Info->About Delta Chat->(pushing text)".
Is this existing in IOS version too? When: How to use it? It would help to see if IMAP server is connected.

csb0730:
What about "Contact Requests" ? Is this existing meanwhile? This would be very helpful too :-)

r10s:
@jonasreinsch i have no recent web.de experiences.

r10s:
would also be interesting if this issue is really related to ios - maybe the same happens on android or desktop.

csb0730:
@jonasreinsch @r10s 
We found the reason for not receiving any mail from web.de: There seems to be a function at web.de account which separates incoming mails to known persons and unknown persons. All not known senders (me) are automatically moved to unknown persons and are not signaled to the device.
What we did is to move testmails from my address from _unknown persons_ folder to _known persons_ (by Web interface). After that, mails from me arrived at phone!

I got the information too, that there is not folder which is called INBOX or Posteingang or similar! This sounds strange to me and I want to double check this with my friend again.

So by now I can say that problem was caused by web.de account behaviour.

But with current IOS version it's very difficult to decide where the problem is because there seems to be no possibility to see interface behaviour (logring buffer). Will it be implemented soon?

r10s:
thanks you for figuring this out. i've added a hint to the weird web.de inbox to the provider-overview in the support-forum.

wrt log: you can view the log by _shaking_ the device :)

as the initial question is solved and there is no actionable item left, i think we can close this issue.

PR #39, opened 2019-01-15
Update deps

dignifiedquire:


r10s:
very cool, all os will have the same defaults wrt move and watch now :)

dignifiedquire:
@jonasreinsch also added a bugfix, sent messages weren't being shown properly. If you could push a new testflight once this is merged that would be great

jonasreinsch:
@dignifiedquire thank you, will do!

https://github.com/deltachat/deltachat-ios/issues/38 should be unrelated to this bugfix (because incoming msgs are concerned), correct?

jonasreinsch:
@dignifiedquire Testflight build should be out now.

PR #42, opened 2019-01-25
Screenshots and new core

dignifiedquire:


hpk42:
Can you add a "alice in wonderland" picture somehow? 
that's one feature ios supports, doesn't it? 
and an audio message? 

the screenshots are also to show what technically currently works already ;) 

holger

PR #50, opened 2019-04-01
Added imap/smpt getter setter

nayooti:


r10s:
we had no idea why the [DC_LP_*_FLAGS](https://github.com/deltachat/deltachat-core/blob/master/src/deltachat.h#L777) are not accessible while the other [DC_LP_*](https://github.com/deltachat/deltachat-core/blob/master/src/deltachat.h#L722) flags are.
however, as the flags probably won't change in the future (they're also written to the database), for now, it seems okayish to use the belonging values directly.

PR #53, opened 2019-04-11
Account setup flow

nayooti:
- Split Setting into Account-Setup (LoginCredentials, Advanced Setup) and Settings (Rest)
- App-start will show Account-Setup if not logged in
- oAuth2Feature -login possible
- implemented Event DC_EVENT_HTTP_POST


dignifiedquire:
@nayooti seems there is a conflict with master

deltachat-ios.xcodeproj/project.pbxproj
```diff
@@ -954,6 +969,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
+				"HEADER_SEARCH_PATHS[arch=*]" = /Users/nayooti/iOSDeveloper/rpgp_temp;
```
dignifiedquire: this is not going to make things easier for others

deltachat-ios.xcodeproj/project.pbxproj
```diff
@@ -971,7 +987,7 @@
 				CODE_SIGN_ENTITLEMENTS = "deltachat-ios/deltachat-ios.entitlements";
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				CODE_SIGN_STYLE = Automatic;
-				DEVELOPMENT_TEAM = EEQW58QXHC;
+				DEVELOPMENT_TEAM = 5287229JS5;
```
dignifiedquire: why does the team change?

deltachat-ios/AppDelegate.swift
```diff
@@ -17,293 +17,305 @@ var mailboxPointer: UnsafeMutablePointer<dc_context_t>!
...
-          name: dcNotificationViewChat,
-          object: nil,
-          userInfo: userInfo
-        )
-      }
-    }
-
-    completionHandler()
-  }
+	static let appCoordinator = AppCoordinator()
```
dignifiedquire: is this formatting change needed? makes it hard to see if anything actually changed here

deltachat-ios/TextFieldCell.swift
```diff
@@ -1,6 +1,6 @@
 //
 //  TextFieldCell.swift
-//  deltachat-ios
+//  deltachat-ios<
```
dignifiedquire: I think this can be deleted ;)

deltachat-ios/TextFieldCell.swift
```diff
@@ -9,85 +9,113 @@
...
+    }
 
-    return nameCell
-  }
+    static func makeNameCell(delegate: UITextFieldDelegate? = nil) -> TextFieldCell {
+        let cell = TextFieldCell(description: "Name", placeholder: "new contacts nickname")
 
-  static func makeConfigCell(label: String, placeholder: String) -> TextFieldCell {
-    let nameCell = TextFieldCell(description: label, placeholder: placeholder)
+        cell.textField.autocapitalizationType = .words
```
dignifiedquire: there seems to be a lot of whitespace changes in here, that are not really necessary

nayooti:
I changed status to WIP. Had a rebase-issue that I need to fix.

deltachat-ios.xcodeproj/project.pbxproj
nayooti: Didn't know my absolute path will show up in the project-file.

I had to do this, because since the last release (update code and rpgp) the project does not compile anymore due to a missing rgpg-header file.

dignifiedquire: If you run `make install` in the `rpgp/pgp-ffi` directory, it should place the header in an accessible location. Alternatively maybe we should simply check it in, and use a project local version of it.

nayooti: Where exactly should I run `make install`? 
In my case `deltachat-ios/deltachat-ios/libraries/rpgp` does not contain a `pgp-ffi`-file. Only librgpg.a is listed.

dignifiedquire:
```
> git clone https://github.com/rpgp/rpgp
> cd rpgp/pgp-ffi
> make install
```

should do the trick

dignifiedquire: but really we should copy the header file into the project, I can do that in a bit

nayooti: that would be nice. tx

dignifiedquire:
@nayooti there is no "Create Backup" option anymore in the settings screen on the latest version

dignifiedquire:
And once I delete my account the create backup button appears, and it does not redirect me to the login screen

PR #57, opened 2019-05-03
Contacts import

nayooti:
- contacts can be imported from device 
- added dialogue to confirm contacts access (only in NewChatController - tab with adressbook will be eventually dropped)
- added searchbar in NewChatViewController that searches through all contacts 
- added functionality to handle contacts search
- search results will be filtered and highlighted (search pattern is incremental - need to be tested for larger address books - for now only tested on simulator with 5 contacts )
- changed primary color to systemblue color to give more native look and feel (temporarily until dc color scheme is there)
- added UNIT-test environment !!!


dignifiedquire:
please don’t change the colors :( :(

r10s:
we have files some colors in use on android/desktop at https://github.com/deltachat/interface , not sure how this can work on ios, however, i'm also happy to improve this :)

deltachat-ios/AccountSetupController.swift
```diff
@@ -222,18 +221,17 @@ class AccountSetupController: UITableViewController {
...
 
   override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
-
-		guard let tappedCell = tableView.cellForRow(at: indexPath) else { return }
-		// handle tap on password -> show oAuthDialogue
+    guard let tappedCell = tableView.cellForRow(at: indexPath) else { return }
+    // handle tap on password -> show oAuthDialogue
     if let textFieldCell = tappedCell as? TextFieldCell {
       if textFieldCell.accessibilityIdentifier == "passwordCell" {
         if let emailAdress = textFieldCell.getText() {
```
cyBerta: seems to me like unnecessary nesting of if conditions instead of using &&, but it's rather a code style issue from my point of view

deltachat-ios/Constants.swift
```diff
@@ -21,9 +21,9 @@ struct Constants {
     static let deltachatImapPasswordKey = "__DELTACHAT_IMAP_PASSWORD_KEY__"
   }
 
-  static let primaryColor = UIColor(red: 81 / 255, green: 73 / 255, blue: 255 / 255, alpha: 1)
-  static let messagePrimaryColor = Constants.primaryColor.withAlphaComponent(0.5)
-  static let messageSecondaryColor = UIColor(red: 245 / 255, green: 245 / 255, blue: 245 / 255, alpha: 1)
+  // static let primaryColor = UIColor(red: 81 / 255, green: 73 / 255, blue: 255 / 255, alpha: 1)
```
cyBerta: Can't this be removed?

deltachat-ios/AccountSetupController.swift
nayooti: I know looks awful.

Explanation:
237 and 239 are actually for unwrapping optionals, so you are assigning a value to a variable that is guaranteed to be not null. 
238 is a conditional.
Both statements can't be combined unfortunately.

deltachat-ios/Constants.swift
nayooti: Hell yeah!

PR #88, opened 2019-05-29
Switch to deltachat-rust-core

dignifiedquire:
All building is integrated and should happen automatically, once the dependencies mentioned in the readme are installed.


nayooti:
Thanks for PR. 

Tried to run on my machine but failed to compile - got this:

`
error: no such subcommand: `lipo`
`
Screenshot 2019-05-31 at 10 51 40

dignifiedquire:
did you follow the new instructions in the Readme?

nayooti:
I ran 
`git submodule update --init --recursive` 
over my existing project and re-opened xCode. 

Maybe that was wrong approach, because in library-folder I have a rust-deltachat-core next to my deltachat-core.

dignifiedquire:
there is a new section on installing rust and cargo-lipo in this PR, which you need

dignifiedquire:
the submodules are fine to exists in parallel

nayooti:
my bad! will try!

nayooti:
in cargo-lipo: did you go through the xcode-integration process (section before installation on cargo-lipo - README)?

nayooti:
In the xcode-integration section at cargo-lipo I am struggling at 3) + 4). Do you know what 
`your-cargo-project/target/universal/`
means in our project?

dignifiedquire:
The integration part is all done (it needs to point to the rust repo, which has a `target` folder generated on build. 

You should only need to make sure things are installed and it should work

nayooti:
Ok, now I get this:

deltachat-ios xcworkspace 2019-05-31 16-15-32

nayooti:
I will write any further step here in comments, so that I will be able to adjust the install instructions in README afterwards.

I set rust on nightly by:

`rustup install nightly
rustup default nightly`

Can be set back to stable by:
`rustup default stable`
[Source](https://jonathanmh.com/rust-nightly-stable-rustup-may-not-used-error)

Now I get this:

deltachat-ios xcworkspace 2019-05-31 16-36-22

nayooti:
did: 
`rustup target add x86_64-apple-ios`

Now I have: 
deltachat-ios xcworkspace 2019-05-31 16-54-58

dignifiedquire:
ah yes, the submodule needs to be updated to latest master, that should fix the assert issue

nayooti:
what exactly shall I do?

nayooti:
Ok, cool. Can build now.

Maybe we should update in README with:
`rustup install nightly`
`rustup default nightly`

We do use nightly atm, don't we?

dignifiedquire:
> We do use nightly atm, don't we?

We do, but the latest changes don't require to set the default toolchain, as it explicitly uses the one specified in `rust-toolchain`

nayooti:
@dignifiedquire 
By the updated README deltachat-ios won't build.
There is a **'** that needs to be deleted. And the given path that links to rust-core doesn't exist.

dignifiedquire:
> And the given path that links to rust-core doesn't exist.

Are you sure your submodule is updated? This command works for me

nayooti:
Sorry, my bad. Deleted my comment right after I realised I was wrong.

PR #93, opened 2019-06-03
fix: always build rust-core in release mode

dignifiedquire:
note: first time build will be longer now


nayooti:
Hey, I just tested:

regular login is much faster - no noticeable difference to pre-Rust-core.

But:
- oAuth still returns an error.
- canceling on-going login procedure still freezes

But I guess fixing these wasn't the intention of this PR, was it?

PR #105, opened 2019-06-07
integrated photo compression before sending

nayooti:
- uploaded images will be compressed to 1280 max(width, height).


r10s:
cool, happy to check this out later on.

PR #128, opened 2019-07-06
fix: do not use ringbuf feature, to make app store reviewers happy

dignifiedquire:


hpk42:
\o/

r10s:
great, will try an upload just now :)

PR #136, opened 2019-07-10
Fix autoconfig

### LINKED ISSUE #132: Autoconfig broken on master
(updated, also with input from @dignifiedquire): iOS current master does not properly trigger autoconfig but rather always assumes one has done advanced config (and thus trying https-autoconfig urls does not happen).  This breaks testrun but other providers might still work if they have settings that are guessed/tried by core automatically (imap.X, smtp.X etc.).

r10s:
closes #132


hpk42:
lgtm, great ;)

PR #146, opened 2019-07-16
Translations

### LINKED ISSUE #58: Localization
iOS needs to support localization -- a bit unclear how soon as it's probably more interesting to get some basic features first (as of May 2019). 

_next paragraph is from @r10s, sorry, @hpk42 for writing in your issue. but such feature lists are rather useful ;)_
- [x] target the issue of strings used as IDs as written down [below](https://github.com/deltachat/deltachat-ios/issues/58#issuecomment-510063762)
- [x] check which format is used by ios
- [x] write a shell or python script that pulls strings.xml translations in all languages from transifex and converts them to the format used by ios
- [x] replace hard-coded strings by replaceable strings in the way this is done typically by ios
- [x] make sure, delta chat is shown in the system language

cyBerta:
closes #58


hpk42:
looks good to me -- but none of the translated strings is used anywhere yet, right?

cyBerta:
yes, I'm currently doing this Sisyphus task :)

OzancanKaratas:
Note: Include Info.plist strings in localizable content.

cyBerta:
> Note: Include Info.plist strings in localizable content.

Thanks for the hint!

r10s:
in general, this looks good to me. it was probably a hard job :)

so, for initalisation/updating translations, one have to run `./updateTranslations.sh`? 
nice.

- as there are probably some more scripts needed in the future, (eg. for adding a language), maybe we can move the scripts to a `tools` subdirectory (as on android). we could also give them the same names, so `./tools/tx-pull-translations.sh` then.

- calling `./updateTranslations.sh` creates/updates nl.lproj/Localizable.string, nl.lproj/Localizable.stringsdict, pl.lproj/Localizable.strings, pl.lproj/Localizable.stringsdict - are these languages updated in between - or just missing in the pr?

r10s:
apart from the nitpicking above: works like a charme - delta ios in different languages now - good job!

cyBerta:
>     * as there are probably some more scripts needed in the future, (eg. for adding a language), maybe we can move the scripts to a `tools` subdirectory (as on android). we could also give them the same names, so `./tools/tx-pull-translations.sh` then.

yes, the name of the script is actually misleading and should be altered. I like the idea of changing  the name into tx-pull-translations.sh, also putting it into tools sounds good to me. 
If you want to add new languages you will only have to add empty language files to the project. As long as those languages are translated in transifex they will be pulled automatically afterwards. 

>     * calling `./updateTranslations.sh` creates/updates nl.lproj/Localizable.string, nl.lproj/Localizable.stringsdict, pl.lproj/Localizable.strings, pl.lproj/Localizable.stringsdict - are these languages updated in between - or just missing in the pr?

I think they are updated. But it was strange that they are empty now :O. I need to investigate where it's going wrong.

cyBerta:
> But it was strange that they are empty now :O. I need to investigate where it's going wrong.

OK, so some languages got updated in between. And there seems to be an issue with nl_NL. For some reason the tx client doesn't recognize the corresponding file. I assume it fetches 'nl' instead, which is an empty translation on transifex.

cyBerta:
if you pull the translation twice nl_NL gets fetched again with all translations... I would consider it as a bug in tx and leave it like that for now. Just pull twice.

cyBerta:
I've fixed the naming and location of the scripts

r10s:
hm, nl is still strange, the nl-strings are deleted when i run the script.

OzancanKaratas:
> hm, nl is still strange, the nl-strings are deleted when i run the script.

@r10s I can resolve this problem. Please add me as project maintainer and team manager into Transifex project. Some languages had been configured wrong.

r10s:
@OzancanKaratas maybe you can just give us a hint what to change to fix the problem?

OzancanKaratas:
Some languages had been set as language+locale. Whereas source is set as language only. For example, source is English. Japanese is Japanese (Japan). This is a wrong usage. Correct usage is Japanese only. Automated processes may not be work correctly.

r10s:
yeah, nl_NL seems to be wrong, however, other language+locale combination are needed and are correct. anyway, this is targeted by https://github.com/deltachat/deltachat-ios/blob/master/tools/.tx/config

but maybe we should change nl_NL to just nl to make sure this is really not the problem.

OzancanKaratas:
If it works, no problem. Can you update the translations now? I want to preview and localize the iOS contents.

r10s:
yip, apart from the weired nl-stuff, things are pulled and converted by `./tools/tx-pull-translations.sh´ - when checking out the reop, however, this step is not needed as all locale files are already part of the repo and are located at https://github.com/deltachat/deltachat-ios/tree/master/deltachat-ios

(i think we will update them in the repo from time to time, esp. for releases. for your local repo, of course, you can update as you like)

PR #149, opened 2019-07-23
Remove “beta” tag

OzancanKaratas:
iOS already displays a beta tag. Not requires to display beta two times.

### DIFF
--- a/deltachat-ios/Info.plist
+++ b/deltachat-ios/Info.plist
@@ -5,7 +5,7 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>$(DEVELOPMENT_LANGUAGE)</string>
 	<key>CFBundleDisplayName</key>
-	<string>Delta Chat Beta</string>
+	<string>Delta Chat</string>
 	<key>CFBundleExecutable</key>
 	<string>$(EXECUTABLE_NAME)</string>
 	<key>CFBundleIdentifier</key>


cyBerta:
Hi @OzancanKaratas 
On my device there's not beta tag shown. Where do you find it? Can you provide a screenshot please?

OzancanKaratas:
Yes, but can I ask what your phone model is it? I have been using iOS 12.4 on iPhone SE.

cyBerta:
I'm using pretty much the same, iPhone SE and i0S 12.3.1 but I've build the app from source so that might be the reason why I cannot see the beta tag?

OzancanKaratas:
Definitely yes! iOS displays beta tag if the app was installed by TestFlight. But it is need to instal via TestFlight. I removed beta tag thinking the TestFlight users. Normal installations checks Info.plist but TestFlight not checks.

Simon-Laux:
Maybe there is a way to have a different Icon for the dev test builds?

OzancanKaratas:
> Maybe there is a way to have a different Icon for the dev test builds?
I think not required. What happens if beta does not shows in the screen? This tag will be already removed in the next times.

OzancanKaratas:
Hello again,

I want to edit some languages and locales. Do you add me as project maintainer into Transifex project? I saw a discussion. I want to change a fiew thing.

cyBerta:
I'm fine with merging it for now. If anyone disagrees - the change is easy to revert :)

PR #164, opened 2019-08-05
Qr code scanner

cyBerta:


r10s:
- in the simulator, leaving the qr-code-scan-viewcntroller takes some time. it's fine on a native device, however, this might be worth a look. maybe sth. is wrong.

- i think the qr-code-scan-viewcntroller should have a title and the same hint as on android (currently just "place you camera ober the QR code", but we will probably refine this at some point)

deltachat-ios/Helper/UIView+Extension.swift
```diff
@@ -0,0 +1,97 @@
...
+//  deltachat-ios
+//
+//  Created by Macci on 05.08.19.
+//  Copyright © 2019 Jonas Reinsch. All rights reserved.
+//
+
+import UIKit
+
+extension UIView {
+	func makeBorder(color: UIColor = UIColor.red) {
```
r10s: also here, please indent by 4 spaces

deltachat-ios/Helper/UIView+Extension.swift
```diff
@@ -0,0 +1,96 @@
+//
```
r10s: please remove the boilerplate

deltachat-ios/Helper/UIView+Extension.swift
```diff
@@ -0,0 +1,96 @@
...
+//  File.swift
```
r10s: no boilerplate, please

deltachat-ios/Helper/UIView+Extension.swift
cyBerta: fixed

deltachat-ios/Helper/UIView+Extension.swift
cyBerta: fixed

cyBerta:
>     * i think the qr-code-scan-viewcntroller should have a title and the same hint as on android (currently just "place you camera ober the QR code", but we will probably refine this at some point)

done

cyBerta:
For now only contact verification is implemented, not group invites

r10s:
got a black screen after a qr code is recognized. however, merging as this pr also provides some other tools and cleanup.

r10s:
> got a black screen after a qr code is recognized. however, merging as this pr also provides some other tools and cleanup.

@cyBerta not sure what was going on there, maybe i had not checked out the latest commits. works like a charme, very nice :)

Issue #171, opened 2019-08-15
target warnings


r10s:
currently, building the ios app results in ~150 warnings, partly more or less unimportant linter stuff, but anyway.

we should bring these warnings close to 0 so that real problems become visible.

r10s:
due to #170, #172, #173 and https://github.com/deltachat/deltachat-core-rust/pull/382 there are now zero warnings left :)

PR #172, opened 2019-08-15
avoid force casts

r10s:
this pr tries to remove all `as!` casts that are potentially dangerous as, when failing, the app panics. esp. when it comes to calling into UIKit or so, we cannot really guarantee that sth. is of a given type.

swiftlint also treats this as bad practice and throws warnings

so, if needed, we'll cast using `as?` and check for `nil`.

cmp https://developer.apple.com/swift/blog/?id=23

targets #171

### DIFF
--- a/deltachat-ios/Coordinator/AppCoordinator.swift
+++ b/deltachat-ios/Coordinator/AppCoordinator.swift
@@ -100,11 +100,12 @@ class AppCoordinator: NSObject, Coordinator {
 
     func showChat(chatId: Int) {
         showTab(index: 3)
-        let navController = self.chatListController as! UINavigationController
-        let chatVC = ChatViewController(dcContext: dcContext, chatId: chatId)
-        let coordinator = ChatViewCoordinator(dcContext: dcContext, navigationController: navController, chatId: chatId)
-        chatVC.coordinator = coordinator
-        navController.pushViewController(chatVC, animated: true)
+        if let navController = self.chatListController as? UINavigationController {
+            let chatVC = ChatViewController(dcContext: dcContext, chatId: chatId)
+            let coordinator = ChatViewCoordinator(dcContext: dcContext, navigationController: navController, chatId: chatId)
+            chatVC.coordinator = coordinator
+            navController.pushViewController(chatVC, animated: true)
+        }
     }
 
     func presentLoginController() {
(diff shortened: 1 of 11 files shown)


r10s:
does this make sense? @dignifiedquire i do not want to pull you away from rust, but maybe you have a stance on this?

deltachat-ios/DC/events.swift
```diff
@@ -48,9 +48,9 @@ public func callbackSwift(event: CInt, data1: CUnsignedLong, data2: CUnsignedLon
             if done {
                 UserDefaults.standard.set(true, forKey: Constants.Keys.deltachatUserProvidedCredentialsKey)
                 UserDefaults.standard.synchronize()
-                let appDelegate = UIApplication.shared.delegate as! AppDelegate
-                // appDelegate.appCoordinator?.showTab(index: 3)
-                AppDelegate.lastErrorDuringConfig = nil
+                if let appDelegate = UIApplication.shared.delegate as? AppDelegate {
```
jonasreinsch: @r10s you probably can leave out the cast/if let line here since appDelegate is not used inside the if let block (AppDelegate.lastErrorDuringConfig only accesses the static variable lastErrorDuringConfig https://github.com/deltachat/deltachat-ios/blob/f15345b46159f4a268da649e9e39930c34d76a8b/deltachat-ios/AppDelegate.swift#L24 )

deltachat-ios/Coordinator/AppCoordinator.swift
```diff
@@ -100,11 +100,12 @@ class AppCoordinator: NSObject, Coordinator {
 
     func showChat(chatId: Int) {
         showTab(index: 3)
-        let navController = self.chatListController as! UINavigationController
-        let chatVC = ChatViewController(dcContext: dcContext, chatId: chatId)
-        let coordinator = ChatViewCoordinator(dcContext: dcContext, navigationController: navController, chatId: chatId)
-        chatVC.coordinator = coordinator
-        navController.pushViewController(chatVC, animated: true)
+        if let navController = self.chatListController as? UINavigationController {
```
jonasreinsch: @r10s here it might be better to use a `guard let` instead of `if let` in combination with an `assertionFailure` (this fails in debug mode) - because it would be a programming error / something would be seriously wrong if the `chatListController` would not cast to a `UINavigationController` here:

    guard let navController = self.chatListController as? UINavigationController else {
        assertionFailure("huh? why no nav controller?")
    }

And the guard let makes the control flow more clear IMO.
See https://www.swiftbysundell.com/posts/picking-the-right-way-of-failing-in-swift

deltachat-ios/DC/events.swift
r10s: right, i was too deep into the as-pattern now seeing that the statement is void at all ;)

r10s:
@jonasreinsch thanks for reviewing this pr 👍

PR #219, opened 2019-09-02
core strings

r10s:
this pr supplies translated strings to the core.

in some future, we may get rid of DC_EVENT_GET_STRING, however, currently it is there and this is the way to have a fully-translated ui.

as the core does no longer care about plural forms currently, i had to do a small hack. also this may change in some future :)

### DIFF
--- a/deltachat-ios/View/ChatTitleView.swift
+++ b/deltachat-ios/View/ChatTitleView.swift
@@ -47,6 +47,17 @@ class ChatTitleView: UIView {
         subtitleLabel.textColor = baseColor.withAlphaComponent(0.95)
         titleLabel.textColor = baseColor
         titleLabel.text = title
-        subtitleLabel.text = subtitle
+        subtitleLabel.text = hackPluralsString(string: subtitle)
     }
+}
+
+func hackPluralsString(string: String?) -> String? {
+    // the rust-core does not care about plural forms (there is just once case that was not worth the effort up to now)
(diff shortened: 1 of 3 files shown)


deltachat-ios/View/ChatTitleView.swift
```diff
@@ -47,6 +47,17 @@ class ChatTitleView: UIView {
...
-        subtitleLabel.text = subtitle
+        subtitleLabel.text = hackPluralsString(string: subtitle)
     }
 }
+
+func hackPluralsString(string: String?) -> String? {
+    // the rust-core does not care about plural forms (there is just once case that was not worth the effort up to now)
+    // therefore, we check if the returned string has the form "N member(s)" and localized the corrently to one/few/many/other
+    guard var string = string else { return nil }
+    if string.hasSuffix(" member(s)") {
```
cyBerta: this string (chat.subtitle) has always the english form so that you can compare it with " member(s)", right?

r10s: yip, the string `N members(s)` is used by the core as a fallback if the ui does provide translations.

as we do not do this by intention at https://github.com/deltachat/deltachat-ios/pull/219/files#diff-61af89ae1ce21d3427b4a9fe947a7660R186 we can be pretty sure to get this exact english string back - and can convert it to local plural forms then.

of course, this is a hack. sooner or later there might be a proper solution for that - or maybe the need for plural forms in the core goes away completely :)

Issue #218, opened 2019-09-02
Adaptions to Settings/Signature text 


cyBerta:
By tapping on Settings -> Signature text, a new view opens that includes both entries 'Your name' and 'Signature text'. Only the latter one should be visible in this context. 
Moreover this Signature text entry should include an expandable multiline textfield because at the moment it is really hard to write an email footer within an about 25 character long single line entry.

r10s:
totally agree with the multi-line issue.

however, i think having name+signature both editable in the same viewcontroller makes  sense. i'd also add the selection of a profile image there (see #153, as on android)

but, of course, the navigation to that viewcontroller may be improved (eg. sth. as on android, or a more clearly named button). but i think for _now_ it is fine and i would not call this a bug :) maybe a ux glitch :)

r10s:
found a nice solution in #221, later on, this should also show the profile image then.

we should keep this issue open until someone makes the signature multiline, however.

r10s:
closing this in favor to the more clear #225

PR #221, opened 2019-09-02
combine name and signature row

### LINKED ISSUE #217: Settings/Your name includes signature text
currently, by tapping on settings (section YOUR PROFILE INFO) -> 'Your name', a new view opens including two entries 'Your name' and 'Signature text'.
The latter should be deleted from that view.

r10s:
this pr combines the name- and signature row, to avoid the ux glitch of two rows navigating to the same screen.

later on, this should also contain the profile image and in the EditSettingsController the profile image should be selectable.

closes #217 
tackles #218


cyBerta:
Nice!

Issue #192, opened 2019-08-27
qr-code in my-profile broken


r10s:
not totally sure about the reason, might be that the updated core does not profile a qr-code-string.

<img src=https://user-images.githubusercontent.com/9800740/63786190-dfdd3300-c8f1-11e9-856a-8882ab4a7ae0.png width=320>

cyBerta:
fixed in latest core

r10s:
yeah, things are getting better now :)

PR #223, opened 2019-09-03
fix layout in profile view if the QR code can't be generated (#192)

cyBerta:
in a future MR we should consider to add an error text if the QR-Code is missing. We don't have text resources for it at the moment. 

Screen Shot 2019-09-03 at 13 09 12
Screen Shot 2019-09-03 at 13 10 11

### DIFF
--- a/deltachat-ios/Controller/NewProfileViewController.swift
+++ b/deltachat-ios/Controller/NewProfileViewController.swift
@@ -238,6 +238,7 @@ class NewProfileViewController: UIViewController, QrCodeReaderDelegate {
         label.font = UIFont.systemFont(ofSize: 14)
         return label
     }
+    
     private func createQRCodeScannerButton() -> UIView {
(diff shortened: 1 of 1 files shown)


cyBerta:
fixes #192

r10s:
> in a future MR we should consider to add an error text if the QR-Code is missing. We don't have text resources for it at the moment.

such strings can be added in english-only, similar to log entries. i do not think it is worth to annoy translators with that. normally, this should just not happen - unless the core is buggy as currently :)

i think, in general, messages for very rare errors are also better in english as they're easer to websearch this way.

cyBerta:
OK, then I'll add an entry in untranslated.xml and a label showing the error.

r10s:
i would not even add it to untranslated.xml, just a hardcoded string,

r10s:
the core-problem is targeted by https://github.com/deltachat/deltachat-core-rust/pull/428 - as soon as this pr gets merged, the fingerprint should reappear :)

PR #227, opened 2019-09-03
fix linter warning

r10s:


deltachat-ios/Controller/NewProfileViewController.swift
```diff
@@ -67,13 +67,6 @@ class NewProfileViewController: UIViewController, QrCodeReaderDelegate {
         return DcContact(id: Int(DC_CONTACT_ID_SELF))
     }
 
-    var fingerprint: String? {
-        if !DcConfig.configured {
-            return nil
-        }
-        return dcContext.getSecurejoinQr(chatId: 0)
-    }
-
```
r10s: i removed this indirection as the variable "fingerprint" is actcually no fingerprint but the content of a qrcode to render - which contains things needed for a securejoin - this may or may not include a fingerprint.

deltachat-ios/Controller/NewProfileViewController.swift
```diff
@@ -256,16 +249,13 @@ class NewProfileViewController: UIViewController, QrCodeReaderDelegate {
     private func createQRCodeView() -> UIView {
         let width: CGFloat = 130
         let frame = CGRect(origin: .zero, size: .init(width: width, height: width))
-        var imageView: UIView
-        if let fingerprint = self.fingerprint {
-            imageView = QRCodeView(frame: frame)
-            (imageView as! QRCodeView).generateCode(
```
r10s: avoid a linter warning wrt `as!`, this also simplifies the code.

Issue #231, opened 2019-09-03
remove contact tab


r10s:
the contact-tab should be removed as already discussed offline several times.

however, _before_ it can be removed, we should make sure that _all functionality_ goes to the "create chat" options (eg. the "edit" option.

~~this feature has a lower priority.~~ i have to correct myself: as this view adds complexibiy to other prs, maybe we should not postpone that too much.

PR #236, opened 2019-09-05
Fix contact search

### LINKED ISSUE #233: contact search erroneous
the current contact search is just a filter about the whole list, done in the ui (the core is not used for searching contacts here)

this is fine for now, however, the search is not incremental as expected, but or-searches for every letter entered. so "hans" searches for all contacts containing "h" "a" "n" or "s"

cyBerta:
fixes #233

### DIFF
--- a/deltachat-ios/Controller/NewChatViewController.swift
+++ b/deltachat-ios/Controller/NewChatViewController.swift
@@ -165,7 +165,7 @@ class NewChatViewController: UITableViewController {
                 return cell
             }
             if row == 1 {
-                // new contact row
+                // scan QR code row
                 let cell: UITableViewCell
                 if let c = tableView.dequeueReusableCell(withIdentifier: "scanGroupCell") {
                     cell = c
@@ -192,6 +192,7 @@ class NewChatViewController: UITableViewController {
                 return cell
             }
         } else if section == 1 {
+            // import device contacts section
             if deviceContactAccessGranted {
                 let cell: ContactCell
                 if let c = tableView.dequeueReusableCell(withIdentifier: "contactCell") as? ContactCell {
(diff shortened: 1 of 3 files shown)


r10s:
i think the contact-list reachable from the tab-bar is ContactListController? if so, we can skip any tweaks for that as this will go away anyway, see https://github.com/deltachat/deltachat-ios/issues/231

cyBerta:
you find the search bar I'm tweaking if you're in the chat list and click on the upper right edit icon.

cyBerta:
relates to #233

deltachat-ios/Controller/NewChatViewController.swift
```diff
@@ -192,6 +192,7 @@ class NewChatViewController: UITableViewController {
                 return cell
             }
         } else if section == 1 {
```
r10s: when touching code like that, i suggest to switch to use constants as `sectionImport`, `rowScanQr` etc (as eg. in https://github.com/deltachat/deltachat-ios/commit/74f255d6b1ede1b18d1d494e1a6ccf66cb722c60 )

if would not change the whole logic and/or use a better controller base class for now, i would just like to get rid of the numbers.

cyBerta: I totally agree. In this case section 2 has not always the same meaning though. section 2 can include the "import Devices Contacts" button or the the contact list. That's why I'm not sure how rename section 2 without being super clumsy (like section 2 importOrContactListSection vs. section 3 unimportedDeviceContactListSection). Any suggestions?

r10s: naming is hard :) i think, the point is not a super-nice-name but that the names make it possible to add sections etc. without breaking things. 

maybe use sth. has `<sectionName>Section` and then inside the section `<sectionName>Section<rowName>Row`, so eg. `unimportedSectionImportRow`. but you will know better :)

r10s:
lgtm + tested

PR #238, opened 2019-09-09
reset backup image color if there's an avatar image

### LINKED ISSUE #235: random background colors for partly transparent group/profile images in chatlist
if a profile image is partly transparent (eg. not round, eg. a delta logo), in the chatlist-controller, a "random" background color is shown for the parts not drawn (the "random" proably comes from the recycled view)

so you see eg. a group-icon with green background beside a chat. scrolling up and down turns the color to blue. or to red.

in the chat-controller-title, the image is always drawn fully transparent on the background, whatever this is - this is totally fine and on pair with android.

i wound suggest to do the same in the chatlist-controller.

cyBerta:
fixes #235


r10s:
lgtm + works :)

PR #241, opened 2019-09-09
show correct alert dialog on qr-group-join (#166)

### LINKED ISSUE #166: wrong question displayed on qr-group-join
qr group-joining works like a charme, however, the initial question should read "do you want to join group $groupname" and not "do you want to chat with $person"

cyBerta:
fixes #166


r10s:
the "Merge branch 'master'" was not really what i wanted to do when using "Resolve conflicts" from github. anyway ...

PR #242, opened 2019-09-09
fix navigation bar title and back button in new chat context

### LINKED ISSUE #237: wrong "back" label
in chats-tab, select "new chat" and then "new group".

now the back button is labled "chats" where it should read "new chat".

clicking that back-button, the title of thw "new group" viewcontroller is now "chats" (where it was correctly "new chat" before)

cyBerta:
fixes #237


r10s:
lgtm - do you have an idea what the hack was really about? cc @nayooti

PR #243, opened 2019-09-09
Add alert before starting new chat

### LINKED ISSUE #240: ask before creating a chat
it is very easy to create chats accidentially from the "new chat" viewcontroller.

on android, we had a question "chat with XY?" before really creating the chat. i suggest to add the same on ios.

cyBerta:
fixes #240


r10s:
i think, the seconds alert is not needed. removed the seconds commit in #244 and closing this one.

PR #246, opened 2019-09-10
remove contacts tab

cyBerta:
implements #231


cyBerta:
Oh it's important to mention that I've implemented both swipe actions from right to left. The reason is, that the swipe from left to right also pops the whole screen from the view stack if the swipe is done from too far left. It happend to my a dozen times, so it felt like bad UX. Now it looks similar to how the delete action has been implemented in the chat list.

r10s:
k. so we spare swipe-from-left-to-right for another time. maybe it is not needed at all, thought it is kind of standard (as used here and there). duckduckgo'ing a bit shows up some _incredibly_ complicated solutions, including the need of external libs, i think this is nothing we really want to go for.

PR #248, opened 2019-09-10
tweak alerts

r10s:
this pr changes the `UIAlertController` to use `preferredStyle: .actionSheet` for user actions and confirmations, cmp. https://developer.apple.com/design/human-interface-guidelines/ios/views/action-sheets/

`preferredStyle: .alert` stays for showing system states errors etc., see https://developer.apple.com/design/human-interface-guidelines/ios/views/alerts/

nb: we're using the _title_ instead of the _message_ if text gets displayed - the latter shows the text in bold when only a message is given. for whatever reason.

### DIFF
--- a/deltachat-ios/Controller/ContactDetailViewController.swift
+++ b/deltachat-ios/Controller/ContactDetailViewController.swift
@@ -30,7 +30,7 @@ class ContactDetailViewController: UITableViewController {
         let cell = ActionCell()
         cell.accessibilityIdentifier = CellIdentifiers.chat.rawValue
         cell.actionColor = SystemColor.blue.uiColor
-        cell.actionTitle = String.localizedStringWithFormat(String.localized("ask_start_chat_with"), contact.nameNAddr)
+        cell.actionTitle = String.localizedStringWithFormat(String.localized("menu_new_chat"), contact.nameNAddr)
         cell.selectionStyle = .none
         return cell
     }()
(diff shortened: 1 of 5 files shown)


deltachat-ios/Controller/ContactDetailViewController.swift
```diff
@@ -31,7 +30,7 @@ class ContactDetailViewController: UITableViewController {
         let cell = ActionCell()
         cell.accessibilityIdentifier = CellIdentifiers.chat.rawValue
         cell.actionColor = SystemColor.blue.uiColor
-        cell.actionTitle = String.localizedStringWithFormat(String.localized("ask_start_chat_with"), contact.nameNAddr)
+        cell.actionTitle = String.localizedStringWithFormat(String.localized("menu_new_chat"), contact.nameNAddr)
```
cyBerta: menu_new_chat has no additional string parameters -> remove `String.localizedStringWithFormat`
Should also a action message be added?

r10s: i've removed that, thanks.
wrt "action message" - i think the button is just fine, but what exactly is the "action message"?

cyBerta: oh, I now see that this title is not within an alert but part of a table view cell :O 
Nevermind.

PR #264, opened 2019-09-16
Language optimizations

OzancanKaratas:
- Some upper and lower cases was fixed.

- Use original strings for meaning same translations.


cyBerta:
Hey @OzancanKaratas, could you change your PR so that it will to merge into deltachat:master instead of deltachat:send-asm? Or is there a specific reason you choose the send-asm feature branch as merge target?

cyBerta:
besides that, it looks good and can be merged

tools/untranslated.xml
```diff
@@ -4,25 +4,24 @@
...
     <string name="reset">Reset</string>
     <string name="edit_contact">Edit contact</string>
-    <string name="custom_port">Custom Port</string>
+    <string name="custom_port">Custom port</string>
     <string name="automatic">Automatic</string>
-    <string name="complete">Complete</string>
+    <string name="completed">Completed</string>
     <string name="info">Info</string>
-    <string name="qr_code_title">QR scan/show</string>
+    <string name="qr_code_title">Scan/Show QR code</string>
```
cyBerta: `Scan/Show QR code` This string might be too long for the tab bar where it is used

r10s:
thanks for the pr!

one note to `Localizable.strings`: this file is generated, so changes there will be overwritten anyway. (the files is generated from untranslated.xml and from the translations pulled from transifex)

deltachat-ios/Controller/ChatViewController.swift
```diff
@@ -251,7 +251,7 @@ class ChatViewController: MessagesViewController {
 
         if disableWriting {
             menuItems = [
-                UIMenuItem(title: String.localized("start_chat"), action: #selector(MessageCollectionViewCell.messageStartChat(_:))),
+                UIMenuItem(title: String.localized("welcome_start_messaging"), action: #selector(MessageCollectionViewCell.messageStartChat(_:))),
```
r10s: no, welcome_start_messaging is the wrong string here. this string is used on the inital welcome string.

tools/untranslated.xml
```diff
@@ -4,25 +4,24 @@
     <!-- iOS specific untranslated strings -->
     <string name="hide">Hide</string>
     <string name="login_error_title">Unable to login!</string>
-    <string name="start_chat">Start Chat</string>
-    <string name="dismiss">Dismiss</string>
+        <string name="dismiss">Dismiss</string>
```
r10s: format

r10s:
i think i will close this pr and move the correct string changes to a new pr with a new commit, will credit @OzancanKaratas from there then :)

Issue #253, opened 2019-09-12
click on contact request does nothing


Simon-Laux:
Expected: open dialog "wanna start a chat with this person?"

cyBerta:
I've added a listener for tabs, but there was already a conext menu on long click, that I found myself accidentally 😄 

![Screen Shot 2019-09-17 at 11 26 56](https://user-images.githubusercontent.com/2614900/65029532-37455080-d93e-11e9-9ede-ae66007c97df.png)

r10s:
i think we should use the `UIAlertController` with `preferredStyle: .actionSheet` for the contact request question.

PR #265, opened 2019-09-17
Click on contact request

cyBerta:
targets #253


cyBerta:
group contact request handling is still buggy

Issue #258, opened 2019-09-16
Tap on Scan QR code does nothing


cyBerta:
In the New Chat context `Scan QR code` (chats -> upper right write button -> Scan QR code) doesn't trigger any action. 
Expectation: The scan QR code View appears.

r10s:
we should just remove the button. there is a qr-code scanner in the tab-bar.

this is also the flow on android, this may change in the future, but for now it should be removed.

PR #266, opened 2019-09-17
remove 'scan qr code' entry in NewChatViewController

cyBerta:
targets #258


deltachat-ios/Controller/NewChatViewController.swift
```diff
@@ -135,7 +134,7 @@ class NewChatViewController: UITableViewController {
 
     override func tableView(_: UITableView, numberOfRowsInSection section: Int) -> Int {
         if section == sectionNew {
-            return 3
+            return 2
```
r10s: can we please have a sectioNewRowCount at line 11 for that?

cyBerta: sure

Issue #252, opened 2019-09-12
search-field is missing in add-group-members


r10s:
the search-field to filter the contact-list inside the new-chat view-controller should also be used in the add-group-members screen.

PR #269, opened 2019-09-17
add search bar to add group members and new group context

cyBerta:
targets #252 

- drawback: code duplication. There are refactorings needed to reuse code between the GroupMembersViewController and NewChatViewController.


r10s:
great!

PR #270, opened 2019-09-17
fix line-length warning

r10s:


r10s:
we should have swiftlint in the ci sooner or later ...

Issue #211, opened 2019-09-01
unblock blocked contacts


r10s:
the settings should have a TapActionRow names "Blocked contacts" (needed strings are already present) that shows a list of all currently blocked contacts and the option to unblock them.

see android to get an idea.

i think the normal contact list and/or items should be reused for this purpose.

r10s:
great, also though that we need this somehow soonish :)

i would add the `Block contacts >` entry as the first item just below settings/communication (before settings/read-receipts as the footer refers just to that item)

cyBerta:
ok

r10s:
closed by #275

PR #275, opened 2019-09-18
Unblock blocked contacts

cyBerta:
targets #211 
- includes the same alert as in android
- search is disabled

Screen Shot 2019-09-18 at 18 59 35

### DIFF
--- a/deltachat-ios/Controller/GroupMembersViewController.swift
+++ b/deltachat-ios/Controller/GroupMembersViewController.swift
@@ -99,6 +99,15 @@ class AddGroupMembersViewController: GroupMembersViewController {
 
 class BlockedContactsViewController: GroupMembersViewController, GroupMemberSelectionDelegate {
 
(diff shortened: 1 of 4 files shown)


cyBerta:
I could add the name of the selected contact to the alert title...

r10s:
looks great! a search is not really needed, at least not for now, this is fine. i will check this out.
the "< back" button looks android-ish, should be "< settings" or so.

r10s:
> I could add the name of the selected contact to the alert title...

yeah, this makes sense here as doable without changing the already translated strings.

r10s:
some nitpicking: the checkmarks are not needed in the list of blocked contacts.

cyBerta:
I'll adjust that tomorrow.

r10s:
great, now i can contact all the ppl again i blocked for testing reasons the last days :)

PR #276, opened 2019-09-18
simplify callbacks

r10s:
instead of calling a function that creates a callback using another identifier,
we specify a callback directly that "knows" which switch to change.
this also makes the additional list of identifiers superfluous.


cyBerta:
lgtm

PR #277, opened 2019-09-18
handling received autocrypt-setup-messages

r10s:
the input field is very basic, a simple improvement might be to add a mask and insert a space or so every 4th character, doing so, we can also switch to UIKeyboardType. decimalPad.

an alternative would be to show the 3x3 input fields, not sure if this is possible in the UIAlertController (this would be my favourite) or in a separate view.

however, all in all, things are working, and the layout can be improved on an interation and should not prevent this pr from being merged :)


cyBerta:
lgtm

PR #296, opened 2019-09-24
move contact-requests out of focus

r10s:
the idea is that contact-requests pop up directly in the chatlist.
so, the contact-request-view is only needed
when the user discards a request and decides to come back to that later.
in a typical usage, there is often no need to open the contact requests,
therefore, a whole tab, at eye level of "chats" is overrated.

on android, contact-requests are used in the same way,
being not directly in the eyes of the user.

there are still some things to do on contact requests,
eg. they are still a bit ugly and not always refreshed in the chatlist
and the options show_emails=all should display all incoming mails
as contact-requests directly in the chatlist.

### DIFF
--- a/deltachat-ios/Controller/SettingsController.swift
+++ b/deltachat-ios/Controller/SettingsController.swift
@@ -110,6 +110,9 @@ internal final class SettingsViewController: QuickTableViewController {
             Section(
                 title: String.localized("pref_communication"),
                 rows: [
+                    NavigationRow(text: String.localized("menu_deaddrop"),
+                              detailText: .none,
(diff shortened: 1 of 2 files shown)


r10s:
one thing to mention: on ios it is very usual to have such "more menu entry items" as part of the settings dialog, probably due to the lack of a real global menu.

eg. whatsapp as their "starred messages" here, telegram their "saved messages" - both in the menu on android.

cyBerta:
I have a small change request: the constants here: https://github.com/deltachat/deltachat-ios/blob/a3d9e0990d52ff8535d614438286466a48c435da/deltachat-ios/Coordinator/AppCoordinator.swift#L9 should be adapted accordingly.

Issue #82, opened 2019-05-28
GroupChatDetail - tap on GroupMember - show ContactDetail


nayooti:

nayooti:
@r10s Can I mute or block single group members? If not, this Issue is a pretty bad idea.

PR #303, opened 2019-09-25
open contact details on tap on group member's avatar (#82)

cyBerta:
implements #82

### DIFF
--- a/deltachat-ios/Controller/GroupChatDetailViewController.swift
+++ b/deltachat-ios/Controller/GroupChatDetailViewController.swift
@@ -1,6 +1,6 @@
 import UIKit
 
-class GroupChatDetailViewController: UIViewController, ContactCellDelegate {
+class GroupChatDetailViewController: UIViewController {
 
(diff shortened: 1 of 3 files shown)


r10s:
@cyBerta if you write `closes #NNN` or `fixes #NNN`, the corresponding issue is automatically closed once the pr is merged. this always saves some clicks and we do not forget to close issues. github shows this by a dotted line below the word "closes" and also says in the issue that the issue is maybe closed by a pr.

however, if you _know_ a pr will not close an issue completely, just link the issue as here without the keyword. i tend to write `tackles #NNN` in this case - but everything else is fine as well.

cyBerta:
ok :) 

Hello Github: "closes #82" 

I've changed the behavior but kept the avatar tap listener in the contact cell as it seems to be reusable e.g. if we implement the selection of avatar images.

r10s:
> Hello Github: "closes #82"

i fear this works only in the initial description *duckundweg*

PR #304, opened 2019-09-25
fix flickering in NewChatViewController (#300)

### LINKED ISSUE #300: flicker issue when opening new-chat
when opening the new-chat viewcontroller, for a moment, there is a separator between the search field and the new-group entry.

a moment later, this separator is scrolled away.

this is a but ugly, would be great if we can fix this as these things are very much in the face of the user.

if would prefer to _always_ have the separator.

cyBerta:
closes #300


cyBerta:
r10s please try this PR out :). After that change I couldn't reproduce the bug anymore, but still...

PR #310, opened 2019-09-25
add self avatar beside settings profile (#302)

### LINKED ISSUE #302: add self-avatar beside settings/profile
in settings, the first row below "your profile info" should show the self-avatar at the left side, similar to other messengers.

the [`icon:` attribute in the QuickTableViewController](https://github.com/bcylin/QuickTableViewController) should be sufficient for that so that we do not need to play around with subviews and whatnot.

if no self-avatar is set, it would be great if we can have the stub-avatar (color+firstletter). if that gets too complicated, however, another placeholder would also do the job.

~~nb: once done, maybe we will put the "Password and account settings" navigation link into the profile info (as on android), it is typically not that important - when the user is logged in, he already managed the things to do there.~~ EDIT: this is already done in https://github.com/deltachat/deltachat-ios/pull/309

cyBerta:
finally managed to add the avatar into the profile table view cell. It was impossible to use an Icon and to use the QuickTableView api, I had to add the views for the first cell to the cell's contentView.

might :) closes #302

### DIFF
--- a/deltachat-ios/Controller/SettingsController.swift
+++ b/deltachat-ios/Controller/SettingsController.swift
@@ -106,7 +106,7 @@ internal final class SettingsViewController: QuickTableViewController {
         cell.contentView.addSubview(nameLabel)
         cell.contentView.addSubview(signatureLabel)
 
-        let badgeConstraints = [badge.constraintAlignLeadingTo(cell.contentView, paddingLeading: 12),
+        let badgeConstraints = [badge.constraintAlignLeadingTo(cell.contentView, paddingLeading: 16),
                                 badge.constraintCenterYTo(cell.contentView),
                                 badge.constraintAlignTopTo(cell.contentView, paddingTop: 8),
                                 badge.constraintAlignBottomTo(cell.contentView, paddingBottom: 8)]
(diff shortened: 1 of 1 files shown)


cyBerta:
The problem was that I could not render the badge view to an UIImage (to wrap it into an Icon) without having rendered the badge view already at least once.

cyBerta:
after merging the PR  #309 and rebasing onto master, the subtitle looks like that: 
Screen Shot 2019-09-26 at 14 33 55

r10s: Should we have 2 lines for the subtitle or change the subtitle text? Having _Signature text:_ takes a little bit too much space from my point of view.

cyBerta:
I think my preferred version would be to remove completely all additional information except the email address. This way it would be also similar to Android

r10s:
> I think my preferred version would be to remove completely all additional information except the email address. This way it would be also similar to Android

yip. this is probably better. i do a commit to this pr, i also tweaked the offset a bit.

bit yeah: this looks great!

r10s:
nb: at some point we should switch to UIFont.preferredFont(forTextStyle:) instead of system-font, but the code is full of system-fonts, so, for now, this is fine. the colors in darkmode are more important.

PR #311, opened 2019-09-26
Verified icons

### LINKED ISSUE #305: use the correct 'verified-group' icon
currently, the verified-group icon is a dark blue badge with a transparent check.

we should use the same icon as on android here, a light-blue badge with a white check.

the source file is available at https://github.com/deltachat/interface/blob/master/icons/verified.xcf (gimp format)

preview: Unknown

### LINKED ISSUE #306: show 'verified-group' icon in chat-title
currently, the verfied group icon is shown only in the chatlist, it is missing in the chat.

on android, we show the icon beside the title, however due to the centered titles on ios, this may look a bit weird. and async.

we can try to show the icon on the avatar, it would be pretty small there, but we could give it a try. eg. it could be half-height and half-weidth of the avatar, this should be large enough (the relative size in the chatlist is smaller, i would not touch that)

this should be targeted after #305 so that we directly use the correct icon :)

the source file is available at https://github.com/deltachat/interface/blob/master/icons/verified.xcf (gimp format)

preview: Unknown

### LINKED ISSUE #307: show verified-contacts as such
currently, the verified-contacts are not recognizable.

we should add a icon to them, on android, we use the same verified-icon as for the groups for that.

this should be targeted after #305 and after #306 so that we directly use the correct icons and have an idea about a good placement (i think we can put the icon on the profile-avatar, same as on the group-avatar, looks a bit quiter than having them right of the name (on android, we could not add them to the avatar because that clashes with the selection marks some time))

the source file is available at https://github.com/deltachat/interface/blob/master/icons/verified.xcf (gimp format)

preview: Unknown

cyBerta:
closes #307 
closes #306 
closes #305


r10s:
@cyBerta great, really great!

PR #316, opened 2019-09-28
fix redraw error

### LINKED ISSUE #314: redraw error in new profile
if you click on the new my-profile-link (the one with the avatar) and go back and click again some times, the my profile font gets bolder and bolder.

seems as if each time the settings screen is opened, a new view for the name+subtitle is created that are drawn all together then.

introduced with https://github.com/deltachat/deltachat-ios/pull/310

r10s:
the subviews are added each time the settings view is opened;
by this, the same texts are drawn on top of each other
which lead to a "bold" font.
as a quick fix, to not handle changes etc.,
we just remove all existing subviews before adding new ones.

fixes #314 

moreover, the second commit in this pr tweaks the profile-info size and position a bit.


cyBerta:
lgtm

PR #327, opened 2019-10-07
Custom text media cells

cyBerta:

PR #330, opened 2019-10-08
Custom video cell

cyBerta:
based on pr #327


r10s:
i think, MessageKit is far too complicated here, there should be just one type of cell. however, it works generally - still without the thumbnail, i assume, but anyway: merging :)

Issue #281, opened 2019-09-19
useless error message on login failure


r10s:
not sure, where this comes from and if this is a core or an ios problem:

![Screen Shot 2019-09-19 at 16 53 13](https://user-images.githubusercontent.com/9800740/65255461-38bd7700-dafe-11e9-9e57-c47967866c3e.png)

r10s:
btw, also here the green-red-connection-line is a bit confusing here.
maybe, for now, we should disable the line until the core provides information about real connectivity.

r10s:
closed by https://github.com/deltachat/deltachat-ios/pull/331

PR #331, opened 2019-10-08
Tweak onboarding errors

### LINKED ISSUE #328: Need more helpful error messages during onboarding
- Device: iPhone 8 plus
- iOS version: Latest
- Delta Chat version: Latest
- Expected behavior: 
  - Log in, or get an error message that tells me that I need to change some email settings in order for it to work.
- Actual behavior: Receive bizarre error messages. No clear fixes that I can provide. User can only just close the app and give up on it.

Do not change any app advanced settings or default email settings for these providers. Just simply enter your emails and passwords.

- Screenshots:

Here is the error I got when I tried my iCloud email:
icloud

Here is the error I got when I tried my Gmail email:
gmail

r10s:
closes #328

### DIFF
--- a/deltachat-ios/DC/events.swift
+++ b/deltachat-ios/DC/events.swift
@@ -13,6 +13,13 @@ let dcNotificationContactChanged = Notification.Name(rawValue: "MrEventContactsC
 @_silgen_name("callbackSwift")
 
 public func callbackSwift(event: CInt, data1: CUnsignedLong, data2: CUnsignedLong, data1String: UnsafePointer<Int8>, data2String: UnsafePointer<Int8>) -> UnsafePointer<Int8>? {
+    if event >= DC_EVENT_ERROR && event <= 499 {
+        let s = String(cString: data2String)
+        AppDelegate.lastErrorString = s
+        logger.error("event: \(s)")
+        return nil
+    }
+
     switch event {
 
     case DC_EVENT_INFO:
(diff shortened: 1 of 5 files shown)


deltachat-ios/DC/events.swift
```diff
@@ -13,6 +13,13 @@ let dcNotificationContactChanged = Notification.Name(rawValue: "MrEventContactsC
 @_silgen_name("callbackSwift")
 
 public func callbackSwift(event: CInt, data1: CUnsignedLong, data2: CUnsignedLong, data1String: UnsafePointer<Int8>, data2String: UnsafePointer<Int8>) -> UnsafePointer<Int8>? {
+    if event >= DC_EVENT_ERROR && event <= 499 {
```
cyBerta: I guess the idea here is to suppress the error events that are in the range of 400 - 499 to be shown to the user, right? 
This condition includes not only DC_EVENT_ERROR_NETWORK, but also also DC_EVENT_ERROR_SELF_NOT_IN_GROUP. Is that ok?

r10s: yip, that is okay for now. it has to be settled in deltachat.h, however, i do not want to wait for that here.

r10s: and: the errors are not suppressed.
they are logged as before logger.error("event: \(s)") - and moreover, the AppDelegate.lastErrorString is set for _all_ errors (was not true for DC_EVENT_ERROR_NETWORK before), this was one of the reasons for #328

trosel:
@r10s This is an improvement, but the end user still has no actionable takeaway from this error.
```
DeltaChat: "Could not log in"
User:      "Okay, what did I do wrong?"
DeltaChat: ...
User:      *deletes app*
```

r10s:
also fixes https://github.com/deltachat/deltachat-ios/issues/281

PR #337, opened 2019-10-10
display audio chat messages

cyBerta:
based on an messageKit example
extended with text functionality for audio cells
includes some refactorings wrt the example (e.g. implementing a seperate audio player view)


r10s:
great! really great!

now we can listen to voice messages on ios :)

PR #345, opened 2019-10-13
tweak and fix delivery-state indicator

### LINKED ISSUE #319: delivery-checkmarks must not be shown for incoming messages
the checkmarks for "message sent" (on checkmark) and "message read" must be shown only for outgoing messages, not for incoming messages.

for incoming messages, this place is just empty or may show the number of unread messages.

r10s:
fixes #319

### DIFF
--- a/deltachat-ios/Controller/ChatViewController.swift
+++ b/deltachat-ios/Controller/ChatViewController.swift
@@ -618,8 +618,24 @@ extension ChatViewController: MessagesDataSource {
             let text = NSMutableAttributedString()
             text.append(NSAttributedString(string: m.formattedSentDate(), attributes: timestampAttributes))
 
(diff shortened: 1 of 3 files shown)


deltachat-ios/Controller/ChatViewController.swift
```diff
@@ -618,8 +618,24 @@ extension ChatViewController: MessagesDataSource {
             let text = NSMutableAttributedString()
             text.append(NSAttributedString(string: m.formattedSentDate(), attributes: timestampAttributes))
 
+            // TODO: this should be replaced by the respective icons,
+            // for accessibility, the a11y strings should be added
```
cyBerta: also, the state strings should be localized in the future

PR #349, opened 2019-10-16
add simple layout to show forwarded messages

cyBerta:
adds a line to the message top label, showing Forwarded Message if dc_msg_is_forwarded returns 1

There's a core bug though that prevents returning ever 1, tested with the current core on Android and on iOS. I'll file a bug for that. 

Screen Shot 2019-10-16 at 16 08 36


r10s:
great, looks good! also great you figured out the rust issue, filed it etc. :)

merging this now, as soon as https://github.com/deltachat/deltachat-core-rust/pull/725 and maybe https://github.com/deltachat/deltachat-core-rust/pull/726 gets merged, the ui will pick up the correct states as expected.

PR #357, opened 2019-10-21
Tweak member list screens

### LINKED ISSUE #355: fix add-/edit-members screen
- the order of the add-member list seems to be broken, at least it is not alphabetivally
- avatar color not updated on edit-name. if this is not easily fixable, we can remove the avatar in the edit-name screen at all
- there is a call to fatalError() if dc_add_contact_to_group() failed (see https://github.com/deltachat/deltachat-ios/blob/master/deltachat-ios/Controller/GroupNameController.swift#L37 ) - i think, this is a bit harsh as this terminates the app totally afaik

cyBerta:
closes #355 
could not reproduce that issue:

> avatar color not updated on edit-name. if this is not easily fixable, we can remove the avatar in the edit-name screen at all

when changing the name of a group, also the avatar color changed. There might be collisions though as there are a limited amount of colors. Sometimes 2 different strings will result in the same color.

### DIFF
--- a/deltachat-ios/Controller/GroupMembersViewController.swift
+++ b/deltachat-ios/Controller/GroupMembersViewController.swift
@@ -147,10 +147,9 @@ class AddGroupMembersViewController: GroupMembersViewController {
     }
 
     func loadMemberCandidates() -> [Int] {
-        var contactIds = Set(Utils.getContactIds()) // turn into set to speed up search
-        for member in chatMemberIds {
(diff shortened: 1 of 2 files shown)


r10s:
> when changing the name of a group, also the avatar color changed. There might be collisions though as there are a limited amount of colors. Sometimes 2 different strings will result in the same color.

yeah, might be :) thanks for re-checking.

Issue #163, opened 2019-08-02
ios dark mode support is very poor


dignifiedquire:
Some parts are inverted, some are not. Some have very poor contrast, or strange artifacts.

r10s:
okay, official xcode 11 is out and ii had a closer look. due to the recent cleanups, at least the title in most viewcontrollers is fine. same for the settings.

todo: 

- in chatlist-view and contact-view (that is also used for member-view and group-creation-view) the title of the chat resp. contact is rendered in a dark color. there is probably some "foreground" color or so that can be used. the subtitle is fine, probably by coincidence as rendered in 50% gray.

- the chat-view with the MessageKit thingie is currently not respecting dark-mode at all, so, at least things stay visible. wondering if there is an update for that or so, but as we've defined many colors on our own, we need to adapt this probably (also) manually. see the android-dark-mode for some nice-working color for the chat-bubbles etc.

- the title in the chat-view is drawn with. a transparent grey.

- in the qr-show-view and in the group-info-view, there is a row with own-avatar (for the headline) this is also rendered in light-mode. however, wondering if we want to have this row at all here (just in case it is hard to fix)

r10s:
oh, and, not to forget: the input field shows light-white on white text ;)

nayooti:
Here is a temporary solution:

Disabling darkmode support for a single ViewController or disable darkmode support for the whole app (2 lines of code).
https://stackoverflow.com/questions/56546267/ios-13-disable-dark-mode-changes

r10s:
@nayooti thanks for pointing to that, very welcome hint!

PR #359, opened 2019-10-22
Message counter

### LINKED ISSUE #358: show number of unread messages in chatlist
by https://github.com/deltachat/deltachat-ios/issues/344 we have now the number of total unread messages on the app icon - however, when opening the app, there is no hint, where the unread messages are.

- we should add a unread counter at the place where the checkmarks are for outgoing messages (this is mutual exclusive - either the last message is outgoing and there are not unread messages therefore of the last message is incoming and there may be unread messages)  
on android, we show the unread messages in a green badge, however, as these badges are red on ios, maybe we should prefer this color. but i do not have a string opinion on that.

- when on this, maybe also show the title in bold, and when changing this, maybe also fix the color for the darkmode (this would target #163 then)

cyBerta:
closes #358
... except of the dark mode issue, which will be handled separately here: #163


r10s:
looks great!

PR #360, opened 2019-10-23
marknoticed chat on opening

r10s:
this marks all messages in the chat as being notices
so that they do no longer appear as fresh
and are eg. not shown in the counter badges.

this does not send out an mdn;
this is done still only for the messages really shown on the display


cyBerta:
nice

Issue #347, opened 2019-10-15
disable underlining of words that describe a date or a location


r10s:
currently words as "here", "morning" and so on are underlined an link to the calendar or whatever.

this is more irritating than helping, in many contexts, this does not really make sense, esp. when word of other languages seems are underlined ("hier"="here" in german, "hier"="yesterday" in french)

underlining links seems to be sufficient.

cyBerta:
similar to that: currently you will be routed to the default emailing app if you click on a mail address in the chat view. I would like to disable this behavior as a first step.

 As a second step it would be nice to open a action sheet asking to chat with the contact. On tapping Start Chat, the corresponding chat opens or a new one will be generated.
@r10s would it be fine to implement it? Should not take long.

r10s:
@cyBerta as discussed on irc: sounds good!

PR #362, opened 2019-10-23
Dark mode

### LINKED ISSUE #163: ios dark mode support is very poor

cyBerta:
closes #163

### DIFF
--- a/deltachat-ios/Controller/ChatViewController.swift
+++ b/deltachat-ios/Controller/ChatViewController.swift
@@ -312,13 +312,10 @@ class ChatViewController: MessagesViewController {
         messageInputBar.delegate = self
         messageInputBar.inputTextView.tintColor = DcColors.primary
         messageInputBar.inputTextView.placeholder = String.localized("chat_input_placeholder")
-        messageInputBar.isTranslucent = true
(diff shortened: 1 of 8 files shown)


r10s:
looks great!

one thing i noticed: the pr changes the font-size of the centered chat-system-messages as "member xy added" or "setup changed for xyz". seems as if this is the same size as the in/out bubbles now - but is should be a bit smaller as before.

cyBerta:
oh, ok. Thanks for pointing it out

r10s:
another thing: also the hyperlinks are always black (when it crosses your way: also some works as yesterday, gestern etc. are turned into links. we should stop that, it is not reliable and not really helpful) (if it not crosses your way, we'll do an issue :)

cyBerta:
> another thing: also the hyperlinks are always black (when it crosses your way: also some works as yesterday, gestern etc. are turned into links. we should stop that, it is not reliable and not really helpful) (if it not crosses your way, we'll do an issue :)
there's already an issue: https://github.com/deltachat/deltachat-ios/issues/347
I'll do it there

cyBerta:
besides that, it's ready for review round 2

PR #363, opened 2019-10-23
Link detection

### LINKED ISSUE #347: disable underlining of words that describe a date or a location

cyBerta:
closes #347


cyBerta:
for now there's no option to copy only the link on longpress as this needs a little bit more work and it doesn't seem to be high prio to me.

r10s:
yip, great!

PR #364, opened 2019-10-23
Delete single msgs

### LINKED ISSUE #181: allow to delete a single message
i think this should be done using a UIAlertController - which can then also be used for "info" or other options (do not show two controls as telegram does currently (one abottom of the screen, one beside the message))

cyBerta:
closes  #181

... but can be seen as a first iteration

### DIFF
--- a/deltachat-ios/Controller/ChatViewController.swift
+++ b/deltachat-ios/Controller/ChatViewController.swift
@@ -502,7 +502,7 @@ class ChatViewController: MessagesViewController {
     }
 
     private func askToDeleteMessage(id: Int) {
-        let alert = UIAlertController(title: String.localized("delete_message_ask"),
+        let alert = UIAlertController(title: String.localizedStringWithFormat(String.localized("ask_delete_messages"), 1),
                                          message: nil,
                                          preferredStyle: .actionSheet)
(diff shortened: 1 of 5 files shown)


deltachat-ios/Controller/ChatViewController.swift
```diff
@@ -472,6 +481,20 @@ class ChatViewController: MessagesViewController {
             super.collectionView(collectionView, performAction: action, forItemAt: indexPath, withSender: sender)
         }
     }
+
+    private func askToDeleteMessage(id: Int) {
+        let alert = UIAlertController(title: String.localized("delete_message_ask"),
```
r10s: i think we should re-use `String.localizedStringWithFormat(String.localized("ask_delete_messages"), 1)`. already available in all languages, less work for translators and gives hint that the message is also deleted from the server.

r10s:
apart from the hint above (and the conflict, of course :) this looks good to me and works as exected.

deltachat-ios/Controller/ChatViewController.swift
cyBerta: fixed.

r10s:
looks great! i think it is just fine using the nearby-tap-menu (how is it called?)

PR #368, opened 2019-10-28
marknoticed and update badge on entering chat

r10s:
this pr updates the app icon's badge-counter as soon a chat is entered and the messages are noticed - before, the counter was updated when the chat was left again.

moreover, the pr delays moves marknoticed call also to viewDidAppear to avoid the one being executed without the other and for performance reasons (no need to delay entering the view)


cyBerta:
lgtm

PR #370, opened 2019-10-29
streamline qr-codes

r10s:
there are still some rust-related bugs https://github.com/deltachat/deltachat-core-rust/issues/758 and https://github.com/deltachat/deltachat-core-rust/issues/759 and some types are not yet supported (shown as text then), but this should not block this pr from being merged :)

### DIFF
--- a/deltachat-ios/Controller/QrViewController.swift
+++ b/deltachat-ios/Controller/QrViewController.swift
@@ -140,11 +140,18 @@ class QrViewController: UITableViewController, QrCodeReaderDelegate {
         switch state {
         case DC_QR_ASK_VERIFYCONTACT:
             let nameAndAddress = DcContact(id: qrParsed.id).nameNAddr
-            joinSecureJoin(alertMessage: String.localizedStringWithFormat(String.localized("qrscan_ask_fingerprint_ask_oob"), nameAndAddress), code: code)
+            joinSecureJoin(alertMessage: String.localizedStringWithFormat(String.localized("ask_start_chat_with"), nameAndAddress), code: code)
+
         case DC_QR_ASK_VERIFYGROUP:
-            if let group = qrParsed.text1?.replacingOccurrences(of: "+", with: " ") {
-                joinSecureJoin(alertMessage: String.localizedStringWithFormat(String.localized("qrscan_ask_join_group"), group), code: code)
-            }
+            let groupName = qrParsed.text1 ?? "ErrGroupName"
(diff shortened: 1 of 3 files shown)


deltachat-ios/Controller/QrViewController.swift
```diff
@@ -140,17 +140,28 @@ class QrViewController: UITableViewController, QrCodeReaderDelegate {
...
         case DC_QR_ASK_VERIFYCONTACT:
             let nameAndAddress = DcContact(id: qrParsed.id).nameNAddr
-            joinSecureJoin(alertMessage: String.localizedStringWithFormat(String.localized("qrscan_ask_fingerprint_ask_oob"), nameAndAddress), code: code)
+            joinSecureJoin(alertMessage: String.localizedStringWithFormat(String.localized("ask_start_chat_with"), nameAndAddress), code: code)
+
         case DC_QR_ASK_VERIFYGROUP:
-            if let group = qrParsed.text1?.replacingOccurrences(of: "+", with: " ") {
-                joinSecureJoin(alertMessage: String.localizedStringWithFormat(String.localized("qrscan_ask_join_group"), group), code: code)
-            }
+            let groupName = qrParsed.text1 ?? "ErrGroupName"
```
cyBerta: ErrGroupName will displayed to the user if text1 is empty? Why not something localized like "unknown group"?

tools/untranslated.xml
```diff
@@ -2,14 +2,11 @@
 <resources>
 
     <!-- iOS specific untranslated strings -->
-    <string name="hide">Hide</string>
-    <string name="start_chat">Start Chat</string>
+    <string name="start_chat">Start chat</string>
```
cyBerta: We could move that to the android strings.xml (in addition)

cyBerta:
@r10s: WIP is there still on purpose?

deltachat-ios/Controller/QrViewController.swift
r10s: because it is no error the user should be aware of at all. when we get the info "group" but no name, this should be fixed in the core, it normally cannot happen, no need to give work to the translators for that.

tools/untranslated.xml
r10s: yip, but let's do that at some point in another pr. the untranslated.xml has always be shrinking the last days, btw, so it is fine it was not done yet :)

r10s:
no, it is no longer wip

deltachat-ios/Controller/QrViewController.swift
cyBerta: ok

PR #379, opened 2019-11-10
tweak support for device-chat

r10s:
this pr removes the input-field depending on dc_chat_can_send() (false for device-chat and deaddrop currently).

moreover, it adds the correct subtitle for device-chats.

Screen Shot 2019-11-10 at 14 11 04


cyBerta:
lgtm

Issue #287, opened 2019-09-20
Smtp on iOS delta chat beta app Cuban nauta email


avivador:
- Device:
- iOS version: 12.4.1
- Delta Chat version: the last beta
- Expected behavior: just chat
- Actual behavior: smtp doesn’t work
- Steps to reproduce the problem:
I am using an android emulator so installed delta chat and export the account 
I opened the iOS beta delta chat and import the file. I saw the app but downs work. 
I wrote in avance configuration. imap.nauta.cu port 143 and smtp.nauta.cu port 26 (I realized the check mark was port 465). 
Right now I saw how the imap works but no smtp 
- Screenshots:
![DC1B9906-E217-4087-B91F-3E663BD9248A](https://user-images.githubusercontent.com/55469763/65350173-1f2a3700-dbb3-11e9-9bb2-b0bd1e6cf64c.jpeg)
![B340E74B-6A52-4442-A102-6C4A42246C07](https://user-images.githubusercontent.com/55469763/65350174-1f2a3700-dbb3-11e9-9b44-252fb6e5978c.png)
![12D1687E-485A-4908-B562-0BC43075F3AE](https://user-images.githubusercontent.com/55469763/65350175-1fc2cd80-dbb3-11e9-94fe-bf8f89bd8513.png)

- Logs:
Device model: iPhone 6s
System version: iOS 12.4.1
Console output:
2019-09-20 14:05:09.523 deltachat-ios[617:95530] <CATransformLayer: 0x282dc7ee0> - changing property masksToBounds in transform-only layer, will have no effect
2019-09-20 14:05:09.536 deltachat-ios[617:95530] <CATransformLayer: 0x282dc37e0> - changing property masksToBounds in transform-only layer, will have no effect
14:05:09.259 💙 INFO AppDelegate.application():66 - launching
14:05:09.339 💙 INFO AppDelegate.open():144 - open: /var/mobile/Containers/Data/Application/4987E45E-EEEC-4D84-8699-4E16EDB26391/Library/messenger.db
14:05:09.339 💙 INFO events.callbackSwift():27 - event: Opened "/var/mobile/Containers/Data/Application/4987E45E-EEEC-4D84-8699-4E16EDB26391/Library/messenger.db".
0
14:05:09.450 💙 INFO AppDelegate.start():171 - ---- start ----
14:05:09.455 💙 INFO AppDelegate.getCoreInfo():42 - deltachat_core_version=v1.0.0-alpha.1
[...]

PR #382, opened 2019-11-12
 improve advanced settings

cyBerta:
* adapts UI to core fallback behavior described here: https://github.com/deltachat/deltachat-ios/issues/287#issuecomment-552522216 (chosen ports will be shown in the UI after a successful login)
 * fixes flag configuration
* adapts smtp/imap port and security selection


r10s:
not totally sure about if my suggestion with ok/cancel thingie in the port settings and in the security settings is the way to go. maybe a normal "back" key is sufficient.

anyway, its an overall code improvement, great!

PR #384, opened 2019-11-15
streamline login

r10s:


deltachat-ios/Controller/AccountSetupController.swift
```diff
@@ -749,24 +767,10 @@ class AccountSetupController: UITableViewController {
         dismiss(animated: true, completion: nil)
         let appDelegate = UIApplication.shared.delegate as! AppDelegate
         appDelegate.registerForPushNotifications()
-        if (!DcConfig.configuredMailPort.isEmpty) {
```
cyBerta: What happens, if you enter an invalid port and click on login now? The deleted lines here updated the UI (irrelevant for login, but not for the settings screen) after the core guessed the right port.

r10s: tbh, i am not sure what should be achieved by the deleted lines at all.

but it is not needed to access the configured-values from the ui at all.

- if the port is invalid, nothing should happen.
the entered "wrong" data stay in the input field "as is", nothing is written in there programmatically, so the user can just correct the data - maybe a typo? - and try over.

- if the dialog is left, data are reset automatically https://c.delta.chat/classdc__context__t.html#adfe52669a5bed893df78a620566dd698

cyBerta: >     * if the port is invalid and the dialog is not left after configure(), nothing should happen.
>       the entered "wrong" data stay in the input field "as is", nothing is written in there programmatically, so the user can just correct the data - maybe a typo? - and try over.

The problem with the current core implementation is that it fixes for example wrong ports by itself by guessing typical ones. This leads to a inconsistent UI state in the settings screen (I'm not talking about the initial login process, but the advanced settings changes).  While the UI shows an invalid port for example, the app is actually using a different, valid one. To handle that inconsistency, the UI got updated on a successful login (by the deleted lines)

cyBerta: you can try it out in advanced settings with an invalid port.

r10s: > The problem with the current core implementation is that it fixes for example wrong ports by itself by guessing typical ones 

yip, but that will be changed in the core at some point.  
and the issue is true for android as well.

the idea of the settings is, that they stay consistent in what the user has entered. that way, subsequent changes and logins are doable.

"view log" on android or "debug toolkit" on ios show the currently selected settings (the debug toolkit has to be tweaked, however, this is out of scope for now)

r10s: in other words, there should be no mix of the "entered" and the "used" settings in the input controls.

this is how android/desktop work and we should do the same on ios.

r10s:
the last commit shows the current settings (if any):

Screen Shot 2019-11-15 at 17 01 08

cyBerta:
perfect!

hpk42:
On Fri, Nov 15, 2019 at 07:08 -0800, björn petersen wrote:

> r10s commented on this pull request.
> 
> 
> 
> > @@ -749,24 +767,10 @@ class AccountSetupController: UITableViewController {
>          dismiss(animated: true, completion: nil)
>          let appDelegate = UIApplication.shared.delegate as! AppDelegate
>          appDelegate.registerForPushNotifications()
> -        if (!DcConfig.configuredMailPort.isEmpty) {
> 
> > The problem with the current core implementation is that it fixes for example wrong ports by itself by guessing typical ones 
> 
> yip, but that will be changed in the core at some point.  

do we want to maybe go for this in the next days? 
i think it's anyway a good idea and if it's causing UI problems
we should change dcc's configure guessing logic to respect
manual settings ...

r10s:
> do we want to maybe go for this in the next days?

yip, we can go for that. i think the idea is mainly to leave things manually set by the user along? eg. if port is 123 do not try other ports and when the server is foo.bar do not try imap.foo.bar?

hpk42:
On Fri, Nov 15, 2019 at 12:10 -0800, björn petersen wrote:

> > do we want to maybe go for this in the next days?
> yip, we can go for that. i think the idea is mainly to leave things manually set by the user along? eg. if port is 123 do not try other ports and when the server is foo.bar do not try imap.foo.bar?

yes, exactly.  manually chosen settings should not be "auto-varied"  :)

PR #386, opened 2019-11-15
avatar image picker

### LINKED ISSUE #380: picker for avatar image
- iOS lacks currently the possibility to set images for your own profile as well as for group images

cyBerta:
closes #380 for self avatar.
setting group avatars will be done in a subsequent PR

known issue: in debug builds, xcode creates a new app folder on each build. Thus, rebuilding the app will lead to a reset - the chosen image is not shown anymore. This could be fixed by using relative paths, but it seems like there's some magic in the core, that allows to set a relative path as a string for 'selfavatar', but getting the value for 'selfavatar' still results in an absolute path including the app folder and it's document directory. I didn't investigate here further yet. In release builds, the app folder should remain, so this issue should not appear).


r10s:
looking forward to see that in action :)

deltachat-ios/Controller/EditSettingsController.swift
```diff
@@ -31,14 +37,24 @@ class EditSettingsController: UITableViewController {
...
-        cell.accessibilityIdentifier = "accountSettingsCell"
+        cell.tag = tagAccountSettingsCell
         return cell
     }()
 
+
+    private lazy var pictureAndNameCell: AvatarEditTextCell = {
+        let contact = DcContact(id: Int(DC_CONTACT_ID_SELF))
+        let cell = AvatarEditTextCell(context: dcContext, defaultImage: defaultImage)
+        cell.inputField.text = contact.displayName
```
r10s: this is wrong.

contact.displayName is typically "Me" as displayed in lists etc. when setting this to the inputField, recipient will get the text "Me" as the sender name.

to get/set the configured displayname that is actually sent to others, use dcContext.get|setConfig("displayname"). it is okay, if this string is empty for "no name set"

cyBerta: thanks! fixed

OzancanKaratas:
Please recreate the strings using genstrings command. Because @r10s deleted some strings in untranslated.xml file. So some strings is missing in source language file and untranslated.xml. We should fill again.

r10s:
@OzancanKaratas which strings do cause troubles? 

i removed some strings from untranslated.xml in https://github.com/deltachat/deltachat-ios/commit/6a89cb1ec7b5a1405ff53e3f306ed1793b1ff494#diff-0588248b083cc7c52c16dad389e1a130 but these strings are just moved to normal translatable transifex strings or are replaced by other existing string.

r10s:
however, @cyBerta, probably a rebase to master is needed

deltachat-ios/Helper/MediaPicker.swift
```diff
@@ -0,0 +1,59 @@
...
+            }
+        } else {
+            self.presentPhotoLibrary(delegate: delegate)
+        }
+    }
+
+    private func presentPhotoLibrary(delegate: MediaPickerDelegate) {
+        if UIImagePickerController.isSourceTypeAvailable(.photoLibrary) {
+            let photoPicker = UIImagePickerController()
+            photoPicker.title = String.localized("photo")
```
OzancanKaratas: This is not included by any xml file.

r10s: ahh, i see. @cyBerta this should be either "camera" or "gallery" then.

OzancanKaratas: Other non-included strings may also exist. Please use genstrings. For example, on the backup screen, I saw the loading string that could not be translated.

cyBerta: Thanks for the hints! I rebased onto master and fixed the string issue

cyBerta:
group and self avatar in different contexts as it is implemented now: 
self avatar, new group, edit group

Screen Shot 2019-11-21 at 09 39 16
Screen Shot 2019-11-21 at 09 41 10
Screen Shot 2019-11-21 at 09 42 41

PR #397, opened 2019-11-20
replace dismiss contact request by delete option (#365)

### LINKED ISSUE #365: Chat message menu item 'dismiss' doesn't work as expected
In the contact requests view it is possible to do a long-press on a message.
The opening menu contains an entry "Dismiss". If you tap on it, there's no visible change. Internally, this action marks a message as noticed. 

From a users point of view, the message should disappear from the contact request list.

cyBerta:
fixes #365


r10s:
nice :)

PR #398, opened 2019-11-20
do not show double items for startChat and block

r10s:
this pr removes "start chat" and "block" from the long-click menu; the options are already available in the single-click action-sheet.

also the is-writable-check was not really correct as it would add a startChat/block also in the device-chat.

the long-click-menu in the contact-requests is now the same as in a normal chat; this also simplifies future-improvements.


cyBerta:
lgtm

PR #408, opened 2019-11-26
Avatar picker for groups

### LINKED ISSUE #395: image picker for groups
On basis of PR https://github.com/deltachat/deltachat-ios/pull/386, the image picker component should be reused in the edit group context.

cyBerta:
- [x] pick avatar on group creation
- [x] change avatar for existing groups

closes #395

### DIFF
--- a/deltachat-ios/DC/Wrapper.swift
+++ b/deltachat-ios/DC/Wrapper.swift
@@ -153,6 +153,10 @@ class DcContext {
        }
        return nil
     }
+
+    func saveChatAvatarImage(chatId: Int, path: String) {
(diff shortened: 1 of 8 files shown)


deltachat-ios/DC/Wrapper.swift
```diff
@@ -153,6 +153,10 @@ class DcContext {
        }
        return nil
     }
+
+    func saveChatAvatarImage(chatId: Int, path: String) {
```
r10s: in general, i'd give these one-to-one wrappers for the core functions the same name as the core function. the case can be adapted. so eg. setChatProfileImage() in this case.

however, this should not block merging.

PR #409, opened 2019-11-27
Update avatars in lists

### LINKED ISSUE #394: Replace avatar in contact cell by InitialsBadge
This is a refactoring task.
The benefit from it will be that avatars will be scaled exactly the same way in all places. At the moment, ContactCell uses a workaround and adds an imageView to an UILabel as an attributedText property.

I consider the current state as a bug, as the badge can look like a stripe in contact cell, if the aspect ratio is not 1:1

### LINKED ISSUE #403: rename "profile photo" to "profile picture", use in contact lists
first off all, the new profile photo image picker is great! good job, @cyBerta :)

some nitpicking i found out during testing:

- the thing should be called "profile picture", this more general and fits to the footer text ("your profile picture and name will be shown alongside your messages ...") as well as eg. whatsapp.

- the set profile picture should be shown in the contact list - eg. group members as well as for "new chat" (see android). to find out the profile image for a contact to be shown in a list, use [dc_contact_get_profile_image()](https://c.delta.chat/classdc__contact__t.html#a2ebee8ac3729e4db85041eaf35bfce73) (not: dc_get_config()!) - currently, this is implemented for SELF only, however, we will spead the profile image from other user soonish [in the core](https://github.com/deltachat/deltachat-core-rust/issues/807) and it is great if the uis are prepared for that.

on android, all this is already implemented.

nb: the contact-image is also be shown as chat-image for one-to-one-chats, however, this is also managed by the core already.

@Simon-Laux can you please check what the current state is on desktop? and file an issue for the missing items (select SELF-image via dc_get|set_config(), use all contact images via dc_contact_get_profile_image())

cyBerta:
based on #408 (please review/merge before)
closes #403 
closes #394

### DIFF
--- a/deltachat-ios/Controller/ContactDetailViewController.swift
+++ b/deltachat-ios/Controller/ContactDetailViewController.swift
@@ -111,6 +111,13 @@ class ContactDetailViewController: UITableViewController {
         present(alert, animated: true, completion: nil)
     }
 
+    override func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
+        if section == 0 {
+            return ContactDetailHeader.cellHeight
+        }
+        return 0
+    }
+    
     override func tableView(_: UITableView, viewForHeaderInSection section: Int) -> UIView? {
         if section == 0 {
             let header = ContactDetailHeader()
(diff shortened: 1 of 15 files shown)


OzancanKaratas:
Hello @cyBerta,

I noticed an error when login.

If the password is not correct, Delta Chat shows an localized error text and a “fail: No” message. This may feel like a technical message. May you fix this?

PR #412, opened 2019-11-28
implement multiline text cell for signature (#404, #225)

### LINKED ISSUE #225: make signature input field multi-line
successor of https://github.com/deltachat/deltachat-ios/issues/218

- we should keep the UITableViewController to stay flexible.
- i think just showing 2-3 lines of input should do the job.

i would not see this as a high-priority issue.

### LINKED ISSUE #404: Signature can't be empty
- Device: iphone SE
- iOS version: 13.2
- Delta Chat version: 0.960
- Expected behavior: Be able to set the signature in settings to be empty
- Actual behavior: If set empty, the field will set back to the default value
- Steps to reproduce the problem: Open settings, clear content of signature field, close the view, open it again.

I guess this is due to TextFieldCell.getText() returning nil if the input is empty? https://github.com/deltachat/deltachat-ios/blob/cbd0b1b4598755504918a40b29c3020dd1ec3d32/deltachat-ios/View/TextFieldCell.swift#L63-L73

But I'm new to Swift, so please treat my hint as just an idea.

cyBerta:
closes #404 
closes #225


cyBerta:
ups. thanks for the hints!

cyBerta:
Screen Shot 2019-12-02 at 12 58 51
Screen Shot 2019-12-02 at 12 59 16

fixed

PR #413, opened 2019-11-30
add welcome device-message, show saved-messages

r10s:
counterpart on android: https://github.com/deltachat/deltachat-android/pull/1121

### DIFF
--- a/deltachat-ios/DC/Wrapper.swift
+++ b/deltachat-ios/DC/Wrapper.swift
@@ -37,6 +37,7 @@ class DcContext {
         return chatlist
     }
 
+    @discardableResult
     func createChat(contactId: Int) -> Int {
         return Int(dc_create_chat_by_contact_id(contextPointer, UInt32(contactId)))
     }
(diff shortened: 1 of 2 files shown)


deltachat-ios/DC/Wrapper.swift
```diff
@@ -157,6 +158,28 @@ class DcContext {
...
+
+    @discardableResult
+    func addDeviceMessage(label: String, msg: DcMsg) -> Int {
+        return Int(dc_add_device_msg(contextPointer, label.cString(using: .utf8), msg.cptr))
+    }
+
+    func addWelcomeDeviceMessages() {
+        // create the device-chat by adding a message to it
+        let msg1 = DcMsg(type: DC_MSG_TEXT)
+        msg1.text = "Welcome to Delta Chat! – Delta Chat looks and feels like other popular messenger apps, "
```
cyBerta: Why are you hardcoding the string here?

cyBerta:
besides the open question here: https://github.com/deltachat/deltachat-ios/pull/413/files#r352537770, it looks good to me. 
Personally, I would probably encapsulate the core calls in a DcMsg object, but that's not a big deal.

deltachat-ios/DC/Wrapper.swift
r10s: it's just now. once the text is settled, it will go to transifex as usual.

r10s:
turned this into wip, we decided yesterday that the welcome messasges etc. are added by the core, the ui can add messages on their own, however, just now this seems not to be needed.

so, we'll use sth. as dc_update_device_messages() that will do most of the things, see https://github.com/deltachat/deltachat-core-rust/issues/937

PR #415, opened 2019-12-02
Streamline attach menu

### LINKED ISSUE #389: streamline attach-menu
currently, the attach menu contains 2 options:

- camera (formerly "photo")
- video

the first is okay, however, the second in fact shows a gallery and the menu item should be names as such.
currently, the gallery shows only videos - that are recoded befored attached iirc, what is great!
but it would be even nicer when the gallery also shows the normal photos that can be attached as photos (also recoded) then.

sooner or later, we will also have a "Document" button, see https://github.com/deltachat/deltachat-ios/issues/376, however, this is out of scope of this issue.

cyBerta:
closes #389


r10s:
great!

the input-text-field is a bit nervous and fades in and out maybe one time too often :)
but this was also the case before and seems to be a more general issue that can be targeted somewhat later ;)

PR #421, opened 2019-12-07
hide incoming-messages-usernames in one-to-one-chats

r10s:
in these chats, the chatname is always the same as the sender's name


cyBerta:
@r10s adding the core update to this PR was done on purpose?

r10s:
> @r10s adding the core update to this PR was done on purpose?

no, thanks for the hint, i remove it

PR #423, opened 2019-12-09
fix suppression of background music

### LINKED ISSUE #420: deltachat stops playing music when opened
reported by @dignifiedquire on irc

@cyBerta maybe you can have a look at this, seems due to the voice-messages, you are more deep into audio currently :)

cyBerta:
closes #420 

But _if_ an audio message was selected, the audio background playback mutes and the user needs to turn it on manually after the audio message was finished.


r10s:
> But if an audio message was selected

"selected" means "played" here?

cyBerta:
> > But if an audio message was selected
> 
> "selected" means "played" here?

yes

PR #425, opened 2019-12-10
voice recording

cyBerta:
reimplementing IQAudioRecorderController in swift (only parts we need)


r10s:
recording does not work for me, did not manage to get the "done" button clickable in the record window, but might be i did not understand the logic fully :)

but in general, the approach to use a library is a reasonable approach for now.

would be nice if we can make the appearing window a bit easier to understand, eg. it could just start recording when opened - this way, the "done" button should be available always and the ui would be dead-simple :)

deltachat-ios/Controller/ChatViewController.swift
```diff
@@ -786,6 +790,14 @@ extension ChatViewController: MessagesDataSource {
         }
     }
 
+    private func sendVoiceMessage(url: NSURL) {
+        DispatchQueue.global().async {
+            let msg = DcMsg(viewType: DC_MSG_AUDIO)
```
r10s: that should be DC_MSG_VOICE, the difference is mainly that for voice messages the filename is not shown and they are not mixed with DC_MSG_AUDIO (which are eg. mp3 songs) if dc_get_next_media() is used at some point.

cyBerta:
> just for understanding: the main view is in swift now, and the other things are in objc, which, however, is no big deal as we probably do not touch these things. if needed, we can do an objc-to-swift conversion later. is this correct? i do not request changes, it is great as it is, i am just interested in the objc-to-swift state :)

so porting objc to swift is not so easy going as I thought. You _can_ use Siftify, but you have to pay more than 100 bugs to convert a project like the one I wanted to. The price depends on the size of the codebase I think. That's why I decided to invest some time to write the important things in swift, leaving out everything not really necessary. 

Still, the waveform library is in objc. I wouldn't touch that, I've seen already swift ports, but I don't know if they are as reliable as the one we're using (seems to have a good reputation).

deltachat-ios/Controller/ChatViewController.swift
cyBerta: Ups, iOS ignores that differnence at the moment (wrt. the filename)

PR #429, opened 2019-12-11
UI improvements for voice recording

cyBerta:


r10s:
i have also some nitpicking:
- the cancel button needs two touches. the first touch should just discard the thing
- can we have the cancel button always blue? as for the erroneous second-tap-cancel?
- can we have the waveform and the pause/record button always in blue? looks better and there is no need for this alarming red

cyBerta:
> i have also some nitpicking:
> 
>     * the cancel button needs two touches. the first touch should just discard the thing
> 
>     * can we have the cancel button always blue? as for the erroneous second-tap-cancel?
> 
>     * can we have the waveform and the pause/record button always in blue? looks better and there is no need for this alarming red

there are 2 cancel options: 
1: during the recording - cancel the recording, you can redecide how you phrase something and start again, by pressing on the mic button
2: if the recording was cancelled: dismisses the view.

I like the idea of having the option to dismiss everything with one button click. 
On the other hand I like to have the option to cancel only the recording.

blue and red are showing different states: red - a recording is ongoing or was started and is paused; blue: no recording is ongoing
the colors of the cancel button refer to both states

cyBerta:
What about this: 
we transfer the recording cancel button to the tabbar (as a trash bin symbol)?
This way the functionality is there and self explaining, and aborting the whole process can be done within 1 click

r10s:
> I like the idea of having the option to dismiss everything with one button click.
On the other hand I like to have the option to cancel only the recording.

so the first cancel cancels the whole recording? but you can start over just by attach -> voice-message.

i would not put too much controls to the ui for now, i think it is not really needed might raises the barriers if we decide to rework that at some point. otoh, i could live with a trash-bin beside the pause button if you really like this feature :)

ah, and one more thing: i would rename "Done" to "Send"

cyBerta:
ok sounds like we've found a compromise :)
* Cancel will be blue 
* 1 cancels everything (recording + dismiss ViewController)
* the "Done" button will be renamed as "Send"
* trash bin goes to the bottom

cyBerta:
active recording:

Screen Shot 2019-12-12 at 12 25 43


paused recording:

Screen Shot 2019-12-12 at 12 26 56


canceled recording (trash bin tapped):

Screen Shot 2019-12-12 at 12 22 20


same for dark mode:

Screen Shot 2019-12-12 at 12 22 52
Screen Shot 2019-12-12 at 12 23 21
Screen Shot 2019-12-12 at 12 23 30

PR #431, opened 2019-12-12
ui fixes attachment button

### LINKED ISSUE #430: do not show the attach-paper-clip disabled
the attach-paper-click in the chat always looks as if it is disabled - although it is never in this state. we should give it a "normal" gray color.

cyBerta:
* closes #430 
* increases the size of the send button and attachment button a little bit so that it's easier to hit


OzancanKaratas:
@r10s and @cyBerta, would you label the buttons for VoiceOver? Both are not labeled.

Issue #426, opened 2019-12-11
can't compile backtrace-sys v0.1.32


homdx:
Error compile with Xcode 10.2.1
error: failed to run custom build command for `backtrace-sys v0.1.32` (also and v.0.1.29 can't compile, then I setup it in Cargo.toml)

```
cargo --version
cargo 1.39.0 (1c6ec66d5 2019-09-30)
```

```
   /bin/sh -c /Users/devuser/Library/Developer/Xcode/DerivedData/deltachat-ios-fakgerxccwyvblbjnwoftnwijmcf/Build/Intermediates.noindex/deltachat-ios.build/Debug-iphonesimulator/deltachat-ios.build/Script-787D6698229F20B8000A7A9D.sh

info: component 'rust-std' for target 'aarch64-apple-ios' is up to date
info: component 'rust-std' for target 'x86_64-apple-ios' is up to date
[INFO  cargo_lipo::meta] Will build universal library for ["deltachat_ffi"]
[INFO  cargo_lipo::lipo] Building "deltachat_ffi" for "aarch64-apple-ios"
   Compiling backtrace-sys v0.1.29
   Compiling unicase v2.6.0
error: failed to run custom build command for `backtrace-sys v0.1.29`

Caused by:
  process didn't exit successfully: `/Users/devuser/Documents/Code/deltachat-ios/deltachat-ios/libraries/deltachat-core-rust/target/release/build/backtrace-sys-7e9279f59c3a20bd/build-script-build` (exit code: 1)
```

```
   Compiling backtrace-sys v0.1.32
   Compiling ring v0.16.9
   Compiling clear_on_drop v0.2.3
   Compiling libsqlite3-sys v0.16.0
   Compiling url v1.7.2
   Compiling url v2.1.0
   Compiling flate2 v1.0.13
   Compiling debug_stub_derive v0.3.0
   Compiling syn v1.0.11
   Compiling bytes v0.4.12
   Compiling rand_core v0.5.1
   Compiling futures-cpupool v0.1.8
   Compiling mio v0.6.21
   Compiling parking_lot v0.10.0
   Compiling uuid v0.6.5
   Compiling tempdir v0.3.7
   Compiling crossbeam-channel v0.4.0
   Compiling num-integer v0.1.41
   Compiling charset v0.1.2
   Compiling regex v1.3.1
   Compiling tokio-current-thread v0.1.6
   Compiling tokio-timer v0.2.12
   Compiling digest v0.8.1
[...]

PR #432, opened 2019-12-13
Setup Xcode command line tools

homdx:
Update instruction for setup missing headers (then give error not found)

#426


r10s:
@homdx great, thanks!

PR #435, opened 2019-12-16
implement empty states (no messages) #428

### LINKED ISSUE #428: show some text when a chat is empty
when a chat is empty, currently the whole screen is empty as well.

instead showing an empty screen, we should show some helpful text.

the text snippets are already there:

- for an **empty 1:1 chat:** `chat_no_messages_hint`
- for an **empty group:** if unpromoted: `chat_new_group_hint` - otherwise: `chat_no_messages`
- for an **empty contact request view:** `chat_no_contact_requests` if show_emails!=DC_SHOW_EMAILS_ALL - otherwise just use `chat_no_messages `
- for an **empty saved-messages chat:** `saved_messages_explain` 
- for an **empty device-chat:** `device_talk_explain` - (may happen when the user deletes all messages)

this is the android code for figuring out the different states: https://github.com/deltachat/deltachat-android/blob/master/src/org/thoughtcrime/securesms/ConversationFragment.java#L161

not sure what the easiest way is to realize that. on android we use some view that is shown instead of the list; this view shows the hints horizontally and vertically centered. but i would also be fine with a different solution - whatever is a simple way :)

cyBerta:
closes #428 

- implementation according to android


cyBerta:
Screen Shot 2019-12-18 at 22 24 45
Screen Shot 2019-12-18 at 22 25 42
Screen Shot 2019-12-18 at 22 26 02

r10s:
great :)

PR #443, opened 2019-12-18
set draft if app goes to background

### LINKED ISSUE #439: draft not saved when app pushed to background
if you type a draft and push the homebutton and do _not_ repoen the app immediately, the draft is gone.

we should call dc_set_draft() before the app goes to background or so.

cyBerta:
closes #439 

* note since iOS 9 there's no need to remove the implemented observer on viewDidDisappear()


r10s:
great!

PR #448, opened 2019-12-19
Date labels

### LINKED ISSUE #440: add self-containing date below messages
it is fine, to have relative dates as "now" or "2 minutes ago", however, when a time is shown it is sometimes unclear on which day this was.

in android, we always show a date that makes some use without scrolling up or down, we should do the same on ios.

the "combining" of messages with similar dates, however, is fine, this should stay (iirc, i had a different opinion some time ago :)

cyBerta:
* closes #440

### DIFF
--- a/deltachat-ios/Controller/ChatViewController.swift
+++ b/deltachat-ios/Controller/ChatViewController.swift
@@ -47,6 +47,8 @@ class ChatViewController: MessagesViewController {
     var msgChangedObserver: Any?
     var incomingMsgObserver: Any?
 
+    weak var timer: Timer?
+
     lazy var navBarTap: UITapGestureRecognizer = {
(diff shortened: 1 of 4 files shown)


cyBerta:
@r10s do we want to have a Timer, that updates the list every minute? This way the date labels will update automatically if the user doesn't scroll the list for a while.

r10s:
either a timer or we should not use relative time wordings.

r10s:
to the relative times:
- in the chatview, we should either show, "now", "1-59 minutes", the time or time plus date. sth. as "2 hours" is too rough here, see getExtendedRelativeTimeSpanString() in android. i think we should do it as there.
- in the chatlistview, we are more rough on android, also showing only the day of the week etc. , see getBriefRelativeTimeSpanString() in android

PR #452, opened 2019-12-22
remove some unneeded scaling that results in pixelated images

### LINKED ISSUE #451: avatars are too pixelated
avatars are sent in 192x192 pixels, but they look much more pixelated on ios and much worse than on android. seems as if they are scaled to 32x32, 48x48 or sth. like that before output :)

IMG_290290BFB45B-1

r10s:
closes #451 

before:

IMG_290290BFB45B-1

after:

IMG_76BE325AB147-1


cyBerta:
Did you check how the avatar selection cell looks like in settings -> your profile (with the default camera image?)

r10s:
@cyBerta not exactly sure what you mean :) can you check that, please?

cyBerta:
Screen Shot 2019-12-23 at 15 15 16
The scaling functionality was needed to have the camera icon placed correctly. We can also change the camera icon png and add with more empty space as padding. Otherwise I would like to propose a different programmatical approach (default scaling to 1.0, set scaling explicitly for profile selection cell).

r10s:
@cyBerta thanks for pointing this out, i will do a fix.

PR #453, opened 2019-12-22
fix deaddrop layout

### LINKED ISSUE #346: tweak contact requests
- background should be gray (take care of dark mode - android can give some hints)
- the shown avatar should be the one of the user requesting contact, not the "contact-request-group-icon" "CR"
- swiping should be disabled to switched to useful things

r10s:
closes #346

~~there is a little issue with the wrong background color in dark mode, maybe it is easy to fix, however, if not, this should not prevent things from being merged.~~ EDIT: fixed by the last commit 

this is how things look like now:

Screen Shot 2019-12-23 at 13 44 21


cyBerta:
nice!

PR #460, opened 2019-12-29
add option to open urls scanned from qr-codes, recognize more qr-codes

r10s:
~~did not test it as i forgot my lightning cable and the emulator cannot scan :)~~
tested, works as expected :)

test url eg. https://duckduckgo.com/?q=qr+https%3A%2F%2Fget.delta.chat


cyBerta:
nice!

PR #466, opened 2020-01-04
remove unused assets folder

r10s:
./Assets.xcassets seems not to be needed.
there is a similar folder, ./deltachat-ios/Assets.xcassets,
which content overlaps and this one seems to be the only used one.

however, might also be that i have missed something :) cc @cyBerta @nayooti


cyBerta:
It should be fine to delete this assets folder. This folder was not even referenced in the project.pbxproj file and thus not visible in xcode.

PR #467, opened 2020-01-04
say clearer what is done on configure-wait

r10s:
before, there was only "one moment" and a progress wheel that may be confused with a download indicator. now we say, that we're logging in to "your" server:

Screen Shot 2020-01-04 at 23 14 21

maybe not perfect, but it is an improvement. also that gets rid of the weird construction with three empty lines to make space for a progress-thingie.


cyBerta:
lgtm :)

PR #471, opened 2020-01-05
reorder new-chat options

r10s:


cyBerta:
looks better like that! +1

PR #472, opened 2020-01-05
[wip] update memberlist on qr-scan

r10s:
on group creation, when a qr-code was show at some point, the memberlist may change at any time (https://github.com/deltachat/deltachat-ios/pull/464/commits/c91d67576e731ab98fca953dbd2dd2507afe0d54 is needed, but not sufficient), in these cases, we get the event [DC_EVENT_CHAT_MODIFIED](https://c.delta.chat/group__DC__EVENT.html#ga27512e465c573fcf295014f8e0075adf) and we have to update the list.

there is still a bug in this pr, however, in theory, it should work like this. checked android, there it works as expected and the list is updated, so this seems to be no core bug.  
i will dive into the details at a later point.


cyBerta:
This can be closed, see https://github.com/deltachat/deltachat-ios/pull/473#issuecomment-571352257

PR #473, opened 2020-01-06
Fix dc event chat modified

cyBerta:
This PR only demonstrates a core bug (in case I understood the general qr code group join process correctly):

steps to reproduce:
* on device A: tap on new chat -> new group or new verified group 
* enter a group name
* tap on QR invite code
* on device B: scan QR code
* on device A: move back to new group / new verified group
* on device B: confirm alert "do you want to join group <XYZ>?"
* grep logs for "new created group chat id"

it will show that the chat id sent by DC_EVENT_CHAT_MODIFIED always stays 40. Instead the chat id should be the same as the return value of `dc_create_group_chat`

### DIFF
--- a/deltachat-ios/Controller/NewGroupController.swift
+++ b/deltachat-ios/Controller/NewGroupController.swift
@@ -74,8 +74,7 @@ class NewGroupController: UITableViewController, MediaPickerDelegate {
         ) { notification in
             if let ui = notification.userInfo {
                 if let chatId = ui["chat_id"] as? Int {
-                    logger.warning("new created group chat id: \(self.groupChatId) vs. DC_EVENT_CHAT_MODIFIED chat Id: \(chatId)")
-                    if self.groupChatId == 0 /*|| chatId != self.groupChatId*/ {
+                    if self.groupChatId == 0 || chatId != self.groupChatId {
                         return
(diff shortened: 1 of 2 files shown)


r10s:
the `40` is not static, it is the chat-id of the belonging one-to-one-chat. DC_EVENT_CHAT_MODIFIED is sent for the one-to-one-chat because the verification-status has changed.

but the DC_EVENT_CHAT_MODIFIED for the joined group is missing, so there is a bug anyway. i will do a pr to rust.

r10s:
did a fix in https://github.com/deltachat/deltachat-core-rust/pull/1141

i think we can close this pr now and see if https://github.com/deltachat/deltachat-ios/pull/472 works as expected then.

r10s:
reopened, i think either this one or https://github.com/deltachat/deltachat-ios/pull/472 or a mix of both should do the job then. @cyBerta can you continue on this?

cyBerta:
tested workaround with current submodule commit as well as 
e5dbb6a785a6581cb46bc79e11d62c909247c4c8 with https://github.com/deltachat/deltachat-core-rust/commit/8a394fb08f31bf75434295d90b6d0fa16727819d

PR #475, opened 2020-01-06
[WIP] Share with dc

cyBerta:


r10s:
link on this topic provided by @cyBerta on irc:  
https://medium.com/@wireapp/the-challenge-of-implementing-ios-share-extension-for-end-to-end-encrypted-messenger-dd33b52b1e97
https://www.avanderlee.com/swift/core-data-app-extension-data-sharing/

cyBerta:
so from what I understood, Signal has 
* a shared database
* does share most of the UI and messaging code between the containing app and the extension
* uses for sharing the code a framework (like described here for example: http://atinyfish.com/2014/06/27/ios-8-extensions-sharing-code-with-an-embedded-framework/)
* doesn't IPC communication between extension and containing app
* sends messages from within the extension and tries to ensure the extension runs as long as needed to send the message.

So the extension looks like the containing app, but actually it's not the same.

cyBerta:
There are hacks using a custom URL scheme, but it's not the apple way to do it: https://stackoverflow.com/questions/27506413/share-extension-to-open-containing-app

r10s:
we target that later

Issue #486, opened 2020-01-11
Address book integration is somewhat broken (incomplete list of contacts)


grossbart:
- Device: iPhone X
- iOS version: 13.3
- Delta Chat version: 1.0.2

### Expected behavior

When I give Delta Chat access to my address book, I expect to search for email addresses within all my contacts.

### Actual behavior

Out of my 2600 contacts (300 in one account, the rest in another), only about 20 showed up. And these were mostly people I don't interact with much. After this, I have found no other way to re-sync the contacts or to add more from the address book, so I had to add them manually.

I'm not sure on what basis these 20 contacts were selected, most of my other contacts have email addresses in them, too.

On the other hand, I'm no glad that not all have been added, as it creates a contact for each email, which leads to some people showing up four times.

### Steps to reproduce the problem

This is on a first-time install and it asked me only once whether I want to connect my contacts, to which I said yes. (I would want to remove them again and try this again, but there is no functionality for this).

### Logs

It seems that the logs don't go back enough (I tried this in the morning and now it's afternoon and I can't see any log messages from the morning anymore in the Console).

nayooti:
Thanks for submitting this issue.

I created this PR #488  to fix the issue. If the problem still exists with the next release we can reopen this issue.

FYI: DeltaChat v1.0x imports Contacts from the Default Account. To change your Default Account on your iOS-Device go to Settings -> Contacts -> DefaultAccount.

The fix will (hopefully) import contacts from all available accounts.

PR #488, opened 2020-01-13
Fix contact import (issue: #486)

nayooti:
Fix for issue: #486 

We only the defaultContainer to fetch device contacts. Now we iterate through all available containers.
I also moved the deviceContact-Fetching to a background thread as recommended.

### DIFF
--- a/deltachat-ios/Handler/DeviceContactsHandler.swift
+++ b/deltachat-ios/Handler/DeviceContactsHandler.swift
@@ -4,6 +4,11 @@ import UIKit
 class DeviceContactsHandler {
     private let store = CNContactStore()
     weak var contactListDelegate: ContactListDelegate?
+    let dcContext: DcContext
+
+    init(dcContext: DcContext) {
+        self.dcContext = dcContext
+    }
 
     private func makeContactString(contacts: [CNContact]) -> String {
         var contactString: String = ""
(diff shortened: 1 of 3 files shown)


deltachat-ios/Handler/DeviceContactsHandler.swift
```diff
@@ -18,35 +18,50 @@ class DeviceContactsHandler {
...
 
     private func addContactsToCore() {
-        let storedContacts = fetchContactsWithEmailFromDevice()
-        let contactString = makeContactString(contacts: storedContacts)
-        dc_add_address_book(mailboxPointer, contactString)
-        contactListDelegate?.deviceContactsImported()
+        fetchContactsWithEmailFromDevice() { contacts in
+            DispatchQueue.main.async {
+                let contactString = self.makeContactString(contacts: contacts)
+                dc_add_address_book(mailboxPointer, contactString)
```
cyBerta: you could move dc_add_address_book to the Wrapper.swift. I tried to ged rid of these unwrapped calls in the past when I was working on code that used them.

nayooti: fixed

r10s:
nitpicking: for the future, i would name the wrapper functions closer to the core-functions, so `addAddressBook()` instead of  `addContacts()` in this case.

i know there are several functions not following this pattern, however, we can try to improve this in the future.

i am not suggesting to change anything just now, it's a great pr!

PR #500, opened 2020-01-16
search bar chatlist

nayooti:
when the search is active the ChatList that before only displayed Chats actually displays Chats, Contacts, Messages.
We use the same CellType for that, that’s why I **renamed the cell `ContactCell` into `AvatarTextCell`** .
I also renamed `nameLabel`, `emailLabel` into `titleLabel` and `subtitleLabel`.

The Filtering-logic takes place in `ChatListViewModel`. 

The `AvatarTextCell` has a function called `update(cellViewModel)` that can be used in cellForRowAt in other ViewControllers.


nayooti:
- I transformed ChatListController into a UITableViewController because it natively supports searchbar hiding when scrolling and it makes sure that no cells are covered by the keyboard.

- my Iphone7-Simulator (iOS 12.4) has no issues launching DC
- cells are now white 
- regarding navigationBar-colors: as it looks now it is like native apps do supposed to look (WhatsApp + Telegram) have the same colors in the chatList-tab.

r10s:
it still does not boot up in my "real" account. this needs to be targeted very first.

just tested it in the emulator: nitpicking:
- the expanded search bar has the wrong color (some gray where it should be white)
- the line between the hidden search-bar and the chatlist seems to be too fat

r10s:
closing this for now, we'll come to this back later.

PR #507, opened 2020-01-31
Update InfoPlist.strings

neilzone:
Fix a typo and propose more detail for use of microphone


r10s:
hi @neilzone thanks for taking care!

do you know if these strings are actually visible somewhere?

i thought that the strings are overwritten by the localisation from  https://github.com/deltachat/deltachat-ios/blob/master/deltachat-ios/en.lproj/InfoPlist.strings - so maybe we should better use the same wording.

neilzone:
Hello Bjoern

The one I saw when I ran Delta Chat this morning was the "This is used to find your contacts who [are] using Delta Chat." string.

I didn't see the other one, but, since I was in the file, thought I'd fix it anyway.


IMG_1397

r10s:
thanks for the info, great! so, i think we can merge that right-away, it is an obvious improvement. we  can iterate from there anyway :)

Issue #491, opened 2020-01-13
add shared contacts to one-to-one-chat-profile


r10s:
at the position where group-chats show the members, we should show the "shared chats" in one-to-one-profiles. 

the headline for that is in the string `profile_shared_chats` and the list of shared chats can be retrieved using `dc_get_chatlist(context, 0, 0, contact_id)`

PR #508, opened 2020-01-31
Shared chats in profile

nayooti:
Implementation of Enhancement #491

### DIFF
--- a/deltachat-ios/View/ContactCell.swift
+++ b/deltachat-ios/View/ContactCell.swift
@@ -191,4 +191,43 @@ class ContactCell: UITableViewCell {
     required init?(coder _: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+
+
+    /*
+     this method can be deleted after searchBarContactList-branch was merged into master
+     */
+
+    func updateCell(cellViewModel: AvatarCellViewModel) {
+        // subtitle
+        emailLabel.attributedText = cellViewModel.subtitle.boldAt(indexes: cellViewModel.subtitleHighlightIndexes, fontSize: emailLabel.font.pointSize)
+
+        switch cellViewModel.type {
+        case .CHAT(let chatData):
+            let chat = DcChat(id: chatData.chatId)
+
+            // text bold if chat contains unread messages - otherwise hightlight search results if needed
+            if chatData.unreadMessages > 0 {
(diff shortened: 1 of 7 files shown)


deltachat-ios/View/ContactCell.swift
```diff
@@ -191,4 +193,43 @@ class ContactCell: UITableViewCell {
     required init?(coder _: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+
+
+    /*
+     this method can be deleted after searchBarContactList-branch was merged into master
```
r10s: i think, this can just be done by the searchBarContactList-branch, if then still needed.
it's always hard to predict the future :)

deltachat-ios/ViewModel/ContactDetailViewModel.swift
```diff
@@ -0,0 +1,84 @@
...
+    func getSharedChatIdAt(indexPath: IndexPath) -> Int {
+        let index = indexPath.row
+        // assert(sections[indexPath.section] == .sharedChats)
+        return sharedChats.getChatId(index: index)
+    }
+
+    func update(sharedChatCell cell: ContactCell, row index: Int) {
+        let chatId = sharedChats.getChatId(index: index)
+        let summary = sharedChats.getSummary(index: index)
+        let unreadMessages = context.getUnreadMessages(chatId: chatId)
```
r10s: while this is probably fine for the "shared chats" view, however, this may make problems when used at the "normal" chatlist. imu, these two functions are called for all chats in the chatlist - even if they are not in view and never will be in view - assume a list with 500 chats, this is a lot of overhead.

again, for 20 or so shared chats, this is not an issue, however, maybe this is already a good moment to think this over and use eg. a function that loads the data as really needed - the recycler view already works this way afaik, so that should be easy.

nayooti: this method is called as part of the cell-recycling process. it will be called for each cell that appears on screen right before it is scrolled in. The length of the chatList has no effect on how often this method is called.

r10s: i see, that would be just fine :)

i probably mixed that up with the loop (is it a loop?) `let _ = (0..<chatList.length).map` from https://github.com/deltachat/deltachat-ios/pull/500/files#diff-5e38f3d38776c60d414d5a59520ba530R114 - that would be bad then?

deltachat-ios/View/ContactCell.swift
nayooti: deleted the comment

Issue #509, opened 2020-01-31
Color of ContactCell transparent -  Color Definitions are a bit misleading 


nayooti:
The ContactCell background color is currently transparent. 

Look at `init` in `ContactCell.swift` you will see: 
`backgroundColor = DcColors.chatBackgroundColor`
`contentView.backgroundColor = DcColors.chatBackgroundColor`

and in `Colors.swift` it is defined as:
`static let chatBackgroundColor = UIColor.themeColor(light: UIColor(red: 255, green: 255, blue: 255, alpha: 0), dark: .black)`

The `alpha`-value is set to zero, meaning the color is transparent.

Is this intended?
Or should the Contact Cells-background not just be plain white? 

Also the ContactCell is used to display chats in the chatlist, where they appear white.

![Simulator Screen Shot - iPhone 11 Pro - 2020-01-31 at 17 53 48](https://user-images.githubusercontent.com/5249565/73558086-c1f09d00-4452-11ea-8265-9a798727c72d.png)

r10s:
> Or should the Contact Cells-background not just be plain white?

yip, they should be plain white, the transparency can be considered as a little ui bug :)

nayooti:
ok, I'll fix that quickly

PR #510, opened 2020-01-31
Changed color of contact cell

nayooti:
fix for #509


nayooti:
I will look into this.

nayooti:
I looked at other competitor apps and at how Apple uses darkmode in their products. 

We're using the ContactCell for 

1.  displaying chats in the `ChatList`

2.for displaying contacts in some other lists. 

I would definitely stick to black for 1).
Look how Apple does this:
[Apple Guidelines](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/dark-mode/) 
I would make them gray for 2). 
How Telegram does it in their ChatDetail.

r10s:
k, we can merge this, it is already an improvement.

whether to use "gray" or "back" on darkmode also depends a bit if we want to have some section headers or not, however, we can decide on that later, threre are lots of changes in this area anyway :)

PR #517, opened 2020-02-05
Local help

nayooti:
- added Local HTML files that are rendered in `HelpViewController` reachable from Settings (on bottom).
- html files are in project under Assets/Help/...
- added safe_fatalError and safe_assertions which can be safely used in code because they are only executed in Debug mode

**Issues:**
- CSS is rendered but the output looks very basic and IMO the font is too small (in landscape it is ok). It can be changed by adjusting the help.css file. I am not good with CSS.
- It does not respect darkmode. If this is required I will look into this, but I don't know if this actually is possible (because the styling is kinda hardcoded in the CSS file).


r10s:
great! i think, this is close to being mergable :)

however, we should not depend on the Desktop-Repo being at some place.
instead, we should just have all html files in this repo (this is already the case imu)  and we can add an update script later to the pages repo.
wrt dark mode: the css is "ours" no need to use the same as on desktop/android.

deltachat-ios.xcodeproj/project.pbxproj
```diff
@@ -341,6 +345,10 @@
...
 		AE0D26FC1FB1FE88002FAFCE /* ChatListController.swift */ = {isa = PBXFileReference; indentWidth = 4; lastKnownFileType = sourcecode.swift; path = ChatListController.swift; sourceTree = "<group>"; tabWidth = 4; };
 		AE18F293228C602A0007B1BE /* SecuritySettingsController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SecuritySettingsController.swift; sourceTree = "<group>"; };
+		AE19887423EB264000B4CD5F /* HelpViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = HelpViewController.swift; sourceTree = "<group>"; };
+		AE1988A423EB2FBA00B4CD5F /* Errors.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Errors.swift; sourceTree = "<group>"; };
+		AE1988A623EB382A00B4CD5F /* Help */ = {isa = PBXFileReference; lastKnownFileType = folder; path = Help; sourceTree = "<group>"; };
+		AE1988AA23EB3C7600B4CD5F /* Assets */ = {isa = PBXFileReference; lastKnownFileType = folder; name = Assets; path = ../../Desktop/Assets; sourceTree = "<group>"; };
```
r10s: this  `../../Desktop/Assets` should be removed, this prevents a successful compiling or requires adapting the environment. the assets should be in this repo, we can add do updates later directly from the deltachat-pages repo as we do for android and desktop

nayooti:
Fixed.

PR #523, opened 2020-02-12
Archive delete in chat detail

### LINKED ISSUE #490: add "Delete chat" and "Archive chat" options to chat-profile-info
currently, "Delete chat" and "Archive chat" are available only from within the chatlist.

so, if you're in a chat and decide, it can be archived/deleted, you have to go back to the chatlist and search the chat in the list which is a bit laborious.

android has a menu with these options inside the chat. as ios has no menu, i think we should add these options to the end of the chat-profie-info (where already "Leave group" and "Block contact" is) as follows:

Screen Shot 2020-01-13 at 14 25 35
Screen Shot 2020-01-13 at 14 20 48

maybe a bit nicer, but you got the idea :)

the corresponding strings are `menu_delete_chat` and `menu_archive_chat`/`menu_unarchive_chat`. before deletion, there should be a quesion with an safeActionSheet as it is in the chatlist. the archive option does not need a quesion for now.

between "archive" and the red-options, i would add a gap and maybe title the red options with DANGER (there is string `danger` for that)

nayooti:
closes #490 

Adds **Delete Chat** and **Archive Chat** Options to SingleChatDetail .

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -64,7 +64,7 @@ class ChatListController: UIViewController {
 
         deleteChatObserver = nc.addObserver(forName: dcNotificationChatDeletedInChatDetail, object: nil, queue: nil) { notification in
             if let chatId = notification.userInfo?["chat_id"] as? Int {
-                self.deleteChat(chatId: chatId)
+                self.deleteChat(chatId: chatId, animated: true)
             }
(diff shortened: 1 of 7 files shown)


r10s:
closes #490  - @nayooti if you add the string `closes #issuenumber` to the pr description, this results is back and forward references between prs and also the issue will be closed automatically on merging :)

PR #542, opened 2020-02-27
[wip] move help up

r10s:
this pr is about making the help-button more visible, by moving it up in the settings page (just after "my profile")


r10s:
just tested that in real-live with some ppl, it does not work.

ppl expect the "help" button being the last entry in the settings screen.

so, we'll keep this pr for reference, but close it without merging. if at some point, the settings dialog get much larger, we can overthing this, however, even then it is probably better to summarize other options.

PR #543, opened 2020-03-02
Manage keys

### LINKED ISSUE #511: implement 'manage private keys'
see android and desktop to get an idea.

cyBerta:
closes #511


r10s:
i would like to tune this down a bit, one of autocrypts findings is "do not talk about keys, ever" - of course, this is not completely doable when trying to be compatible to classic-pgp :)

in android, we've moved that to an "advanced" sections, however, we currently do not have that. but maybe we can move the button to the export section and maybe also name if "advanced" there. i will have a look at that.

PR #548, opened 2020-03-04
SettingsController now native

### LINKED ISSUE #519: Settings - Help-Button click event doesn't always go through
Realised it while working on the new HelpViewController:

When tapping in Settings-Tab on Help-Button it always gives optical feedback but the underlying event is not always triggered in 1 out of 10 cases. 
It's really hard to reproduce. It's seem to happen randomly on my Iphone 11 Pro and I could not trigger the error at all from my Iphone SE, both on iOS 13.3.  My feelings say it happens more often after restarting the app.

I tried setting a print-statement inside 
`openHelp`-method and it is indeed not printing anything when the error occurs. 

I assume it has something to do with the `QuickTableViewController`-library  and the way it triggers events. 

Btw: Originally I thought it is caused by the new HelpViewController that fetches HTML-data from disk. But it's not.

nayooti:
- removed QuickTableViewController from Pods because of #519 
- re-implemented SettingsController natively
- profileCell was created via factory method and the whole table was reloaded to manage updates - now profileCell is class and can be updated without reloading table

- After implementation I realized that the **help button still sometimes does not trigger action** - really weird.

- I suggest we merge this anyway and continue from there. Apart from refactoring the whole thing it has minor usability improvements.


r10s:
i just merged the "manage keys" button in #543, so, this needs a rebase now :)

the last requests in #543 renamed the manage key button to "advanced" and moved it to the autocrypt" section. moreover, i added some explanations before actually doing the import/export. @cyBerta maybe you can help on the rebase.

r10s:
also, the submodule should not be part of the pr.

r10s:
apart from that @nayooti nice, well readable pr. i also find the model with the sections pretty nice readable and also like that the stuff is self-contained and does not need some external files.

cyBerta:
I don't know (yet) if it is releated to the changes in this PR, but the pressed state of the ProfileCell is still weird - see the right darker part.

Screen Shot 2020-03-05 at 14 04 17

deltachat-ios/Controller/SettingsController.swift
```diff
@@ -20,15 +38,163 @@ internal final class SettingsViewController: QuickTableViewController {
...
+    private lazy var profileCell: ProfileCell = {
+        let displayName = DcConfig.displayname ?? String.localized("pref_your_name")
+        let email = (DcConfig.addr ?? "")
+        let selfContact = DcContact(id: Int(DC_CONTACT_ID_SELF))
+        let cell = ProfileCell(contact: selfContact)
+        cell.tag = CellTags.profile.rawValue
+        return cell
+    }()
+
+    private var contactRequestCell: UITableViewCell = {
```
cyBerta: what is the rationale behind the fact that some cells are marked as lazy and others not?

deltachat-ios/Controller/SettingsController.swift
```diff
@@ -90,222 +256,90 @@ internal final class SettingsViewController: QuickTableViewController {
...
-        }))
-        alert.addAction(UIAlertAction(title: String.localized("pref_managekeys_import_secret_keys"), style: .default, handler: { _ in
-            let msg = String.localizedStringWithFormat(String.localized("pref_managekeys_import_explain"), self.externalPathDescr)
-            let alert = UIAlertController(title: nil, message: msg, preferredStyle: .alert)
-            alert.addAction(UIAlertAction(title: String.localized("ok"), style: .default, handler: { _ in
-                self.startImex(what: DC_IMEX_IMPORT_SELF_KEYS)
-            }))
-            alert.addAction(UIAlertAction(title: String.localized("cancel"), style: .cancel, handler: nil))
-            self.present(alert, animated: true, completion: nil)
+            let documents = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true)
```
cyBerta: you can use startImex here, reduces code duplication

cyBerta:
nice, I like the PR, I made just some minor comments

deltachat-ios/Controller/SettingsController.swift
nayooti: I removed two unneeded 'lazy' statements.

By declaring objects lazy you 'delay' the instantiation of the corresponding object. It is needed in case the corresponding object/code block uses methods / attributes of the surrounding class (which need to be instantiated first).

As far as I know a lazy statement does not harm.

deltachat-ios/Controller/SettingsController.swift
nayooti: Fixed

r10s:
@cyBerta is the issue you mentioned at https://github.com/deltachat/deltachat-ios/pull/548#issuecomment-595219064 fixed? or was the issue invalid - at least i failed to reproduce it.

PR #549, opened 2020-03-04
Image and document gallery

### LINKED ISSUE #546: implement basic document and image gallery in profiles
in a chat profile there should be two buttons for image gallery and documents.
pressing on them should open the list of all documents / images

for now the idea is to keep it simple, using no tabbar, reusing already existing components etc.

cyBerta:
* implements a media section in contact and group profiles
* closes #546 
* drawback for now using existing components: tapping on gallery/documents opens the single preview first, a second tap on the list icon opens the item overview, there's no api to change that behavior of QLPreviewController (a native iOS component), thus custom views should be implemented in another PR

Screen Shot 2020-03-05 at 13 19 03
Screen Shot 2020-03-05 at 13 19 29
Screen Shot 2020-03-05 at 13 19 41

### DIFF
--- a/deltachat-ios/Controller/ContactDetailViewController.swift
+++ b/deltachat-ios/Controller/ContactDetailViewController.swift
@@ -49,19 +49,17 @@ class ContactDetailViewController: UITableViewController {
         return cell
     }()
 
-    private lazy var galleryCell: ActionCell = {
-        let cell = ActionCell()
-        cell.actionTitle = String.localized("gallery")
-        cell.actionColor = SystemColor.blue.uiColor
-        cell.selectionStyle = .none
+    private lazy var galleryCell: UITableViewCell = {
+        let cell = UITableViewCell(style: .value1, reuseIdentifier: nil)
+        cell.textLabel?.text = String.localized("gallery")
+        cell.accessoryType = .disclosureIndicator
         return cell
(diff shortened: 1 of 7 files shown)


r10s:
i think, this is a good, pragmatic approach to get this functionality in :)

however, i would like to tweak the buttons in the profile:

- the style of the buttons must not be a "centered blue link" but "left-aligned with arrow on the right" (as eg. for "show class emails" in the settings)

- i would move the two buttons up, just directly below the header with the avatar and put it to that section, so the "MEDIA" header can go away

- the "Add members" and "QR Invite code" buttons should go atop of the members section then (the "centered blue link" style is correct here)

for groups:
```
(A) GROUP NAME
[ Gallery             > ]
[ Documents           > ]

MEMBERS
[      Add members      ]
[    QR Invite code     ]
... memberlist & rest
```

for contacts:
```
(A) CONTACT NAME
[ Gallery             > ]
[ Documents           > ]

[     Send message      ]

SHARED CHATS
... chatlist & rest
```

"Send message" can be a separate own section without title or can just be the last item in the first section, not totally sure here, so i would chose whatever is simpler :)

cyBerta:
Screen Shot 2020-03-06 at 15 46 17

PR #553, opened 2020-03-06
replace textual sending state representation of messages by icons (#550)

### LINKED ISSUE #550: show sending status icons below messages
Instead of showing a textual representation of the current sending state, we should display the same icons as in the chat list.

cyBerta:
closes #550 

* I colored the icons so that we don't need to tint them programmatically while scrolling through lists. Tinting using UIImageViews is done easily but tinting UIImages directly requires more efford. Since we additionally need to scale down the UIImages for messages in the chat view, I preferred to reduce the programmatical complexity. But we can also decide to keep the images and tinting the UIImages, it is possible.


r10s:
@cyBerta so, the icons in the chatlist and in the chat-window are the same now, both pre-colored? this makes sense, great!

but i am wondering why the icons in the chat-window are a bit unsharp and appear more pixelated than in the chatlist.

cyBerta:
> @cyBerta so, the icons in the chatlist and in the chat-window are the same now, both pre-colored? 
yes they are the same

> but i am wondering why the icons in the chat-window are a bit unsharp and appear more pixelated than in the chatlist.

probably because the UIImage in the chat window is scaled down. I'll check if there's a way to get a better image quality

cyBerta:
There was a bug in the scaling extension.

before:
scale_old

now:
scale_now

r10s:
great you've figured that out!

PR #554, opened 2020-03-10
Location streaming

cyBerta:
closes #554


deltachat-ios/Assets.xcassets/ic_location.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "filename" : "ic_location_on_white_24dp-2.png",
```
r10s: please use a suffix that represents the scale, see most other images for an example,

cyBerta: The image assets are a mess with regard to that requirements. Anyway, I will try to get the icon with appropriate resolutions and name then correctly

cyBerta: fixed

PR #555, opened 2020-03-10
less clicks on save-message and start-chat

r10s:
this pr streamlines the ui, by removing some superfluous alerts and by re-ordering the forward-dialog so that "saved messages" is always one of the first entries.

counterpart of https://github.com/deltachat/deltachat-android/pull/1266 and https://github.com/deltachat/deltachat-core-rust/pull/1336/ , the core-part needs to be merged first


cyBerta:
@r10s due to other merged PRs this one has a conflict now

r10s:
@cyBerta i resolved all conflicts

cyBerta:
UI-wise I think having the saved messages below the pinned chats would be nicer.
Besides that it's a really nice feature.

r10s:
> UI-wise I think having the saved messages below the pinned chats would be nicer.
Besides that it's a really nice feature.

i fact, this was my first approach. but, after using it some days this way, i found out that it is better to have the "saved messages" _always_ at the same, first position.
also as this lacks the confirmation-dialog, it is maybe a bit easier to understand that the first item is a bit different.

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -157,8 +161,9 @@ extension ChatListController: UITableViewDataSource, UITableViewDelegate {
         }
 
         let chatId = chatList.getChatId(index: row)
+        let chat = DcChat(id: chatId)
         if chatId == DC_CHAT_ID_ARCHIVED_LINK {
-            return getArchiveCell(tableView)
+            return getArchiveCell(tableView, title: chat.name)
```
cyBerta: wow, the chat name has now the translated string for "Archived chats (#noOfChats)"?

r10s: yip, exactly :)

PR #559, opened 2020-03-12
Fix transparent navigation bar glitch

### LINKED ISSUE #518: Navbar color glitch during Push- and Pop-transition
On push- and pop-transition between two ViewControllers the NavigationBar has a darker color glitch during the Animation.

IMG_D1DC29B54889-1

nayooti:
fixes #518

NavigationBar is translucent by default, which causes the ui glitch.

NavigationBars no longer translucent.


nayooti:
I found a better approach. Should work now :)

PR #558, opened 2020-03-12
Message input bugfix

nayooti:
closes #521

Please read my comment to #521.

This PR contains rather a workaround than a fix.

PR #560, opened 2020-03-13
Fix inputBar (without Rust-core changes).

### LINKED ISSUE #521: input field hides and appears again very slowly
steps to reproduce:

- open a chat containing some images
- open an image
- close the image by _swiping down_
- => the input field is hidden _but_ reappears slowly in a minute or so

thanks @paulaluap for figuring that out :)

nayooti:
Got lost when attempting to extract Rust-core changes from #558, so I had to create a new branch.
closes #521 

PreviewController is now actually a QLPreviewController (before it was a object/class holding a QLPreviewController instance).

EDIT: fix link

### DIFF
--- a/deltachat-ios/Controller/ChatViewController.swift
+++ b/deltachat-ios/Controller/ChatViewController.swift
@@ -65,7 +65,6 @@ class ChatViewController: MessagesViewController {
     private var showNamesAboveMessage: Bool
     var showCustomNavBar = true
     var previewView: UIView?
-    var previewController: PreviewController?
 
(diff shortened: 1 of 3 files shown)


cyBerta:
nice. works as expected :)
Even though I don't fully understand how this fix does the trick

cyBerta:
I need to investigate if that issue is related to the PR at all.

nayooti:
I still have no clue, what exactly caused the issue. I assume it is some iOS-Bug.
@r10s idea is to fix the inputBar at the bottom, but I haven't found a solution for this yet.

PR #562, opened 2020-03-14
Bugfix for contact list search sync issue

### LINKED ISSUE #557: contact-list and contact-search can get out of sync
contact-list and contact-search get out of sync when:
- type some characters to the search, eg. `anne`
- tap on any result
- hit "cancel"
=> now the contact-list shows again all contacts while the contact-search shows still "anne". 

to fix that, the contact-list should not be changed - so, a user can revert an accidental click and select another result.

nayooti:
fixes #557 

Short explanation:
To present this Ask-To-Chat-With-Alert we have to close / deactivate the searchbar first.
Now when hitting cancel there was nothing that re-activated the searchbar.
This solved the issue.

There was still one problem. When deactivating the searchBar the chatList reloaded (showing all contacts). To fix this I had to change the `isFiltering`-Condition.

### DIFF
--- a/deltachat-ios/Controller/NewChatViewController.swift
+++ b/deltachat-ios/Controller/NewChatViewController.swift
@@ -39,7 +39,7 @@ class NewChatViewController: UITableViewController {
 
     // searchBar active?
     func isFiltering() -> Bool {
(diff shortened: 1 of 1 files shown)


r10s:
nice!

PR #575, opened 2020-03-18
WIP: SearchBar in ChatList (2nd Attempt)

nayooti:
fixes #500 

ChatList is now managed by ChatListViewModel.

ChatList registers on ViewModel to listen to changes (such changes occur during search/filtering).
ChatList also can trigger a table reload by calling `viewModel.refreshData`.

The ViewModel holds:
- chatList 
or 3 sections with SearchResults (updated as the user types):
- chats
- contacts
- messages

The ViewModel provides all relevant data to make `UITableViewDelegate/Datasource` happy.
The cell's data is stored in a class called `AvatarCellViewModel` which can be used for contacts and chats. 

Before this PR ChatListController was a normal `UIViewController` with a `tableView` as subview. I turned ChatListController into a `UITableViewController`, because the scrolling reacts better to Keyboard-events.

Also since SearchResults are grouped in sections, I changed the UITableView-style to `.grouped`, because it looks a lot better.

I did see no reason to add a SearchBar to ArchivedChatsList, because Archived Chats will appear in the normal search results, and it would have complicated things a lot. Therefor the ArchivedChatList's style is not `.grouped`.

**Issues:**
* when active search, tap search result, navigate back, there is a scroll-glitch. (edit: caused by KeyBoard-reapperance).

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -1,42 +1,67 @@
 import UIKit
 
-class ChatListController: UIViewController {
+class ChatListController: UITableViewController {
     weak var coordinator: ChatListCoordinator?
+    let viewModel: ChatListViewModelProtocol
+
+    private let chatCellReuseIdentifier = "chat_cell"
+    private let deadDropCellReuseIdentifier = "deaddrop_cell"
+    private let contactCellReuseIdentifier = "contact_cell"
+
+    private lazy var searchController: UISearchController = {
+        let searchController = UISearchController(searchResultsController: nil)
+        searchController.searchResultsUpdater = viewModel
+        searchController.obscuresBackgroundDuringPresentation = false
+        searchController.searchBar.placeholder = String.localized("search")
+        searchController.searchBar.delegate = self
+        return searchController
+    }()
 
     private var dcContext: DcContext
     private var chatList: DcChatlist?
     private let showArchive: Bool
-
-    private lazy var chatTable: UITableView = {
-        let chatTable = UITableView()
-        chatTable.dataSource = self
-        chatTable.delegate = self
-        chatTable.rowHeight = 80
-        return chatTable
-    }()
 
     private var msgChangedObserver: Any?
     private var incomingMsgObserver: Any?
     private var viewChatObserver: Any?
     private var deleteChatObserver: Any?
 
-    private var newButton: UIBarButtonItem!
+    private lazy var newButton: UIBarButtonItem = {
+        let button = UIBarButtonItem(barButtonSystemItem: UIBarButtonItem.SystemItem.compose, target: self, action: #selector(didPressNewChat))
+        button.tintColor = DcColors.primary
+        return button
+    }()
 
-    lazy var cancelButton: UIBarButtonItem = {
+    private lazy var cancelButton: UIBarButtonItem = {
         let button = UIBarButtonItem(barButtonSystemItem: .cancel, target: self, action: #selector(cancelButtonPressed))
         return button
     }()
 
-    init(dcContext: DcContext, showArchive: Bool) {
+    private lazy var archiveCell: UITableViewCell = {
+        let cell = UITableViewCell()
+        cell.textLabel?.textColor = .systemBlue
+        return cell
+    }()
+
+    init(viewModel: ChatListViewModelProtocol, dcContext: DcContext, showArchive: Bool) {
         self.dcContext = dcContext
+        self.viewModel = viewModel
         self.showArchive = showArchive
         dcContext.updateDeviceChats()
         super.init(nibName: nil, bundle: nil)
+        viewModel.onChatListUpdate = handleChatListUpdate
     }
 
     required init?(coder _: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+
+    // MARK: - lifecycle
+    override func viewDidLoad() {
+         super.viewDidLoad()
+         navigationItem.rightBarButtonItem = newButton
+         configureTableView()
+     }
 
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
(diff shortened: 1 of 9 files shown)


cyBerta:
I've played a little bit with it and overall I really like it. 
if you click on a message out of the search result, you don't get scrolled to the right position in the chat. This should be still implemented.

r10s:
> if you click on a message out of the search result, you don't get scrolled to the right position in the chat. This should be still implemented.

i would suggest to to this in another pr, let's first concentrate on the search.

cyBerta:
In dark mode, the search results and the section headers have the same background colors. On other screens the cells do have a slightly different color so that they differ from the header sections. Might be nice to have that here as well

cyBerta:
If you click on a contact out of the search results with whom you already created a chat, you should be directed to the chat instead of asked "Chat with <contact>?"

cyBerta:
Tapping on Contact Requests (out of search results) doesn't open the corresponding view, yet.

nayooti:
I think you are right that search results could look better in darkmode.
The problem about the darkmode:
Our ChatCells/ContactCells are black by default, so are the section headers.
I would not change default behaviours here in this PR.

nayooti:
> If you click on a contact out of the search results with whom you already created a chat, you should be directed to the chat instead of asked "Chat with ?"

nice! I take care of this! Tx!

nayooti:
> Tapping on Contact Requests (out of search results) doesn't open the corresponding view, yet.

I'll look into this. Tx!

r10s:
i tried it on my 1gb database - again hanging or crashing :( but, iirc the first attempt already crashed on app opening, so we're on the right way. but this has to be targeted very first, before adding more stuff to the pr

cyBerta:
Not all search results appear in the chats section:
currently on my device only the contact requests are the only search result for chats if I enter more than 1 character even though multiple chat names include them

deltachat-ios/ViewModel/ChatListViewModel.swift
```diff
@@ -0,0 +1,299 @@
...
+            filteredContactCellViewModels.append(viewModel)
+        }
+        searchResultsContacts.cellData = filteredContactCellViewModels
+
+        // #3 messages with searchPattern (filtered by dc_core)
+        let msgIds = dcContext.searchMessages(searchText: searchText)
+        var filteredMessageCellViewModels: [ChatCellViewModel] = []
+
+
+        for msgId in msgIds {
```
r10s: i think, this loop is one issue for hanging on larger databases: assume, dcContext.searchMessages() returns 5000 ids - the loop will load all details for all messages then. an even all chats for each messages and check the unread state for each message.

this should only happen for the few messages in view.

deltachat-ios/ViewModel/ChatListViewModel.swift
```diff
@@ -0,0 +1,299 @@
...
+            let msg: DcMsg = DcMsg(id: msgId)
+            let chatId: Int = msg.chatId
+            let chat: DcChat = DcChat(id: chatId)
+            let summary: DcLot = msg.summary(chat: chat)
+            let unreadMessages = dcContext.getUnreadMessages(chatId: chatId)
+
+            let viewModel = ChatCellViewModel(
```
r10s: also here: we create a view model for every search result. this should only happen for the few messages in view.

deltachat-ios/ViewModel/ChatListViewModel.swift
```diff
@@ -0,0 +1,299 @@
...
+        let contactIds: [Int] = dcContext.getContacts(flags: DC_GCL_ADD_SELF, queryString: searchText)
+
+        let contacts = contactIds.map { return DcContact(id: $0) }
+
+        for contact in contacts {
+            let nameIndexes = contact.displayName.containsExact(subSequence: searchText)
+            let emailIndexes = contact.email.containsExact(subSequence: searchText)
+
+            // contact contains searchText
+            let viewModel = ContactCellViewModel(
```
r10s: this may easily create a 1000 cells - we should only create cells that scroll to view.

deltachat-ios/ViewModel/ChatListViewModel.swift
```diff
@@ -0,0 +1,299 @@
...
+            )
+            filteredChatCellViewModels.append(viewModel)
+        }
+        searchResultsChats.cellData = filteredChatCellViewModels
+
+        // #2 contacts with searchPattern in name or in email
+        var filteredContactCellViewModels: [ContactCellViewModel] = []
+        let contactIds: [Int] = dcContext.getContacts(flags: DC_GCL_ADD_SELF, queryString: searchText)
+
+        let contacts = contactIds.map { return DcContact(id: $0) }
```
r10s: this loads all contacts where only contacts in view should be loaded

deltachat-ios/ViewModel/ChatListViewModel.swift
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    private func filterAndUpdateList(searchText: String) {
+
+        // #1 chats with searchPattern in title bar
+        var filteredChatCellViewModels: [ChatCellViewModel] = []
+        var flags: Int32 = 0
+        flags |= DC_GCL_NO_SPECIALS
+        let filteredChatList = dcContext.getChatlist(flags: flags, queryString: searchText, queryId: 0)
+        _ = (0..<filteredChatList.length).map {
```
r10s: compared to contacts and messages, this loading of all chats is probably a smaller issue, but still, we should not load 100 chats when only 8 are in view

deltachat-ios/ViewModel/ChatListViewModel.swift
nayooti: The msgIds-array in line 272 only consists of the search results.

nayooti: Ok, I will try to create the cell data on the fly.

deltachat-ios/ViewModel/ChatListViewModel.swift
nayooti: I think my last commits should have fixed it, but I am not able to test it since my DB is a little less than 1GB.

nayooti:
I think I am done fixing.

r10s:
great stuff! finally it works with my >1gb database - and fast :)

some tiny things:
- the background color for pinned chats is lost somehow
- the "Archived chats (123)" links should be centered again

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -85,231 +118,147 @@ class ChatListController: UIViewController {
...
+        case .contact(let contactData):
+            let contactId = contactData.contactId
+            if let chatId = contactData.chatId {
+                coordinator?.showChat(chatId: chatId)
             } else {
-                cell.setBackupImage(name: chat.name, color: chat.color)
+                self.askToChatWith(contactId: contactId)
             }
-
-            if chat.visibility == DC_CHAT_VISIBILITY_PINNED {
```
r10s: this condition is lost on refactoring,

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -85,231 +118,147 @@ class ChatListController: UIViewController {
...
     }
 
-    func getArchiveCell(_ tableView: UITableView) -> UITableViewCell {
-        let archiveCell: UITableViewCell
-        if let cell = tableView.dequeueReusableCell(withIdentifier: "ArchiveCell") {
-            archiveCell = cell
-        } else {
-            archiveCell = UITableViewCell(style: .default, reuseIdentifier: "ArchiveCell")
-        }
-        archiveCell.textLabel?.textAlignment = .center
```
r10s: this textAlignment also lost on refactoring

nayooti: Ok, tx. It's now fixed.

r10s:
closed in favour to #577

PR #581, opened 2020-03-21
WIP: Counter search headers

### LINKED ISSUE #579: add counter to search-headers
the search-headers currently read only "CHATS", "CONTACTS", "MESSAGES".

it would be great, if we can add the number of results to each header, so "5 CHATS", "6 CONTACTS", "7 MESSAGES".

the strings needed for that are already available and translated as `n_chats`, `n_contacts` and `n_messages` (the strings have different singular/plural forms, grep for n_members to get an idea how to use it)

btw, this is what it looks like on android:

nayooti:
fixes #579 

Implemented counting titles.

Somehow the Master complains about 
DC_GCL_FOR_FORWARDING 
in ChatListViewModel.swift Line 78.


r10s:
superseded by #583

PR #585, opened 2020-03-21
Search and chatlist fixes

### LINKED ISSUE #578: search and chatlist fixes
the new search from #577  introduces a new few issues that need to be targeted:

- the contact request cell does not show the timestamp (maybe this was always missing) (the unread counter, however, should stay hidden)

- when a message is forwarded, this is done by just going back to the chatlist, refreshing the list and scrolling to the top, https://github.com/deltachat/deltachat-ios/blob/master/deltachat-ios/Controller/ChatListController.swift#L72  
however, in addition to this, an existing search should be ended and the archive should not be shown by default, i tried to add just an viewModel.endSearch() , however, this was not sufficient :)

nayooti:
fixes #578

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -72,6 +72,7 @@ class ChatListController: UITableViewController {
         viewModel.refreshData()
 
         if RelayHelper.sharedInstance.isForwarding() {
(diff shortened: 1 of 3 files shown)


r10s:
yeah! good job, @nayooti, @cyBerta!

PR #587, opened 2020-03-21
Close item appearance condition fixed

### LINKED ISSUE #570: gallery is no longer full screen
introduced by one of the most recents merges, the gallery (opened from a chat) is no longer fullscreen. or is this intended? then it looks still weird ;) cc @cyBerta

nayooti:
fixes #570 

The condition that determines the appearance of the close-item was wrong. Fixed that.

Not sure, what you mean with full-screen?
Since iOS13 a modally presented (animated from bottom) UIViewController is not fullscreen by default anymore.  But we do get this nice swipe-to-dismiss-feature which we use here.

### DIFF
--- a/deltachat-ios/Controller/PreviewController.swift
+++ b/deltachat-ios/Controller/PreviewController.swift
@@ -23,7 +23,7 @@ class PreviewController: QLPreviewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
-        if navigationController != nil && isBeingPresented {
+        if navigationController?.isBeingPresented ?? false {
             /* QLPreviewController comes with a done-button by default. But if is embedded in UINavigationContrller we need to set a done-button manually.
             */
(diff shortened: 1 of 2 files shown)


r10s:
> Since iOS13 a modally presented (animated from bottom) UIViewController is not fullscreen by default anymore. But we do get this nice swipe-to-dismiss-feature which we use here.

well, but for a gallery, one really wants fullscreen, cmp the other apps using a gallery, also the gallery available from the profile.

but let me try to get a hand on #521 first, maybe we can do other things.

nayooti:
I changed it to fullscreen. As said above, it comes with the price of losing swipe-to-dismiss.

r10s:
>  it comes with the price of losing swipe-to-dismiss.

this is okay, and not expected for fullscreen i'd say. everything's fine now ;) great job :)

PR #589, opened 2020-03-22
AppStart also respects last active chat

### LINKED ISSUE #541: save app state between starts
currently, the app always starts in the chatlist, but it should start where it was left, as a minimum the last chat.

nayooti:
fixes #541 

* renamed `TabBarRestorer` to `AppStateRestorer`
* `AppStateRestorer` is now a singleton, so it can be accessed from everywhere 
* added methods to store and restore chatIds (will be set/reset in ChatView-lifecycle-delegate-methods).
* AppCoordinator.start restores chat if possible
* `AppCoordinator.showChat` can be run without animation (otherwise the user would see the animation from `chatList` to `chatView` right at app start).

### DIFF
--- a/deltachat-ios/Handler/AppStateRestorer.swift
+++ b/deltachat-ios/Handler/AppStateRestorer.swift
@@ -40,10 +40,8 @@ class AppStateRestorer: NSObject, UITabBarControllerDelegate {
         let activeTab = tabBarController.selectedIndex + offsetKey
 
         if let tab = Tab(rawValue: activeTab), tab != .chatTab {
-            // reset last active chat
             resetLastActiveChat()
         }
-
 
         UserDefaults.standard.set(activeTab, forKey: lastActiveTabKey)
         UserDefaults.standard.synchronize()
--- a/deltachat-ios/Handler/AppStateRestorer.swift
+++ b/deltachat-ios/Handler/AppStateRestorer.swift
@@ -44,13 +44,11 @@ class AppStateRestorer: NSObject, UITabBarControllerDelegate {
         }
 
         UserDefaults.standard.set(activeTab, forKey: lastActiveTabKey)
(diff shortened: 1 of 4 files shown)


deltachat-ios/Handler/AppStateRestorer.swift
```diff
@@ -33,9 +38,34 @@ class TabBarRestorer: NSObject, UITabBarControllerDelegate {
...
+        }
+
         UserDefaults.standard.set(activeTab, forKey: lastActiveTabKey)
         UserDefaults.standard.synchronize()
     }
 
+    private func storeChat(chatId: Int?) {
+        let value = chatId ?? -1
+        UserDefaults.standard.set(value, forKey: lastActiveChatId)
+        UserDefaults.standard.synchronize()
```
cyBerta: From the docs: 
 
-synchronize is deprecated and will be marked with the API_DEPRECATED macro in a future release.
-synchronize blocks the calling thread until all in-progress set operations have completed. This is no longer necessary. Replacements for previous uses of -synchronize depend on what the intent of calling synchronize was. If you synchronized...
     - ...before reading in order to fetch updated values: remove the synchronize call
     - ...after writing in order to notify another program to read: the other program can use KVO to observe the default without needing to notify
     - ...before exiting in a non-app (command line tool, agent, or daemon) process: call CFPreferencesAppSynchronize(kCFPreferencesCurrentApplication)
     - ...for any other reason: remove the synchronize call


I tested the PR without this method call and it works fine on iOS 13. I didn't test with older devices though

nayooti: Oh wow. Didn't know. That's actually good news :) 

I will delete the .synchronize thing,

r10s:
great!

PR #592, opened 2020-03-23
relayout message cells on screen orientation change

### LINKED ISSUE #556: fix info message layouting
The info messages in a chat view (like "Location streaming enabled", "Location streaming disabled" should be centered in the chat view. For some reason these messages are after a orientation change alternating between left- and right-alignment but they're never centered. Scrolling the list up and down "fixes" the issue.

infomessagebug

cyBerta:
closes #556


nayooti:
Nice!
FYI: There is also 
[traitcollectiondidchange](https://developer.apple.com/documentation/uikit/uitraitenvironment/1623516-traitcollectiondidchange).
It is called on device-rotations + Ipads split screen events.

cyBerta:
> Nice!
> FYI: There is also
> [traitcollectiondidchange](https://developer.apple.com/documentation/uikit/uitraitenvironment/1623516-traitcollectiondidchange).
> It is called on device-rotations + Ipads split screen events.

is traitcollectiondidchange preferred?

nayooti:
notifications can be handled from anywhere and it has a one-to-many-relationship. It has it's merits when the listener is not directly linked to the notifier.

With `traitCollediontDidChange` you can be a bit more specific since the arguments give you info about device classes and changes etc.
And we will probably support split-screen on ipads at some point. Maybe we should give it a try.

cyBerta:
@nayooti  have a look at https://stackoverflow.com/a/37353850
I double checked the implementation and realized we already use `viewWillTransition` to detect layout changes. I removed the observers and used that method instead.

nayooti:
Maybe this is not related to this PR (because it is completely independent of device rotation). 
But when I have location streaming on, switch it off, and on again, I get this:
C39058B4-1A18-44F8-B9CF-9C46C98B6DC0

cyBerta:
> Maybe this is not related to this PR (because it is completely independent of device rotation).

it is related. What I proposed here is a workaround rather that a fix. 
It is like a heisenbug: if I turn on the debugger, the layouting works as expected, but without, these layout issues appear. I think it is related to parallel event handling, but I couldn't really pin down the cause

cyBerta:
@nayooti ready for review round 2

nayooti:
Sorry the merge was by accident when I tried out a little fix.

It's all aligned well! Nice!

Only thing I realised, that after switching from portrait to landscape the message list would never scroll to bottom. I tried something out, which seemed to worked. 

You can have a look at:
[Fix](https://github.com/deltachat/deltachat-ios/pull/592/commits/3ab2321c39cec06e0f8f02e0c0a2b11ae4e5efeb)

After you had a look, I will gladly approve the PR.

cyBerta:
Yes codewise it looks good. Didn't test it, but you did it already.

PR #593, opened 2020-03-23
Fixed contact list scroll bug

### LINKED ISSUE #591: Contact List scrolling is lagging
.... apparently true on larger contact lists.

nayooti:
fixes #591 

@r10s reported the contact list was lagging on larger contact list.

I don't have such a long contact list on my device but I found the potential bug and fixed that.
We had a computed property that virtually created cell-data arrays on the fly, and this thing was actually only used to specify the length of the list. 

I also updated the way the cells are updated, using our new ContactCell.update function.


r10s:
> i compared the contact lists with android and I realized that the verified icons needs to be displayed as well, if a contact is verified. This is still missing.

they were displayed before, just checked that on current master.

PR #594, opened 2020-03-23
scroll to top if filtering of contact list changed

### LINKED ISSUE #588: In Contact List  search events (input, cancel etc) should scroll up list
see title

cyBerta:
closes #588 
I don't like the resulting UX though. What I at least sometimes do is entering some characters, scrolling a little and entering more characters if I don't see contact in question yet. In this case jumping again to the beginning of the list feels unsatisfying as I need to orientate newly.
@r10s @nayooti what do you think?

### DIFF
--- a/deltachat-ios/Controller/GroupMembersViewController.swift
+++ b/deltachat-ios/Controller/GroupMembersViewController.swift
@@ -414,6 +414,7 @@ class GroupMembersViewController: UITableViewController, UISearchResultsUpdating
 
         filteredContacts = contactsWithHighlights.filter { !$0.indexesToHighlight.isEmpty }
         tableView.reloadData()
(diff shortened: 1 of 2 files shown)


r10s:
what are other apps doing in this case on ios? i quickly checked the address book and telegram - they also scroll up.

i'd say this makes some sense - if you leave the scrolling position, the items changed after changing the search query are totally random, it may or may not help on orientating or may add some confusion. whereas always scrolling to the top always results in a reliable behavior.

nayooti:
Approved :)

We could add a heuristic:
if the user in the middle of the list, and the user changes the input -> we can check, if one contact item is still in the list (not filtered out), and scroll there. Otherwise scroll to top.

PR #604, opened 2020-03-30
Remove mailbox pointer

cyBerta:


nayooti:
singletons are usually using the keyword `shared` on iOS.
`UIApplication.shared. ...`

cyBerta:
> singletons are usually using the keyword `shared` on iOS.
> `UIApplication.shared. ...`

Thanks, I've changed that.

deltachat-ios/View/ContactCell.swift
```diff
@@ -231,7 +231,7 @@ class ContactCell: UITableViewCell {
     }
 
     // use this update-method to update cell in cellForRowAt whenever it is possible - other set-methods will be set private in progress
-    func updateCell(dcContext: DcContext, cellViewModel: AvatarCellViewModel) {
+    func updateCell(cellViewModel: AvatarCellViewModel) {
```
nayooti: Yeah!

We could also consider adding the needed DcChat-object to the cellViewModel.cellData, in case we want to avoid accessing DcContext.shared here.  Maybe in another PR.

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -41,8 +42,9 @@ class ChatListController: UITableViewController {
         return cell
     }
 
-    init(viewModel: ChatListViewModelProtocol) {
+    init(dcContext: DcContext, viewModel: ChatListViewModelProtocol) {
```
nayooti: I would probably prefer to not inject `dcContext` here, since we do have a `viewModel` in this class that already holds the `dcContext`-object.
Instead I would just alter the viewModel by  
- `func getChat(chatId: Int)`
- `func createChatByMessageId()`
and use the viewModel instead.

nayooti:
My comments are just based on personal preferences and could also be done in an upcoming PR. 

So therefor: Approved.

deltachat-ios/DC/Wrapper.swift
```diff
@@ -18,6 +19,10 @@ class DcContext {
         dc_context_unref(contextPointer)
     }
 
+    static var shared: DcContext {
+        return .dcContext
```
r10s: i would suggest to get rid of this, it seems to be a "better hidden" global, but will still has its disadvantages, eg. it is unclear about the timing when it is created, there cannot be two instances of DcContext etc.

also, code-wise, i think it does not make things easier, but adds some complexity.

i would suggest to just create a DcContext object in AppDelegate as in the past and pass around the dcContext pointer - in most cases, this is already done this way.

deltachat-ios/AppDelegate.swift
```diff
@@ -17,7 +16,7 @@ enum ApplicationState {
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterDelegate {
-    private let dcContext = DcContext()
+    private let dcContext = DcContext.shared
```
r10s: i would just create a normal object a before here, this is much simpler, more understandable and also allows other usage of `DcContext` in the future

deltachat-ios/Controller/EditGroupViewController.swift
```diff
@@ -77,7 +77,7 @@ class EditGroupViewController: UITableViewController, MediaPickerDelegate {
         if let groupImage = groupImage, let dcContext = coordinator?.dcContext {
             AvatarHelper.saveChatAvatar(dcContext: dcContext, image: groupImage, for: Int(chat.id))
         }
-        dc_set_chat_name(mailboxPointer, UInt32(chat.id), newName)
+        _ = DcContext.shared.setChatName(chatId: chat.id, name: newName ?? "")
```
r10s: we should get the dcContext through the `init()` function and use this object. i think, we should get rid of `DcContext.shared`

deltachat-ios/Controller/GroupMembersViewController.swift
```diff
@@ -325,6 +324,7 @@ class GroupMembersViewController: UITableViewController, UISearchResultsUpdating
     var selectedContactIds: Set<Int> = []
 
     init() {
+        self.dcContext = DcContext.shared
```
r10s: better pass `dcContext` to `init()` and get it from the parameter

deltachat-ios/Controller/NewChatViewController.swift
```diff
@@ -47,7 +47,7 @@ class NewChatViewController: UITableViewController {
     var hud: ProgressHud?
 
     lazy var deviceContactHandler: DeviceContactsHandler = {
-        let handler = DeviceContactsHandler(dcContext: DcContext())
```
r10s: wow, this was really a bug. this created new context with new connections, database handles and everything - as a side-effect, this seems to prove, that the same database can be opened twice at the same time :) cc @flub :)

however, instead of using DcContext.shared, we should just use `self.dcContext`

deltachat-ios/DC/Wrapper.swift
```diff
@@ -497,7 +586,7 @@ class DcChat {
 
     // use DcContext.getChat() instead of calling the constructor directly
     init(id: Int) {
-        if let p = dc_get_chat(mailboxPointer, UInt32(id)) {
+        if let p = dc_get_chat(DcContext.shared.contextPointer, UInt32(id)) {
```
r10s: we should pass the `dcContext` through `init()`

deltachat-ios/DC/Wrapper.swift
```diff
@@ -919,37 +1008,32 @@ class DcMsg: MessageType {
...
     }
 
-    func createChat() -> DcChat {
-        let chatId = dc_create_chat_by_msg_id(mailboxPointer, UInt32(id))
-        return DcChat(id: Int(chatId))
-    }
-
     func sendInChat(id: Int) {
-        dc_send_msg(mailboxPointer, UInt32(id), messagePointer)
+        dc_send_msg(DcContext.shared.contextPointer, UInt32(id), messagePointer)
```
r10s: we should pass `dcContext` as a parameter instead of using global `DcContext.shared`. this is also true for the other usages of `DcContext.shared.contextPointer` in this file.

deltachat-ios/DC/Wrapper.swift
nayooti: Depends on what we want. If we want to use DcContext as a singleton,  this is the way to go. There won't be more than one instances (because the initialiser is `private`). 

We still could instantiate and set the dcObject in AppDelegate, so it will be there at the very app start.
I would generally like to avoid the pattern:
`(UIApplication.shared.delegate as! AppDelegate).dcContext`

r10s: i think, we do not want to use DcContext as a singleton, i think, we should take the chance to use it as a normal object, this makes usage more flexible.

a "singleton" still is a global pointer under a different name, and this "shared" thingie adds some complexity and hidden things.

eg. it is completely unclear _when_ the object is actually created - in fact, even before AppDelegate is created and it is a bit luck which other objects (Bundle, ...) are available at that point - and even more bad, this may even change. i would not depend on that - also as there is no real reason for that, even if we keep a global pointer.

nayooti: Ok.
We could instantiate it in `AppDelegate.didFinishLaunchingWithOptions`.

r10s: instancing it at the current position is okay, it is a member of AppDelegate and AppDelegate is create at some well defined point.

however, using a global value for the shared thing makes it unclear, when the shared pointer is really initialized - it may be at the moment where it is used at AppDelegate - but it may als be much sooner. even if this is documented somewhere, i think this is some complexity we should avoid. also, these kind of things are obviously hard to understand and to communicate.

that's my point.

getting rid of the global pointer will solve this issue - with the additional advantage that the use of DcContext is more flexible.

r10s: for convenience, we may come to the point where we want to save the dcContext somewhere globally, however, this should be done outside DcContext somewhere.

cyBerta:
> ~i think, we can still merge that, but i would like to get rid of DcContext.shared in a subsequent pr~
> 
> while maybe we can leave the DcContext.shared, the creation time is unlucky - it is created before even AppDelegate is created, as a global value. it is _not_ created in AppDelegate as the code suggests. even it this works now, this hidden flow is unneeded and a call for trouble :)

Hm. Unfortunately, I don't get the point about the initialization yet. Static variables are always initialized lazily. As soon as DcContext.shared was called the first time, DcContext initializes 'itself'. 

My idea with the singleton was to 1.) encapsulate all native calls, 2.) avoid having too many changes in the code. DcContext needs to be injected nearly everywhere otherwise. 

My first approach was actually to inject DcContext, but at some point it felt like a bad, messy idea. I'm fine to go this path though :)

r10s:
maybe you're right, but still, this kind of things seems not to be obvious, at least not to me.

i agree that a global pointer can keep the number of code changes low, however, i would not add that to DcContext then - this would make future use of multiple DcContext impossible - and i think when we're on it, it makes sense to make this right.

deltachat-ios/Controller/NewChatViewController.swift
flub: :)

PR #605, opened 2020-03-30
Welcome screen

### LINKED ISSUE #580: add welcome screen
on first startup and after deleting the account, there should be a "welcome" screen shown before forwarding to the account creation.

the screen shall show a title, a short message and the delta-logo.

as "login to your server" button, we agreed with @nayooti to use a standard-system-blue-button.

the needed strings are `welcome_desktop` (sic!), `welcome_intro1_message` and `welcome_start_messaging`

this is what it looks like on android:

the buttons "Scan QR code" and "Import backup" will be part of a subsequent pr, this issue is about changing the flow to a welcome screen and implementing the welcome screen.

nayooti:
fixes #580 

To trigger: logout - run DC via Xcode
-  without qr_code and import_backup-Buttons as specified (buttons + layout is implemented but currently not showing). 
- welcome screen's layout is highly responsive. It is put into a scrollview that kicks in once screens get too small (for example on most Iphones in landscape).
- currently the loginController is modally presented over the welcome screen with the default iOS-13-swipe-to-dismiss-style. I like it, but if @r10s prefers fullscreen, I can change that quickly. 

the login button currently looks a little lost. It will look ok, once the missing buttons will be added. They are centered-y-aligned from bottom edge to subtile label.


r10s:
looks good - at least as good as the original :) (see below for ppl that cannot build the app on their own)

Screen Shot 2020-03-30 at 16 43 58

r10s:
and: dark-mode needs tweaking. 

i would be fine with completely skipping dark-mode for the welcome screen, otoh, adaption is probably very simple.

currently, the font is changed to light-gray on darkmode, that's it.

deltachat-ios/Controller/WelcomeViewController.swift
```diff
@@ -0,0 +1,298 @@
...
+        let buttonStackBottomAnchor = buttonStack.bottomAnchor.constraint(equalTo: buttonContainerGuide.bottomAnchor, constant: -50)
+
+        _ = [buttonStackTopAnchor, buttonStackBottomAnchor].map {
+            $0.priority = .defaultLow
+            $0.isActive = true
+        }
+
+/*
+         // alternative approach without stackview
+        loginButton.centerXAnchor.constraint(equalTo: container.centerXAnchor).isActive = true
```
cyBerta: can be removed, right?

deltachat-ios/Assets.xcassets/ic_launcher.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
+{
```
cyBerta: You don't use the ic_launcher imageset from what I can see. I think it can be removed?!

cyBerta:
Besides these comments I really like the PR and also the design. Looks great in dark mode :)

nayooti:
Glad you like it.

The code in comments describes an approach without stackview. I left it, because after we add qr- and backup-buttons, there might be the case that we don't want to use the stackview anymore. My intention was to leave it there until the design is final.

r10s:
nice!

PR #606, opened 2020-03-30
Scroll chat to search result

### LINKED ISSUE #602: scroll chat to search result

nayooti:
fixes #602


cyBerta:
you could squash both commits to one, since the first one is basically reverted with the second commit, but if it's too cumbersome then nevermind :)

deltachat-ios/Controller/ChatViewController.swift
```diff
@@ -381,6 +390,17 @@ class ChatViewController: MessagesViewController {
...
 
+    func scrollToMessage(id: Int, animated: Bool) {
+        for (index, msg) in messageList.enumerated() {
+            if msg.id == id {
+                let indexPath = IndexPath(row: 0, section: index)
+                messagesCollectionView.scrollToItem(at: indexPath, at: .top, animated: animated)
+                return
+            }
+        }
+        safe_fatalError("message could not be found")
```
cyBerta: this is too harsh. On my first try, the app already crashed. The reason was that the message was not within this first ones that are loaded, but already 'hidden'. I would have to pull to refresh in the chat view to have the searched message in the list.

cyBerta: Ok, this will not crash in production but still it's annoying a bit.

r10s: doesn't that always happen when clicking on a search result that is not in the last 100 or so messages?

r10s:
> It would be super nice if the scroll feature would also work for older messages

this is why i pointed out that we may need '"real virtual" list of messages to target' at #602 - but yes, as a first step, this is already great.

deltachat-ios/Controller/ChatViewController.swift
nayooti: Ok, I probably got this wrong. I didn't know/considered, that messages are incrementally fetched

r10s:
closing this for now, i think sometimes scrolling to a messages an sometimes not is worse than just always showing the chat at the end.  
we can come back to that when we actually have a virtual message list and can scroll to random positions.

PR #611, opened 2020-04-01
Qr login welcome

### LINKED ISSUE #608: allow account creation by qr-code-scan
- add a button "Scan QR code" to the new welcome screen

- on tapping, open a new screen with the option to scan a qr-code, check the scanned qr-code as usual with [`dc_check_qr()`](https://c.delta.chat/classdc__context__t.html#a34a865a52127ed2cc8c2f016f085086c)  
of course, the scan-screen also needs a "cancel" button as the manual-setup screen.

- if a `checkQr()` returns with `lot.state` set to `DC_QR_ACCOUNT`, ask the user if he wants to create a new account on the domain given as `lot.text1`

- if the user wants to do so, pass the whole scanned qr-code to [`dc_set_config_from_qr()`](https://c.delta.chat/classdc__context__t.html#ac95d7275f01862fba70169b8cbb937c0) and continue with `dc_configure()` as usual.

- if things work out, the user ends in the chatlist as usual - on errors, the user should be redirected to the welcome or, if possible, to the scan-qr screen.

this is how things look like on android - the alert is shown over the welcome-screen for $technicalReasons, i would prefer them laying over the scan-screen, if possible on ios:

### LINKED ISSUE #610: QRCodeReaderController does not like device rotations
54BCAB7B-6A06-4506-B50A-B0282FC94612
B4B229B7-C736-40EE-BF81-8F6A856D2E9A

nayooti:
fixes #610 
fixes #608 

Realised that our QRReaderController had some bugs, so fixed that. 
I also changed that the QRReaderController is presented on top of the QR-tab (instead of being pushed).  It is presented on WelcomeScreen also, and it is a lot easier to reuse, if on each case the reused VC is either pushed or presented. And presenting makes a lot more sense too.

Whats need to be tested:
scan non valid qr-code (use a random non valid qr-code) -> should show Error message over qrReader (translations are still missing)
scan valid qr-code ->   should show Confirm-Alert (cancel, ok). 

Cancel will currently redirect to WelcomeView (is this wanted?). 
Ok, will redirect to WelcomeController (showing Spinner) while trying to create account. 

If creation fails, Error message is shown (**need some additional text here**). On success user will be redirected to Chat-Tab. (On next iteration we will add options to add Username, Image etc.). 

To test this PR you need a QR-Code (can be provided by @r10s) .


deltachat-ios/DC/Wrapper.swift
```diff
@@ -139,6 +139,16 @@ class DcContext {
...
     }
 
+    func configureAccountFromQR(qrCode: String) -> Bool {
+        let state = checkQR(qrCode: qrCode).state
+        if state != DC_QR_ACCOUNT {
+            return false
+        }
+        let success = dc_set_config_from_qr(contextPointer, qrCode)
+        configure()
+        return success == 1
```
r10s, line 149: i think, this is wrong. configure() always returns immediately and error/success is reported by DC_EVENT_CONFIGURE_PROGRESS, see AccountSetupController to get an idea, maybe copy/paste the code from there ;)

nayooti:
I added Progress Alert (AccountSetupController and WelcomeViewController use now the same ProgressAlert-code).

In combination with progress alert I wasn't happy with the spinner, so I changed that, too. Spinner only replaces the buttons on the welcome view  now.

Hope you like it.

deltachat-ios/Controller/QrViewController.swift
```diff
@@ -318,3 +240,81 @@ class QrViewController: UITableViewController, QrCodeReaderDelegate {
         navigationController?.pushViewController(chatVC, animated: true)
     }
 }
+
+// MARK: - QRCodeDelegate
+extension QrViewController: QrCodeReaderDelegate {
```
cyBerta, line 245: out of curiosity: this is a style question to move delegate implementations to the extension, right?

cyBerta:
very cool that you implemented the ProgressAlertHandler this way so that every UIViewController can make use of it's functionality now.

I have to other things to comment: 
1. the string for qraccount_ask_create-and_login is still missing, but I think you are aware of that and there's probably a plan for it, right?
2. I tried an invalid QR code and what I got was this Alert Dialog. I don't know if this is intended.
Screen Shot 2020-04-06 at 13 11 41

deltachat-ios/DC/Wrapper.swift
nayooti: this has been fixed.

nayooti:
@cyBerta As stated in Issue-specs the localised strings are still missing. 

I'll look into the Error-alert. I possibly mixed two alert-calls (calling the wrong alert function).
Thank you!

deltachat-ios/Controller/QrViewController.swift
nayooti: Technically it makes no difference, so yes, it's a style question.

IMO it improves readability by separating a class after concerns. And if you use "MARK-statements" you can find them very easily. 

If you decide later to extract the delegate to a different class (for example if you want the class' viewModel to implement the delegate instead of the class), you can do so by basicly changing one word (`extension ViewModel: QRCodeReaderDelegate`).

nayooti:
@cyBerta 
I checked the error-alert you mentioned. If you scan a non-dc-related QRcode you get this warning (translations are still missing):
IMG_5EB752BAB5FB-1

I assume you scanned a dc-related qr code where the account creation failed. Did you?

Here it does show a plain-error alert as you showed in your screenshot. I stated in the PR-description above (in bold text), that we need a more specified error-message for this case.

cyBerta:
G

> I stated in the PR-description above (in bold text), that we need a more specified error-message for this case.

👍 Then let's create a string key for it and add that to the alert!
In general, we have the option to add yet untranslated strings to the untranslated.xml which you can find in deltachat-ios/tools/untranslated.xml . If you run` ./tools/tx-pull-translations.sh ` afterwards the untranslated strings will be shown on runtime (instead of the key).

nayooti:
@cyBerta Cool! Did just that.

PR #612, opened 2020-04-02
Shared db

### LINKED ISSUE #595: Create shared database
* we need a shared database if we want to send messages with shared data to contacts
* create app groups, as an identifier for both host and extension app
* check app version and check if database exists, start database migration to shared persistent containter if necessary
* change database file path to shared persistent container
* main app should work without any issues like before, no further extension work needs to be done within the scope of this task

https://www.avanderlee.com/swift/core-data-app-extension-data-sharing/

cyBerta:
closes #595 

f581fd6b545f27c85cf843180beef1d559e4f7f7 can be checked out to move the database and blobs back and forth between the local and the shared storage (for testing). These methods are removed in the last commit.

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -125,16 +125,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         NotificationCenter.default.removeObserver(self, name: .reachabilityChanged, object: reachability)
     }
 
-    func dbfile() -> String {
-        let paths = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true)
-        let documentsPath = paths[0]
-
-        return documentsPath + "/messenger.db"
-    }
-
     func open() {
-        logger.info("open: \(dbfile())")
-        dcContext.openDatabase(dbFile: dbfile())
+        guard let databaseLocation = DatabaseHelper().testMove(toShared: true) else {
+            fatalError("Database could not be opened")
+        }
+        logger.info("open: \(databaseLocation)")
+        dcContext.openDatabase(dbFile: databaseLocation)
     }
 
     func setStockTranslations() {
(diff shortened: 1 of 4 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -125,16 +125,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         NotificationCenter.default.removeObserver(self, name: .reachabilityChanged, object: reachability)
     }
 
-    func dbfile() -> String {
```
nayooti, line 128: I would also use a computed property here as you did DatabaseHelper.

cyBerta: This method has been deleted :).

deltachat-ios/Controller/AccountSetupController.swift
```diff
@@ -815,6 +815,7 @@ class AccountSetupController: UITableViewController {
             appDelegate.stop()
             appDelegate.close()
             do {
+                //TODO: remove also blobs folder?!
```
cyBerta, line 818: @r10s @nayooti what do you think about this todo? I think we should clear the blobs folder if the user deleted the account, right?

cyBerta: I'll do a seperate PR for this, as it will add some other changes in AccountSetupController that are out of scope of this PR

deltachat-ios/Helper/DatabaseHelper.swift
```diff
@@ -0,0 +1,94 @@
+import Foundation
+class DatabaseHelper {
+
+    var sharedDbFile: String {
+        guard let fileContainer = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: "group.eu.merlinux.group.chat.delta.ios") else {
```
r10s, line 5: can you please move `group.eu.merlinux.group.chat.delta.ios` to a constant and add a comment about where this string comes from and why it is needed this way. also this "double" `group.` part needs some explanations.

cyBerta: I've moved the group id to a constant. The double group is unnecessarey IMO. I only reused the preconfigured one that was already available.

cyBerta: I also added the explanation

PR #620, opened 2020-04-08
ui for autodelete

r10s:
this pr adds the ui needed to drive the new autodelete-features:

Screen Shot 2020-04-11 at 02 24 26 Screen Shot 2020-04-11 at 02 30 20 

corresponding android pr: https://github.com/deltachat/deltachat-android/pull/1257

### DIFF
--- a/deltachat-ios/Controller/SettingsAutodelOverviewController.swift
+++ b/deltachat-ios/Controller/SettingsAutodelOverviewController.swift
@@ -1,0 +1,88 @@
+import UIKit
+
+class SettingsAutodelOverviewController: UITableViewController {
+
+    var dcContext: DcContext
+
+    private struct SectionConfigs {
+        let headerTitle: String?
+        let footerTitle: String?
+        let cells: [UITableViewCell]
+    }
+
+    private enum CellTags: Int {
+        case autodelDevice = 0
+        case autodelServer = 1
+    }
+
+    func autodelSummary() -> String {
+        return String.localized("off")
+    }
+
+    private lazy var autodelDeviceCell: UITableViewCell = {
+        let cell = UITableViewCell(style: .value1, reuseIdentifier: nil)
+        cell.tag = CellTags.autodelDevice.rawValue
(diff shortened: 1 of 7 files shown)


deltachat-ios/Controller/SettingsAutodelOverviewController.swift
```diff
@@ -0,0 +1,110 @@
...
+        guard let cell = tableView.cellForRow(at: indexPath), let cellTag = CellTags(rawValue: cell.tag) else {
+            safe_fatalError()
+            return
+        }
+        tableView.deselectRow(at: indexPath, animated: false) // to achieve highlight effect
+
+        switch cellTag {
+        case .autodelDevice:
+            let controller = SettingsAutodelSetController(dcContext: dcContext, fromServer: false)
+            navigationController?.pushViewController(controller, animated: true)
```
cyBerta, line 107: I would prefer if the single source for pushing ViewController is the coordinator. It would help to keep the architecture consistent

cyBerta: This would imply to implement a coordinator for this class

cyBerta:
Why do the numbers of mails-to-delete-immediately (in the alert) change depending on the selected time?

cyBerta:
Besides the coordinator - architecture aspect the PR looks good to me. 👍

deltachat-ios/Controller/SettingsAutodelOverviewController.swift
r10s: yip, i also thought about that. but what would be the advantage? this would requires a class, indirections of calls etc. tbh, i also do not get the point on other palces in the code.

r10s:
> Why do the numbers of mails-to-delete-immediately (in the alert) change depending on the selected time?

because the selected time defines which messages should be deleted. if you select eg. "after 4 weeks", 3 weeks old messages are not delete - but these 3 weeks old messages would be deleted if you select "after 1 day"

PR #623, opened 2020-04-09
DcCore framework

cyBerta:
- there is still some code duplication in Utils.swift and one or two extensions that I would like to discuss with you. Besides that the core is separated into an own framework


nayooti:
Somehow I find the DcCore.framework twice in my project pane (in deltachat-ios). Is this intended?
image

DcCore/DcCore/Extensions/UIColor+Extensions.swift
```diff
@@ -0,0 +1,64 @@
+import UIKit
+
+internal extension UIColor {
```
nayooti, line 3: What's the reason of having this here? Don't see where we are using it.

If we want to have `UIColor+Extensions` in `DcCore` for later use, we should (try to) declare these methods in here public (and stop declaring the whole thing `internal`)so we can get rid of `deltachat-ios/Extensions/UIColor+Extensions.swift`.

deltachat-ios/AppDelegate.swift
```diff
@@ -51,6 +52,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
         let console = ConsoleDestination()
         logger.addDestination(console)
+        dcContext.logger = DcLogger()
```
nayooti, line 55: We're still having a global `logger` in `AppDelegate.swift line 9` which we are using a lot.

DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -667,37 +669,37 @@ class DcChat {
                 let image = UIImage(data: data)
                 return image
             } catch {
-                logger.warning("failed to load image: \(filename), \(error)")
+                DcContext.shared.logger?.warning("failed to load image: \(filename), \(error)")
```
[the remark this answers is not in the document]
cyBerta: There's only one logging object in this file. It's a member of DcContext. Other classes in this file reference this logging interface via DcContext.shared.logger?.<METHOD> while DcContext itself references it like this: logger?.<METHOD>

DcCore/DcCore/Helper/Utils.swift
```diff
@@ -0,0 +1,107 @@
+import Foundation
+import UIKit
+import AVFoundation
```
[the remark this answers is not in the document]
cyBerta: fixed

cyBerta:
> Somehow I find the DcCore.framework twice in my project pane (in deltachat-ios). Is this intended?
fixed

deltachat-ios/AppDelegate.swift
cyBerta: That's right. I didn't want to have SwiftyBeaver as a depenency in DcCore. That's why I've defined a Logger interface in DcCore. The main module implements that Logger interface in the class DcLogger and delegates all calls to the global Swifty beaver instance (the `logger` you mentioned).

DcCore/DcCore/Extensions/UIColor+Extensions.swift
cyBerta: DcChat uses the extension

cyBerta: I've declared the methods and the extension public and reduced the code duplication now

nayooti:
LGTM

PR #633, opened 2020-04-15
Share extension

cyBerta:


r10s:
tried to test it: "use of undeclared type 'ChatListDelegate'" - i think just a file that was forgotten to check in :)

cyBerta:
sharing other content than text is still missing, I can add it here or in a subsequent PR

r10s:
> sharing other content than text is still missing

if sharing-other-content-to-delta is offered by ios but not handled by Delta Chat yet, i'd say, do it in this pr to avoid getting a master that only partly works.

if share-to-delta is only available for content that can be handled by delta, i would be fine with merging it now.

@cyBerta i am failing testing this pr, can you please rebase to master? thanks :)

r10s:
@cyBerta great! i could also build that now :)

i'd say, if we actually send out the images/whatever already shown in the ui, we can merge+ship.

i would like to streamline some things in the ui, also wondering if the "quick links" in the share dialog (the ones with preselected recipients, shown by apple before the share-app is stated) are already supported.  
but both can be separate pr then.

nayooti:
I just shared images! Yeah!
But I think I found a bug:

When I sent first image it was shown in the share window. 
Then I tried to send another image and the image was not shown in the share window. 

IMG_18CB2260A180-1

cyBerta:
Can you forward me the example file?

nayooti:
> Can you forward me the example file?

I've been playing around with it a bit.
Strangely, I can only trigger this if I use GooglePhotos. If I use Apple's Photo app it all works flawless.

Open GooglePhotos -> select image -> share -> DeltaChat -> Send -> back in GooglePhotos with same Photo -> share -> DeltaChat -> Images are not shown.

cyBerta:
> Open GooglePhotos -> select image -> share -> DeltaChat -> Send -> back in GooglePhotos with same Photo -> share -> DeltaChat -> Images are not shown.

And the first time the thumbnail is shown?

nayooti:
> > Open GooglePhotos -> select image -> share -> DeltaChat -> Send -> back in GooglePhotos with same Photo -> share -> DeltaChat -> Images are not shown.
> 
> And the first time the thumbnail is shown?

Yes.

cyBerta:
> When I sent first image it was shown in the share window.
> Then I tried to send another image and the image was not shown in the share window. It did still sent the image, so prob a UI-thing.

I've fixed that issue by customizing the preview instead of using the default mechanism in 
a6d0891

cyBerta:
>     * images and shared urls are not forwarded
> 
>     * the forward dialog is a bit weird, it is a bit larger and has some shading around
> 
>     * when forwarding images, you cannot forward them without entering some text (the "send" button is disabled until text is entered)

@r10s can you please double check that you checked out the latest commit? 
as a fast proof have a look at line https://github.com/deltachat/deltachat-ios/blob/aa35a98f8972ea0d57f48196767b180bbcde6c17/DcShare/Controller/ShareViewController.swift#L83 in your local sources. From what you describe ("you cannot forward them without entering some text") your code looks probably different atm.

PR #634, opened 2020-04-15
Debug toolkit moved to Settings

### LINKED ISSUE #496: move log and debug-toolkit to Settings/Advanced/View log
currently, the DBDebugToolkit opens when the device is shaked - this conflicts with the text-undo-function - and also happens easily by accident. todo:

- do not open the toolkit on device shaking

- instead, add the entry "View log" (String `pref_view_log`) to the alert shown at "Settings/Advanced" ("View log" is not a 100% match, however, it is the wording used on other uis, it is good enough for now)

- on opening, the DBDebugToolkit screen should be shown. if possible, i would prefer to also create the object just in time - this way, startup will be faster, and, more important, we do not pull-in problems of the DBDebugToolkit (there were already some in the past) to the general startup routine.

- finally, on opening, the custom variables should be updated - eg. `display_name` should represent the name acutally in the database, not only at the moment the DBDebugToolkit is created. maybe this can be achieved easily by just recreating the toolkit on hitting Settings/Advanced/View log.

nayooti:
fixes #496 

The DBToolkit no longer can be triggered by shaking the device. 
Instead go to:
**Settings-Tab** -> **Advanced**->**View Log**

Basically moved all the DBToolkit from AppDelegate to SettingsCoordinator, so also the instantiation is setup there.

Unfortunately the DBDebugToolkit doesn't provide any presentation options at least not without forking the repo. At least on my device I don't see any animations when presenting the whole thing (even though internally it actually has animations turned on).

### DIFF
--- a/deltachat-ios/Coordinator/AppCoordinator.swift
+++ b/deltachat-ios/Coordinator/AppCoordinator.swift
@@ -327,10 +327,14 @@ class SettingsCoordinator: Coordinator {
     }
 
     func showDebugToolkit() {
-
-        let vc = UIViewController()
-        vc.view.backgroundColor = .red
-        navigationController.present(vc, animated: true, completion: nil)
+        DBDebugToolkit.setup(with: [])  // emtpy array will override default device shake trigger
+        DBDebugToolkit.setupCrashReporting()
+        let info: [DBCustomVariable] = dcContext.getInfo().map { kv in
+            let value = kv.count > 1 ? kv[1] : ""
+            return DBCustomVariable(name: kv[0], value: value)
+        }
+        DBDebugToolkit.add(info)
+        DBDebugToolkit.showMenu()
     }
 }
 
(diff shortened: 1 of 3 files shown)


deltachat-ios/Coordinator/AppCoordinator.swift
```diff
@@ -321,8 +325,20 @@ class SettingsCoordinator: Coordinator {
         let helpViewController = HelpViewController()
         navigationController.pushViewController(helpViewController, animated: true)
     }
+
+    func showDebugToolkit() {
+        DBDebugToolkit.setup(with: [])  // emtpy array will override default device shake trigger
+        DBDebugToolkit.setupCrashReporting()
```
cyBerta, line 331: This means that crash reporting will only work if the user has opened the DebugToolkit once.  Maybe for reproducible bugs it's ok though...

PR #635, opened 2020-04-17
Name and Avatar Selection after QR-Account-Creation

### LINKED ISSUE #609: ask the user for a name an an avatar after qr-code-login
this should be targeted when #608 is done.

- before showing the chatlist, the user should be asked for a name and a qr-code, together with some string as "login successful, blabla, this is your email address".

- on android, i've reused the account settings for that, might be that this is the easiest approach for ios as well (of course, the other options (signature, account details) needs to be hidden then)

currently, the screen should only be shown if actually a qr-code was scanned before, however, in the future, we may want to show this also for manual account creation.

this is how it looks like on android:

nayooti:
fixes #609

No issues, but ....

... I am not too happy with the transition from our WelcomeScreen to the MainScreen (TabBar).
Since we're switching the UIWindow when we navigate from WelcomeScreen to MainScreen  there is no chance to add nice animations to the login flow. 

For example after the account creation is completed, I had to switch windows first, before I could present the Avatar/NamePicker. So there are a few milliseconds where the user can actually see the MainScreen. So this makes the login flow a bit wonky from the animation perspective. 

Also keep in mind, that this PR is based on master, there is another PR  #632 which also fix minor things. I suggest to merge #632 first. I can do a rebase after. 

Also have a look on Ipad when selecting an Avatar how the UIAlertController is presented. If this is preferred, I would do the same on other occasions, because that's how it is usually implemented on Ipad.


r10s:
closed in favor to #642

PR #636, opened 2020-04-17
use dcContext.lastErrorString instead of AppDelegate.lastErrorString

r10s:
use dcContext.lastErrorString instead of AppDelegate.lastErrorString; the latter is just never set, maybe it was used at some time in the past and was forgot on some refactoring

### DIFF
--- a/deltachat-ios/Controller/QrViewController.swift
+++ b/deltachat-ios/Controller/QrViewController.swift
@@ -141,9 +141,9 @@ class QrViewController: UITableViewController {
             // execute blocking secure join in background
             DispatchQueue.global(qos: .background).async {
                 self.addSecureJoinProgressListener()
-                AppDelegate.lastErrorString = nil
+                self.dcContext.lastErrorString = nil
                 let chatId = self.dcContext.joinSecurejoin(qrCode: code)
-                let errorString = AppDelegate.lastErrorString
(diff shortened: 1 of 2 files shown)


deltachat-ios/Controller/QrViewController.swift
```diff
@@ -141,9 +141,9 @@ class QrViewController: UITableViewController {
             // execute blocking secure join in background
             DispatchQueue.global(qos: .background).async {
                 self.addSecureJoinProgressListener()
-                AppDelegate.lastErrorString = nil
+                self.dcContext.lastErrorString = nil
                 let chatId = self.dcContext.joinSecurejoin(qrCode: code)
-                let errorString = AppDelegate.lastErrorString
+                let errorString = self.dcContext.lastErrorString
```
nayooti, line 146: To get the logic: Does `dcContext.joinSecureJoin(qrCode: code)` update/manipulate `dcContext.lastErrorString` in case of an error?

r10s: in directly, yes. in case of errors, the core sends DC_EVENT_ERROR that handled by https://github.com/deltachat/deltachat-ios/blob/master/DcCore/DcCore/DC/events.swift#L20 - and sets self.dcContext.lastErrorString.

AppDelegate.lastErrorString was just _never_ set or changed.

PR #641, opened 2020-04-19
restore chat only if existant

### LINKED ISSUE #630: Gracefully handle non existing chatId
Currently delta chat crashes if the last used chat id saved in UserDefaults doesn't exist. While switching forth and back during the development it happend to me that the chatid was not existent and I needed to comment out the logic in  AppStateRestorers `restoreLastActiveChatId`. For me this looks like a potential call for trouble in production when the user updates DC or recovers DC from a backup. Therefore we should check on start up if a  chat id to restore exists and then try to navigate to the according chat. 

The crash comes from initializer in DcChat: 
```
init(id: Int) {
        if let p = dc_get_chat(DcContext.shared.contextPointer, UInt32(id)) {
            chatPointer = p
        } else {
            fatalError("Invalid chatID opened \(id)")
        }
    }
```

r10s:
closes #630 

we could also let  getChat() return a `DcChat?` - however, that would require lots of changes in unrelated code parts, also in most cases, the object is created by just returned ids, so there would be much overhead for very few corner cases. also, an uninitialized chat-object works, there is no getter or setter that causes problems, on android, it is also used this way.

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -64,7 +64,8 @@ public class DcContext {
     }
 
     public func getChat(chatId: Int) -> DcChat {
-        return DcChat(id: chatId)
+        let chatPointer = dc_get_chat(contextPointer, UInt32(chatId))
+        return DcChat(chatPointer: chatPointer)
     }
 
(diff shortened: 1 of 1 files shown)


r10s:
opps, i forgot to push the second commit

PR #645, opened 2020-04-20
restore chat only if existant, the missing changes

### LINKED ISSUE #630: Gracefully handle non existing chatId

r10s:
successor of #641, closes #630


r10s:
thanks for testing!

PR #652, opened 2020-04-21
ChatView: Scroll to search result 

### LINKED ISSUE #602: scroll chat to search result

nayooti:
**Please review #651 first**.

fixes #602 

This PR is basicly #606 squashed down to one commit. Rebasing old PR was virtually impossible. 

It is based on #651 so I could test it (with success).


r10s:
see comment at https://github.com/deltachat/deltachat-ios/pull/651#issuecomment-617275741

PR #653, opened 2020-04-21
Tweak qr view

### LINKED ISSUE #335: tweak qr-code view
- add a "segmented control" with two items, show+scan, if possible, in the [blue flavor](https://www.ioscreator.com/tutorials/segmented-control-ios-tutorial)
- add option two swipe between the tabs, if possible
- default to show
- remember the last selected segment on re-entering

this is my rough idea:

nayooti:
fixes #335 

`QrPageController` is container for QrView + QrReader and handles all the logic like Coordinator-actions, Progress-Alerts etc. 

I also adjusted `ProgressHandler-protocol? to make it more generic, since `QrPageController` is also using it now.

Currently after reading someone's QrCode we're redirected to the ChatView-List. Is this for purpose? I think we should redirect straight into the related chat.


nayooti:
Not a big fan of having the tabBar visible on the bottom.

IMG_90FC2F0412BD-1

r10s:
> Not a big fan of having the tabBar visible on the bottom.

i do not understand your comment here in this context :) most (all?) messengers i've seen have the tabBar on the bottom.

cyBerta:
> @cyBerta what do you think?

Nice, looks good!

nayooti:
> > Not a big fan of having the tabBar visible on the bottom.
> 
> i do not understand your comment here in this context :) most (all?) messengers i've seen have the _main_ tab-bar on the bottom.

I mean, having the QR-Reader option selected, I prefer not having the TabBar on the bottom.
This is how it look on WhatsApp.
IMG_469EA0D47823-1
On Telegram:
IMG_60A35FBB329C-1

r10s:
> I mean, having the QR-Reader option selected, I prefer not having the TabBar on the bottom.
This is how it look on WhatsApp.

i see.

however, telegram and whatsapp are different here as they do not have a segmented-control with the "show qr invite".

"show qr invite" and "scan qr" should be on the same level, assume the following:

- the segmented-control remembers the last segment
- if the last action was "scan" and you then come back to the qr-code tab later, "scan" is shown again
- if we would then hide the TabBar, there won't be an easy way to go back eg. to "chats"

therefore, the TabBar is needed as it is now.

PR #654, opened 2020-04-21
avoid flashing up the chatlist between qr-scan and enter-name

r10s:
targets https://github.com/deltachat/deltachat-ios/pull/642#pullrequestreview-397728054:

instead of directly creating the tab-view-controller with enter-name opend (which led to flashing the chatlist shortly), go from the welcome-screen to the enter-name-view-controller and when "done" is pressed there, finally go to the tab-view-controller.

@nayooti i tested it and it works, however, i am not that familiar with all these coordinator, can you please check if this is not terribly wrong in the model-logic :)

### DIFF
--- a/deltachat-ios/Controller/ProfileInfoViewController.swift
+++ b/deltachat-ios/Controller/ProfileInfoViewController.swift
@@ -132,7 +132,6 @@ class ProfileInfoViewController: UITableViewController {
 
     @objc private func doneButtonPressed(_ sender: UIBarButtonItem) {
         dcContext.displayname = displayName
-        self.dismiss(animated: false, completion: nil) // not sure if this is needed
         onClose?()
     }
 
--- a/deltachat-ios/Controller/ProfileInfoViewController.swift
+++ b/deltachat-ios/Controller/ProfileInfoViewController.swift
@@ -131,6 +131,10 @@ class ProfileInfoViewController: UITableViewController {
     @objc private func doneButtonPressed(_ sender: UIBarButtonItem) {
         dcContext.displayname = displayName
         self.dismiss(animated: true, completion: nil)
+
+        if let appDelegate = UIApplication.shared.delegate as? AppDelegate {
+            appDelegate.appCoordinator.presentTabBarController()
+        }
     }
 
     private func galleryButtonPressed(_ action: UIAlertAction) {
--- a/deltachat-ios/Controller/ProfileInfoViewController.swift
+++ b/deltachat-ios/Controller/ProfileInfoViewController.swift
@@ -4,6 +4,8 @@ import DcCore
 class ProfileInfoViewController: UITableViewController {
 
     weak var coordinator: EditSettingsCoordinator?
+    var onClose: VoidFunction?
+
     private let dcContext: DcContext
 
     private var displayName: String?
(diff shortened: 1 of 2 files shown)


deltachat-ios/Controller/ProfileInfoViewController.swift
```diff
@@ -131,6 +131,10 @@ class ProfileInfoViewController: UITableViewController {
     @objc private func doneButtonPressed(_ sender: UIBarButtonItem) {
         dcContext.displayname = displayName
         self.dismiss(animated: true, completion: nil)
+
+        if let appDelegate = UIApplication.shared.delegate as? AppDelegate {
```
nayooti: Yep, I would not pull in appDelegate.appCoordinator here. 
I'll provide fix for this.

deltachat-ios/Controller/ProfileInfoViewController.swift
```diff
@@ -132,7 +132,6 @@ class ProfileInfoViewController: UITableViewController {
 
     @objc private func doneButtonPressed(_ sender: UIBarButtonItem) {
         dcContext.displayname = displayName
-        self.dismiss(animated: false, completion: nil) // not sure if this is needed
```
r10s: why is that not needed?

i am pretty sure you are right, but i like to understand that.

is it because of onClose() calls presentTabBarController() that discards the welcomeViewController together with the ProfileInfoViewController indirectly somehow? where does this happen? i also think, we should comment a bit on this stuff, it is not obvious at least to me :)

deltachat-ios/Controller/ProfileInfoViewController.swift
```diff
@@ -4,6 +4,8 @@ import DcCore
 class ProfileInfoViewController: UITableViewController {
 
     weak var coordinator: EditSettingsCoordinator?
+    var onClose: VoidFunction?
```
r10s, line 7: i would prefer a name that does not look like a system-event here, maybe sth. as onProfileInfoDone.

deltachat-ios/Controller/ProfileInfoViewController.swift
nayooti: I tried out, on both versions if ProfileInfoController is de-initialised, and it seems, that either way, the ProfileInfoController stays in memory. So I was right, in the term, that it doesn't make a difference :(. 

I will try to fix this. Have already an idea.

r10s: these things turn out to be complicated. ProfileInfoController stays in memory because it is referenced somehow?

nayooti: I assume we have a retain cycle somewhere. I have to look deeper into this.

nayooti: I found the memory leak, and created a PR (based on master).  I will rebase this PR after review and go from there.

deltachat-ios/Controller/ProfileInfoViewController.swift
r10s: no need to block this pr, though, just for the future, it is better to be a bit more precise.

PR #656, opened 2020-04-23
avoid flashing up the welcome-screen between login-view and chatlist

r10s:
there was a short, annoying flashing of the welcome-screen (that was long left...) and the chatlist that is shown on login success.

this pr fixes that by removing a one-second-timeout and also by removing a probably superfluous dismiss (probably similar to the removing at https://github.com/deltachat/deltachat-ios/pull/654/commits/59aafeafafbed7787a30c20236d44ec9026002d9 )

the two account-created-functions handleLoginSuccess() and handleProfileInfoClosed() are exactly the same now, which looks quite correct to me :)

### DIFF
--- a/deltachat-ios/Coordinator/AppCoordinator.swift
+++ b/deltachat-ios/Coordinator/AppCoordinator.swift
@@ -178,21 +178,17 @@ extension AppCoordinator: WelcomeCoordinator {
         welcomeController?.present(loginController, animated: true, completion: nil)
     }
 
-    func handleLoginSuccess() {
-        presentTabBarController()
-    }
-
     func handleQRAccountCreationSuccess() {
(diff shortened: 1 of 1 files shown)


deltachat-ios/Coordinator/AppCoordinator.swift
```diff
@@ -179,11 +179,7 @@ extension AppCoordinator: WelcomeCoordinator {
     }
 
     func handleLoginSuccess() {
-        DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
```
r10s, line 182: i have no idea why this was needed at all.

nayooti:
since `handleLoginSuccess()` and `handleProfileInfoClosed()` do the same thing I suggest to reduce it to one method.

PR #663, opened 2020-04-30
set voiceover text on chat avatar

### LINKED ISSUE #629: Missing AccessibilityLabel for Info button in a chat view
- Device: iPhone or iPad
- iOS version: 13.4
- Delta Chat version: 1.3 app store or 1.5 TestFlight
- Expected behavior: When opening a chat window, like the greeting message, and exploring the items with VoiceOver, the Info button should be labeled with "Information" or "Contact details" or something appropriate.
- Actual behavior: When reaching the button with VoiceOver, VoiceOver says "button", sometimes followed by "May be Information".
- Steps to reproduce the problem: Install Delta Chat, start VoiceOver via Settings/Accessibility/VoiceOver, or by configuring the Accessibility Shortcut to launch and shut down VoiceOver, open a chat and swipe left and right through the controls.
- Screenshots: N/A
- Logs: Not applicable, no crash, just missing information (AccessibilityLabel on the info button).

nayooti:
fixes #629

There was no accessibilityLabel set when using an image as avatar.
Fixed. The text is pure optional and can be omitted (so it does not affect cell-avatars).

### DIFF
--- a/deltachat-ios/Controller/ChatViewController.swift
+++ b/deltachat-ios/Controller/ChatViewController.swift
@@ -274,9 +274,9 @@ class ChatViewController: MessagesViewController {
 
         let badge: InitialsBadge
         if let image = chat.profileImage {
-            badge =  InitialsBadge(image: image, size: 28, accessibilityLabelText: chat.name)
+            badge =  InitialsBadge(image: image, size: 28, accessibilityLabel: String.localized("menu_view_profile"))
         } else {
-            badge =  InitialsBadge(name: chat.name, color: chat.color, size: 28)
+            badge =  InitialsBadge(name: chat.name, color: chat.color, size: 28, accessibilityLabel: String.localized("menu_view_profile"))
             badge.setLabelFont(UIFont.systemFont(ofSize: 14))
         }
         badge.setVerified(chat.isVerified)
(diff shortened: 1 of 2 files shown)


deltachat-ios/Controller/ChatViewController.swift
```diff
@@ -274,7 +274,7 @@ class ChatViewController: MessagesViewController {
 
         let badge: InitialsBadge
         if let image = chat.profileImage {
-            badge =  InitialsBadge(image: image, size: 28)
+            badge =  InitialsBadge(image: image, size: 28, accessibilityLabelText: chat.name)
```
r10s, line 277: this is the avatar in the title bar above an opened chat?

a tap on it opens the profile - so maybe better just use `menu_view_profile` / "View profile"?

deltachat-ios/View/InitialsBadge.swift
```diff
@@ -90,11 +90,14 @@ class InitialsBadge: UIView {
...
     }
 
-    func setImage(_ image: UIImage) {
+    func setImage(_ image: UIImage, _ accessibilityLabelText: String? = nil) {
         self.imageView.image = image
         self.imageView.contentMode = UIView.ContentMode.scaleAspectFill
         self.imageView.isHidden = false
         self.label.isHidden = true
+        if let text = accessibilityLabelText {
+            accessibilityLabel = "avatar \(text)"
```
r10s, line 99: if this function is only called by https://github.com/deltachat/deltachat-ios/pull/663/files#diff-e140c04537dd6b083c550047ebd58506R277 i would not prepend "avatar" here

PR #670, opened 2020-05-04
clearer error message on login error if no internet connection available

### LINKED ISSUE #503: Login when offline gives imap error
The dialog should rather tell the user to first connect to wifi or mobile data connection.

cyBerta:
closes #503

it also includes a fix for a potential bug where changing network connectivity is not recognized while the app is in the background leading to a wrong state when reentering the foreground.

The framework we're using has 2 event handling options: using a callback or notifications and I switched to the callback option while fixing the mentioned bug.

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -22,7 +22,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     var locationManager: LocationManager!
     private var backgroundTask: UIBackgroundTaskIdentifier = .invalid
     var reachability = Reachability()!
-    var hasNetwork = false
     var window: UIWindow?
     var state = ApplicationState.stopped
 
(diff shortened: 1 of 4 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -25,6 +25,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     private var backgroundTask: UIBackgroundTaskIdentifier = .invalid
 
     var reachability = Reachability()!
+    var hasNetwork = false
```
r10s, line 28: why can't we use appDelegate.reachability.connection directly? instead of caching the state on our own. at least, this seems to compile.

cyBerta: I fixed that and found also another bug in the already existing code which is fixed by b6f3631 as well

deltachat-ios/AppDelegate.swift
```diff
@@ -73,6 +73,28 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         UIApplication.shared.setMinimumBackgroundFetchInterval(UIApplication.backgroundFetchIntervalMinimum)
         start()
         setStockTranslations()
+
+        reachability.whenReachable = { reachability in
+            logger.info("could not start reachability notifier")
```
r10s, line 78: this message does not make sense to me

cyBerta: neither to me :P. I'll fix that

cyBerta:
> does reachabilityChanged call when(Un)Reachable by default? etc.

the callbacks will be only called after `startNotifier()` was successfully called. rechability has already the right state when initializing it

PR #673, opened 2020-05-05
improve accessibility of voice message recording #493

### LINKED ISSUE #493: VoiceOver silences when voice message button is pressed
In a chat window, VoiceOver silences after voice message button is pressed. I don't know why, but it does not happens in the WhatsApp and Skype.

cyBerta:
fixes #493

### DIFF
--- a/deltachat-ios/Controller/AudioRecorderController.swift
+++ b/deltachat-ios/Controller/AudioRecorderController.swift
@@ -87,6 +87,8 @@ class AudioRecorderController: UIViewController, AVAudioRecorderDelegate {
                                            target: self,
                                            action: #selector(recordingButtonAction))
         button.tintColor = UIColor.themeColor(light: .darkGray, dark: .lightGray)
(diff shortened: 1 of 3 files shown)


cyBerta:
> if this is possible, maybe we can move the "double tap again to send recorded message" to the attach-menu?

Can you please explain that a little bit more? Do you mean sth. like hold-to-record - as it is done in Android?

cyBerta:
> i just played around with whatsapp: they just do not stop voice-over on recording, wondering, if that wouldn't be a better fix for us.

I can try to figure out that

cyBerta:
> however, there is no way to cancel the sending, or?
> (the cancel button is there, but you cannot navigate to it by voice-over)

that's correct. It's more like hold-to-record -> if you started once you cannot cancel sending it anymore.

r10s:
i'll do an alternative pr that describes what i mean.

cyBerta:
closing in favor of #690

PR #676, opened 2020-05-05
fix garbage collection issue leading to inability to add attachments 

### LINKED ISSUE #675: Addition of attachments not working
Steps to reproduce:

1. open Delta Chat
2. navigate to a chat
3. kill the app
4. restart the app -> the last opened chat appears
5. try to add any attachment -> corresponding view controllers don't open

This issue is related to the weak coordinators in the view controllers. The coordinator is null in the described scenario und thus doesn't push a new viewcontroller onto the stack.

cyBerta:
fixes #675

### DIFF
--- a/deltachat-ios/Coordinator/AppCoordinator.swift
+++ b/deltachat-ios/Coordinator/AppCoordinator.swift
@@ -120,10 +120,10 @@ class AppCoordinator: NSObject, Coordinator {
             assertionFailure("huh? why no nav controller?")
             return
         }
-        let chatVC = ChatViewController(dcContext: dcContext, chatId: chatId)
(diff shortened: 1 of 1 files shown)


r10s:
@cyBerta so without this fix, the attach-dialog, voice-message dialogs are not working, or?

asking, because somtimes, i recognized the same with the profile-view - also that does sometime not open - exiting back to the chatlist and re-entering the chat, however, opens it then.

is this also fixed by this pr? if not, can you do a similar fix, please?

cyBerta:
It is also fixed with this pr

r10s:
yeah! great! this is also one of the (few :) issues i had from time to time but i could not really reproduce

Issue #660, opened 2020-04-30
qr-view nitpicking


r10s:
great progress on the qr-view, some nitpicking from my side:

- although i stated things differently in the past, i think it is worth remembering the selected section (show or scan) between program starts, so it would be great if it can be persisted in addition to the main selection (qr, chats, settings)  
the reason for that is that "scan" is typically done more often than "show", but, anyway, the default should stay unchanged as "show"

- i would add some padding between the hint "Hold your camera over the QR code" and the lower screen border

- can we have the qr code and the hint on the show-section centered vertically, if room? if there is no room, of course, scrolling is required, esp. as the hint may get longer, depending on the locale. @nayooti iirc you figured out a similar smart solution for the welcome sceen :)

for the first point, however, i would be a bit careful when the tab is stored - there may be situations where the camera cannot be accessed - and this may easily turn delta chat into becoming unusable (easily reproducible in the emulator). if there is no easy way to check this, i would better leave things as they are now.

r10s:
k, we can close this as we have more concrete issues now.

Issue #679, opened 2020-05-06
QrTab nitpicking #1 - persist last selected segment


nayooti:
This is the first part of #660:

although i stated things differently in the past, i think it is worth remembering the selected section (show or scan) between program starts, so it would be great if it can be persisted in addition to the main selection (qr, chats, settings)
the reason for that is that "scan" is typically done more often than "show", but, anyway, the default should stay unchanged as "show"

I had two failed attempts implementing this, therefor a separate issue.

r10s:
> I had two failed attempts implementing this, therefor a separate issue.

it is not that important, let's give that up for now. maybe it is even better to have the qr-view by default, we can do this later, if really needed.

PR #687, opened 2020-05-06
Qr tab ui enhancement

### LINKED ISSUE #680: QRTab nitpicking #2 - QrCode in vertically centered etc ...
UI-part of #660

- i would add some padding between the hint "Hold your camera over the QR code" and the lower screen border

- can we have the qr code and the hint on the show-section centered vertically, if room? if there is no room, of course, scrolling is required, esp. as the hint may get longer, depending on the locale. @nayooti iirc you figured out a similar smart solution for the welcome sceen :)

nayooti:
fixes #680 

This is the UI-part of #660.

UI elements are vertically centered. On small devices the content is scrollable (can be triggered on SE landscape). Very same principle as WelcomeViewController.

### DIFF
--- a/deltachat-ios/Controller/QrPageController.swift
+++ b/deltachat-ios/Controller/QrPageController.swift
@@ -31,12 +31,6 @@ class QrPageController: UIPageViewController, ProgressAlertHandler {
     // MARK: - lifecycle
     override func viewDidLoad() {
         super.viewDidLoad()
-        
-        // temp fix - ToDo: check if this is needed for >iOS13 devices
-        if #available(iOS 13, *) {}
-        else {
-            self.view.backgroundColor = .white
-        }
         dataSource = self
         delegate = self
         navigationItem.titleView = qrSegmentControl
--- a/deltachat-ios/Controller/QrPageController.swift
+++ b/deltachat-ios/Controller/QrPageController.swift
@@ -31,9 +31,10 @@ class QrPageController: UIPageViewController, ProgressAlertHandler {
     // MARK: - lifecycle
     override func viewDidLoad() {
         super.viewDidLoad()
+        
+        // temp fix - ToDo: check if this is needed for >iOS13 devices
         if #available(iOS 13, *) {}
         else {
-            // temp fix - ToDo: check if this is needed for >iOS13 devices 
             self.view.backgroundColor = .white
         }
         dataSource = self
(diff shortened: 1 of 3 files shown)


deltachat-ios/Controller/QrPageController.swift
```diff
@@ -33,6 +33,12 @@ class QrPageController: UIPageViewController, ProgressAlertHandler {
     // MARK: - lifecycle
     override func viewDidLoad() {
         super.viewDidLoad()
+        
+        // temp fix - ToDo: check if this is needed for >iOS13 devices
+        if #available(iOS 13, *) {}
```
cyBerta, line 38: lint is complaining here. can you put the else into the same line?

deltachat-ios/Controller/QrViewController.swift
```diff
@@ -23,93 +36,144 @@ class QrViewController: UITableViewController {
...
         label.numberOfLines = 0
         label.textAlignment = .center
         label.font = UIFont.systemFont(ofSize: 14)
         return label
+    }()
+
+    private let container = UIView()
+
+    var minContainerHeight: CGFloat = 0 {
+        didSet {
```
cyBerta, line 110: cool, I didn't know that this is possible

cyBerta:
I think this part 

> although i stated things differently in the past, i think it is worth remembering the selected section (show or scan) between program starts, so it would be great if it can be persisted in addition to the main selection (qr, chats, settings)

is not implemented yet. Probably on purpose to have a fallback for not working cameras?

nayooti:
> I think this part
> 
> > although i stated things differently in the past, i think it is worth remembering the selected section (show or scan) between program starts, so it would be great if it can be persisted in addition to the main selection (qr, chats, settings)
> 
> is not implemented yet. Probably on purpose to have a fallback for not working cameras?

As said. I split up #660 into #680 + #679  because somehow I had two failed attempts in implementing #679. I will have another try in #679 after #696 has been merged. I also will take care on #688 later.

deltachat-ios/Controller/QrPageController.swift
nayooti: The code block became obsolete with lastest master

nayooti:
I removed a temp fix, rebased and forced push.

cyBerta:
lgtm :) yeah

PR #692, opened 2020-05-07
bugfix ios 12 - deallocation bug 

### LINKED ISSUE #686: iOS 12.4 - AccountSetupController behind UITabBarController
I realised this when I implemented #679 and tested on a iOS 12.4 device on simulator. On fresh install I logged in manually via ProfileInfo. After I logged in successfully and the app redirected me to the chats tab as normal. But then when I switched to the QR Tab I got this: 

screenshot

I assume that ScrollView has no background color in iOS 12 and this way the underlying viewController shines through. But obviously there should be nothing behind the TabBarViewController. 

I did some analysing.
Here is what you get when you inspect the view layers:

image

It seems that behind the UITabBarController is still some AccountSetupController somehow.

I haven't looked into this too deeply, but it seems that under iOS 13 the issue does not occur.

nayooti:
fixes #686 

This bug **only happened for iOS12** or older devices. 

I realised that after logging in  manually via `WelcomeController` the `AccountSetupController`  was still behind or UIWindow, which is really strange. 
The first thing I did, I made the `AccountSetupController`-instance **optional** in `AppCoordinator`, and made sure it was set to nil after the user logged in.

Then things became even more weird:
Instead of the `AccountSetupController` there was now the `WelcomeContentView` behind `UIWindow`. `WelcomeViewController` was clearly deallocated, and there is no object in our code, that actually references `WelcomeContentView` except for `WelcomeViewController`. The fact that `WelcomeViewController` was clearly de-allocating implies also that there is no retain cycle.
Also on iOS13 this issue did not exist at all.

The solution now I found that worked:
I explicitly remove `WelcomeContentView` from superview during `WelcomeController.deinit` (on iOS 12).

To test:
To reproduce the issue **on Master**:
- use a simulator with iOS 12. Make sure delta chat is not installed. (Delete every time between tests). 
- on WelcomeScreen -> Login to your Server -> Login with credentials -> After redirect to TabBar -> Go to Qr-tab -> select Scan QR-Code -> You see AccountSetupController shine through

### DIFF
--- a/deltachat-ios/Controller/WelcomeViewController.swift
+++ b/deltachat-ios/Controller/WelcomeViewController.swift
@@ -43,6 +43,16 @@ class WelcomeViewController: UIViewController, ProgressAlertHandler {
         fatalError("init(coder:) has not been implemented")
     }
 
+    deinit {
+        if #available(iOS 13, *) {}
+        else {
+            /*
+             for pre-iOS 13 versions there was an issue that welcomeView was not deallocated after login - reason unknown
+             */
+            welcomeView.removeFromSuperview()
+        }
+    }
+
     // MARK: - lifecycle
     override func viewDidLoad() {
         super.viewDidLoad()
--- a/deltachat-ios/Coordinator/AppCoordinator.swift
+++ b/deltachat-ios/Coordinator/AppCoordinator.swift
@@ -26,22 +26,7 @@ class AppCoordinator: NSObject, Coordinator {
         return tabBarController
     }()
 
-    private lazy var loginController: UIViewController = {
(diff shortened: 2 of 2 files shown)


r10s:
hm, can someone test if this issue is real _outside_ the emulator?

is is really a viewController - or just some bits from a randomly shown offscreen bitmap?

at least for me, also for iOS 13 in the emulator the qr-scanner does not work at all.

r10s:
i just tested this, this seems to be fixed by #696 , the "shine through" effect is gone. but would be interested if scanning as such works.

PR #694, opened 2020-05-08
Hint if blocked contacts list is empty

### LINKED ISSUE #661: empty blocked-contacts list should show some hints
the empty blocked-contacts list should show a hint.

desktop already uses the string `none_blocked_desktop` - we should just also use this this.

(once done, i'd like to reword the hint to "If you block contacts, they are shown here" or so - but that is independent from ios then)

nayooti:
fixes #661

### DIFF
--- a/deltachat-ios/Controller/GroupMembersViewController.swift
+++ b/deltachat-ios/Controller/GroupMembersViewController.swift
@@ -228,13 +228,13 @@ class AddGroupMembersViewController: GroupMembersViewController {
 
 class BlockedContactsViewController: GroupMembersViewController, GroupMemberSelectionDelegate {
 
-    private var emptyStateLabel: UILabel = {
-        let label = UILabel()
-        label.text = String.localized("none_blocked_desktop")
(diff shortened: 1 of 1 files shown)


cyBerta:
For more consistency, it might make sense to have a similar layouts for the empty contact request list as well as for blocked contacts? Designwise I prefer the minimal design of the blocked contacts (so would rather tend to change the layout of the empty contact requests), otoh the contact requests layout is more similar to the android design. 

 @r10s do you have a opinion about that? 
Besides that the code looks good to me and can be merged.

Screen Shot 2020-05-11 at 16 22 11

Screen Shot 2020-05-11 at 16 24 06

nayooti:
Label needs some additional leading+trailing constraints.

r10s:
if we can easily can get the same layout for both, contact-request and for blocked-contacts, that would be nice.

can we resule the view used for the contact-requests accessible?

cyBerta:
The PaddingLabel needs to be replaced by sth. smarter. But this is out of scope.

PR #695, opened 2020-05-08
Simplify edit contact

r10s:


deltachat-ios/Coordinator/AppCoordinator.swift
```diff
@@ -861,32 +849,6 @@ class EditGroupCoordinator: Coordinator {
...
-
-    func navigateBack() {
-        navigationController.popViewController(animated: true)
-    }
-
-    func showChat(chatId: Int) {
-        let chatViewController = ChatViewController(dcContext: dcContext, chatId: chatId)
-        let coordinator = ChatViewCoordinator(dcContext: dcContext, navigationController: navigationController, chatId: chatId)
-        coordinator.chatViewController = chatViewController
-        childCoordinators.append(coordinator)
```
r10s, line 884: i think, this was wrong anyway, why append a coordinator that is just discarded to the list of childCoordinators?

deltachat-ios/Controller/NewContactController.swift
```diff
@@ -108,6 +107,26 @@ class NewContactController: UITableViewController {
...
+    func navigateBack() {
+        if let navigationController = self.parent as? UINavigationController {
+            navigationController.popViewController(animated: true)
+        }
+    }
+
+    func showChat(chatId: Int) {
+        let chatViewController = ChatViewController(dcContext: dcContext, chatId: chatId)
+        if let navigationController = self.parent as? UINavigationController {
+            let coordinator = ChatViewCoordinator(dcContext: dcContext, navigationController: navigationController, chatId: chatId)
```
r10s, line 123: these three lines would go away when continue removing the coordinators.

deltachat-ios/Controller/NewContactController.swift
```diff
@@ -108,6 +107,26 @@ class NewContactController: UITableViewController {
         super.didReceiveMemoryWarning()
         // Dispose of any resources that can be recreated.
     }
+
+
+    // MARK: - coordinator
```
r10s, line 112: we could move stuff from the coordinators to such a `// MARK: -` - this way, we can easily create coordinators as needed - if really needed.

deltachat-ios/Coordinator/AppCoordinator.swift
```diff
@@ -768,9 +759,6 @@ class ContactDetailCoordinator: Coordinator, ContactDetailCoordinatorProtocol {
 
     func showEditContact(contactId: Int) {
         let editContactController = EditContactController(dcContext: dcContext, contactIdForUpdate: contactId)
-        let coordinator = EditContactCoordinator(dcContext: dcContext, navigationController: navigationController)
-        childCoordinators.append(coordinator)
-        editContactController.coordinator = coordinator
```
r10s, line 773: 4 times the same code duplicated could be removed - one could treat this as a hint, we do not really need this pattern currently.

PR #701, opened 2020-05-13
replace potentially harmful 'unowned' by 'weak'

### LINKED ISSUE #700: stop using unowned
`unowned` references are similar to `weak` optional-references and allow values to become nil at any point in the future.

but unlike `weak`, the `unowned` references are forced-unwrapped at the point where the optional is accessed.

if the accessed optional is nil at that point, the app crashed.

defining a reference as being `unowned`, we give a guarantee that the reference can never become nil during the lifetime it is referenced.

this can be guaranteed by a strong pointer, however, it also can be done easily done wrong. and you have to think over it all the time.

- eg. https://www.hackingwithswift.com/articles/179/capture-lists-in-swift-whats-the-difference-between-weak-strong-and-unowned-references says "You should use unowned very carefully indeed."  
- i grep'd the signal source code, and there is not a single unowned reference.

i am not saying, that we are actually doing things wrong, but we have a few crashes here and there, and a general precaution, we should avoid `unowned` as we avoid `as!` and replace existing `unwrap` by `weak` - which is doable in an hour or so, i just tried it. swift makes it pretty simple to deal with weak pointers.

not using `unwrap` eliminates one source of possible troubles.

wrt losing references, this change does not make a difference as unowned references are handled more or less the same by [arc](https://docs.swift.org/swift-book/LanguageGuide/AutomaticReferenceCounting.html).

r10s:
closes #700

### DIFF
--- a/deltachat-ios/Controller/GroupChatDetailViewController.swift
+++ b/deltachat-ios/Controller/GroupChatDetailViewController.swift
@@ -401,21 +401,20 @@ extension GroupChatDetailViewController: UITableViewDelegate, UITableViewDataSou
             !isMemberManagementRow(row: row) &&
             getGroupMemberIdFor(row) != currentUser.id {
             // action set for members except for current user
-            let delete = UITableViewRowAction(style: .destructive, title: String.localized("remove_desktop")) { [unowned self] _, indexPath in
-
-                let contact = self.getGroupMember(at: row)
+            let delete = UITableViewRowAction(style: .destructive, title: String.localized("remove_desktop")) { [weak self] _, indexPath in
+                guard let strongSelf = self else { return }
+                let contact = strongSelf.getGroupMember(at: row)
                 let title = String.localizedStringWithFormat(String.localized("ask_remove_members"), contact.nameNAddr)
                 let alert = UIAlertController(title: title, message: nil, preferredStyle: .safeActionSheet)
                 alert.addAction(UIAlertAction(title: String.localized("remove_desktop"), style: .destructive, handler: { _ in
-                    let success = self.dcContext.removeContactFromChat(chatId: self.chat.id, contactId: contact.id)
+                    let success = strongSelf.dcContext.removeContactFromChat(chatId: strongSelf.chat.id, contactId: contact.id)
                     if success {
-                        self.removeGroupMemberFromTableAt(indexPath)
+                        strongSelf.removeGroupMemberFromTableAt(indexPath)
                     }
                 }))
                 alert.addAction(UIAlertAction(title: String.localized("cancel"), style: .cancel, handler: nil))
-                self.present(alert, animated: true, completion: nil)
-
- }
+                strongSelf.present(alert, animated: true, completion: nil)
+            }
             delete.backgroundColor = UIColor.red
             return [delete]
         }
--- a/deltachat-ios/Controller/GroupChatDetailViewController.swift
+++ b/deltachat-ios/Controller/GroupChatDetailViewController.swift
@@ -402,18 +402,18 @@ extension GroupChatDetailViewController: UITableViewDelegate, UITableViewDataSou
             getGroupMemberIdFor(row) != currentUser.id {
             // action set for members except for current user
             let delete = UITableViewRowAction(style: .destructive, title: String.localized("remove_desktop")) { [weak self] _, indexPath in
-                guard let strongSelf = self else { return }
-                let contact = strongSelf.getGroupMember(at: row)
+                guard let self = self else { return }
+                let contact = self.getGroupMember(at: row)
                 let title = String.localizedStringWithFormat(String.localized("ask_remove_members"), contact.nameNAddr)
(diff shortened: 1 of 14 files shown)


deltachat-ios/Controller/GroupChatDetailViewController.swift
```diff
@@ -401,21 +401,20 @@ extension GroupChatDetailViewController: UITableViewDelegate, UITableViewDataSou
             !isMemberManagementRow(row: row) &&
             getGroupMemberIdFor(row) != currentUser.id {
             // action set for members except for current user
-            let delete = UITableViewRowAction(style: .destructive, title: String.localized("remove_desktop")) { [unowned self] _, indexPath in
-
-                let contact = self.getGroupMember(at: row)
+            let delete = UITableViewRowAction(style: .destructive, title: String.localized("remove_desktop")) { [weak self] _, indexPath in
+                guard let strongSelf = self else { return }
```
cyBerta, line 405: In swift 4.2 the syntax `guard let self = self else { return } ` was introduced to check if self is nil or not

cyBerta: personally I like that more than a `strongSelf` variable

deltachat-ios/Controller/NewChatViewController.swift
```diff
@@ -228,19 +228,21 @@ class NewChatViewController: UITableViewController {
         if indexPath.section == sectionContacts {
             let contactId = contactIdByRow(indexPath.row)
 
-            let edit = UITableViewRowAction(style: .normal, title: String.localized("info")) { [unowned self] _, _ in
-                if self.searchController.isActive {
-                    self.searchController.dismiss(animated: false) {
-                        self.showContactDetail(contactId: contactId)
+            let edit = UITableViewRowAction(style: .normal, title: String.localized("info")) { [weak self] _, _ in
+                guard let strongSelf = self else { return }
```
cyBerta, line 232: same here

deltachat-ios/Controller/WelcomeViewController.swift
```diff
@@ -14,21 +14,22 @@ class WelcomeViewController: UIViewController, ProgressAlertHandler {
...
-            self.navigationController?.pushViewController(accountSetupController, animated: true)
+            strongSelf.navigationController?.pushViewController(accountSetupController, animated: true)
         }
-        view.onScanQRCode  = { [unowned self] in
+        view.onScanQRCode  = { [weak self] in
             let qrReader = QrCodeReaderController()
             qrReader.delegate = self
-            self.qrCodeReader = qrReader
-            self.navigationController?.pushViewController(qrReader, animated: true)
+            self?.qrCodeReader = qrReader
```
cyBerta, line 31: we can use `guard let self = self else { return }` as well. it is very unlikely that between the first and the second `self?` the reference disappears but still it might make sense to have a common pattern for cases where more than once `self?` is called

r10s: but even when it disappears, things would be fine. in both cases the QrCodeReaderController would not be pushed.

most usages of [weak self] are one-liners, and i would vote for `self?` where possible to keep things short and readable. this is also the way signal uses (also the stringSelf comes from there)

deltachat-ios/Controller/GroupChatDetailViewController.swift
r10s: copied the strongSelf pattern from signal, did not know `guard let self = self else { return }` is possible - but in, fact, we were already using it at two places :)

yip, i will do a commit that changes that.

ftr, this is the discussion about it https://github.com/apple/swift-evolution/blob/master/proposals/0079-upgrade-self-from-weak-to-strong.md

deltachat-ios/Controller/WelcomeViewController.swift
cyBerta: maybe exact this occourence was not the best example. but there are others in the PR where popping a view controller or calling a callback would not happen while other logic was already happening.

cyBerta: I also wouldn't do it for oneliners

r10s: i agree, it is better to use the guard if self? would be used multiple times, this avoid thinking about what will happen if self disappears in between, you're right.

deltachat-ios/Handler/DeviceContactsHandler.swift
```diff
@@ -85,15 +85,15 @@ class DeviceContactsHandler {
         case .denied:
             contactListDelegate?.accessDenied()
         case .restricted, .notDetermined:
-            store.requestAccess(for: .contacts) { [unowned self] granted, _ in
+            store.requestAccess(for: .contacts) { [weak self] granted, _ in
                 if granted {
                     DispatchQueue.main.async {
-                        self.addContactsToCore()
-                        self.contactListDelegate?.accessGranted()
+                        self?.addContactsToCore()
```
cyBerta, line 91: here it's still missing

PR #713, opened 2020-05-14
Wrapped PaddingView into FlexView

### LINKED ISSUE #702: Throw away PaddingLabel
PaddingLabel is a rather poor attempt to build a multiline label conaining padding and a background with rounded corners. 
The main weakness is that it doesn't adapt it's width correctly to the actual width of the rendered text. Instead it uses a fixed width including a padding and adapts the height so that all multiline text is visible. 
What would be nice is to have a UI element that fulfits the folliwing
- includes left, right, top, bottom padding
- rounded corners
- multiline text
- having a max width
- if the rendered text block is narrower than the max width, adapt the width to the actual size of the rendered text

nayooti:
fixes #702 

I created a `FlexLabel` class that contains an UILabel (with padding). The idea is to setup the FlexView with constraints. It should have a leading and trailing or a width constraint. 

To display the text the label will horizontally grow within its available horizontal space and line wraps if needed. It is also able to increase the FlexLabel's total height (if constraints allow it).

I then created `EmptyStateLabel` as a subclass of `FlexLabel` with styling like rounded corners, background colors etc.


deltachat-ios/View/FlexLabel.swift
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    private func setupSubviews() {
+        addSubview(label)
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.centerYAnchor.constraint(equalTo: centerYAnchor, constant: 0).isActive = true
+        label.centerXAnchor.constraint(equalTo: centerXAnchor, constant: 0).isActive = true
+        label.topAnchor.constraint(equalTo: layoutMarginsGuide.topAnchor).isActive = true
+        label.bottomAnchor.constraint(equalTo: layoutMarginsGuide.bottomAnchor).isActive = true
+        label.widthAnchor.constraint(lessThanOrEqualTo: widthAnchor, multiplier: 0.95).isActive = true
```
cyBerta, line 56: Why is the multiplier needed?

nayooti: it adds some "padding" effect. probably better to remove the multiplier. 

I will check if it has some effect on the label insets/ corner radius.

nayooti:
I just tested: I indeed added that multiplier for a reason. 

It is needed because of the padding in PaddingLabel. Without it the label get's truncated and please don't ask me why. It's a quite common issue with labels with additional horizontal insets.

With multiplier:

image

Without:
image

PR #718, opened 2020-05-15
Remove hud handler

### LINKED ISSUE #697: backup progress alert is hidden
Currently the progress alert is 'hidden' when a backup has been triggered. It is attached to the top of the settings views table list so that you need to scroll up to see the progress alert. 
Instead the progress alert should be always visible and not be able to scrolled away

nayooti:
fixes #697 

As suggested in #697 I removed HudHandler from the project. 

SettingsController now uses ProgressAlertHandler, which is also used for login-process in WelcomeViewController/AccountSetupController.

I would like to tweak ProgressAlertHandler in a continuing issue/PR.


cyBerta:
I think  the pod JGProgressHUD can be removed as well now

r10s:
yeah, removing code and dependencies is fun!

nayooti:
> I think the pod JGProgressHUD can be removed as well now

done.

PR #724, opened 2020-05-18
Bugfix - qr reader no longer induces crashes on simulators

### LINKED ISSUE #688: On Simulator - app crashes when toggling between qr segments
To reproduce:

On simulator -> qr Tab -> select Qr-Code scan segment -> back to Qr-invitation segment -> back to Qr-Code scan segment -> App freezes + crashs.

I think this has something to do that on Simulators there is no camera access.

nayooti:
fixes #688 

Easiest solution was just to not start the video capture session on simulators.

### DIFF
--- a/deltachat-ios/Controller/QrCodeReaderController.swift
+++ b/deltachat-ios/Controller/QrCodeReaderController.swift
@@ -13,13 +13,6 @@ class QrCodeReaderController: UIViewController {
         return videoPreviewLayer
     }()
 
-    private lazy var cameraAccessButton: UIButton = {
-        let button = UIButton()
-        button.setTitle("Settings", for: .normal)
-        button.addTarget(self, action: #selector(cameraAccessButtonPressed(_:)), for: .touchUpInside)
-        return button
-    }()
-
     private var infoLabel: UILabel = {
         let label = UILabel()
            label.translatesAutoresizingMaskIntoConstraints = false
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/QrCodeReaderController.swift
```diff
@@ -95,32 +95,37 @@ class QrCodeReaderController: UIViewController {
...
 
     // MARK: - actions
     func startSession() {
+        #if targetEnvironment(simulator)
+        // ignore if run from simulator
+        #else
         captureSession.startRunning()
-    }
+        #endif
+        }
```
r10s, line 120: i think this curly bracket is not corrected indented.

deltachat-ios/Controller/QrCodeReaderController.swift
```diff
@@ -95,32 +95,37 @@ class QrCodeReaderController: UIViewController {
...
 
     func stopSession() {
         captureSession.stopRunning()
     }
+
+    @objc private func cameraAccessButtonPressed(_ sender: UIButton) {
+        print("Camera button pressed")
+    }
```
r10s, line 128: why is that overided now? how is that related to simulator or not?

nayooti: forgot to delete.

PR #726, opened 2020-05-18
Search: show empty state banner

### LINKED ISSUE #720: message 'nothing found' missing in search
when a search is started, and nothing is found, there should be a hint, that nothing was found. 

otherwise, the user does not know if the search is still going on or if nothing was found.

currently, just a white screen is shown.

we have the string `search_no_result_for_x` for that purpose.

i suggest to do that similar to android.

nayooti:
fixes #720 

First of all:
I had issues with the `"No results found for \"%s\""`-String (runtime error). 
I changed the english version to `"No results found for \"%1$@\""` and this now works. But I did not change all the other strings because I didn't know if this is correct.

So to test - you need to set the device language to English.

The empty state banner will be shown in:

- Chat list
- NewChatController
- GroupMemberViewController


deltachat-ios/en.lproj/Localizable.strings
```diff
@@ -240,8 +240,7 @@
 // search
 "search" = "Search";
 "search_explain" = "Search for chats, contacts, and messages";
-"search_no_result_for_x" = "No results found for \"%s\"";
-
+"search_no_result_for_x" = "No results found for \"%1$@\"";
```
r10s, line 243: we cannot change this here, the string will arrive as %s again, esp. in translations. 
but i think this can be handled by ./tools/tx-pull-translations.sh

cyBerta: The issue is that the android tranlsation (strings.xml) uses a wrong placeholder. It should be `%1$s` instead of `%s` and therefore the regex doesn't recognizes the placeholder in order to replace it by the iOS conform `%1$@`. I will create a pull request in the Android repository for it.

cyBerta: I think these kind of edge cases shouldn't be covered by ./tools/tx-pull-translations.sh. I would prefer to have a valid input file as a precondition for the translation process.

AndreazLattmann:
When will it be available on transifex for translation?

r10s:
> When will it be available on transifex for translation?

not sure if i understand the question :)

however, ios is already available for translations on transifex since the very beginning - ios uses the same strings as android and the same strings as desktop - so all 5 systems are translated at https://www.transifex.com/delta-chat/public/

deltachat-ios/en.lproj/Localizable.strings
r10s: i disagree here. most input files come from transifex and maybe other sources, we cannot guarantee "clean input" here.

for robustness, we should adapt the regex and accept %s as an equivalent to %1$s - even if this is not 100% correct - it is the format used in printf() and easily done wrong therefore. and, of course, we should not crash.

be strict what you write and accepting what you read :)

nayooti:
I undid  the commit where I edited the translations.

cyBerta:
I will add the changes for #743 here, if that's ok. This way, we make sure we don't introduce a new crash with merging this before adapting the transifex pull script.

cyBerta:
Screen Shot 2020-06-02 at 17 50 00

I added the changes to the pull translations script. The UI works like intended.

deltachat-ios/en.lproj/Localizable.strings
cyBerta: done

PR #729, opened 2020-05-19
Fix image message layout

### LINKED ISSUE #685: images are too large in landscape mode
when using the device in landscape mode, landscape-images are shown too large, they might be even larger than the screen.

this happens a lot on ipad-usage - but also on iphone in landscape.

i suggest, to use the calculated-portrait-image sizes for both, landscape and portrait mode.

cyBerta:
closes #685

### DIFF
--- a/deltachat-ios/MessageKit/Layout/MediaMessageSizeCalculator.swift
+++ b/deltachat-ios/MessageKit/Layout/MediaMessageSizeCalculator.swift
@@ -27,15 +27,28 @@ import UIKit
 
 open class MediaMessageSizeCalculator: MessageSizeCalculator {
 
+    var maxMediaItemHeight: CGFloat {
+        return UIScreen.main.bounds.size.height * 0.8
+    }
+
     open override func messageContainerSize(for message: MessageType) -> CGSize {
         let maxWidth = messageContainerMaxWidth(for: message)
         let sizeForMediaItem = { (maxWidth: CGFloat, item: MediaItem) -> CGSize in
+            var imageWidth = item.size.width
+            var imageHeight = item.size.height
             if maxWidth < item.size.width {
                 // Maintain the ratio if width is too great
-                let height = maxWidth * item.size.height / item.size.width
-                return CGSize(width: maxWidth, height: height)
+                imageHeight = maxWidth * item.size.height / item.size.width
+                imageWidth = maxWidth
             }
-            return item.size
+
+            if self.maxMediaItemHeight < imageHeight {
+                // Maintain the ratio if height is too great
+                imageWidth = self.maxMediaItemHeight * imageWidth / imageHeight
+                imageHeight = self.maxMediaItemHeight
+            }
+
+            return CGSize(width: imageWidth, height: imageHeight)
         }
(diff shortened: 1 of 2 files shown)


deltachat-ios/MessageKit/Layout/MediaMessageSizeCalculator.swift
```diff
@@ -27,15 +27,28 @@ import UIKit
 
 open class MediaMessageSizeCalculator: MessageSizeCalculator {
 
+    var maxMediaItemHeight: CGFloat {
```
nayooti, line 30: can be set `private`

deltachat-ios/MessageKit/Layout/TextMediaMessageSizeCalculator.swift
```diff
@@ -27,6 +27,12 @@ import UIKit
 
 open class TextMediaMessageSizeCalculator: MessageSizeCalculator {
 
+    var maxMediaItemHeight: CGFloat {
```
nayooti, line 30: `private`

deltachat-ios/MessageKit/Layout/TextMediaMessageSizeCalculator.swift
```diff
@@ -27,6 +27,12 @@ import UIKit
...
+        return UIScreen.main.bounds.size.height * 0.8
+    }
+
+    let minTextWidth: CGFloat = 180
```
nayooti, line 34: `private`

deltachat-ios/MessageKit/Layout/TextMediaMessageSizeCalculator.swift
```diff
@@ -27,6 +27,12 @@ import UIKit
 
 open class TextMediaMessageSizeCalculator: MessageSizeCalculator {
 
+    var maxMediaItemHeight: CGFloat {
+        return UIScreen.main.bounds.size.height * 0.8
```
nayooti, line 31: On smaller devices in landscape the images still are larger than the screen. Why not use something like this:
Screenheight - NavbarHeight (~44) - InputViewHeight (~100). 
This will never accurate (because inputView does respect safeArea), but this way it is guaranteed  to be smaller than the chat window.

cyBerta: yes we can do that. Still it can be that the image is larger than the screenHeight if it has a bad ratio (say 1:8) and text is below. But yor suggestion will fix other cases, +1

cyBerta: The problem is that the navigationbar height differs between landscape and portatrait, leading to a unnecessarily small image in landscape if I use a constant subtracted from the screen height. I also wonder if the navigation bar height differs between devices, e.g. iPads and iPhones. I would rather reduce the multiplier a little bit so that the media height keeps being fully relative to the screen size.

nayooti: alright.

PR #735, opened 2020-05-19
Chatview - landscape navbar fix

### LINKED ISSUE #734: ChatView Navbar does not like landscape on "notch"-devices
image

on landscape you'll get this:

image

meh ...

### LINKED ISSUE #736: In chat view: consider removing titleview font adjustment
Discussion:

Do we want this (current state):

image

or (iOS-default):

image

Using the `adjustsFontSizeToFitWidth`-attribute on UILabels is very uncommon. In fact I can't remember that I've ever seen this on an Apple-app or any other top rated app.

I 100% prefer using the default behaviour.

nayooti:
closes #734 
closes #736  

Our titleView-container is actually "self-sizing", so it's width constraint is not needed.

Also removed automatic font size adjustment as discussed in #736

### DIFF
--- a/deltachat-ios/View/ChatTitleView.swift
+++ b/deltachat-ios/View/ChatTitleView.swift
@@ -9,7 +9,6 @@ class ChatTitleView: UIView {
         titleLabel.backgroundColor = UIColor.clear
         titleLabel.font = UIFont.systemFont(ofSize: 16, weight: .semibold)
         titleLabel.textAlignment = .center
-        titleLabel.adjustsFontSizeToFitWidth = true
         return titleLabel
     }()
 
@@ -42,6 +41,8 @@ class ChatTitleView: UIView {
         addConstraints([ containerView.constraintAlignTopTo(self),
                          containerView.constraintAlignBottomTo(self),
                          containerView.constraintCenterXTo(self),
+                         containerView.constraintAlignLeadingTo(self),
+                         containerView.constraintAlignTrailingTo(self)
         ])
 
         containerView.addSubview(titleLabel)
--- a/deltachat-ios/View/ChatTitleView.swift
+++ b/deltachat-ios/View/ChatTitleView.swift
@@ -21,17 +21,6 @@ class ChatTitleView: UIView {
         return subtitleLabel
     }()
 
-    private let locationStreamingIndicator: UIImageView = {
-        let view = UIImageView()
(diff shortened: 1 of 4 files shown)


cyBerta:
can you please check the title with location streaming enabled?

nayooti:
> can you please check the title with location streaming enabled?

wasn't aware that. 

I fixed it by moving the location indicator out the title view. It has it's place now as a navigation item next to the badge. 

I had to make a LocationIndicator class because just assigning a plain image to the navigation item did not work (too small).

r10s:
> But I can't find this feature on master either.

you can enable it at "Settings / Advanced / On-demand location streaming".

once done, you can send you current location to any chat using "Location" from the paper-clip-attach menu.

nayooti:
> > But I can't find this feature on master either.
> 
> you can enable it at "Settings / Advanced / On-demand location streaming".
> 
> once done, you can send you current location to any chat using "Location" from the paper-clip-attach menu.

What I meant: after the active location streaming has ended, the location stream indicator is still visible in the navigation bar.

r10s:
> the location stream indicator is still visible in the navigation bar.

might be some update event from the core is missing, however, at least "Attach / Stop sharing location" is working and that can be tested.

deltachat-ios/View/ChatTitleView.swift
```diff
@@ -49,10 +38,11 @@ class ChatTitleView: UIView {
         let containerView = UIView()
         containerView.translatesAutoresizingMaskIntoConstraints = false
         addSubview(containerView)
+
         addConstraints([ containerView.constraintAlignTopTo(self),
                          containerView.constraintAlignBottomTo(self),
                          containerView.constraintCenterXTo(self),
-                         containerView.constraintWidthTo(UIScreen.main.bounds.width - CGFloat(paddingNaviationButtons + sizeStreamingIndicator)) ])
```
cyBerta, line 55: I think the removal of this line introduces the following bug:
Screen Shot 2020-05-27 at 12 27 18

nayooti: Turns out this ticket produces more bugs than it actually fixes.

nayooti: Thanks for finding this bug. I fixed it.

PR #737, opened 2020-05-20
Play gifs

### LINKED ISSUE #683: play gifs in chat
currently gifs are shown as files in a chat
instead they should be displayed and automatically play in a loop

optionally, as discussed with @r10s: 
in the context menu (long tap) we can implement an option to stop the animation 
the message id's of chats with stopped animations are held in a list on runtime but are not persisted. During the cell updates the list will be checked for the cell's new message id.

cyBerta:
closes #683 
implements also sharing of (locally stored) gif files


nayooti:
I managed to share a gif via DcShare, nice! 
I could not find how to share from our ChatView-menu? When I select it via  gallery it is uploading as a video file, which in turn does not auto play etc.

DcCore/DcCore/Helper/DcUtils.swift
```diff
@@ -83,6 +83,22 @@ public struct DcUtils {
...
 
+    public static func saveAnimatedImage(data: Data, suffix: String) -> String? {
+        guard let directory = try? FileManager.default.url(for: .documentDirectory, in: .userDomainMask,
+                                                           appropriateFor: nil, create: false) as NSURL else {
+                                                            return nil
+        }
+        do {
+            let timestamp = Double(Date().timeIntervalSince1970)
+            let path = directory.appendingPathComponent("\(timestamp).\(suffix)")
+            try data.write(to: path!)
```
nayooti, line 94: probably better to unwrap `path`.

In general it is possible to unwrap several optionals in a single `guard let` statement.
So you could youse the `guard let` in line 87 like this.

`guard let directory = try ? ..., let path = directory ... else { return }`

DcShare/Helper/ShareAttachment.swift
```diff
@@ -63,6 +66,37 @@ class ShareAttachment {
...
+                let pixelSize = result.imageSizeInPixel()
+                msg.setDimension(width: pixelSize.width, height: pixelSize.height)
+                self.messages.append(msg)
+                self.delegate?.onAttachmentChanged()
+                if self.imageThumbnail == nil {
+                    self.imageThumbnail = result.scaleDownImage(toMax: self.thumbnailSize)
+                    self.delegate?.onThumbnailChanged()
+                }
+                if error != nil {
+                    self.dcContext.logger?.error(error?.localizedDescription ?? "Could not load share item as image")
```
nayooti, line 94: minor thing, but the `"Could not load share item as image"` is unreachable.
Mabye better:
`if let error = error`, so we don't need the optional chaining.

DcShare/Helper/ShareAttachment.swift
```diff
@@ -63,6 +66,37 @@ class ShareAttachment {
         }
     }
 
+    // for now we only support GIF
+    func createAnimatedImageMsg(_ item: NSItemProvider) {
```
nayooti, line 70: `private`  ... (I guess all `create...`-methods in this file can be private).

cyBerta: fixed

r10s:
i think the commit https://github.com/deltachat/deltachat-ios/pull/737/commits/3ea2722a248c313d9975614861d412fa9c4d38a9 can be removed from the pr, it is already on master as another commit (wondering that there is no merge conflict)

cyBerta:
The commit was made here first, and then cherry-picked for the other PR. I did that in order to be able to work on the PR while the master was broken. Since the git history shows the reference to 3ea2722 (sth like "cherry-picked from 3ea2722") in the first commit on master, I would keep it rather in. This way the history keeps being comprehensible (for git nerds at least)

r10s:
thanks for explanation!

PR #739, opened 2020-05-23
Gallery for chat detail

### LINKED ISSUE #682: improve gallery layout
- implement a gallery grid similar to android containing section headers for different dates (last week, last month, month XYZ). 
- Tapping on a thumbnail should open the QLPreviewController allowing to see the file in detail and being able to swipe to the next or previous one. 
- if possible hide the list icon in the QLPreviewController as it is not needed in our case

nayooti:
fixes #682 

Implemented classic responsive grid layout to display chat images. The layout (number of columns) can be easily changed.

I struggled with the TimeBuckets a bit, but I think the buckets are accurate.
I did not find a translation key for older than month X. So this is currently missing. 

Needs further tweaks if we also want to use the Gallery for documents also. I would prefer to have this in a separate issue/PR. 

Removing the List-item on PreviewController is possible but not super easy neither. I would like to move this sub-task into a separate ticket (since this PR has become quite large already.).

[UPDATE]:
Our core currently doesn't provide any message fetches for "time buckets", so grouping the gallery into time sections is not very easy to achieve, because it would require creating DcMsg-Objects from all available gallery related MessageIds to get the date.

After short discussion with @cyBerta we agreed to handle dates as Telegram (iOS) does for now. This can be changed later if we decide that we want to have time groups. 
The approach now:
While scrolling it shows a centered label with the date of the first visible cell. 

I personally like this approach a lot better and the implementation was quite easy and fits nicely with our DcCore.

Also consider section headers in iOS UICollections (what we use in our gallery) are not as easy to implement compared to UITableViews. There is no default label, and no grouping etc. and no default native look.


cyBerta:
Hey @nayooti, the gallery already looks really nice. Some nitpicking from my side:

- The section header of could have a light gray background in order to differentiate the time buckets better. Personally I would also prefer to have not capital letters, but I think this is an iOS convention.

- There's an empty state for the gallery missing. Iirc @r10s implemented some weeks ago a check that disables the gallery/documents button if there are no common images/files. It seems to be missing here. This way we don't need an extra string saying "the gallery is empty" or something like that. 

- Some images are not displayed correctly in the grid view. It looks like there's a dummy image in the background, and if there's an image with transparency, you can see it in the background. Also some images are not displayed at all
I will attach an example: 
gallery1

the image that is not shown in the gallery from above example
Screen Shot 2020-05-27 at 11 41 25

- Videos are also displayed incorrectly. Instead of an thumbnail, the dummy image from the axample above is shown. I think either videos should be excluded here for now or shown as a thumbnail image. 

- On Android the gallery shows also gifs. Since we now use the SDWebImage framework, it can probably easily archieved on iOS. But this can also be part of another PR.

deltachat-ios/Helper/TimeBucket.swift
```diff
@@ -0,0 +1,59 @@
...
+        if date.isInThisMonth {
+            return .thisMonth
+        }
+
+        if date.isInLastMonth {
+            return .lastMonth
+        }
+
+        // TODO: handle months
+        return .lastMonth
```
cyBerta, line 40: this probably needs another enum value

deltachat-ios/Extensions/Date+Extension.swift
```diff
@@ -0,0 +1,45 @@
...
+        }
+        return isEqual(to: lastMonthDate, toGranularity: .month)
+    }
+
+
+    var isInYesterday: Bool { Calendar.current.isDateInYesterday(self) }
+    var isInToday: Bool { Calendar.current.isDateInToday(self) }
+    var isInTomorrow: Bool { Calendar.current.isDateInTomorrow(self) }
+
+    var isInTheFuture: Bool { self > Date() }
```
cyBerta, line 41: some of these variables are unused.

cyBerta:
did you retest with videos?

deltachat-ios.xcodeproj/project.pbxproj
```diff
@@ -949,7 +963,12 @@
 				305FE03523A81B4C0053BE90 /* EmptyStateLabel.swift */,
 				AEFBE22E23FEF23D0045327A /* ProviderInfoCell.swift */,
 				AEB54C7E246DBA610004624C /* FlexLabel.swift */,
+<<<<<<< HEAD
```
cyBerta, line 966: This needs to be deleted, remains of a merge conflict

deltachat-ios/Helper/TimeBucket.swift
nayooti: > Hey @nayooti, the gallery already looks really nice. Some nitpicking from my side:
> 
> * The section header of could have a light gray background in order to differentiate the time buckets better. Personally I would also prefer to have not capital letters, but I think this is an iOS convention.

Solved by changing the concept. See PR-description update.

> * There's an empty state for the gallery missing. Iirc @r10s implemented some weeks ago a check that disables the gallery/documents button if there are no common images/files. It seems to be missing here. This way we don't need an extra string saying "the gallery is empty" or something like that.

I looked into this: I will add an empty state label for consistency. 

> * Some images are not displayed correctly in the grid view. It looks like there's a dummy image in the background, and if there's an image with transparency, you can see it in the background.
>   Also some images are not displayed at all

Hopefully fixed. :)

>   [EDIT: The dummy image in the background is not true. There captured image has exactly that table view cells in the background :P. That other images are not displayed at all is still true though.]
> 
> I will attach an example:
> gallery1
> 
> the image that is not shown in the gallery from above example
> Screen Shot 2020-05-27 at 11 41 25
> 
> * Videos are also displayed incorrectly. Instead of an thumbnail, the dummy image from the axample above is shown. I think either videos should be excluded here for now or shown as a thumbnail image.

Fixed by showing thumbnails.

> * On Android the gallery shows also gifs. Since we now use the SDWebImage framework, it can probably easily archieved on iOS. But this can also be part of another PR.

Fixed.

nayooti:
I think rendering thumbnails from videos is still wonky (scrolling hangs) - should be addressed very soonish in #748

cyBerta:
Unfortunately when tapping on the preview the wrong images are shown. The image ordering between the grid view controller and the second image controller (showing the expanded image) seem to differ.

deltachat-ios.xcodeproj/project.pbxproj
nayooti: > There's a mismatch between the image ordering of the grid view controller and the detail image controller

Yeah true. I reverted the messageIds ordering (show newest on top). This might be the reason. Thank you!

cyBerta:
the empty state string has no default yet in the english Localizable.strings. This should be done as well and it is fine in this case to add it manually to the bottom of the Localizable.strings (as long as you _also_ add it to untranslated.xml)

cyBerta:
When there's only one image and thus the whole view doesn't scrolls, there's no way to figure out when the image was sent. If this I'm not sure if this is easy to fix though using the current implementation, @nayooti ?

nayooti:
> When there's only one image and thus the whole view doesn't scrolls, there's no way to figure out when the image was sent. I'm not sure if this is easy to fix using the current implementation, @nayooti ?

I don't give up so easily.

nayooti:
> the empty state string has no default yet in the english Localizable.strings. This should be done as well and it is fine in this case to add it manually to the bottom of the Localizable.strings (as long as you _also_ add it to untranslated.xml)

nayooti:
> the empty state string has no default yet in the english Localizable.strings. This should be done as well and it is fine in this case to add it manually to the bottom of the Localizable.strings (as long as you _also_ add it to untranslated.xml)

fixed

nayooti:
> great stuff!
> 
> i also like the clean code, pretty straigt-forward, thanks a lot :)
> 
> what's missing at a first check:
> 
> * a "play icon" if the item is a video, please re-use the icon and the style from the chat-view-controller.
> * wrt columns count: they are static? then i would use 3 ones in portrait-mode (and at least 4 in landscape).
> * if a single image is selected, there is still the old document-like gallery-icon in the lower-right corner; this is no longer needed.  however, we can also remove that in a subsequent pr.

 As stated in the PR description - I created a new issue for removing the list item (#785). Will take care of this right after this PR.

PR #744, opened 2020-05-26
remove unused progress hud from NewChatViewController

### LINKED ISSUE #742: still references to JGProgressHUD
the JGProgressHUD pod was removed in https://github.com/deltachat/deltachat-ios/pull/718, which is great.

however, there is a reference left at View/ProgressHud.swift which is probably no longer used and another reference at Controller/SettingsController.swift

this leads to linker issues on rebuild-all or so, got that when trying #740

cyBerta:
closes #742 

* I didn't find any indication the observer gets ever triggered, that's why I removed it completely.

### DIFF
--- a/deltachat-ios/Controller/NewChatViewController.swift
+++ b/deltachat-ios/Controller/NewChatViewController.swift
@@ -42,9 +42,6 @@ class NewChatViewController: UITableViewController {
         return searchController.searchBar.text?.isEmpty ?? true
     }
 
-    var syncObserver: Any?
-    var hud: ProgressHud?
-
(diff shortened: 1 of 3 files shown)


deltachat-ios/Controller/NewChatViewController.swift
```diff
@@ -42,9 +42,6 @@ class NewChatViewController: UITableViewController {
         return searchController.searchBar.text?.isEmpty ?? true
     }
 
-    var syncObserver: Any?
-    var hud: ProgressHud?
```
r10s, line 46: even if the handler is triggered for some reason, this `hud` seems to be never set.

PR #746, opened 2020-05-27
[WIP] Virtual chat list

### LINKED ISSUE #614: Create virtual list for messages
As a precondition to fix #602  we need to be able to refer to the whole list of chat messages, not only the last N as currently implemented using a pull-to-refresh mechanism. 
The pull-to-refresh can be removed completely.

cyBerta:
closes #614 
similarly to the Android implementation (ConversationAdapter) it uses a cache for faster loading of DcMsgs.
The [cache](https://developer.apple.com/documentation/foundation/nscache) gets automatically cleared/resized by the OS if the memory is low.

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -31,6 +31,10 @@ public class DcContext {
         return .dcContext
     }
 
+    public func getMessageCount(chatId: Int) -> Int {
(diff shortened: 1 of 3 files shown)


cyBerta:
closing this PR for now. Explanation can be found here: https://github.com/deltachat/deltachat-ios/issues/614#issuecomment-635435748

PR #753, opened 2020-05-29
Share-Extension-To-Main-App-Communication

### LINKED ISSUE #732: sync main app after database changes from share extension
Currently the main app doesn't updates it's UI if it is still running in the background while a message was sent using the share extension.

An easy way is to write to shared UserDefaults after the extension has sent a message. On reentering the foreground the main app needs to check the UserDefaults and send a DC_MESSAGE_CHANGED event if a database change is indicated. Afterwards the flag needs to be reset.

cyBerta:
closes #732

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -107,6 +107,26 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         if reachability.connection != .none {
             self.dcContext.maybeNetwork()
         }
+        
+        let userDefaults = UserDefaults.init(suiteName: Constants.sharedUserDefaults)
+        if userDefaults?.bool(forKey: Constants.hasExtensionAttemptedToSend) ?? false {
+            let nc = NotificationCenter.default
+
+            DispatchQueue.main.async {
+                nc.post(
+                    name: dcNotificationChanged,
+                    object: nil,
+                    userInfo: [
+                        //TODO: Use correct message_id and chat_id
+                        "message_id": Int(0),
+                        "chat_id": Int(0),
+                        "date": Date(),
+                    ]
+                )
+            }
+
+            userDefaults?.removeObject(forKey: Constants.hasExtensionAttemptedToSend)
+        }
     }
 
     func applicationDidEnterBackground(_: UIApplication) {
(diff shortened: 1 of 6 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -111,6 +111,25 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
+        let userDefaults = UserDefaults.init(suiteName: Constants.sharedUserDefaults)
+        if userDefaults?.bool(forKey: Constants.hasExtensionAttemptedToSend) ?? false {
+            let nc = NotificationCenter.default
+
+            DispatchQueue.main.async {
+                nc.post(
+                    name: dcNotificationChanged,
+                    object: nil,
+                    userInfo: [
+                        //TODO: Use correct message_id and chat_id
```
r10s, line 123: i think, this is not needed, DC_EVENT_MSGS_CHANGED is also sent from the core without these information, the app has to handle that (an actually does afaik)

deltachat-ios/AppDelegate.swift
```diff
@@ -111,6 +111,25 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
+                        "message_id": Int(0),
+                        "chat_id": Int(0),
+                        "date": Date(),
+                    ]
+                )
+            }
+
+            userDefaults?.removeObject(forKey: Constants.hasExtensionAttemptedToSend)
```
r10s, line 131: maybe move this line before the notification is sent, as the first line after the "if": that would make race-conditions less likely (i know, they are already unlikely as the user cannot really do both), but, more important it keeps things closer together and improve readability.

deltachat-ios/AppDelegate.swift
cyBerta: sending an empty hash table for user info is sufficient, indeed. omitting userInfo completely won't work.

deltachat-ios/AppDelegate.swift
```diff
@@ -103,10 +103,23 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         logger.info("---- foreground ----")
         appIsInForeground = true
         startThreads()
-
         if reachability.connection != .none {
             self.dcContext.maybeNetwork()
         }
+
+        let userDefaults = UserDefaults.shared
+        if userDefaults?.bool(forKey: UserDefaults.hasExtensionAttemptedToSend) ?? false {
```
nayooti, line 111: How about:
`if let userDefaults = UserDefaults.shared, userDefaults.bool(forKey ....) { /* if block*/}`

cyBerta: yes that's nicer, thanks

cyBerta: fixed

PR #759, opened 2020-06-02
remove location icon when location streaming ended

### LINKED ISSUE #747: Location streaming indicator not removed from Chat title
After location streaming ended and the user has opend the corresponding chat at that time the location indicator doesn't get removed from a chat title. Leaving a chat and reentering it "fixes" the situation.

cyBerta:
fixes #747

### DIFF
--- a/deltachat-ios/Controller/ChatViewController.swift
+++ b/deltachat-ios/Controller/ChatViewController.swift
@@ -300,7 +300,7 @@ class ChatViewController: MessagesViewController {
         if chat.isSendingLocations {
             navigationItem.rightBarButtonItems = [badgeItem, locationStreamingItem]
         } else {
-            navigationItem.rightBarButtonItem = badgeItem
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/ChatViewController.swift
```diff
@@ -300,7 +300,7 @@ class ChatViewController: MessagesViewController {
         if chat.isSendingLocations {
             navigationItem.rightBarButtonItems = [badgeItem, locationStreamingItem]
         } else {
-            navigationItem.rightBarButtonItem = badgeItem
+            navigationItem.rightBarButtonItems = [badgeItem]
```
nayooti, line 303: Cool!

cyBerta:
> So I think there is some other issue behind.

Yes the problem is the implementation of location streaming. You will need to walk around (location messages will be sent) or write messages in order to trigger the validation if location streaming should be disabled. There's no timer implemented in the core.

PR #740, opened 2020-05-24
use async core

r10s:
this pr shifts the ios-ui towards the new async-core-api - core-threads are handled in the core from now on and the ui will use an event-emmiter-object to get events.

~~the pr requires the core-submodule checked out at https://github.com/deltachat/deltachat-core-rust/pull/1356, before this can finally be merged, https://github.com/deltachat/deltachat-core-rust/pull/1356 has to be merged first.~~ EDIT: async is merged to core  

counterpart of https://github.com/deltachat/deltachat-android/pull/1383

PR #760, opened 2020-06-02
fix app delegate

r10s:
this pr calls `maybeStartIo()` without checking ApplicationState and removes the now unused application-quad-state completely; it was already replaced by `appIsInForeground` in #740 

moreover, according to [[1]](https://developer.apple.com/library/archive/documentation/NetworkingInternetWeb/Conceptual/NetworkingOverview/CommonPitfalls/CommonPitfalls.html), `SIGPIPE` exception is ignored unconditionally, afaik in the past it was ignored as a side-effect of `DBDebugToolkit.setupCrashReporting()`, not sure, though.

finally `DBDebugToolkit.setupCrashReporting()` is called on startup, not sure if and which problems this can/could cause, however, at least for the next testflight, this is seems to make some sense :)

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -28,10 +28,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     var appIsInForeground = false
 
     func application(_: UIApplication, didFinishLaunchingWithOptions _: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+        // explicitly ignore SIGPIPE to avoid crashes, see https://developer.apple.com/library/archive/documentation/NetworkingInternetWeb/Conceptual/NetworkingOverview/CommonPitfalls/CommonPitfalls.html
+        // setupCrashReporting() may create an additional handler, but we do not want to rely on that
+        signal(SIGPIPE, SIG_IGN)
+
         DBDebugToolkit.setup(with: []) // empty array will override default device shake trigger
         DBDebugToolkit.setupCrashReporting()
 
-        // main()
         let console = ConsoleDestination()
         logger.addDestination(console)
         dcContext.logger = DcLogger()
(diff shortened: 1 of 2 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -24,13 +17,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     private var backgroundTask: UIBackgroundTaskIdentifier = .invalid
     var reachability = Reachability()!
     var window: UIWindow?
-    var state = ApplicationState.stopped
     var appIsInForeground = false
 
     func application(_: UIApplication, didFinishLaunchingWithOptions _: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+        // explicitly ignore SIGPIPE to avoid crashes, see https://developer.apple.com/library/archive/documentation/NetworkingInternetWeb/Conceptual/NetworkingOverview/CommonPitfalls/CommonPitfalls.html
+        // setupCrashReporting() may create an additional handler, but we do not want to rely on that
```
cyBerta, line 24: indeed, DBDebugToolkit is exactly doing that 
```
    // SIGPIPE signal
    _previousSIGPIPEHandler = signal(SIGPIPE, SIG_IGN);
    signal(SIGPIPE, handleSIGPIPESignal);
```

deltachat-ios/AppDelegate.swift
```diff
@@ -24,13 +17,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
+        signal(SIGPIPE, SIG_IGN)
```
cyBerta, line 25: Did you test if DBDebugToolkit is still saving exceptions correctly?

deltachat-ios/AppDelegate.swift
r10s: yi., however, if we disable/move the setupCrashReporting or maybe DBDebugToolkit gets an update and does things differently, i think it is good to have this disabled explicitly.

deltachat-ios/AppDelegate.swift
r10s: it does.

r10s:
merged that for now, a 1.9.2 is on the way, we can improve things as needed.

PR #763, opened 2020-06-03
add last chats to share suggestions

### LINKED ISSUE #731: last contacts in share context
iOS' share context allows to show the last contacts/group a user wrote messages to. It is ment as a shortcut and would complement the current share extension implementation.

cyBerta:
closes #731 

we might want to have this feature optional. The documentation is quite very vague about the privacy implications of using SiriKit intents needed to implement the suggestion shortcuts. 

Wire and Signal don't have implemented this feature at all, whereas Telegram does. 
From what I understood, reading various third party articles is that at least anonymized data will be sent to Siri servers.

Another open discussion point is if we want to ignore location messages (as currently implemented) from being taken into account for the the share suggestion data aggregation. 

Also missing in this PR is a default image generation in case a user has no own user image set. In a subsequent PR the image generation could create an UIImage based on the InitialsBadge layout containing the users Initials and default background color.

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -408,7 +408,6 @@ public class DcContext {
 
     public func sendLocationsToChat(chatId: Int, seconds: Int) {
         dc_send_locations_to_chat(contextPointer, UInt32(chatId), Int32(seconds))
-        //TODO: discuss if we want to omit the message intent donation here here
     }
 
     public func setLocation(latitude: Double, longitude: Double, accuracy: Double) {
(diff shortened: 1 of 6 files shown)


nayooti:
I am not going to merge, since you mentioned "second thoughts" about this PR in description.

r10s:
yip, please wait a bit wrt merging, i did not find the time to dive into the details yet. let's do an a/v on that next week.

r10s:
> From what I understood, reading various third party articles is that at least anonymized data will be sent to Siri servers.

@cyBerta can you recommend some links for that?

in general, maybe we could do the INInteraction.donate() call only when a different contact on sharing is selected. maybe this is even nicer, as the selection is more focused on ppl you already shared to.

cyBerta:
it is quite hard to find any discussions about the privacy implications. This sources claims siri shortcuts are fine to use privacy wise: https://www.imore.com/siri-shortcuts-how-apple-keeping-your-data-private-and-secure

cyBerta:
@r10s I've updated the PR according to our agreements: shortcuts producing INSendMessageIntents will be only donated while sharing. Since self chat is already the default selected item when tapping on the general delta chat icon I excluded that from the short cuts.

r10s:
ftr: this is the apple reference: https://developer.apple.com/documentation/foundation/app_extension_support/supporting_suggestions_in_your_app_s_share_extension

PR #766, opened 2020-06-04
send gifs from attach menu

### LINKED ISSUE #750: send gifs via attach menu
currently gifs are interpreted as images if they are attached from the gallery.
as a follow-up of #737, gifs should be sent as messages with the viewtype gif instead.

cyBerta:
closes #750

### DIFF
--- a/deltachat-ios/Controller/ChatViewController.swift
+++ b/deltachat-ios/Controller/ChatViewController.swift
@@ -1053,7 +1053,7 @@ extension ChatViewController: MessagesDataSource {
         if url.pathExtension == "gif" {
             sendAnimatedImage(url: url)
         } else if let data = try? Data(contentsOf: url as URL),
-            let image = UIImage(data: data) {
+                  let image = UIImage(data: data) {
             sendImage(image)
         }
     }
--- a/deltachat-ios/Controller/ChatViewController.swift
+++ b/deltachat-ios/Controller/ChatViewController.swift
@@ -4,6 +4,7 @@ import UIKit
 import InputBarAccessoryView
 import AVFoundation
 import DcCore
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/ChatViewController.swift
```diff
@@ -1036,10 +1051,11 @@ extension ChatViewController: MessagesDataSource {
...
 
     private func sendImage(url: NSURL) {
-        if let data = try? Data(contentsOf: url as URL) {
-            if let image = UIImage(data: data) {
-                sendImage(image)
-            }
+        if url.pathExtension == "gif" {
+            sendAnimatedImage(url: url)
+        } else if let data = try? Data(contentsOf: url as URL),
+            let image = UIImage(data: data) {
```
r10s, line 1057: this is a weird linebreak. it was not obvious to me at a first glance that this line belongs to the if-statement. not a blocker, however, for the future, maybe use one line or align the let-statements horizontally.

cyBerta:
> iiuc, the whole point of the pr is to handle gif-files separated as `dcCompress()` cannot recode gif files.

not exactly. The point is to load and save the gif data as files in the apps document folder, so that they can be attached. If gifs were loaded in a UIImage the data would be trunkated and only the fist image of the gif can be saved. That dcCompress is not called is rather a side effect and removing that from sendImage() will be done with #773

deltachat-ios/Controller/ChatViewController.swift
cyBerta: changed

r10s: thanks!

PR #767, opened 2020-06-04
show share error view if account was not configured yet

### LINKED ISSUE #761: Show an error view if sharing was tried while account was not configured
If the user has not configured an account yet, the share extension appears currently without any hint that sharing won't work. 
As a quick fix, the sending button should be disabled, the preview view and the contact selection button should be hidden and the text input field should show an error message.

cyBerta:
closes #761

### DIFF
--- a/DcShare/Controller/ShareViewController.swift
+++ b/DcShare/Controller/ShareViewController.swift
@@ -34,7 +34,7 @@ class ShareViewController: SLComposeServiceViewController {
     var selectedChat: DcChat?
     let dbHelper = DatabaseHelper()
     var shareAttachment: ShareAttachment?
-    var isAccountConfigured: Bool = false
+    var isAccountConfigured: Bool = true
 
     lazy var preview: UIImageView? = {
         let imageView = UIImageView(frame: .zero)
--- a/DcShare/Controller/ShareViewController.swift
+++ b/DcShare/Controller/ShareViewController.swift
@@ -68,7 +68,6 @@ class ShareViewController: SLComposeServiceViewController {
             dcContext.openDatabase(dbFile: dbHelper.sharedDbFile)
             isAccountConfigured = dcContext.isConfigured()
             if isAccountConfigured {
-                isAccountConfigured = true
                 selectedChatId = dcContext.getChatIdByContactId(contactId: Int(DC_CONTACT_ID_SELF))
                 if let chatId = selectedChatId {
                     selectedChat = dcContext.getChat(chatId: chatId)
@@ -144,7 +143,7 @@ class ShareViewController: SLComposeServiceViewController {
                 self.pushConfigurationViewController(chatListController)
             }
         } else {
-            item?.value = String.localized("share_account_not_configured")
+            item?.title = String.localized("share_account_not_configured")
         }
 
         return [item as Any]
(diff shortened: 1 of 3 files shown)


DcShare/Controller/ShareViewController.swift
```diff
@@ -65,11 +66,16 @@ class ShareViewController: SLComposeServiceViewController {
         if dbHelper.currentDatabaseLocation == dbHelper.sharedDbFile {
             dcContext.logger = self.logger
             dcContext.openDatabase(dbFile: dbHelper.sharedDbFile)
-            selectedChatId = dcContext.getChatIdByContactId(contactId: Int(DC_CONTACT_ID_SELF))
-            if let chatId = selectedChatId {
-                selectedChat = dcContext.getChat(chatId: chatId)
+            isAccountConfigured = dcContext.isConfigured()
+            if isAccountConfigured {
+                isAccountConfigured = true
```
r10s, line 71: `isAccountConfigured` is always `true` here.

DcShare/Controller/ShareViewController.swift
```diff
@@ -65,11 +66,16 @@ class ShareViewController: SLComposeServiceViewController {
...
+                selectedChatId = dcContext.getChatIdByContactId(contactId: Int(DC_CONTACT_ID_SELF))
+                if let chatId = selectedChatId {
+                    selectedChat = dcContext.getChat(chatId: chatId)
+                }
             }
             reloadConfigurationItems()
+            validateContent()
```
r10s, line 78: just to get an idea about the flow: why was the call to `validateContent()` not needed before?

DcShare/Controller/ShareViewController.swift
```diff
@@ -124,16 +130,21 @@ class ShareViewController: SLComposeServiceViewController {
...
+
+            item?.title = String.localized("forward_to")
+            item?.value = selectedChat?.name
+            logger.debug("configurationItems chat name: \(String(describing: selectedChat?.name))")
+            item?.tapHandler = {
+                let chatListController = ChatListController(dcContext: self.dcContext, chatListDelegate: self)
+                self.pushConfigurationViewController(chatListController)
+            }
+        } else {
+            item?.value = String.localized("share_account_not_configured")
```
r10s, line 147: if possible, i'd use `item?.title` here. the right-aligned gray text is easily overseen.

DcShare/Controller/ShareViewController.swift
cyBerta: copy and paste error, thanks

DcShare/Controller/ShareViewController.swift
cyBerta: because validateContent gets called earlier in the life cycle when we don't yet check if the account was configured or not. 
Before this PR changes we didn't check the account configuration at all. 
We don't move the configuration check to an earlier slot in the life cycle, because 'heavier' processings should not happen before `presentationAnimationDidFinish()`

DcShare/Controller/ShareViewController.swift
cyBerta: done

DcShare/Controller/ShareViewController.swift
r10s: thanks for explanations!

PR #776, opened 2020-06-09
allow to select images in the document picker

### LINKED ISSUE #774: allow attaching images as files
if possible, also allow attaching images by the "Attach files" function of the paperclip-attach-menu.

when we continue sending the selection as DC_MSG_FILE, the images would not be recoded and we would have a way to attach original images.

this is also the way, android handles this currently. at some point, however, we might introduce some per-chat-settings or so.

follow-up of https://github.com/deltachat/deltachat-ios/issues/733

cyBerta:
closes #774

### DIFF
--- a/deltachat-ios/Helper/MediaPicker.swift
+++ b/deltachat-ios/Helper/MediaPicker.swift
@@ -65,7 +65,7 @@ class MediaPicker: NSObject, UINavigationControllerDelegate, UIImagePickerContro
     }
 
     func showDocumentLibrary(delegate: MediaPickerDelegate) {
-        let types = [kUTTypePDF, kUTTypeText, kUTTypeRTF, kUTTypeSpreadsheet, kUTTypeVCard, kUTTypeZipArchive]
+        let types = [kUTTypePDF, kUTTypeText, kUTTypeRTF, kUTTypeSpreadsheet, kUTTypeVCard, kUTTypeZipArchive, kUTTypeImage]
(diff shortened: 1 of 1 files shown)


cyBerta:
I also noticed that but haven't found a fast way to add the photos directory. I can try and research a little bit more though.

cyBerta:
After researching a little bit more I don't think it is possible to access the photos directory through UIDocumentPickerViewController (It is only possible with the help of [photokit](https://developer.apple.com/documentation/photokit) framework) 
@nayooti do you know any special tricks :)?

nayooti:
> After researching a little bit more I don't think it is possible to access the photos directory through UIDocumentPickerViewController (It is only possible with the help of [photokit](https://developer.apple.com/documentation/photokit) framework)
> @nayooti do you know any special tricks :)?

I will look into this tomorrow.

nayooti:
Yep, the *Photokit*-approach seems to be the way to go. But what is stopping us from using *Photokit*?

cyBerta:
Photokit lets us only select photos and only from iCloud photos and the local photo library location. We want to access the documents folders and attach different kind of files when tapping on the option 'Documents'

So either we merge this PR as it is (only photos saved in the document directory is accessible) or we should reconsider the UX for the underlying problem we tried to solve with it - sending photos without compression.

r10s:
yip, i think, we can merge that in. i also think, PhotoKit won't help on that.

the idea is to have a per-chat-image-quality setting at some point, however, nothing to start just now. it is not a worsening, so, this is not super-urgent.

PR #787, opened 2020-06-15
Async thumbnail loading in chatView

### LINKED ISSUE #748: async thumbnail generation
when scrolling through chat messages, the lists starts to become laggy if one ore more videos are about to appear on the screen. The reason is that the thumbnail generation happens currently on the main UI thread. It should be delegated to a background thread. Also caching thumbnails would be an additional option to discuss.

nayooti:
closes #748 

In ChatViewController the thumbnails of video cells are now created in background thread and are no longer blocking the UI. Created thumbnails are cached and restored if possible.

When the table is loaded for the first time, there is now chance to calculate the video cells height. In that case we use the max width as default value, so video cells appear squared. 
Once the content has beed setup, the table will reload and the cell sizes are re-calculated.

Technically I wrote an extension of UIImage, so once the thumbnail images has been created it will be loaded into the imageView.

In ShareViewController thumbnails are created in main thread. I will create a separate ticket for it.

### DIFF
--- a/deltachat-ios/DC/DcMsg+Extension.swift
+++ b/deltachat-ios/DC/DcMsg+Extension.swift
@@ -48,13 +48,18 @@ extension DcMsg: MessageType {
     }
 
     internal func createVideoMessage(text: String) -> MessageKind {
-        let thumbnail = DcUtils.generateThumbnailFromVideo(url: fileURL)
         if text.isEmpty {
+            var thumbnail: UIImage?
+            if let fileURL = fileURL {
+                thumbnail = ThumbnailCache.shared.restoreImage(key: fileURL.absoluteString)
+            }
             return MessageKind.video(Media(url: fileURL, image: thumbnail))
         }
-        let attributedString = NSAttributedString(string: text, attributes: [NSAttributedString.Key.font: UIFont.preferredFont(forTextStyle: .body),
-                                                                             NSAttributedString.Key.foregroundColor: DcColors.defaultTextColor])
-        return MessageKind.videoText(Media(url: fileURL, image: thumbnail, text: [attributedString]))
+        let attributedString = NSAttributedString(string: text, attributes: [
+            NSAttributedString.Key.font: UIFont.preferredFont(forTextStyle: .body),
+            NSAttributedString.Key.foregroundColor: DcColors.defaultTextColor]
+        )
+        return MessageKind.videoText(Media(url: fileURL, text: [attributedString]))
     }
 
     internal func createImageMessage(text: String) -> MessageKind {
--- a/deltachat-ios/DC/DcMsg+Extension.swift
+++ b/deltachat-ios/DC/DcMsg+Extension.swift
@@ -48,18 +48,19 @@ extension DcMsg: MessageType {
     }
 
     internal func createVideoMessage(text: String) -> MessageKind {
+        var thumbnail: UIImage?
+        if let fileURL = fileURL {
+            thumbnail = ThumbnailCache.shared.restoreImage(key: fileURL.absoluteString)
+        }
         if text.isEmpty {
-            var thumbnail: UIImage?
-            if let fileURL = fileURL {
-                thumbnail = ThumbnailCache.shared.restoreImage(key: fileURL.absoluteString)
-            }
+
             return MessageKind.video(Media(url: fileURL, image: thumbnail))
         }
         let attributedString = NSAttributedString(string: text, attributes: [
             NSAttributedString.Key.font: UIFont.preferredFont(forTextStyle: .body),
             NSAttributedString.Key.foregroundColor: DcColors.defaultTextColor]
         )
-        return MessageKind.videoText(Media(url: fileURL, text: [attributedString]))
+        return MessageKind.videoText(Media(url: fileURL, image: thumbnail, text: [attributedString]))
     }
 
     internal func createImageMessage(text: String) -> MessageKind {
--- a/deltachat-ios/Extensions/UIImageView+Extensions.swift
+++ b/deltachat-ios/Extensions/UIImageView+Extensions.swift
@@ -1,0 +1,17 @@
+import UIKit
+import DcCore
+
+extension UIImageView {
+
+    func loadVideoThumbnail(from url: URL, placeholderImage: UIImage?, completionHandler: ((UIImage?)->Void)?) {
+
+        self.image = placeholderImage
+        DispatchQueue.global(qos: .background).async {
+            let thumbnailImage = DcUtils.generateThumbnailFromVideo(url: url)
+            DispatchQueue.main.async { [weak self] in
+                self?.image = thumbnailImage
+                completionHandler?(thumbnailImage)
+            }
+        }
+    }
+}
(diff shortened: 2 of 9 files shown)


deltachat-ios/Extensions/UIImageView+Extensions.swift
```diff
@@ -0,0 +1,17 @@
+import UIKit
+import DcCore
+
+extension UIImageView {
+
+    func loadVideoThumbnail(from url: URL, placeholderImage: UIImage?, completionHandler: ((UIImage?)->Void)?) {
```
cyBerta, line 6: lint issue here :)

deltachat-ios/MessageKit/Views/Cells/MediaMessageCell.swift
```diff
@@ -73,12 +81,30 @@ open class MediaMessageCell: MessageContentCell {
             imageView.image = mediaItem.image ?? mediaItem.placeholderImage
             playButtonView.isHidden = true
         case .video(let mediaItem):
-            imageView.image = mediaItem.image ?? mediaItem.placeholderImage
+            if let url = mediaItem.url {
+                if let image = mediaItem.image {
+                    imageView.image = image
+                } else {
+                    // no image in cache
```
cyBerta, line 88: no placeholder is set in this case, it's just the cells background and the play button.

deltachat-ios/MessageKit/Views/Cells/MediaMessageCell.swift
```diff
@@ -73,12 +81,30 @@ open class MediaMessageCell: MessageContentCell {
...
+                    imageView.loadVideoThumbnail(from: url, placeholderImage: mediaItem.placeholderImage, completionHandler: { [weak self] thumbnail in
+                        if let image = thumbnail {
+                            self?.cache(thumbnail: image, key: url.absoluteString)
+                            self?.asyncDelegate?.contentDidLoad()
```
cyBerta, line 92: this reloads the whole dataset instead of the single cell. It still leads to little interruptions when scrolling. It would be nice if there is a method to request redraw of the UIImageView only. Can you check if there's something like that?

deltachat-ios/Extensions/UIImageView+Extensions.swift
```diff
@@ -0,0 +1,17 @@
...
+
+        self.image = placeholderImage
+        DispatchQueue.global(qos: .background).async {
+            let thumbnailImage = DcUtils.generateThumbnailFromVideo(url: url)
+            DispatchQueue.main.async { [weak self] in
+                self?.image = thumbnailImage
+                completionHandler?(thumbnailImage)
```
cyBerta, line 13: referring to https://github.com/deltachat/deltachat-ios/pull/787/files#r441420963, instead of the completion handler maybe setNeedsDisplay() is sufficient here?

nayooti: I am using the completion handler here to cache the image, so I would like to keep the completion handler here.

deltachat-ios/Extensions/UIImageView+Extensions.swift
nayooti: fixed

deltachat-ios/MessageKit/Views/Cells/MediaMessageCell.swift
nayooti: > no placeholder is set in this case, it's just the cells background and the play button.

Do you mean in the `else`-case? In .loadVideoThumbnail I give an optional placeholder as argument and set the placeholder there. Not sure what I should change here?

deltachat-ios/MessageKit/Views/Cells/MediaMessageCell.swift
nayooti: Fixed by giving video cells a fixed size -> therefore a reloading is no longer needed.

deltachat-ios/MessageKit/Views/Cells/MediaMessageCell.swift
cyBerta: oh, you're right. I've overseen that. Thanks for the hint

deltachat-ios/Extensions/UIImageView+Extensions.swift
cyBerta: I think we will also want to cache the images in other cases when we call this extension method, so we don't need to do it in a completion handler. Anyways, it can be refactored later, too.

deltachat-ios/DC/DcMsg+Extension.swift
```diff
@@ -48,13 +48,18 @@ extension DcMsg: MessageType {
...
             return MessageKind.video(Media(url: fileURL, image: thumbnail))
         }
-        let attributedString = NSAttributedString(string: text, attributes: [NSAttributedString.Key.font: UIFont.preferredFont(forTextStyle: .body),
-                                                                             NSAttributedString.Key.foregroundColor: DcColors.defaultTextColor])
-        return MessageKind.videoText(Media(url: fileURL, image: thumbnail, text: [attributedString]))
+        let attributedString = NSAttributedString(string: text, attributes: [
+            NSAttributedString.Key.font: UIFont.preferredFont(forTextStyle: .body),
+            NSAttributedString.Key.foregroundColor: DcColors.defaultTextColor]
+        )
+        return MessageKind.videoText(Media(url: fileURL, text: [attributedString]))
```
cyBerta, line 62: There is a  🐞 
Cells with combined text and video should also provided with the thumbnail here. (`Media(url: fileURL, image: thumbnail, text: [attributedString])`)
TextMediaMessageCells should also handle the situation if there's no thumbnail yet there.

nayooti: Tx! Fixed

deltachat-ios/MessageKit/Layout/TextMediaMessageSizeCalculator.swift
```diff
@@ -81,13 +81,22 @@ open class TextMediaMessageSizeCalculator: MessageSizeCalculator {
...
-            case .photoText(let mediaItem), .videoText(let mediaItem), .animatedImageText(let mediaItem):
+            case .photoText(let mediaItem), .animatedImageText(let mediaItem):
                 if let text = mediaItem.text?[MediaItemConstants.messageText] {
                     let textHeight = text.height(withConstrainedWidth: maxTextWidth)
                     messageContainerSize.height += textHeight
                     messageContainerSize.height +=  self.messageLabelInsets(for: message).vertical
                 }
                 return messageContainerSize
+            case .videoText(let mediaItem):
+                var videoContainerSize = CGSize(width: self.minTextWidth, height: self.minTextWidth)
```
cyBerta, line 92: This line looks suspicious to me. 
The layout of the text image cell is not correct right now.
Screen Shot 2020-06-18 at 16 24 19

cyBerta: The Cooooool is within the picture. You can reproduce it if you share a video + text to a chat

nayooti: Cool thanks, it should be fixed now.

PR #789, opened 2020-06-15
fix string

r10s:
for most ui things, there is already a string in the android repo at strings.xml :)

apart from that, Localizable.strings is generated and changes there have no long-lasting effect.

### DIFF
--- a/deltachat-ios/Controller/GalleryViewController.swift
+++ b/deltachat-ios/Controller/GalleryViewController.swift
@@ -37,7 +37,7 @@ class GalleryViewController: UIViewController {
 
     private lazy var emptyStateView: EmptyStateLabel = {
         let label = EmptyStateLabel()
-        label.text = String.localized("chat_gallery_empty_state")
+        label.text = String.localized("tab_gallery_empty_hint")
         label.isHidden = true
         return label
     }()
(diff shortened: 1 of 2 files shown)


cyBerta:
> apart from that, Localizable.strings is generated and changes there have no long-lasting effect.

But why do you change deltachat-ios/en.lproj/Localizable.strings then? As you said, with the next tx-pull it will be re-added if it was not thrown out from Android's strings.xml

r10s:
well, for some reason, that string is converted to `"chat_gallery_empty_state" = ""No media items in this chat.\"";` (literally) when pulling. so it is kept for some reasons and added with wrong quoting. therefore i reverted it :)

cyBerta:
> well, for some reason, that string is converted to `"chat_gallery_empty_state" = ""No media items in this chat.\"";` (literally) when pulling. so it is kept for some reasons and added with wrong quoting. therefore i reverted it :)

ouch. Looks like more parsing issues :(. I'll have a look at that.

cyBerta:
Ok, the corresponding entry in untranslated.xml file was not correctly formatted. I'll remove that entry there in another PR.

r10s:
> I'll remove that entry there in another PR.

thanks - i've overseen that :)

PR #793, opened 2020-06-17
PreviewController - remove list item when accessed from Gallery

### LINKED ISSUE #785: Gallery follow-up: remove list button in gallery detail
Go to Chat -> Chat Detail -> Gallery -> tap any image -> the Gallery Detail (`PreviewController.swift`) view is now showing:

In the bottom right there is an icon showing:
IMG_7F8B30D09FE1-1

The icon is no longer needed and should be removed.

nayooti:
closes #785 

### Observation: 
- On Iphones the the list button appears in a **bottom toolbar**
- On Ipads the list button appears in the **navigation bar** in the top
- All toolbar items are only accessible via subview traversing and can only be modified to a certain degree
- Toolbar at `viewDidLoad` always nil -> so customisation has to be done on `viewWillAppear`

Unfortunately the QL-toolbars are very resistant to customisation.  The solution was:
- traverse search the NavigationBar / BottomToolbar
- in the toolbar find the list item (can be identified by accessibility identifier)
- now: isHidden, remove etc. don't work, so the only solution was to disable it (or remove action), and hide it by setting `tintColor =.clear`

In case PreviewController is inside a navigationController this approach isn't meant to work, so the customisation won't be visible if the PreviewController is accessed from ChatView. 

[Update:]
for iOS 13.4 and lower the bottomToolbar is not customisable, so we cover it by a fake toolbar.

### DIFF
--- a/deltachat-ios/Controller/PreviewController.swift
+++ b/deltachat-ios/Controller/PreviewController.swift
@@ -169,7 +169,9 @@ private extension PreviewController {
         self.nativeToolbar = NativeToolbar(qlToolbar: qlToolbar, qlShareButton: shareButton)
 
         // this will bind native toolbar's alpha to our fake toolbar (used on swipe down events)
-        observerToken = qlToolbar.observe(\.alpha, changeHandler: { [weak self] toolbar, _ in self?.fakeToolbar.alpha = toolbar.alpha})
+        observerToken = qlToolbar.observe(\.alpha, changeHandler: { [weak self] toolbar, _ in
+            self?.fakeToolbar.alpha = toolbar.alpha * 2}  // use factor 2 to avoid real toolbar items shine through
+        )
 
         view.addSubview(fakeToolbar)
         fakeToolbar.translatesAutoresizingMaskIntoConstraints = false
--- a/deltachat-ios/Controller/PreviewController.swift
+++ b/deltachat-ios/Controller/PreviewController.swift
@@ -3,21 +3,21 @@ import UIKit
 
 class PreviewController: QLPreviewController {
 
-    struct QLToolbar {
+    struct QLBottomToolbar {
         let toolbar: UIToolbar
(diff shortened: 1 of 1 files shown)


cyBerta:
unfortuantely this doesn't work reliable on my iPhone SE. In rare cases the list icon disappears but most of the time it's still visible. It seems to be a timing issue or maybe the changes get overwritten again?

nayooti:
> unfortuantely this doesn't work reliable on my iPhone SE. In rare cases the list icon at the bottom disappears but most of the time it's still visible. It seems to be a timing issue or maybe the changes get overwritten again?

Are you accessing via ChatDetail or directly via ChatView?

nayooti:
> unfortuantely this doesn't work reliable on my iPhone SE. In rare cases the list icon at the bottom disappears but most of the time it's still visible. It seems to be a timing issue or maybe the changes get overwritten again?

it works on my Iphone Pro (100%), but on my SE it doesn't neither.

nayooti:
For iOS 13.4 and lower I added a fake-toolbar, that covers the real toolbar and imitates it's behaviour. Quite hacky and I am not 100% happy with this.
But I tested for Ipad, Iphone 13.4, 12.4 and Iphone 13.5 and the list button is gone now, so I think it's good for now.

For future iOS releases we should have an eye on this, but the worst thing that can happen, that the list button will be visible again.

cyBerta:
Hm... this adds quite a lot of code to hide one single button. I think the complexitiy is over the top for what we want to archieve and that it is reasonable to support only 13.5 and newer.
@r10s can you have a look at it, too?

nayooti:
Ok, I give up.

I went back to the state were the list item should have hidden for iOS13 devices and unfortunately the list button is suddenly back. It disappears when start pulling down the image. 

So at this point I would suggest to live with the list button. So I close this for now.

PR #797, opened 2020-06-18
Bugfix: delete contact on active search in NewChatController

### LINKED ISSUE #758: UI Glitch when deleting a contact
steps to reproduce:
- tap new chat
- search for a contact so that at least 2 candidates remain
- swipe the first one to the left and tap on delete

the result is the following
Screen Shot 2020-06-02 at 16 17 42

tapping on (x) resolves the problem, so there's no permanent empty entry

nayooti:
closes #758 

Solution: 
Deleting a contact while active search did not update filteredContacts.

Extra: 
- Using tableView.deleteRow instead of reloadData to get animation
- moved alert-functions into separate code block (attempt to give NewChatController some structure)

### DIFF
--- a/deltachat-ios/Controller/NewChatViewController.swift
+++ b/deltachat-ios/Controller/NewChatViewController.swift
@@ -227,7 +227,7 @@ class NewChatViewController: UITableViewController {
             let delete = UITableViewRowAction(style: .destructive, title: String.localized("delete")) { [weak self] _, _ in
                 guard let self = self else { return }
                 let contactId = self.contactIdByRow(indexPath.row)
-                self.askToDeleteContact(contactId: contactId, context: self.dcContext)
+                self.askToDeleteContact(contactId: contactId, indexPath: indexPath)
             }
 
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/NewChatViewController.swift
```diff
@@ -248,46 +248,21 @@ class NewChatViewController: UITableViewController {
...
-            present(alert, animated: true, completion: nil)
+    // MARK: - updates
+    private func deleteContact(contactId: Int, indexPath: IndexPath) {
+        if dcContext.deleteContact(contactId: contactId) {
+            contactIds = dcContext.getContacts(flags: DC_GCL_ADD_SELF)
+            if isFiltering {
+                filteredContactIds = dcContext.getContacts(flags: DC_GCL_ADD_SELF, queryString: searchText)
+            }
+            tableView.deleteRows(at: [indexPath], with: .automatic)
+            // self.tableView.reloadData()
```
cyBerta, line 261: can be probably removed, right?

Issue #772, opened 2020-06-06
app gets slower when used for some time


r10s:
as if some viewcontrollers or maybe loaded images are not freed or so. or some large images.

@nayooti you have some experiences with the debug tools, can you please have a look at that? and try to reproduce that?

nayooti:
probably #748.

Before @cyberta created this ticket I didn't notice the lagging etc. but suddenly it's quite significant.

I already started working on this, but unfortunately it's not that easy to fix.

r10s:
hm, not sire, might also be another problem, i mean opening a chat after startup is super-fast - later on it gets slower. i would assume that with already generated thumbnails things are faster :)

r10s:
further observations:

this happens only if the app is running for _several hours without being killed_ because of location streaming enabled. 

so, at least, it is not an initial annoyance as normally background fetch is done only every few minutes and the app is killed all the time.

cyBerta:
we have memory leaks. When entering a chat with large data like video messages the used memory increases significantly with each entrering & leaving.

cyBerta:
I've found some missing weak declarations in closures

nayooti:
@cyBerta Do you need any help with this?
This view here is very helpful:
![image](https://user-images.githubusercontent.com/5249565/85014894-60ae3980-b167-11ea-91a7-fb419f13e36e.png)

cyBerta:
I think this one is fixed with #798, I'll close it for now. If it reappears, we can reopen this ticket.

r10s:
yip, i can confirm, things are much better meanwhile :)

PR #798, opened 2020-06-18
Fix memory leaks

cyBerta:
related to #772


nayooti:
Crucial retain cycle was actionHandler of the messageInputBar. Not sure if the other weak-statements etc. were needed.

cyBerta:
great catch! Thanks @nayooti for continuing this task. 
What is now still needed is work on DBDebugKit.

nayooti:
@cyBerta What exactly make you think there are memory leaks in DBDebugToolkit?

cyBerta:
> @cyBerta What exactly make you think there are memory leaks in DBDebugToolkit?

you need to reactive the out commented code in order to see them in the instruments tool.
Bildschirmfoto 2020-06-22 um 12 30 00

cyBerta:
@nayooti DBDebugToolkit only leaks small amouts compared to the fixed leaks in this PR. I think we can merge that in and handle DBDebugToolkit in a separate PR. 
Can you please approve this PR?

PR #799, opened 2020-06-19
set number of lines in fileviews titleview

### LINKED ISSUE #741: Fix MessageLabel numberOfLines
MessageLabel has the property numberOfLines to define the maximum amount of allowed lines. 
For some reason it is impossible to restrict the numberOfLines to 3 while using linkebreakMode `.byTruncatingTail`. The desired effect would be to ellipsize Text if it exceeds the maximum number of lines. Instead, a single ellipsized line is shown.

nayooti:
related to #741 

@cyBerta Could you please have a quick look if this is sufficient. It actually does restrict the number of lines of our document-cells.


cyBerta:
Sorry, I didn't pay enough attention to the issue description. I've updated it to illustrate the problem better. Nevertheless restricting the amount of lines to 3 without ellipsizing is already an improvement.

PR #804, opened 2020-06-21
ChatList - add visual feedback when cell tapped

### LINKED ISSUE #522: Add visual feedback when cells are tapped
Only a few cells provide visual feedback when they are tapped. Maybe subtile but it does make a difference.

nayooti:
closes #522 

Currently some tables/cells provide visual feedback when tapped (->Settings).

Our tables containing ContactCells don't. Adding tapping feedback gives a very subtile feel of liveliness and is actually native behaviour.

I implemented it by setting ContactCells contents background color to nil (so the default `SelectedView` can shine through in selection state). I also realised that we're currently setting some backgroundColor to the cell, which I believe is not needed (contentview covers it completely), so I removed that.

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -206,7 +206,7 @@ class ChatListController: UITableViewController {
         return viewModel.titleForHeaderIn(section: section)
     }
 
-    override func tableView(_: UITableView, didSelectRowAt indexPath: IndexPath) {
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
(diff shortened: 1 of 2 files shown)


nayooti:
> Pinned conversations don't have a different background color anymore (which they should have). It's probably related to the changes here https://github.com/deltachat/deltachat-ios/pull/804/files#diff-b9a1ec185d87b439e1d753d577216c34L278 and here https://github.com/deltachat/deltachat-ios/pull/804/files#diff-b9a1ec185d87b439e1d753d577216c34L278.

Thanks, apparently contentView did not use bg-color at all. Use cell.backgroundColor now. Not sure why I didn't do that in first place :(

PR #809, opened 2020-06-22
save images to photo library 

### LINKED ISSUE #802: Saving images crashes app
Saving a random picture crashes the app on ios.
How to reproduce:
Clicking on picture so its fullscreen, then clicking on the share symbol, then clicking on save image, then app crashes. The iphone is probably out of disk space, not sure if this is part of the problem.

cyBerta:
fixes app crash when saving images to photo library
fixes #802

### DIFF
--- a/deltachat-ios/Info.plist
+++ b/deltachat-ios/Info.plist
@@ -51,7 +51,7 @@
 	<key>NSPhotoLibraryUsageDescription</key>
 	<string>Delta Chat will let you choose which photos from your library to send.</string>
 	<key>NSPhotoLibraryAddUsageDescription</key>
-	<string>Delta Chat uses this permission to save images to your photo library.</string>
+	<string>Delta Chat wants to save images to your photo library.</string>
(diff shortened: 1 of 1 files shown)


nayooti:
I could not trigger the crash on iOS12.2 Simulator.

cyBerta:
> i do not get why this fixes the error "This media format is not supported." reported at [#802 (comment)](https://github.com/deltachat/deltachat-ios/issues/802#issuecomment-647604305) and why/if this was an issue for 12.3 only.
> 
> can you please explain in a few more details? thanks!

The reason was that the save action is shown on iOS 12 unconditionally, but we didn't request the permission to save to the systems photo library. Selecting the save action raised an exception, but the log output https://github.com/deltachat/deltachat-ios/issues/802#issuecomment-647604305 was just misleading. I tested saving an image without this PR on iOS 12 and was able to reproduce the crash. Fixing the permissions solved the issue with the same content I tried to save.

r10s:
thanks for the additional explanations :)

PR #810, opened 2020-06-24
Adaptive font sizes

### LINKED ISSUE #780: adaptive font sizes
currently, some views adapt their font sizes according to the preferred system settings and others not
this leads to an ugly UI and is of course also an accessibility issue
to test the UI with increased fonts: tap system settings -> Display & Brightness -> Text Size

cyBerta:
closes #780 

* This PR touches most UI parts of the project. UIMessageKit has been untouched though, also because we will replace it soon. Moreover it supports on a basic level adaptive fonts.

* There has been implemented BasicCells as a replacement for normal UITableViewCells which react unreliably on multiple subsequent font size changes leading to broken layouts.
* TextFieldCells have been heavily refactored. Both BasicCells and TextFieldCells change their layout depending on the font size. For extra large fonts available from the accessibility settings they use a two line layout (title above, value below).
* Most cells are now self-sizing with the exception of ContactCells. This can be done in a subsequent PR.
* The document gallery yview and the Welcome screen are untouched yet as well and at least the document gallery view can be adapted in another iteration.

* I added comments to the PR in order to guide a little bit


deltachat-ios/Controller/AccountSetupController.swift
```diff
@@ -79,7 +79,6 @@ class AccountSetupController: UITableViewController, ProgressAlertHandler {
         cell.tag = tagEmailCell
         cell.textField.addTarget(self, action: #selector(emailCellEdited), for: .editingChanged)
         cell.textField.tag = tagTextFieldEmail // will be used to eventually show oAuth-Dialogue when pressing return key
-        cell.textField.delegate = self
```
cyBerta, line 82: -> the delegate is already passed on line 78

deltachat-ios/Controller/AccountSetupController.swift
```diff
@@ -102,9 +100,9 @@ class AccountSetupController: UITableViewController, ProgressAlertHandler {
         return cell
     }()
 
-    private lazy var restoreCell: UITableViewCell = {
-        let cell = UITableViewCell(style: .value1, reuseIdentifier: nil)
-        cell.textLabel?.text = String.localized("import_backup_title")
+    private lazy var restoreCell: BasicCell = {
```
cyBerta, line 103: All normal UITableViewCell have been replaced by a custom cell (BasicCell) that resizes properly if the user changes the system font size.

deltachat-ios/Controller/AddGroupMembersViewController.swift
```diff
@@ -153,20 +154,6 @@ class AddGroupMembersViewController: GroupMembersViewController {
         navigationController?.popViewController(animated: true)
     }
 
-    func getNewContactCell() -> UITableViewCell {
```
cyBerta, line 156: method removed in favor of a lazy variable containing an ActionCell, streamlining to have more consistency with other static cells.

deltachat-ios/Controller/AddGroupMembersViewController.swift
```diff
@@ -99,20 +107,13 @@ class AddGroupMembersViewController: GroupMembersViewController {
...
     override func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
-        switch indexPath.section {
-        case sectionNewContact:
-            return Constants.defaultCellHeight
-        case sectionMemberList:
-            return ContactCell.cellHeight
-        default:
-            return Constants.defaultCellHeight
-        }
+        return indexPath.section == sectionMemberList ? ContactCell.cellHeight : UITableView.automaticDimension
```
cyBerta, line 110: generally all ViewControllers should avoid `tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath)` if we want to have self-sizing cells. The only exception for now is ContactCell. Returning `UITableView.automaticDimension` indicates that iOS will respect the self-sizing of cells

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -169,6 +166,12 @@ class ChatListController: UITableViewController {
         updateTitle()
     }
 
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
```
cyBerta, line 169: all tableViewControllers and UITableViewCells can react on font size changes with the help of `override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?)`

deltachat-ios/Controller/GroupChatDetailViewController.swift
```diff
@@ -237,7 +244,12 @@ class GroupChatDetailViewController: UIViewController {
     }
 
     private func showQrCodeInvite(chatId: Int) {
-        let qrInviteCodeController = QrInviteViewController(dcContext: dcContext, chatId: chatId)
+        var hint = ""
+        let dcChat = dcContext.getChat(chatId: chatId)
+        if !dcChat.name.isEmpty {
+            hint = String.localizedStringWithFormat(String.localized("qrshow_join_group_hint"), dcChat.name)
+        }
+        let qrInviteCodeController = QrViewController(dcContext: dcContext, chatId: chatId, qrCodeHint: hint)
```
cyBerta, line 252: I refactored QRViewController in order to be able to dump QRInviteViewController, which was hard to adjust for dynamic font sizes. Now the hint below the QR code has to be injected.

deltachat-ios/Controller/QrViewController.swift
```diff
@@ -6,13 +6,7 @@ class QrViewController: UIViewController {
 
     private let dcContext: DcContext
 
-    private var contact: DcContact? {
```
cyBerta, line 9: DcContact was only needed to create the hint message below the QR code, which is now injected in instanciation

deltachat-ios/Extensions/Extensions.swift
```diff
@@ -84,3 +84,13 @@ extension UITableView {
         }
     }
 }
+
+extension UIFont {
+    static func preferredFont(for style: TextStyle, weight: Weight) -> UIFont {
```
cyBerta, line 89: extension method that allows to add a weight to a dynamic system font, used in ChatListViewController to keep a medium font weight for the chat names

deltachat-ios/Extensions/UIView+Extensions.swift
```diff
@@ -33,20 +33,36 @@ internal extension UIView {
         self.layer.borderWidth = 2
     }
 
-    func alignLeadingToAnchor(_ anchor: NSLayoutXAxisAnchor, paddingLeading: CGFloat = 0.0) {
-        self.leadingAnchor.constraint(equalTo: anchor, constant: paddingLeading).isActive = true
+    func alignLeadingToAnchor(_ anchor: NSLayoutXAxisAnchor, paddingLeading: CGFloat = 0.0, priority: UILayoutPriority? = .none) {
```
cyBerta, line 36: priorities have been added in order to avoid much boilerplate while adapting self-sizing layouts

deltachat-ios/View/BasicCell.swift
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        if previousTraitCollection?.preferredContentSizeCategory !=
+            traitCollection.preferredContentSizeCategory {
+            updateViews()
+        }
+    }
+
+    public func updateViews() {
```
cyBerta, line 79: BasicCells can show titles and values. For very large font sizes, values are shown below the title in order to be able to read both title and value

deltachat-ios/View/ContactCell.swift
```diff
@@ -136,6 +148,25 @@ class ContactCell: UITableViewCell {
...
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        if previousTraitCollection?.preferredContentSizeCategory !=
+            traitCollection.preferredContentSizeCategory {
+            configureCompressionPriority()
+        }
+    }
+
+
+    private func configureCompressionPriority() {
+        if isLargeText {
```
cyBerta, line 163: For very large font sizes (larger than normal scale in system font size settings, avaliable only via accessibility settings),  the timelabel can be overlapped by the chat title. for small to large font sizes the time label always overlapps the chat title if there's not enough place for both. This was also the default before

deltachat-ios/View/ContactCell.swift
```diff
@@ -191,6 +230,16 @@ class ContactCell: UITableViewCell {
     }
 
     func setStatusIndicators(unreadCount: Int, status: Int, visibility: Int32, isLocationStreaming: Bool, isMuted: Bool) {
+        if isLargeText {
```
cyBerta, line 233: for very large fonts (available from accessibilty menu) only the unread message counter will be shown. all other icons are to small anyways for visually impaired ppl and reduce the amount of readable characters of the chat title and the subtitle (last message)

deltachat-ios/View/ContactCell.swift
```diff
@@ -305,6 +354,23 @@ class ContactCell: UITableViewCell {
                                 visibility: 0,
                                 isLocationStreaming: false,
                                 isMuted: false)
+        case .profile:
```
cyBerta, line 357: the contact cell now also handles the profile cell in the settings.

deltachat-ios/View/TextFieldCell.swift
```diff
@@ -83,13 +137,39 @@ class TextFieldCell: UITableViewCell {
...
+
+    override func prepareForReuse() {
+        super.prepareForReuse()
+        title.text = nil
+        title.attributedText = nil
+        textField.text = nil
+    }
+
+    private func updateViews() {
+        if fontSize <= maxFontSizeHorizontalLayout {
```
cyBerta, line 155: for large font sizes the text field is below the text title in order to have enough space to read both and/or write into the text field

deltachat-ios/View/TextFieldCell.swift
```diff
@@ -1,37 +1,87 @@
...
 
         // see: https://stackoverflow.com/a/35903650
         // this makes the textField respect the trailing margin of
         // the table view cell
         selectionStyle = .none
         setupViews()
-        textField.delegate = delegate
+        textFieldDelegate = delegate
         textField.placeholder = placeholder
+        preferValue = false
```
cyBerta, line 84: in case there is not enough space for both title and text field content, the title is preferred by default. this was also the default before.

deltachat-ios/View/TextFieldCell.swift
```diff
@@ -121,7 +201,72 @@ class TextFieldCell: UITableViewCell {
...
+
+    func textFieldShouldReturn(_ textField: UITextField) -> Bool {
+        if let delegate = textFieldDelegate, let result = delegate.textFieldShouldReturn?(textField) {
+            return result
+        }
+        return true
+    }
+
+    func textFieldShouldBeginEditing(_ textField: UITextField) -> Bool {
+        preferValue = true
```
cyBerta, line 244: if the text field was focussed it gains preference over the title until it is unfocussed again.
This way it is easier to read the whole value string while typing.

deltachat-ios/View/TextFieldCell.swift
```diff
@@ -121,7 +201,72 @@ class TextFieldCell: UITableViewCell {
...
+        if let delegate = textFieldDelegate, let result = delegate.textFieldShouldBeginEditing?(textField) {
+            return result
+        }
+        return true
+    }
+
+    func textFieldShouldEndEditing(_ textField: UITextField) -> Bool {
+        preferValue = false
```
cyBerta, line 252: if the text field looses the focus the title is again preferred over the value

cyBerta:
Since this PR is HUGE, I added some explanatory comments

deltachat-ios/ViewModel/ContactCellViewModel.swift
```diff
@@ -59,6 +60,32 @@ class ContactCellViewModel: AvatarCellViewModel {
     }
 }
 
+class ProfileViewModell: AvatarCellViewModel {
```
nayooti, line 63: :)

deltachat-ios/Controller/AccountSetupController.swift
r10s: replacing `UITableViewCell.textLabel` by `BasicCell.title.text` removes knowledge about the _semantic_ of that field. same for replacing `detailTextLabel` by `value.text` and maybe other occurances. 

this remove of _semantic_ may introduce additional/new accessibility issues as readers etc. does not have an idea about what the custom text fields are.

also, already now, the change introduce a visual difference (different gray levels), so that our settings appear different from the system ones.

also, if with some os update, displays things differently, we would need to adapt our BasicCell.

finally, i just checked that, the `UITableViewCell.textLabel` and `detailTextLabel` already use adaptive-font sizes, so there is no need to change that.

cyBerta: @r10s unfortunately UITableViewCell.textlabel and detailTextLabel are performing very poorly on iOS 13.5 wrt adaptive font-sizes which is why I introduced BasicCell. Just try to switch between scaling up and down the font sizes several times and switch in between to the app with a view containing UITableViewCells. The layout breaks.

cyBerta: I think the point about the accessibility issues are a very good point, indeed, which needs to be fixed. I'll checkout the behavior of the screen reader for BasicCells. 
Wrt the colors, we _can_ check if there are system colors we use. However, we use custom colors all over the place. Also the behavior of BasicCells already differs from normal UITableViewCells (2 line layout for huge fonts). So it's not ment to be a default system cell.

cyBerta:
> in general, a good improvement, things look much nicer when using larger fonts, great 👍
> 
> however, as detailed below, we should not remove semantics, therefore i would remove the BasicCell or use it only where really needed.

Let's have an a/v :). I tested the cells with the screen reader and I didn't find any difference in their VoiceOver behaviour from default UITableViewCells.

deltachat-ios/Controller/AddGroupMembersViewController.swift
```diff
@@ -13,6 +13,14 @@ class AddGroupMembersViewController: GroupMembersViewController {
         return button
     }()
 
+    lazy var newContactCell: ActionCell = {
```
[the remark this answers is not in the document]
cyBerta: fixed

cyBerta:
I addressed @r10s concerns about a custom UITableViewCell implementation to work around an iOS bug that is at least reproducible on a iPhone SE (iOS 13.5.1) but obviously not on other phones like iPhone 7 by undoing the BasicCell 
implementation. With that, also @nayooti's comments about BasicCell implementation details became deprected.

deltachat-ios/Controller/QrViewController.swift
```diff
@@ -21,14 +15,19 @@ class QrViewController: UIViewController {
...
-        let view = QrViewContentView(contact: contact, qrCode: qrCode)
+        let qrCode = dcContext.getSecurejoinQr(chatId: chatId)
+        let view = QrViewContentView(qrCode: qrCode, hint: qrCodeHint)
         view.translatesAutoresizingMaskIntoConstraints = false
         return view
     }()
 
-    init(dcContext: DcContext) {
+    private var qrCodeHint: String
+    private var chatId: Int
```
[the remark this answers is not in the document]
cyBerta: fixed

deltachat-ios/Controller/ContactDetailViewController.swift
```diff
@@ -106,6 +106,8 @@ class ContactDetailViewController: UITableViewController {
             title: String.localized("global_menu_edit_desktop"),
             style: .plain, target: self, action: #selector(editButtonPressed))
         self.title = String.localized("tab_contact")
+        tableView.sectionHeaderHeight =  UITableView.automaticDimension
```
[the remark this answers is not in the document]
cyBerta: fixed

deltachat-ios/Controller/AccountSetupController.swift
```diff
@@ -778,9 +776,9 @@ class AccountSetupController: UITableViewController, ProgressAlertHandler {
     }
 
     private func initSelectionCells() {
-        smtpSecurityCell.detailTextLabel?.text = SecurityConverter.convertHexToString(type: .SMTPSecurity, hex: dcContext.getSmtpSecurity())
-        imapSecurityCell.detailTextLabel?.text = SecurityConverter.convertHexToString(type: .IMAPSecurity, hex: dcContext.getImapSecurity())
-        certCheckCell.detailTextLabel?.text = CertificateCheckController.ValueConverter.convertHexToString(value: dcContext.certificateChecks)
+        smtpSecurityCell.textLabel?.text = SecurityConverter.convertHexToString(type: .SMTPSecurity, hex: dcContext.getSmtpSecurity())
+        imapSecurityCell.textLabel?.text = SecurityConverter.convertHexToString(type: .IMAPSecurity, hex: dcContext.getImapSecurity())
+        certCheckCell.textLabel?.text = CertificateCheckController.ValueConverter.convertHexToString(value: dcContext.certificateChecks)
```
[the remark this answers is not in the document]
cyBerta: Thanks! fixed

deltachat-ios/Controller/NewContactController.swift
```diff
@@ -32,8 +32,8 @@ class NewContactController: UITableViewController {
         self.dcContext = dcContext
         cells = [emailCell, nameCell]
         super.init(style: .grouped)
-        emailCell.textField.delegate = self
-        nameCell.textField.delegate = self
+        emailCell.textFieldDelegate = self
+        nameCell.textFieldDelegate = self
```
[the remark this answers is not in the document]
cyBerta: `TextFieldCell`s are already `UITextFieldDelegate`s of its internally used `UITextField`. They handle the UI behavior if its `UITextField` gets and looses the focus.  
To set a delegate from outside the cell has now an own delegate property which is handled in https://github.com/deltachat/deltachat-ios/pull/810/files#diff-4dd8cf8df767fe20c99b259c28e71dbfR209

cyBerta:
> the font-size in the chatlist is now slightly larger, or? is this on-purpose or just a side-effect?

yes, you're right the subtitle and the time label are slightly bigger now. They use now the system's preferred text height for subheadlines (https://github.com/deltachat/deltachat-ios/pull/810/files#diff-b9a1ec185d87b439e1d753d577216c34R107). It sounded reasonable and according to the style guidelines somehow.
Generally we a have little bit less flexiiblity using system preferred font sizes than if we hardcode them. The set of available font sizes is restricted. However we could check if other categories instead of `.subheadline` fit extactly the 14 dip(?) defined before.

deltachat-ios/Controller/NewContactController.swift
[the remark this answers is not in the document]
r10s: thanks for explanation!

PR #815, opened 2020-06-25
Gallery: video thumbnails are generated in background thread

### LINKED ISSUE #792: Async thumbnail loading in the gallery
follow up for #748

nayooti:
closes #792

### DIFF
--- a/deltachat-ios/View/Cell/GalleryCell.swift
+++ b/deltachat-ios/View/Cell/GalleryCell.swift
@@ -53,7 +53,16 @@ class GalleryCell: UICollectionViewCell {
             imageView.image = msg.image
             playButtonView.isHidden = true
         case .video:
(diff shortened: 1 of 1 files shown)


cyBerta:
It doesn't fix the lagging completely. What works nicely is the Image cache, but on first loading scrolling is not smooth.
I think a reason could be that the loading jobs are not yet finished, but the cells are already about to be recycled. 
Invalidating of an image is not implemented in the current UIView extension.

cyBerta:
It becomes very obvious if you have many videos right after each other in the gallery.

nayooti:
Interesting catch. But definitely makes sense. I‘ll look into this.

nayooti:
So, I tested it and on my device (Iphone SE) the scrolling is very smooth. You can see the thumbnails appear with delay but it does not affect scrolling at all. Also opening the gallery is instant.

nayooti:
I was actually currently working on this PR.

PR #826, opened 2020-07-07
fix disappearing email and password strings during account setup

### LINKED ISSUE #825: account name and credentials are deleted if user enters any advanced account settings
When setting up an account, the account name and credentials get deleted if the user enters any settings subscreen from `advanced`. 
Reentering the screen should not generally result in the deletion of the email address and password. 
This issue was probably introduced by the attempt to fix  #615

cyBerta:
fixes #825

### DIFF
--- a/deltachat-ios/Controller/AccountSetupController.swift
+++ b/deltachat-ios/Controller/AccountSetupController.swift
@@ -660,8 +660,6 @@ class AccountSetupController: UITableViewController, ProgressAlertHandler {
                 } else if ui["done"] as! Bool {
                     self.dcContext.maybeStartIo()
                     self.updateProgressAlertSuccess(completion: self.handleLoginSuccess)
-                    self.emailCell.setText(text: nil)
-                    self.passwordCell.setText(text: nil)
                 } else {
                     self.updateProgressAlertValue(value: ui["progress"] as? Int)
                 }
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/AccountSetupController.swift
```diff
@@ -666,6 +660,8 @@ class AccountSetupController: UITableViewController, ProgressAlertHandler {
                 } else if ui["done"] as! Bool {
                     self.dcContext.maybeStartIo()
                     self.updateProgressAlertSuccess(completion: self.handleLoginSuccess)
+                    self.emailCell.setText(text: nil)
```
nayooti, line 663: Why exactly is this needed?

I find it a bit weird to see the credential fields to be cleared right before being redirected.
If the clearing is necessary I think we can clear it in `handleLoginSuccess `, so it will be executed in the completion handler.

nayooti:
Somehow I only get this error on my Main-Device - and not on my Dev-Device.

deltachat-ios/Controller/AccountSetupController.swift
cyBerta: Thanks, it was indeed not needed.

r10s:
this is also tested on some device? :)

PR #834, opened 2020-07-09
Async gallery follow up

nayooti:
follow up for #815 

I enabled CollectionView's `prefetching`-feature for our gallery, to enhance scrolling performance.

Since results needs to be stored/cached in the ViewController (not in the cell) I created a `GalleryItem that manages all the fetching + data. The GalleryItem is passed to the cell before display. 

The cell can use the `galleryItem.thumbnailImage` if already loaded or register for the incoming thumbnails.

For better testing: our ThumbnailCache has a clear-method, which you can use on GalleryViewController.viewWillAppear.

### DIFF
--- a/deltachat-ios/Controller/GalleryViewController.swift
+++ b/deltachat-ios/Controller/GalleryViewController.swift
@@ -253,7 +253,7 @@ class GalleryItem {
         }
     }
 
-    func loadThumbnail() {
+    private func loadThumbnail() {
         guard let viewtype = msg.viewtype, let url = msg.fileURL else {
             return
         }
--- a/deltachat-ios/Controller/GalleryViewController.swift
+++ b/deltachat-ios/Controller/GalleryViewController.swift
@@ -271,10 +271,15 @@ class GalleryItem {
     }
 
     private func loadGifThumbnail(from url: URL) {
-        guard let imageData = try? Data(contentsOf: url) else {
-            return
+        DispatchQueue.global(qos: .background).async {
+            guard let imageData = try? Data(contentsOf: url) else {
+                return
+            }
+            let thumbnailImage = SDAnimatedImage(data: imageData)
+            DispatchQueue.main.async { [weak self] in
+                self?.thumbnailImage = thumbnailImage
+            }
         }
-        self.thumbnailImage = SDAnimatedImage(data: imageData)
     }
 
     private func loadVideoThumbnail(from url: URL) {
--- a/deltachat-ios/Controller/GalleryViewController.swift
+++ b/deltachat-ios/Controller/GalleryViewController.swift
@@ -5,6 +5,7 @@ class GalleryViewController: UIViewController {
 
     // MARK: - data
     private let mediaMessageIds: [Int]
+    private var items: [Int: GalleryItem] = [:]
 
     // MARK: - subview specs
     private let gridDefaultSpacing: CGFloat = 5
@@ -26,6 +27,8 @@ class GalleryViewController: UIViewController {
         collection.backgroundColor = DcColors.defaultBackgroundColor
         collection.delaysContentTouches = false
         collection.alwaysBounceVertical = true
+        collection.isPrefetchingEnabled = true
+        collection.prefetchDataSource = self
         return collection
     }()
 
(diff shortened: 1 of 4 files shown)


r10s:
> I enabled CollectionView's prefetching-feature for our gallery, to enhance scrolling performance.

tbh, i cannot see any difference on my quite old iphone7 with the rather large xyiv-gallery - so, not sure, if this is worth additional caching.

nayooti:
Currently there is a slight hanging if you mix Gifs and videos. @Cyberta mentioned this in previous PR. I could reproduce the hanging  on newest IPhone Pro.

Also since UIImageView belongs to a cell that gets eventually recycled while generating a thumbnail. I had case where the UIImageView generated more thumbnails asynchronously at the same time ( and I think there is no guarantee which thumbnail will finally be displayed. 

The cache is managed by the system and is cleared once memory is needed.

r10s:
> Currently there is a slight hanging if you mix Gifs and videos. @cyBerta mentioned this in previous PR.

k, i the pr fixes that, i am fine with it :) cc @cyBerta can you please re-check as well and have a look at the code? great!

cyBerta:
This PR creates thumnails for gifs along with the play button as if they were videos. It's indeed scrolling nicely, however I think it's much nicer to have playing gifs as it was before. Actually sdwebimage should be able to load a gif async without a ImageView as well: 
```
if let gifData = try? Data(contentsOf: url) {
        return SDAnimatedImage(data: gifData)
 }
```

deltachat-ios/View/Cell/GalleryCell.swift
```diff
@@ -42,32 +50,13 @@ class GalleryCell: UICollectionViewCell {
...
-            } else {
-                imageView.loadVideoThumbnail(from: fileUrl, placeholderImage: nil) { thumbnail in
-                    if let image = thumbnail {
-                        ThumbnailCache.shared.storeImage(image: image, key: key)
-                    }
-                }
-            }
-            playButtonView.isHidden = false
-        case .gif:
-            imageView.sd_setImage(with: fileUrl, placeholderImage: nil)
```
cyBerta, line 66: instead of working on the imageView we should try to load the gif with the help of `SDAnimatedImage(data: gifData)`

deltachat-ios/Controller/GalleryViewController.swift
```diff
@@ -197,3 +217,112 @@ extension GalleryViewController {
...
+            return true
+        default:
+            return false
+        }
+    }
+
+    init(msgId: Int) {
+        self.msg = DcMsg(id: msgId)
+
+        if let key = msg.fileURL?.absoluteString, let image = ThumbnailCache.shared.restoreImage(key: key) {
```
cyBerta, line 249: ThumnailCache would have to accept `SDAnimatedImage` as values instead of `UIImage`s. `SDAnimatedImage` can be used on any UIImageView and inherits from UIImage

deltachat-ios/Controller/GalleryViewController.swift
```diff
@@ -197,3 +217,112 @@ extension GalleryViewController {
...
+                }
+            }
+        }
+    }
+
+    private func gifThumbnail(url: URL) -> UIImage? {
+        return getGifSequence(url: url)?.first ?? nil
+    }
+
+    private func getGifSequence(url: URL) -> [UIImage]? {
```
cyBerta, line 300: we wouldn't need that if we were using SDAnimatedImage

deltachat-ios/Helper/ThumbnailCache.swift
```diff
@@ -18,4 +18,8 @@ class ThumbnailCache {
     func restoreImage(key: String) -> UIImage? {
         return cache.object(forKey: NSString(string: key))
     }
+
+    func clearCache() {
```
cyBerta, line 22: This method is unused

deltachat-ios/Controller/GalleryViewController.swift
nayooti: Thanks for the input. I changed that (had to replace the `UIImageView` by `SDAnimatedImageView`.

cyBerta:
There was a crash when tapping on the title in a chat view. Maybe it was related to rebasing? Anyways, I rebased again and it vanished. I've also put loading of gifs to the background which improves the scrolling a little bit more.

PR #849, opened 2020-07-15
tweak location streaming indicator icon:

cyBerta:
reduce transparent space to the left while taking care of aspect ratio

before: 
location_icn_old

after:
location_icn_new

the cut out parts differ a little bit, the height is kept in reality the same, the lower image just appears bigger because of the auto-scaling of the cut and paste feature of mac's preview tool.


cyBerta:
> this pr contains the submodule, that should be removed. apart from that, this lgtm.

Thanks! I fixed that.

PR #854, opened 2020-07-16
show name and address below qr-code

r10s:
this pr shows the name _and_ the address below the qr-invite-code; previously, only the address was shown.

### DIFF
--- a/deltachat-ios/Controller/QrPageController.swift
+++ b/deltachat-ios/Controller/QrPageController.swift
@@ -12,10 +12,18 @@ class QrPageController: UIPageViewController, ProgressAlertHandler {
     private lazy var qrCodeHint: String = {
         var qrCodeHint = ""
         if dcContext.isConfigured() {
-            let contact = DcContact(id: Int(DC_CONTACT_ID_SELF))
+            // we cannot use dc_contact_get_displayname() as this would result in "Me" instead of the real name
+            let name = dcContext.getConfig("displayname") ?? ""
+            let addr = dcContext.getConfig("addr") ?? ""
+            var nameAndAddress = ""
+            if name.isEmpty {
+                nameAndAddress = addr
+            } else {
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/QrPageController.swift
```diff
@@ -12,10 +12,18 @@ class QrPageController: UIPageViewController, ProgressAlertHandler {
...
         if dcContext.isConfigured() {
-            let contact = DcContact(id: Int(DC_CONTACT_ID_SELF))
+            // we cannot use dc_contact_get_displayname() as this would result in "Me" instead of the real name
+            let name = dcContext.getConfig("displayname") ?? ""
+            let addr = dcContext.getConfig("addr") ?? ""
+            var nameAndAddress = ""
+            if name.isEmpty {
+                nameAndAddress = addr
+            } else {
+                nameAndAddress = name + " (" + addr + ")"
```
cyBerta, line 22: you could use a imo little bit 'swiftier' syntax: 
`nameAndAddress = "\(name) (\(addr))"`

nayooti: When changing the profile name the qrCodeHint does not get updated. We should remove `lazy` and make this a computed property.

cyBerta: (the syntax highlighting in xcode makes the parenthesis around addr better readable)

nayooti:
There is a bug.
When changing the profile name the qrCodeHint does not get updated. We should remove lazy and make this a computed property.

r10s:
i've turned it into a normal getter-function, in fact, i've overseen this part. in general i am not a fan of variables computed, sometimes computed etc. it is easily overseen, what really happens.

but anyway, this does not solve the problem, the view is not updated on name changes - and this does not only affect the text but also the qr-code (that also contains the name that is shown to the user then)

PR #856, opened 2020-07-16
update QRViewController on profile name change

### LINKED ISSUE #855: qr-code is not updated on name changes
the invite-qr-code contains the name of the inviter, however, currently, this qr-code is not updated immediately on name changes (it is updated on restarts, however).

same for the name shown below the qr-code as added with #854 - in fact, the whole view seems to need an update. maybe just do not cache it.

nayooti:
closes #855 

The hintText in QRViewController is updated on viewWillAppear (only) if there was a change.

### DIFF
--- a/deltachat-ios/Controller/QrPageController.swift
+++ b/deltachat-ios/Controller/QrPageController.swift
@@ -9,7 +9,7 @@ class QrPageController: UIPageViewController, ProgressAlertHandler {
 
     private var selectedIndex: Int = 0
 
-    private var getQrCodeHint: String {
+    private var qrCodeHint: String {
         var qrCodeHint = ""
         if dcContext.isConfigured() {
             // we cannot use dc_contact_get_displayname() as this would result in "Me" instead of the real name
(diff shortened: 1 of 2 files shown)


deltachat-ios/Controller/QrViewController.swift
```diff
@@ -115,7 +126,7 @@ class QrViewContentView: UIView {
 
     init(qrCode: String?, hint: String) {
         super.init(frame: .zero)
-        hintLabel.text = hint
+        hintText = hint
```
cyBerta, line 129: why this change? hintText is used only internally if I understand it correctly and therefore it adds only boilerplate code without benefits.

deltachat-ios/Controller/QrPageController.swift
```diff
@@ -9,7 +9,7 @@ class QrPageController: UIPageViewController, ProgressAlertHandler {
 
     private var selectedIndex: Int = 0
 
-    private func getQrCodeHint() -> String {
+    private var getQrCodeHint: String {
```
r10s: no blocker, this is apple-swift-style, i know. but _i_ find that somewhat confusing, when such a thing is _accessed_, you simply do not know, by just seeing the variable name, what actually happens.

therefore, i prefer just functions.

don't make me think.

nayooti:
This should do it now.

PR #861, opened 2020-07-16
Enlarge profile image

### LINKED ISSUE #831: Enlarge profile image
Profile images are currently shown in a small-scaled fashion as avatars. As an enhancement the user should be able to enlarge the profile image if they tap on the avatar in a contact or group detail controller. The easiest way would be to open a PreviewController showing the enlarged avatar image. This is how it's done on Android.

nayooti:
closes #831 

In GroupsDetail  or in ChatDetail-> tap on Avatar -> if there is a group/contact image the PreviewController displays that.

- I had to alter our core-functions to get profileImageUrls -> because they are needed in PreviewController
- by default PreviewController if there only one url (like here) displays the URL as title which is pretty ugly -> to display a title I had to **subclass QLPreviewItem**. 

Issues:
It seems that the contact avatar images have very low resolution, true? If so, we should save some backup image.

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -759,12 +759,12 @@ public class DcChat {
 	}()
 
 	public var profileImageURL: URL? {
-		guard let cString = dc_chat_get_profile_image(chatPointer) else { return nil }
-		let filename = String(cString: cString)
-		dc_str_unref(cString)
-		let path: URL = URL(fileURLWithPath: filename, isDirectory: false)
-		return path
-	}
+        guard let cString = dc_chat_get_profile_image(chatPointer) else { return nil }
+        let filename = String(cString: cString)
+        dc_str_unref(cString)
+        let path: URL = URL(fileURLWithPath: filename, isDirectory: false)
+        return path
+    }
 
     public var isSendingLocations: Bool {
         return dc_chat_is_sending_locations(chatPointer) == 1
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -760,10 +760,10 @@ public class DcChat {
 
 	public var profileImageURL: URL? {
 		guard let cString = dc_chat_get_profile_image(chatPointer) else { return nil }
-        let filename = String(cString: cString)
-        dc_str_unref(cString)
-        let path: URL = URL(fileURLWithPath: filename, isDirectory: false)
-        return path
+		let filename = String(cString: cString)
+		dc_str_unref(cString)
+		let path: URL = URL(fileURLWithPath: filename, isDirectory: false)
+		return path
 	}
 
     public var isSendingLocations: Bool {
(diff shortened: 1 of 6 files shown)


DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -756,7 +756,15 @@ public class DcChat {
             }
         }
         return nil
-        }()
+	}()
+
+	public var profileImageURL: URL? {
+		guard let cString = dc_chat_get_profile_image(chatPointer) else { return nil }
```
nayooti, line 762: somehow this alignment is broken - on my machine it is actually alright? Not sure what's wrong here?

deltachat-ios/Controller/ContactDetailViewController.swift
```diff
@@ -6,15 +6,16 @@ class ContactDetailViewController: UITableViewController {
     private let viewModel: ContactDetailViewModel
 
     private lazy var headerCell: ContactDetailHeader = {
-        let cell = ContactDetailHeader()
```
nayooti, line 9: just some renaming to be more consistent with GroupDetailHeader

deltachat-ios/Extensions/UIImage+Extension.swift
```diff
@@ -12,29 +12,34 @@ extension UIImage {
...
-
-         if let cgImage = context.makeImage() {
-             let coloredImage = UIImage(cgImage: cgImage)
-             return coloredImage
-         } else {
-             return nil
-         }
-     }
-
+    func maskWithColor(color: UIColor) -> UIImage? {
```
nayooti, line 15: alignment was off

DcCore/DcCore/DC/Wrapper.swift
r10s: seems as if you're using a _tab_ instead of _4 spaces_ for indentation. tabs are rendered as 8 spaces on github - and probably as 4 spaces on your machine.

just use spaces as elsewhere in the file and things should be fine :)

r10s:
> It seems that the contact avatar images have very low resolution, true? If so, we should save some backup image.

that would not help as most profile-images come from extern. but we might increase the resolution in core, maybe depending on "outgoing media quality", however, there is noting ios can do on that, it is totally an core thing (which is great, and by design :)

DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -756,7 +756,15 @@ public class DcChat {
             }
         }
         return nil
-        }()
+	}()
+
+	public var profileImageURL: URL? {
```
r10s, line 761: still wrong indentation

DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -1099,6 +1107,13 @@ public class DcContact {
         return nil
     }()
 
+	public var profileImageURL: URL? {
```
r10s, line 1110: still wrong indentation

r10s:
https://github.com/deltachat/deltachat-ios/pull/861/commits/4f1136a695630602d3075dcb7e21c81020d012d4 did not fix indentation but adds even more tabs instead of spaces. i'll do a fix, however, for the future, it makes sense if you double check that, seems as if your settings go wild from time to time :)

PR #862, opened 2020-07-17
groupDetail / chatDetail -> background colors

### LINKED ISSUE #838: tweak profile background colors
- the background of the shared-chats and of the member-list should be the same background than a button item.  
this is correct for light-member-list, dark-member-list, light-shared-chats but _not correct_ for dark-shared-chats

- the overall header with the avatar in the first row is currently a header afaik and should have the same background color than used below the other headers. this is correct in dark-mode, but not correct in light-mode.

having #831 (enlarge profile images) in mind: if it is easier, i would also be fine with turning the avatar-header to a normal row (if this is not yet the case), iirc, clicks in headers are complicated.

when tweaking, maybe just add an option or so, the items are used at many places and it is so easy to fix this issue and open a new one in another list :)

nayooti:
closes #838 

IMG_E8137DABEF65-1

IMG_5483

Unfortunately there is no defined background color for our actionCells (not sure what makes them gray in darkmode). I had to use the xcode's color tool to grab the color. It looks alright though.

I discovered **color-literals** lately and I added the new color as such. I find them very handy in Xcode, so maybe we want to use them for our color definitions.

### DIFF
--- a/DcCore/DcCore/Helper/DcColors.swift
+++ b/DcCore/DcCore/Helper/DcColors.swift
@@ -1,8 +1,9 @@
 import UIKit
 
 public struct DcColors {
-    private static let white						= #colorLiteral(red: 1, green: 1, blue: 1, alpha: 1)
-    private static let actionCellBackgroundDark		= #colorLiteral(red: 0.1031623408, green: 0.1083367988, blue: 0.1185036376, alpha: 1)
+    private static let white                                  = #colorLiteral(red: 1, green: 1, blue: 1, alpha: 1)
+    private static let actionCellBackgroundDark               = #colorLiteral(red: 0.1031623408, green: 0.1083367988, blue: 0.1185036376, alpha: 1)
+
 
     public static let primary = UIColor.systemBlue
     public static let colorDisabled = UIColor.themeColor(light: UIColor(white: 0.9, alpha: 1), dark: UIColor(white: 0.2, alpha: 1))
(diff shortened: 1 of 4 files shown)


deltachat-ios/ViewModel/ContactDetailViewModel.swift
```diff
@@ -102,7 +102,6 @@ class ContactDetailViewModel {
         let chatId = sharedChats.getChatId(index: index)
         let summary = sharedChats.getSummary(index: index)
         let unreadMessages = context.getUnreadMessages(chatId: chatId)
-
```
r10s, line 105: is that empty line removed on-purpose?

DcCore/DcCore/Helper/DcColors.swift
```diff
@@ -1,6 +1,9 @@
 import UIKit
 
 public struct DcColors {
+    private static let white						= #colorLiteral(red: 1, green: 1, blue: 1, alpha: 1)
```
r10s, line 4: this is also a tab, no spaces, no big deal here, however @nayooti please fix your environment :)

DcCore/DcCore/Helper/DcColors.swift
```diff
@@ -9,6 +12,7 @@ public struct DcColors {
                                                           dark: UIColor.init(hexString: "333333"))
     public static let contactCellBackgroundColor = UIColor.themeColor(light: .white, dark: .black)
     public static let defaultBackgroundColor = UIColor.themeColor(light: .white, dark: .black)
+	public static let sharedChatCellBackgroundColor = UIColor.themeColor(light: white, dark: actionCellBackgroundDark)
```
r10s, line 15: same

nayooti:
On project base I was using spaces (it is defined in `project.pbxproj`). No idea what's wrong with myxcode.  I haven't knowingly changed anything.

PR #868, opened 2020-07-21
remove tap highlighting of members list in NewGroupController

### LINKED ISSUE #863: NewGroupController - tap groupMember - selection stays
in Chats -> New Chat -> New Group -> Add Members (add some members to group) 

Now: tap on Members in list.
- This will induce a visual selection (which personally should only be visible if there it actually induce some action)
- worse the selection state stays. 

I think since tapping on a cell doesn't do anything the cells shouldn't be selectable.

cyBerta:
fixes #863 

I decided to remove the highlighting completely from that list even though there is the option to swipe to left. UX wise I'm not sure if it would be better to just should unhighlight after tapping. Otoh users might then try to tap and cannot trigger any action.

### DIFF
--- a/deltachat-ios/Controller/NewGroupController.swift
+++ b/deltachat-ios/Controller/NewGroupController.swift
@@ -189,6 +189,7 @@ class NewGroupController: UITableViewController, MediaPickerDelegate {
                     contactCell.avatar.setImage(profileImage)
                 }
                 contactCell.setVerified(isVerified: contact.isVerified)
+                contactCell.selectionStyle = .none
             }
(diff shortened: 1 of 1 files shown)


r10s:
> UX wise I'm not sure if it would be better to just should unhighlight after tapping

what are others doing?

cyBerta:
I think the swipe to left is more like a secondary action, whereas tapping is a primary action. On Telegram there aren't secondary actions without primary actions. Tapping on a contact opens the contact detail view. That's why they have highlighting enabled.

PR #871, opened 2020-07-21
add some values to documents, gallery, ephemeral messages cells 

### LINKED ISSUE #839: add values to some profile rows
the rows gallery, document and disappearing-messages all have the .disclosureIndicator instyle, i think it would be useful to make some more value of that:

- gallery, document can show the number of items in them, 0 for none
- disappearing-messages can just get the works On/Off (there are already strings for that) (showing the concrete selection is more tricky as too long, two line style is ugly and not used elsewhere etc. so i would not target that now)

cyBerta:
closes #839 

at some point we should streamline both ViewControllers. ContactDetailViewController uses a viewModel whereas GroupDetailViewController doesn't.

### DIFF
--- a/deltachat-ios/Controller/ContactDetailViewController.swift
+++ b/deltachat-ios/Controller/ContactDetailViewController.swift
@@ -28,8 +28,9 @@ class ContactDetailViewController: UITableViewController {
     }()
 
     private lazy var ephemeralMessagesCell: UITableViewCell = {
-        let cell = UITableViewCell(style: .default, reuseIdentifier: nil)
+        let cell = UITableViewCell(style: .value1, reuseIdentifier: nil)
         cell.textLabel?.text = String.localized("ephemeral_messages")
+        cell.detailTextLabel?.text = String.localized(viewModel.chatIsEphemeral ? "on" : "off")
         cell.selectionStyle = .none
         cell.accessoryType = .disclosureIndicator
         return cell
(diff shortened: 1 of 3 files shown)


r10s:
> at some point we should streamline both ViewControllers. ContactDetailViewController uses a viewModel whereas GroupDetailViewController doesn't.

yip, the ContactDetailViewModel for ContactDetailViewController is useless - and it seems not to help on avoiding code duplication between GroupDetailViewController and ContactDetailViewModel ;)

apart from maybe removing ContactDetailViewModel, however, i would not try to get a separate base class or so for both profile types, at least not now.

cyBerta:
I'm fine with throwing away the view model. it's only nice to have both architectures more similar as we have code duplication and it will make it easier to produce even more :D - it's less thinking involved then

r10s:
looks great,  btw!

- i am wondering, if we shouldn't just show the selected value instead of "on" (although i suggested that myself :) - there should be sufficient space

- in larger chats, there seems to be a noticable delay, currently profiling that, maybe it makes sense to put the computation in a background thread

r10s:
> in larger chats, there seems to be a noticable delay, currently profiling that, maybe it makes sense to put the computation in a background thread

nahh, it's fine, even for my largest chats, after a fresh restart without anything cached, things are not more than 20ms. in most cases >5ms.

Issue #569, opened 2020-03-17
Notifications not shown instantly


tompatulpan:
No notifications are shown until DC is opened. In standby there is no (push) notifications.

- Device: iphone X
- iOS version: 13.3.1
- Delta Chat version: 1.2.1
- Expected behavior: show notification on locked phone.
- Actual behavior: dont show anything until DC is opened, then the red dot shows if new messages.
- Steps to reproduce the problem:
- Screenshots:
- Logs:

Let me know what logs to collect if needed.

croessner:
Same problem here. I get not informed about new messages until I open the app

stefanbesler:
Same here

seehma:
same here, this continuous check does not work. I only see the emails coming in my mailbox, but the app does not check. When i open the app -> the messages are arriving.

Is the mail App also continuously checking the server, or how does the notification of new email work in those mail apps (Outlook, Apple Mail, ...)

iOS 13.6
deltaChat App: 1.10.1
iPhone 11

r10s:
as we do not apples push service, currently, Delta Chat can only check for new mail when the operating-system allows us to do so. if the app is "in frequent use", that may be every 10 minutes or so, however, this timespan may grow endlessly when the app is not in use.

csb0730:
I got the same complaints from a friend of mine.

No chance to do anything? For standard WA users (focus is at WA) this issue makes delta quite unusable :-(

croessner:
What, if ActiveSync was supported? I use SOGo with ActiveSync to have push working. Would it be possible to implement this for delta chat, too?

hpk42:
There recently were some successful experiments with a Delta/iOS version and a 
notification provider that integrates with Apple's notification cloud. 
It worked on dev phones but there is a number of things that need doing
to get this releasable, among them legal/GDPR questions.

csb0730:
Isn't it possible to have at least a fix time interval to wake up DC when sleeping? Maybe 10min or similar? Even an hour would be better than nothing!

r10s:
without the help of the server, as far as i know, a "fix time interval" is not possible, only intervals chosen by the system  as mentioned at https://github.com/deltachat/deltachat-ios/issues/569#issuecomment-663457132

masvil:
I hope a kind of notification will be implemented. In the while we can rely on "standard" incoming email notification in some way.

cyBerta:
[...]

PR #889, opened 2020-07-30
remote push

r10s:
targets https://github.com/deltachat/deltachat-ios/issues/569

- [x] get token from apple and pass it to the [notification server](https://github.com/dignifiedquire/notifiers/)
- [x] wake up on notifications coming from apple, triggered by notification server (should work, but currently the notification server is not working)
- [x] respect "Notification" user setting (do not do token-dance when disabled; when state switches to enabled, do that token-dance at once)

maybe update the token on the notification server only on changes, however, for resilience (server issues), re-pass the token at least every some hours. ~~however, that can also go to a subsequent pr~~ we decided to do that in another pr, if at all.


cyBerta (approved):
nice

Issue #427, opened 2019-12-11
avoid frequent appearing/disappearing of the input field


r10s:
in a chat, when eg. attaching an image or recording a voice-message the input bar disappears.

in between, it shortly reappears just to disappear again, this results in a bit annoying flickering.
also the whole screen is moved up and down.

same eg. if you select "delete" from the context menu and the actionSheet appears to ask for confirmation.

wondering if there are reasons why the input field disappears at all. it is grayed out and all controls lay over it.

cyBerta:
The issue is in the InputBarAccessoryView library and I guess it's related to how the layout updates are done. Anyways, I couldn't find an easy fix for now. Maybe sb else can find out more.

r10s:
this is currently fixed by #560 , however with some side-effects, #570 

also, #560 hides the but more coincidentally, so, might be it pops up again.

so, it may be worth having a closer looks, also as the quick disappearing and  reappearing is annoying. if we can fix this correctly, we can probably revert #560

nayooti:
@r10s Can we close this?

r10s:
it is still annoying and worth fixing. i think, we have a chance to fix it with #888

cyBerta:
https://programming.vip/docs/ios-uses-uimenucontroller-without-hiding-keyboard.html
This might give us an hint how to handle the appearing/disappearing inputfield by tweaking the UIResponder chain.

r10s:
looks complicated - not sure how many new issues that will produce :)
as mentioned, maybe target that after messageKitReplacement, also living with the current state seems to be a reasonable approach meanwhile :)

r10s:
closing this, i just checked, at least it appears less annoying to be. might be because of "gallery" showed in cards, might be that the os is showing the things a bit smoother now.

cyBerta:
https://github.com/MessageKit/MessageKit/issues/620

r10s:
great! so, there is a chance in fixing that? in this case, please re-open :)

cyBerta:
some things need time but will be fixed at some point :)

r10s:
not sure if the re-appearing is really a bug, you'll notive sth. like that also in other apps. otherwise tracked in https://github.com/deltachat/deltachat-ios/issues/1737 now

PR #892, opened 2020-08-05
Message kit replacement

cyBerta:
todo to make the chat-view usable again:
- [x] open the chat at the end of the list (scroll down) (@cyBerta)
- [x] add keyboard/input-field, attach-button, send-button (@cyBerta)
    - [x] maintain scroll position if keyboard expands (@cyBerta )
    - [x] hide keyboard when scrolling down (is there a flag or so for that? other apps seem to hide the keyboard only when scrolling down near the bottom) (currently, it is not possible to hide the keyboard at all) (@cyBerta )
    - [x] scroll down on send (@cyBerta )
- [x] fix empty state view (message is not wrapped, see message after creating a new group) (@cyBerta )
- [x] add image avatars  (@cyBerta)
    - [x] tap on image avatar should open user-profile 👈 (@cyberta)
- [x] add text-view
  - [x] make links clickable 👈 (@cyBerta )
- [x] add image-view, including gif  (@cyBerta)
   - [x] a tap on image-view should go to gallery as before 👈 (@cyBerta )
   - [x]  fix cell size calculation of gif / gif + text messages [*]
- [x] add video-view, incl. play-button 👈 and opening in gallery 👈 as before (@cyBerta )
- [x] add audio-view (@cyBerta )
- [x] add file and other views (@cyBerta)
    - [x] a tap on file-view should open documents as before 👈 (@cyBerta )
- [x] add delivery-line (date/time, checkboxes), should go inside the bubble this time to allow backgrounds in a subsequent pr (@cyBerta )
- [x] add generic context menu (copy/info/delete/forward) 👈 (@r10s)
- [x] add a delegate-or-so to get different events from BaseMessageCell (roughly marked as 👈 above), i would avoid creating different delegate-classes for all types, a single delegate-class added to BaseMessageCell should do the job, it can contain all events for the subclasses then. (@r10s)
- [x] call dc_markseen_msgs() for messages scrolled into view. this also makes "disappearing messages" work again. (@cyBerta)
- [x] autocrypt-setup message handling in saved messages (@cyberta)

things to fix:
- [x] scrolling blocks from time to time (@r10s, might be fixed by #940)
- [x] use adaptive font sizes as before, see https://github.com/deltachat/deltachat-ios/pull/810 (@cyBerta )
- [x] on empty chats, messages get out-of-view when posted as they stuck at the upper border of the screen - and that is shifted out-of-view by the virtual keyboard. see eg. new saved-messages chat, posting a message there will make it invisible first. (@cyBerta )
- [x] scrolling is still flaky, when opening a chat with recent images, things are not really scrolled down. maybe this is related to groups only, or maybe to [*] i could not reproduce that for "saved messages" EDIT: this seems to be vastly improved by #921, however, i found another, fortunately now reproducible issue:
- [x] if the last (?) messages contains media+text, the initial scrolling position is wrong. can be reproduced eg. by sharing an image plus some text to "Saved messages"
- [x] images not shown on ios14 (@cyBerta )
- [x] if the input field disappears and reappears the scroll position is not correct (offset of input field height is not reapplied)
- [x] the scroll position should be retained if the user opens chat message details and goes back to the chat (@cyBerta )
- [x] adapt padding for messages including images
- [x] hide name + avatar for 1:1 chats

optional/later:
- if possible: avoid frequent appearing/disappearing of the input field, see #427
- switching from dark- to lite-mode lacks redraw of some cells
- tweak padding and sizes
- removing MessageKit
- scroll to correct position on search
- background image


r10s:
@cyBerta can you drop a few lines about reasoning to the source code here and there? eg. why the ChatTableView is needed and why/when this becomes a firstResponder. no big explanations, just one line or so top of class declaration.

cyBerta:
@r10s I think we're more or less done with this ticket.

r10s:
> @r10s I think we're more or less done with this ticket.

yeah, looks pretty good. thanks for getting that through :)

i just rebased it and will test it out another day on my phone before finally merging :)

r10s (approved):

r10s:
looks good so far, i'll merge that in now; there are some minor issues but they can be targeted in separated tickets - esp. as also the MessageKit had several issues :)

great job, @cyBerta :)

mlyongardiner:
@r10s and @cyBerta -
I'm not amazing with Git/Github, how can I use this implementation? MessageKit is an amazing project but it doesn't load messages as efficiently as I need in my project. Any help would be great!

PR #893, opened 2020-08-05
Streamline add group members

### LINKED ISSUE #890: streamline "new contact" in add-member views
members to verified groups cannot be added by just entering an email-address, they must be qr-code-verified by some other member.

the added contact will not be added to the group, core will take care of that, however, it is pretty confusing as with #887 non-verified contacts are not shown (which is totally correct), so the just added contact is not visible which is quite confusing.

for verified-group-creation, the button is already hidden.

however, the new-contact-button is also missing for normal-group-creation - where it is useful.

currently, we have NewGroupAddMembersViewController and AddGroupMembersViewController - the view look quite similar to me take and return a list of groups, maybe they can be combined.

cyBerta:
closes #890


cyBerta:
> currently, (a) the already added members are not shown in the list, however, when doing a search, they are shown, however, without a checkmark.

I'll change that.

cyBerta:
What shall happen, if the user unmarks a selected contact that was already part of the group? The option is called "Add members", not "edit members". Rename the label and allow removal of members? Prevent deselection?

cyBerta:
For now, removal of already added contacts is allowed in "Add members" for extisting groups. This way the behavior between new group creation and existing group editing (wrt. member handling) is the same.

r10s:
>  The option is called "Add members", not "edit members". Rename the label and allow removal of members? Prevent deselection?

"Add members" is still fine, it is the primary action. even if you can uncheck already added members and remove members this way (the primary action to remove members is swipe on ios, resp. long-click on android). we have such secondary options in many cases, eg. you can do "contact info" when creating a chat.

PR #907, opened 2020-08-24
implement delivery line

cyBerta:


r10s (approved):
great, let's merge it in, we can refine as needed later, the prs to messageKitReplacement are mostly to avoid conflicts :)

PR #915, opened 2020-08-29
add basic context menu

r10s:


r10s:
@cyBerta this pr has the problem that the context menu does not disappear when tapped outside.

after some research, i figured out that the `self.tableView.becomeFirstResponder()` stops the context menu from working, not totally sure, what the reason for this is, maybe the input does not "give up" the responder state or so.

cyBerta:
shall I have a look at it?

r10s:
@cyBerta that would be great, i am a bit out of ideas. but we can also have an a/v on that these days (as you have time :) (but tonight i am offline :)

r10s:
great, much better now. so one has to hide the menu manually ... tbh, i did not expect that :)

PR #925, opened 2020-09-20
add option to show contact-encryption-info

### LINKED ISSUE #924: show contact encryption info
there should a an "show encryption info" button somewhere in the profile, i think at the bottom, after "archive", before "block" is a good place, it is not an important function and maybe will be replaced by sth. else.

the encryption dialog should just be an alert-dialog that just shows the result of `dc_get_contact_encrinfo()`.

this is not super-sophisticated, but aligns possibilities between all systems, see https://github.com/deltachat/deltachat-pages/pull/375/files#diff-52e34649a6f7233447d978f98af84ad8R235

as this function is referenced from the help, this issue is close to a bug ;)

r10s:
this was just missing - but was mentioned in the help (which also has a pending update-pr, https://github.com/deltachat/deltachat-pages/pull/375)

closes #924

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -178,6 +178,15 @@ public class DcContext {
         return []
     }
 
+    public func getContactEncrInfo(contactId: Int) -> String {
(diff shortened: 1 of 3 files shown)


cyBerta (approved):
lgtm

PR #926, opened 2020-09-22
mark seen messages

cyBerta:
* call markSeenMessages every 10 unread (= not message state DC_STATE_IN_SEEN) messages displayed on screen 
* only in real chats
* debouncing implemented with scolling performance in mind

### DIFF
--- a/deltachat-ios/Chat/ChatViewControllerNew.swift
+++ b/deltachat-ios/Chat/ChatViewControllerNew.swift
@@ -351,6 +351,33 @@ class ChatViewControllerNew: UITableViewController {
         return cell
     }
 
+    public override func scrollViewDidEndDragging(_ scrollView: UIScrollView, willDecelerate decelerate: Bool) {
+        if !decelerate {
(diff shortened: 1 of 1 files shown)


cyBerta:
@r10s for some reason the message state seems not to change after calling `dcContext.markSeenMessages`. I'll checkout android tomorrow and verify if this behavior can be reproduced on android as well.

r10s:
we should do this debouncing only for "one view".

so, if a view with 4 messages is opened, it makes sense to call dcContext.markSeenMessages() for these messages together, maybe delayed by half a second. but it does not make sense to wait maybe minutes until it is actually called.

note, that core is also aggregating multiple mdns-messages into one message sent, so maybe we can also remove debouncing completely if it gets too complicated here. or we leave debouncing as a optimisation thing for later.

cyBerta:
now without debouncing. Messages get marked as seen when scrolling was finished which is what android also does.

r10s:
nice!

PR #928, opened 2020-09-22
adapt to xcode12

### LINKED ISSUE #927: does not build with Xcode 14
warnings/errors are:

- The [iOS Simulator|some Pod name] deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 14.0.99.

- plus a compile error wrt non-exhaustive switch

r10s:
xcode11 does not support latest ios13 nor the new ios14, therefore an update is needed.

see commit messages for details.

~~i think, these changes should be fine for Xcode 11 as well, otherwise @cyBerta probably also has to upgrade sooner than expected.~~ EDIT: update is needed because of PHPhotoLibrary.requestAuthorization .limited

closes #927

### DIFF
--- a/Pods/Pods.xcodeproj/project.pbxproj
+++ b/Pods/Pods.xcodeproj/project.pbxproj
@@ -15,6 +15,7 @@
 			dependencies = (
 			);
 			name = SwiftFormat;
+			productName = SwiftFormat;
 		};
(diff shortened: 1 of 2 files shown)


cyBerta (commented):
lgtm

PR #932, opened 2020-09-24
Update deps for ios14

cyBerta:
- fixes not appearing images on iOS14
- current pod versions generate additional xconfig files for all pods. I added them since they need to be removed with every update otherwise


r10s:
lgtm - however, better merge this to master and rebase messageKitReplacement on top of that.

- images are probably broken on ios14 also without messageKitReplacement
- this keeps the changes in messageKitReplacement low and maintainable

cyBerta:
ok, makes sense

r10s:
great!

r10s:
if in doubt, open a new pr, not sure if once can change the merge-target

PR #952, opened 2020-10-17
Gallery - context menu (delete only)

### LINKED ISSUE #944: add option to delete image from gallery
currently, there is no option to delete an image from "profile / images and videos".

enlarging the image allows swiping and sharing, but no delete button - afaik, this enlarged view is not really under our control, so it is probably hard to get a button in here.

therefore i suggest to show a simple "Delete" button on a long-tap in the overview - just the same as we are using in the chatview.

tappign the button should do the same as the "delete" button in the chat does - let the user confirm via safeActionSheet and call deleteMessage(). refreshing should be done automatically by listening to the DC_EVENT_MSGS_CHANGED (that might already be implemented)

nayooti:
closes #944
This PR adds a context menu to the GalleryViewController. The context menu only shows **delete**-action (as required in #944) but can be easily altered with other options like sharing images etc.

Since iOS13 apple introduced a new type of context menu, so I implemented two context menus that are shown depending on the iOS version. To review I suggest using an iOS12 simulator.

I had to move the `prepareContextMenu`-method in chatList to viewWillAppear. Otherwise the menu manipulation in `GallerViewController` would override the ChatList's context menu.

iOS12-style context menu
image

iOS13-style context menu
image


cyBerta:
I like the new ios13 menu. What's still missing is previews for videos though.
Screen Shot 2020-10-20 at 13 07 57

cyBerta (changes requested):
see my comment

r10s:
looks great the new context menu! and also a good place to make some experiences with them :)

cyBerta:
closing in favor of #985

PR #953, opened 2020-10-18
Preview fix

### LINKED ISSUE #936: opening gallery slow
opening the simple image gallery is slow if there are many images in the gallery.

reason is that all message-objects are loaded before the image is displayed, which, however, is not needed - opening the gallery should be only a call to [dc_get_chat_media()](https://c.delta.chat/classdc__context__t.html#a344a82b9f288b5eb5d39c7cf475cddb7) plus loading the image to display.

all other images should be loaded as needed.

on my iphone7, things get slow on chats with >500 images, it takes half a second to open. so that is not _that_ slow, but i still would consider this as a bug - you may have 2k images and you may have slower phones.

nayooti:
closes #936 

PreviewController now takes msgIds - so messages can be generated on demand.

I had to change the way how PreviewController is initialised. It takes an enum with associated type - this way single previews like we are using on avatar taps take an url, while chat/message related previews take msg-ids.

### DIFF
--- a/deltachat-ios/Chat/ChatViewControllerNew.swift
+++ b/deltachat-ios/Chat/ChatViewControllerNew.swift
@@ -805,8 +805,8 @@ class ChatViewControllerNew: UITableViewController {
         mediaPicker?.showPhotoVideoLibrary()
     }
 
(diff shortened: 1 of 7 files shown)


cyBerta (approved):
cool

PR #955, opened 2020-10-20
switch chat bubble background color on dark/light mode change

### LINKED ISSUE #945: chatview is not redrawn on switching to dark-mode
while the darkmode in the new messageKitReplacement is fine so far, a redraw is missing on switching the modes.

currently, eg. some artefacts of darkmode are shown when switching to lite-mode; scrolling a bit up or down corrects these things.

cyBerta:
closes #945


r10s (approved):
lgtm, tested, great!

PR #956, opened 2020-10-20
[WIP] remove MessageKit 

### LINKED ISSUE #888: Get rid of MessageKit
MessageKit adds a lot of unnecessary complexities in adding new Views. Moreover there are some issues that are not resolvable with MessageKit, like https://github.com/deltachat/deltachat-ios/issues/602 or https://github.com/deltachat/deltachat-ios/issues/614.

For getting rid of MessageKit I propose the following steps:
* implement a ChatViewControllerNew, based on a UITableViewController while keeping ChatViewController as it is
* keep the InputBarAccessoryView for now which is a separate library
* implement a hard-coded switch in order to test the incremental development of ChatViewControllerNew. It will allow us to merge back to master early. Releases will use only the old ChatViewController.
* if we consider ChatViewControllerNew as feature rich and stable enough, we replace ChatViewController with it. 
* afterwards we remove all MessageKit dependencies

cyBerta:
and replace all remaining references
closes #888


r10s:
k, let's move that in, it is an improvement.

i still like to have some more cleanup, also i do not like the shifting of the items when doing a multiple-select, an "always there" context menu would help there and moving COPY there looks reasonable to me (whatsapp does the same btw), however, this can be part of another discussion/pr.

r10s:
/me deleted my comment, was meant for another pr

r10s (approved):
what a fun!!!! throwing away 1k loc 💯

PR #958, opened 2020-10-21
show context of translations errors

r10s:
otherwise, source strings that are not accepted as "foo %s bar %s" are hard to detect (placeholders need to be uniquely addressable  "foo %1%s bar %2$s" would be accepted, %s is only accepted if there is only one placeholder)


cyBerta (approved):
👍

cyBerta:
the folder changed :)

r10s:
> the folder changed :)

ahh :) i did a rebase :)

PR #961, opened 2020-10-21
Tweak date line

### LINKED ISSUE #950: tweak date-line in bubbles
- the date-line in the bubbles should be always right-aligned (this is true currently for outgoing messages, but not for incoming messages)

- the color of the outgoing messages date-line should be green (eg. #70b15c is used on android), in color of incoming messages date-line seems to be fine.

- the space between message and date-line is a bit too large, roughly the half should be fine.

cyBerta:
I addressed each point in #950 in a different commit and that maybe closes #950 
you might tweak it more yourself further @r10s as this is very subjective. 

see also: https://github.com/deltachat/deltachat-ios/issues/950#issuecomment-713914875


r10s (commented):
looks much better, thanks a lot :)

r10s (approved):

PR #963, opened 2020-10-23
add basic quote view 

### LINKED ISSUE #942: display quotes
~~_before this ticket can be targeted, [messageKitReplacement](https://github.com/deltachat/deltachat-ios/pull/892) should be merged in!_~~ EDIT: messageKitReplacement is merged.

current master already has a new functionality for handling quotes, see https://github.com/deltachat/deltachat-core-rust/pull/1975 and android will probably have a quote-functionality in the next release.

_as a first step_, we should start **displaying quotes**. while in the past the sender was somehow aware of adding some context to the text, now one might do that eg. by quoting and just saying "yes". therefore, displaying quotes is pretty urgent to get in (in contrast to **creating quotes** which can also be a release or so later).

displaying a quote is pretty straight forward:

- use [`dc_msg_get_quoted_text()`](https://c.delta.chat/classdc__msg__t.html#a456299eae0b43c4ec6ee6fe812947ed9) and display the returned text in a quoting style as in the android version (slightly smaller text, vertical bar on the left, only display 2~3 lines of text.
- note, that quotes may be added to all types of messages
- if `dc_msg_get_quoted_text()` returns NULL, there is just no quote

this is already the absolute minimum. once that is done,

-  use [`dc_msg_get_quoted_msg()`](https://c.delta.chat/classdc__msg__t.html#a9282d97546388655d99018fd63294bb6) to enrich the quote:

- show quoted contact name above the quote using the contact-color
- also colorize the vertical bar in the contact-color
- a tap on the quote should jump upwards to the original message
- if `dc_msg_get_quoted_msg()` returns 0 while `dc_msg_get_quoted_text()` returns != NULL, the original message is unknown/deleted/not-yet-there, in this case just display the minimum from above

once this is done, we can continue enriching quotes by displaying eg. the quoted image. however, that can also be done in another ticket :)

cyBerta:
closes #942 
accidentally also image preview got implemented here. 

- missing is background highlighting after scrolling to quoted message

### DIFF
--- a/deltachat-ios/Chat/Views/Cells/BaseMessageCell.swift
+++ b/deltachat-ios/Chat/Views/Cells/BaseMessageCell.swift
@@ -111,7 +111,7 @@ public class BaseMessageCell: UITableViewCell {
     lazy var bottomLabel: PaddingTextView = {
         let label = PaddingTextView()
         label.translatesAutoresizingMaskIntoConstraints = false
-        label.font = UIFont.preferredFont(for: .caption1, weight: .regular)
+        label.font = UIFont.preferredFont(for: .caption1, weight: .medium)
         label.setContentHuggingPriority(.defaultHigh, for: .horizontal)
         label.setContentCompressionResistancePriority(.defaultHigh, for: .horizontal)
         label.layer.cornerRadius = 4
--- a/deltachat-ios/Chat/Views/Cells/BaseMessageCell.swift
+++ b/deltachat-ios/Chat/Views/Cells/BaseMessageCell.swift
@@ -192,6 +192,11 @@ public class BaseMessageCell: UITableViewCell {
         let messageLabelGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(handleTapGesture(_:)))
         gestureRecognizer.numberOfTapsRequired = 1
         messageLabel.addGestureRecognizer(messageLabelGestureRecognizer)
+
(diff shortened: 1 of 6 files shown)


deltachat-ios/Chat/Views/Cells/BaseMessageCell.swift
```diff
@@ -277,6 +277,20 @@ public class BaseMessageCell: UITableViewCell {
...
+
+        if let quoteText = msg.quoteText {
+            quoteView.isHidden = false
+            quoteView.quote.text = quoteText
+
+            if let quoteMsg = msg.quoteMessage {
+                let contact = quoteMsg.fromContact
+                quoteView.senderTitle.text = contact.displayName
+                quoteView.senderTitle.textColor = contact.color
+            }
```
r10s, line 300: as cells are reused, there should probably an `else { quoteView.senderTitle.isHidden = true }` or so. (just came over that, i know, it is WIP :)

cyBerta: :+1:

cyBerta: after having a second glance on it, I realized that sender title gets cleaned up in quoteView.prepareForReuse(). if the senderTitle.text is nil the sender title doesn't take any space

r10s (approved):
great stuff!

quotes above images/files/etc. are missing - can be added to this pr or to another one :)

PR #968, opened 2020-10-25
cleanup saved-messages and device-chat profiles

### LINKED ISSUE #965: hide 'disappearing messages' when a chat is not sendable
if one cannot send to a chat,  'disappearing messages' does not make much sense (and does not work, eg. in case of device-chat)

the corresponding profile-option should just be hidden/removed.

counterpart of https://github.com/deltachat/deltachat-android/pull/1690, on desktop, this is already fine :)

### LINKED ISSUE #966: hide/disable 'edit' from device-chat and saved-messages
it just does not make sense here, you cannot really change avatar nor name here.

also, if we enhance the dialog by "protect chat", this won't make much sense for device-messages or saved-messages (maybe for the latter, but this probably opens further issues, so it is fine to just not allow it initially)

### LINKED ISSUE #967: hide 'block contact' from saved-messages and device-messages profile
it does not make sense to block yourself. also blocking device-messages is probably not a good idea (until someone tells a real usecase :)

same for "show encryption info"

### LINKED ISSUE #969: SELF in group member list should not open saved-messages
when tapping on SELF in the group member list, just nothing should happen.

the entry itself should stay as is, however.

this is also what happens on android, opening saved-messages does not make sense here.

r10s:
remove useless things from saved-messages and device-chat profiles. normal group/contact profiles are unchanged.

closes #967, closes #966, closes #965, close #969

counterpart of https://github.com/deltachat/deltachat-android/pull/1692

before/after screenshots:

 

 

EDIT: also remove "mute" from saved-messages - for self-sent-messages, no notifications are sent at all.

### DIFF
--- a/deltachat-ios/Controller/ContactDetailViewController.swift
+++ b/deltachat-ios/Controller/ContactDetailViewController.swift
@@ -111,9 +111,11 @@ class ContactDetailViewController: UITableViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         configureTableView()
-        navigationItem.rightBarButtonItem = UIBarButtonItem(
-            title: String.localized("global_menu_edit_desktop"),
-            style: .plain, target: self, action: #selector(editButtonPressed))
+        if !self.viewModel.isSavedMessages && !self.viewModel.isDeviceTalk {
+            navigationItem.rightBarButtonItem = UIBarButtonItem(
+                title: String.localized("global_menu_edit_desktop"),
+                style: .plain, target: self, action: #selector(editButtonPressed))
+        }
         self.title = String.localized("tab_contact")
     }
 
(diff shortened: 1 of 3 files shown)


cyBerta (approved):
looks great!
I think it could also make sense to disable the image enlargement when tapping on the contact detail avatar for saved messages and device talk. These images don't have any hidden details one can discover by tapping on them.

r10s:
> I think it could also make sense to disable the image enlargement when tapping on the contact detail avatar for saved messages and device talk. These images don't have any hidden details one can discover by tapping on them.

i also thought about that, however, it is useful to "learn" things, having the same behavior in different chats. also android allowed it and it is easier to keep it - so i think it makes some sense to leave that bit as is for now.

PR #972, opened 2020-10-27
fix micro-hangs

r10s:
testing out if that removes the one-second-hang i am getting from time to time on my larger account ...

### DIFF
--- a/DcCore/DcCore/DC/events.swift
+++ b/DcCore/DcCore/DC/events.swift
@@ -105,6 +105,9 @@ public func handleEvent(event: DcEvent) {
                     "chat_id": Int(data1),
                 ]
             )
+
+            let array = DcContext.shared.getFreshMessages()
+            UIApplication.shared.applicationIconBadgeNumber = array.count
         }
 
     case DC_EVENT_CHAT_MODIFIED:
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -268,7 +268,7 @@ class ChatViewController: UITableViewController {
 
         // things that do not affect the chatview
         // and are delayed after the view is displayed
-        DispatchQueue.main.async { [weak self] in
+        DispatchQueue.global(qos: .background).async { [weak self] in
             guard let self = self else { return }
             self.dcContext.marknoticedChat(chatId: self.chatId)
         }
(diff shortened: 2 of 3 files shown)


r10s:
while first commit, "async calls to markseen/marknoticed, update global-badge by event", seems to be useful, it does not solve the problem.

meanwhile i find a way to reprocude it:
- Alice sends a contact request to Bob
- Bob accepts the request -> chat opens - and is unresponsive for a second.

the send message may be a single word only, and it also happens when read-receipts are disabled. needs further investigation.

r10s:
as discussed in a/v, i was using the wrong function to start a background thread :) 

the initially used `DispatchQueue.main.async` just delayes execution in the main/ui thread,  `DispatchQueue.global(qos: .background).async` really does the things in another thread.

at a first glance, this seems to work now, then Alice/Bob scenario described above is no longer an issue with the second commit  :)

DcCore/DcCore/DC/events.swift
```diff
@@ -105,6 +105,9 @@ public func handleEvent(event: DcEvent) {
                     "chat_id": Int(data1),
                 ]
             )
+
+            let array = DcContext.shared.getFreshMessages()
+            UIApplication.shared.applicationIconBadgeNumber = array.count
```
cyBerta, line 110: nice catch

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -265,11 +265,13 @@ class ChatViewController: UITableViewController {
...
         AppStateRestorer.shared.storeLastActiveChat(chatId: chatId)
+
         // things that do not affect the chatview
         // and are delayed after the view is displayed
-        dcContext.marknoticedChat(chatId: chatId)
-        let array = dcContext.getFreshMessages()
-        UIApplication.shared.applicationIconBadgeNumber = array.count
+        DispatchQueue.global(qos: .background).async { [weak self] in
+            guard let self = self else { return }
+            self.dcContext.marknoticedChat(chatId: self.chatId)
```
cyBerta, line 273: you can use self? here, too

r10s: that would result in `self?.dcContext.marknoticedChat(chatId: self?.chatId ?? 0)` - i do not think, this is a big win in readablility.

cyBerta: You're right I oversaw the second self there

cyBerta (approved):
lgtm

PR #979, opened 2020-11-02
tweak aspect ratio for image messages:

### LINKED ISSUE #970: Don't show messages very thin if small or tall images are sent
if an image is pretty small and comes with text attached, the width is calculated by the image width, not by the text.

this results in  very small text-column, where rarely two words fit beside each other. 

we should:

- let the text define the width of a message bubble
- the image is scaled to fit this width
- _landscape_ images are fine that way
- _portrait_ images should be cut atop/abottom so that they are not higher than a square, the square is also a hint that this is is no landscape image.

counterpart of https://github.com/deltachat/deltachat-android/issues/1687 and roughly what other messengers are doing. 

not totally sure, how the situation was before messageKitReplacement.

### LINKED ISSUE #981: weird scrolling
i found a simple way to reproduce the weird scrolling behavior i mentioned in the a/v yesterday:

- open a chat with some more messages in, the chat opens correctly with the last message

- _without scrolling_, rotate the device from portrait to landscape mode

- the chat now jumps to somewhere up and scrolls immediately down to the last message again

if you rotate the device with the message list already scrolled up, this does not happen, things are fine and stable in this case.

i mentioned a similar behavior when entering more lines to the input field, might be that is the same issue, also there, the available height changes, however, this issue is harder to reproduce.

cyBerta:
* show square if image is in portrait, keep aspect ratio for landscape images
* reduce height of image-only messages so that they fit into the visible area
* improve balancing between image size and message text width

closes #970
closes #981 

fixes also an unrelated bug related about gif loading - all ongoing loading processes should be cancelled on prepareForReuse()

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -306,14 +306,17 @@ class ChatViewController: UITableViewController {
                 if self.showCustomNavBar {
                     self.navigationItem.setRightBarButton(self.badgeItem, animated: true)
                 }
+                if lastSectionVisibleBeforeTransition {
+                    self.scrollToBottom(animated: false)
+                }
             },
             completion: {[weak self] _ in
                 guard let self = self else { return }
                 self.updateTitle(chat: self.dcContext.getChat(chatId: self.chatId))
                 if lastSectionVisibleBeforeTransition {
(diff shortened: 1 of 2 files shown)


r10s (approved):
yeah, much better!

PR #983, opened 2020-11-03
fix potentially dangling pointer and bad arraysize

r10s:
this pr fixes two issues:

- using 'UnsafePointer<UInt32>' as before results in a warning saying that this
  results in a dangling pointer;
  this is fixed by using a closure and withUnsafeBufferPointer

- the 'count' argument was never given and defaulted to 1 -
  this results in not all messages beeing marked as unseen -
  and also even in a pointer error when, for whatever reason,
  an empty array as given.
  this is fixed by just removing the superfluous 'count' argument


cyBerta (approved):
great :rocket:

PR #985, opened 2020-11-04
Gallery - context menu - 2nd Attempt

### LINKED ISSUE #891: add context menu for gallery and documents
long-tapping a context menu in gallery or  documents should open a context menu in the same style as the one that appears when a chat-bubble is long-tapped.

the context-menu should offer at least "delete", share would also be nice (later one sth. as "jump to message", however, that requires probably #888 to be ready and is out of scope for now)

### LINKED ISSUE #944: add option to delete image from gallery

nayooti:
closes #944 
closes #891 

Since MessageKit-files were deleted I was not able to rebase properly, so I had to rebase "manually" using this branch.

## This PR:
- context menu now displays (and plays) videos like on Telegram.
- context menu works on different device rotations 
- context menu closes on device rotation -> this is intended native behaviour
- context menu for documents gallery in iOS 13
- `contextMenuProvider` to generate legacy and new styled context menus 
- context menu provides **show in chat** that redirects to the message in chat view
- same menu applied to `DocumentGalleryController` (modern context menu uses default preview)
 
Gallery did not look nice in landscape on devices with notch. I did a quick fix in this PR too.

## To test: 
- use videos+images in landscape and portrait 
- try device in portrait and landscape

replaces #952

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -566,6 +566,14 @@ class ChatViewController: UITableViewController {
         }
     }
 
+    func scrollToMessage(msgId: Int, animated: Bool = true) {
+        guard let index = messageIds.firstIndex(of: msgId) else {
+            return
+        }
+        let indexPath = IndexPath(row: index, section: 0)
+        tableView.scrollToRow(at: indexPath, at: .top, animated: animated)
+    }
+
     private func showEmptyStateView(_ show: Bool) {
         if show {
             let dcChat = dcContext.getChat(chatId: chatId)
@@ -1108,10 +1116,8 @@ extension ChatViewController: BaseMessageCellDelegate {
     @objc func quoteTapped(indexPath: IndexPath) {
         _ = handleUIMenu()
         let msg = DcMsg(id: messageIds[indexPath.row])
-        if let quoteMsg = msg.quoteMessage,
-           let index = messageIds.firstIndex(of: quoteMsg.id) {
-            let indexPath = IndexPath(row: index, section: 0)
-            tableView.scrollToRow(at: indexPath, at: .top, animated: true)
+        if let quoteMsg = msg.quoteMessage {
+            scrollToMessage(msgId: quoteMsg.id)
         }
     }
 
--- a/deltachat-ios.xcodeproj/project.pbxproj
+++ b/deltachat-ios.xcodeproj/project.pbxproj
@@ -109,6 +109,7 @@
 		AE4AEE3522B1030D000AA495 /* PreviewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = AE4AEE3422B1030D000AA495 /* PreviewController.swift */; };
 		AE52EA19229EB53C00C586C9 /* ContactDetailHeader.swift in Sources */ = {isa = PBXBuildFile; fileRef = AE52EA18229EB53C00C586C9 /* ContactDetailHeader.swift */; };
 		AE52EA20229EB9F000C586C9 /* EditGroupViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = AE52EA1F229EB9F000C586C9 /* EditGroupViewController.swift */; };
+		AE57C0802552BBD0003CFE70 /* GalleryItem.swift in Sources */ = {isa = PBXBuildFile; fileRef = AE57C07F2552BBD0003CFE70 /* GalleryItem.swift */; };
 		AE6EC5242497663200A400E4 /* UIImageView+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = AE6EC5232497663200A400E4 /* UIImageView+Extensions.swift */; };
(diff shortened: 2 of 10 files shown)


cyBerta:
I just had a short check on my phone. It looks great, video works too. What is missing is support for gif files. At least the first image of an gif should be shown, if previewing gifs is not possible. Currently there's a white image shown.

cyBerta (changes requested):
see my comment

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -260,6 +259,7 @@ class ChatViewController: UITableViewController {
         if RelayHelper.sharedInstance.isForwarding() {
             askToForwardMessage()
         }
+        prepareContextMenu()
```
r10s, line 262: how is this change related to the gallery context menu? or what does that fix?

nayooti: `UIMenuController` is a singleton (by Apple). I change it's options in Gallery, so I have to reset these options when moving back from Gallery.

r10s:
great!

some ui things:
- the delete-button should be red (see other apps)
- once tapped, there should be a confirmation (also see other apps)

to the general understanding: 
- the menu is a normal ViewController, or? os maybe call it ContextMenuViewController.swift to point out that it is just this. i was a bit confused what a MenuController is :)
- all in all this is some code and complexity - if we want to get the menu to the document view and - even more interestingly - to the normal chatview, we can resuse ContextMenuViewController.swift? (of course, we would need to add renderings for normal messages and files then)

deltachat-ios/Chat/ChatViewController.swift
r10s: thanks for explanation :)

nayooti:
`MenuController` is from Apple to show the old context menu (iOS 12 and lower)
`ContextMenuViewController' to show new context menu (iOS 13 and after).

`ContextMenuViewController' can be easily be reused and extended. For example to be used in `ChatViewController`, but obviously this needs to be extended for different MessageTypes. 

The complexity is only that we have to cover both context-menu-styles as long we support iOS12 and lower.
After all, the complexity isn't too bad IMO.

nayooti:
@r10s I could not find any suitable confirm-delete-string. Any idea?

r10s:
we're getting closer :)

some more things, **for ios 11, 12:**

- when the old-style context menu is opened, a tap outside the context menu should just close it. however, then this tap goes to the image, it also enlarges the image. this enlargement should require a second tap.

you can test this eg. in "iPhone 8 (iOS 12.2)" emulator

things **for ios 13+**:

- when the new-style context menu is opened, a tap outside the context menu should close it. this is correct. a tap _inside, on the preview image_ should directly enlarge the image. please check the normal ios gallery to get the idea.

- the delete confirmation is okayish, however, the ios gallery stays in the context-menu for the confirmation. this is no blocker, however, if this is easily doable by only setting a flag or so, we should just do it the same way. if it adds complexity, however, we should postpone it.

nayooti:
I looked into this, and I came to the conclustion, that none of these "things" seem to be makable without much effort. 
Options are very limited. 

For iOS 13-context menu:
I tried adding gesture recognisers to the preview-view, but all touches were simply ignored. The preview-view closes no matter what.
Out of interest I added a button to that view, and also here I could not get it's action triggered.

In general I think Apple catches all the touches before they hit the view/menus. I assume they heavily use `hitTestOnEvent` here, where you can decide if you let touches pass though or not. 

The only idea I have right now is to override `hitTest`-method by ourself, but this is never beautiful as it will implemented on top of the whole app. 
 ... and it is not guaranteed to work (because I think Apple already is overriding/using it here)

nayooti:
@r10s Ok, I am done for now.

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -162,7 +162,13 @@ class ChatViewController: UITableViewController {
                                        name: UIApplication.willResignActiveNotification,
                                        object: nil)
         notificationCenter.addObserver(self, selector: #selector(keyboardWillShow(_:)), name: UIResponder.keyboardWillShowNotification, object: nil)
+<<<<<<< HEAD
```
cyBerta, line 165: That won't compile, I think.

cyBerta (changes requested):

cyBerta:
There's an unresolved merge conflict and unfortunately also new merge conflcts with latest master that need to be resolved, too.

cyBerta (approved):
lgtm, I fixed the merge conflict and removed the controls from the video preview, because a tap on any item within the preview will open the preview controller, so the controls can never be triggered. It's now the same UI as in the iOS Gallery wrt. video previews.

PR #986, opened 2020-11-04
open chat and scroll to correct message after tap on notification

### LINKED ISSUE #721: tapping notifications do not open the correct chat
tapping a notification just opens the app at the last position - that might be settings, qr-code, whatever ...

nb: currently, notification processing is in DcCode, i think, this is not optimal as this may led to double notifcation. currently this is not the case as share does not perform imap, but that should be cleaned up, i suggest to use a callback for that as routing things through the notification system may just take too much time.

cyBerta:
closes #721 
* showing notifications when the app was in forground was buggy and is fixed by this PR
* foreground notifications are only generated if the user has not opened the corresponding chat of the incoming message
* foreground notifications disappear after 4 seconds
* background notifications keep on the system 'dashboard' 
* tapping on notifications, will open the correct chat and move the list so that the corresponding message is shown
* notifications show group names in group chats
* support for image previews

### DIFF
--- a/DcCore/DcCore/DC/events.swift
+++ b/DcCore/DcCore/DC/events.swift
@@ -139,30 +139,10 @@ public func handleEvent(event: DcEvent) {
             "message_id": Int(data2),
             "chat_id": Int(data1),
         ]
-
         DispatchQueue.main.async {
             nc.post(name: dcNotificationIncoming,
                     object: nil,
                     userInfo: userInfo)
-
-            let chat = DcContext.shared.getChat(chatId: Int(data1))
-            if !UserDefaults.standard.bool(forKey: "notifications_disabled") && !chat.isMuted {
-                let content = UNMutableNotificationContent()
-                let msg = DcMsg(id: Int(data2))
-                content.title = msg.getSenderName(msg.fromContact)
-                content.body = msg.summary(chars: 40) ?? ""
-                content.userInfo = userInfo
-                content.sound = .default
-
-                let trigger = UNTimeIntervalNotificationTrigger(timeInterval: 0.1, repeats: false)
-
-                let request = UNNotificationRequest(identifier: Constants.notificationIdentifier, content: content, trigger: trigger)
-                UNUserNotificationCenter.current().add(request, withCompletionHandler: nil)
-                DcContext.shared.logger?.info("notifications: added \(content)")
-            }
-
-            let array = DcContext.shared.getFreshMessages()
-            UIApplication.shared.applicationIconBadgeNumber = array.count
         }
 
     case DC_EVENT_SMTP_MESSAGE_SENT:
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -279,13 +279,11 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
     }
 
     func userNotificationCenter(_: UNUserNotificationCenter, willPresent notification: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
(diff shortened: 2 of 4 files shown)


r10s:
this pr needs a rebase as  https://github.com/deltachat/deltachat-ios/pull/987 is merged to master in between and also does the highligh-message bit.

r10s (commented):
i rebased the pr and did a functional test, without looking much at the code.

- it seems, that the notifications are less reliable when the app is unfocused. on master, in the first ~30 seconds after unfocus, notifications appear instantly, this does not work reliably with this pr.  
might be some coincidence, however, i could reproduce that several times.

cyBerta:
>     * in the first ~30 seconds after unfocus

So, if the app went to the background?

r10s:
> So, if the app went to the background?

yip. there is usually a small timeframe, where the app is still running before it terminates finally, 30 seconds iirc.

on master, in this timeframe, notifications pop up.

this does not happen with this pr, i could reproduce that several times.

cyBerta:
The system won't move notifications to the background if they are allowed to be shown in foreground. These foreground notifications are defined here:
` func userNotificationCenter(_: UNUserNotificationCenter, willPresent notification: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void)` 
the completion handler there is responsible for the behavoir of the foreround notification.
* The notification type `.alert`  allows to display the content of the noticiation - both as heads up notification and in the notification center you access by swiping the sreen from top to bottom. Foreground notifications of this type needs to be dismissed manually, they won't appear after a timeout. 
* The notification type `.sound` plays a sound when the notification appears but also updates an existing DC notification in the notification center. If there's no notification in the notification center, no new notification will be shown there.
* The notification type `.badge` updates the unread messages counter of the launcher icon. In any case we want that to happen.
* The notification type `.list` is a new property, introduced in iOS 14. It delivers a notification to the notification center, without a heads-up notification. 
* The notification type `.banner` is a new property, introduced in iOS 14 and replaces the type .alert. It has the same behavior.

Having all that in mind what I'm propsing is to implement UNNotificationPresentationOptions `[.list, badge]` for iOS 14 and [.badge] for ios < 14
This way the foreground notification will not disturb the user by requiring each single message notification to swipe away while writing in a different chat. On new iOS versions the last message will be shown in the notification center. The unread messages counter of the launcher will be always updated correctly.

cyBerta:
> yip. there is usually a small timeframe, where the app is still running before it terminates finally, 30 seconds iirc.
> 
> on master, in this timeframe, notifications pop up.
> 
> this does not happen with this pr, i could reproduce that several times.

It looks to me that this is an unrelated issue. Notifications will only be delivered if a background task (fetching new messages etc.) is still running, while the app goes to the background. 

This reference  https://developer.apple.com/documentation/uikit/uiapplication/1623029-backgroundtimeremaining
It explains:

> The value is valid only after the app enters the background and has started at least one task using beginBackgroundTask(expirationHandler:) in the foreground.

This means if the app went to the background without having started a background task, we cannot count on the value returned by `backgroundtimeremaining`: it always returns 1.7976931348623157e+308 if no background task is running 

This is why we're stopping all the threads quite immedeately after going to background. This check always fails in this scenario: 
it looks, we're not using backgroundtimeremaining correctly here  https://github.com/deltachat/deltachat-ios/blob/3c8fa9070c8a038d33bf361ca31b8cb472c99d9a/deltachat-ios/AppDelegate.swift#L144

cyBerta:
We might solve the `backgroundtimeremaining` issue by triggering a dummy task that runs for 30 seconds in the background to ensure we get the whole available background timeframe. Do you have other suggestions?

PR #987, opened 2020-11-11
scroll to search result

### LINKED ISSUE #602: scroll chat to search result

r10s:
i've made the highlightedMsg bit compatible to the pending https://github.com/deltachat/deltachat-ios/pull/986/files

closes #602


r10s:
merging, as i like to do a 1.14 testflight soonish and this was the last open point in https://github.com/orgs/deltachat/projects/30

PR #989, opened 2020-11-12
Empty search -> make  empty state banner always visible

### LINKED ISSUE #807: move empty state labels always to visible area
When searching through the members list, an empty state label appears if no result has been found. It is always centered and thus partly covered by the keyboard. 
The state label should be moved up if the keyboard is opened so that the user can still read the message. 

Screen Shot 2020-06-22 at 13 56 35

nayooti:
closes #807 

Fixed by using the  `emptyStateLabel` as tableViewHeader.

IMG_5962
IMG_5963


cyBerta (approved):
cool

PR #991, opened 2020-11-16
don't invert programatically camera icon, use a white icon set instead

### LINKED ISSUE #990: avatar preview not updated after selection on some phones
eg. on iphone12mini emulator, when selecting eg. an image for a newly created group, the image is not updated in the group selection dialog, instead grey circle is shown - while the image actually is set. 

Screen Shot 2020-11-16 at 11 50 23

also, the "camera" placeholder is not shown. same for self-profile image.

i thought first, this is an emulator issue, however, on iphone7-emulator, this works.

iirc, we had a similar issue in the past.

cyBerta:
closes #990


r10s (approved):
nix - fixed, fewer lines, fewer assets +1

r10s:
i just tested it, works like a charm - is this still wip? if not feel free to merge. this is the last™ open issue for 1.14 then, great! (core50 is on its way, however)

PR #999, opened 2020-11-19
add permanent hint about enabled server-deletion

### LINKED ISSUE #998: show permanent hint about server-delete impact
show permanent hint about server-delete impact on both view controllers, delete-overview and delete-server-details

see https://github.com/deltachat/deltachat-android/pull/1737

r10s:
when server-deletion is enabled, always show a hint about the impact below the option. this way, the user is more aware of the impact if enabling was done some time ago or the option was even enabled by the provider-db.

Screen Shot 2020-11-19 at 23 19 52

Screen Shot 2020-11-19 at 23 20 08

closes #998 
counterpart of https://github.com/deltachat/deltachat-android/pull/1737

### DIFF
--- a/deltachat-ios/Controller/SettingsAutodelOverviewController.swift
+++ b/deltachat-ios/Controller/SettingsAutodelOverviewController.swift
@@ -7,7 +7,6 @@ class SettingsAutodelOverviewController: UITableViewController {
 
     private struct SectionConfigs {
         let headerTitle: String?
-        let footerTitle: String?
         let cells: [UITableViewCell]
     }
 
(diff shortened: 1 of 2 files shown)


nayooti:
To me looks good.

In the longrun we could start using DiffableDataSource for smoother and easier/safer insert-operations.

cyBerta (approved):
it's ok :)
I would consider to just update the footer instead of the whole table view with `tableView.beginUpdate()` and `tableview.endUpdate()` but it's also not a big deal in this case (no performance hits)

cyBerta:
Oh, I didn't see that this PR was already reviewed (Browser cache)

r10s:
> I would consider to just update the footer instead of the whole table view with tableView.beginUpdate() and tableview.endUpdate() but it's also not a big deal in this case (no performance hits)

good to know, in fact, i was searching for sth. like that. but, it would be lot more code and it is no performance issue, so, i would leave it as is.

PR #1000, opened 2020-11-19
fix summary for drafted quotes

### LINKED ISSUE #996: Quotes for Document, Audio and Voice messages
currently there's no graphical representation for audio/voice messages and attached files, only the message text appears in a quote. If a message doesn't contain any message text, there's no indication (besides the sender information) wich message is quoted.

* For document, voice and audio messages without any message text, the message summary should be shown as quote text
* if a preview of a document can be rendered, a small image of the pdf should be shown, otherwise the corresponding document icon
* voice message quotes should have the play button as graphical representation

r10s:
(this should be merged into swipe-to-reply branch)

core guarantees that there is always a quote text if there is a quote,
in QuotePreview::configure(), rely on that,
using a summary fallback is not needed.

moreover, in DraftModel::setQuote(), we use the core to set the quote summary.
this closes #996

(nb: quote summaries were alrady fine before when a
chat with a draft-quote was left and entered again,
this was because of the implicit draft creation that is now
done explicit in DraftModel::setQuote())

### DIFF
--- a/deltachat-ios/Chat/DraftModel.swift
+++ b/deltachat-ios/Chat/DraftModel.swift
@@ -19,8 +19,16 @@ public class DraftModel {
     }
 
     public func setQuote(quotedMsg: DcMsg?) {
-        quoteMessage = quotedMsg
-        self.quoteText = quotedMsg?.text
+        if let quotedMsg = quotedMsg {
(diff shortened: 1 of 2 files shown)


cyBerta (approved):
lgtm

PR #1006, opened 2020-11-20
add dark mode support to quotePreview

### LINKED ISSUE #1002: draft-quotes do not respect darkmode
also, the padding above the upper line seems superfluous.

cyBerta:
fixes #1002


r10s (approved):
once line of code - impressive! look great!

how does the text color "know" it has to be changed? or is it just a 50% grey?

one tiny thing: if the quote is dismissed in dark-mode, there is a white background flickering. but maybe this is sth. we would not want to fix now, just before the release; if that would require changing background of some hidden element, this easily has other side effects.

PR #1014, opened 2020-11-30
remove side-effects and states from sendTextMessage()

### LINKED ISSUE #1013: sometimes messages are sent without text
it happens to me two times that i typed in some text and the message was sent empty, without text.

the second time, i quoted another message - the quote was sent, but not the text. in both cases, the message was encrypted - but as the quote could be decrypted, this is probably not the issue.

the message is shown as empty also on the sending device, so it seems not be related to message/receiving, my current theory is that `dc_msg_set_text()` was given an empty string, for whatever reason.

i am currently failing to find a way how to reproduce that.

maybe the input-field goes away, before the text is grabbed? as the message is sent in a thread? sort of race condition?

i anyone has an idea, this is very welcome :)

r10s:
this pr adds two parameters to sendTextMessage - text and quote - that are used to finally send the message. also, the function does not change state in the ui, this is moved to the caller.

this avoids race-conditions and other weird effects when controls and states are changed by other conditions.

the text to send is taken from didPressSendButtonWith argument, that should be same as messageInputBar.inputTextView.text, however.

fixes #1013

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -911,19 +911,14 @@ class ChatViewController: UITableViewController {
         }
     }
 
-    private func sendTextMessage() {
+    private func sendTextMessage(text: String, quoteMessage: DcMsg?) {
         DispatchQueue.global().async {
-            if let draftText = self.draft.draftText {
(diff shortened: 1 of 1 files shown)


cyBerta (approved):
nice and clear solution!

PR #1016, opened 2020-12-01
Staging area

cyBerta:


cyBerta:
@r10s do you have any comments / recommendations UI wise?
What bothers me is that the draft area covers all space (in addition the the keyboard) if there's a quote and a attachment and there's no way yet to hide the keyboard.

r10s:
> do you have any comments / recommendations UI wise?

- i suggest to left-align the image in the staging area, adding exactly the same amount of padding as for the quote (i have the feeling, left-align looks better than centering horizontally)
- a tap onto the image in the staging area should enlage the image (using the already existing enlage-view as used for avatars) (that are also not part of the real chat images)
- there is some probably unwanted scroll-down effect when there is a draft with an image and you leave the chat and re-enter. i've also seen that in other situations, but this is how it can be reproduced easily

> What bothers me is that the draft area covers all space (in addition the the keyboard) if there's a quote and a attachment and there's no way yet to hide the keyboard in that case.

but the keyboard is only visible if you type or? i'd say as long as there is at least _some_ place left for the chat (you need some space to scroll down and hide the keyboard) it's fine. and i think _some_ space is always left - here is a screenshot from iphone se:

cyBerta:
I would say this is ready for a review now.
In portrait layout the quote and the input text view adapt their size so that there's always enough place to swipe down the whole message bar to be able to read previous messages. In horizontal layout the input text view uses all available space above the keyboard just like in DC Android. The draft area is not visible anymore in that case.

cyBerta:
>     * a tap onto the image in the staging area should enlage the image (using the already existing enlage-view as used for avatars) (that are also not part of the real chat images)

oh wait, I forgot about that 🤦

r10s:
> oh wait

k :)

r10s:
> there is some probably unwanted scroll-down effect when there is a draft with an image and you leave the chat and re-enter. i've also seen that in other situations, but this is how it can be reproduced easily

i still can reproduce that issue. but we can also merge the pr in and do another issue/pr for that.

r10s (approved):

PR #1018, opened 2020-12-02
fix disappearing and reappearing of the message input bar 

### LINKED ISSUE #1017: input bar hides and reappears on scroll down on ios 14.2
Beginnig with the latest ios update (14.2) the keyboard behavior seems to have changed. Scrolling down in a chat leads now to a short animated dis- and reappearance of the message input bar which is not intended and needs to be fixed.

cyBerta:
... on scroll down in a chat

fixes #1017 

tested only with 14.2 so far

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -18,8 +18,6 @@ class ChatViewController: UITableViewController {
     var incomingMsgObserver: Any?
     var ephemeralTimerModifiedObserver: Any?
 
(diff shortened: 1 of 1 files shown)


r10s (approved):
tested that on 14.1 and 14.2 - seems to work :) [keyboardDismissMode is available since ios7](https://developer.apple.com/documentation/uikit/uiscrollview/1619437-keyboarddismissmode) (our target is ios11) - so, also from that side, np :)

the issue is not super-annoying, however, it looks better and more professional with the fix, so i'll do a 1.14.4 now.

PR #1023, opened 2021-01-04
New chat context menu

cyBerta:
follow up of #985


r10s:
although being still "WIP", one comment, not sure if you know about the issue:

if you delete a message, the input bar stays hidden - again the input bar :)

cyBerta:
thanks @r10s I'm already on it :)

r10s (changes requested):
overall, this is a great improvement! some tiny things:

- please reorder the context menu entries to Reply - Forward - Info - Copy- Delete

- the top/bottom padding of the focused-view-bubble is uneven. i know, we cannot change the display here easily, and i do not want to go that way.  
however, i assume, this uneven padding is also present in the normal-view - but it is just not visible because the top/bottom padding melt together there.  
so, i suggest to use the same amount of top/bottom padding, but use exactly one half atop and one abottom.  
that should not affect the normal-view but will improve the focused-view

- left/right padding is okay

if i am wrong, and the padding cannot be fixed this simple way, please only do the reorder and merge :)

r10s (approved):
lgtm, i rebased and needed to add a class as otherwise compiling fails, see https://github.com/deltachat/deltachat-ios/pull/1023/commits/ca5b88d8e670e25d5274494cf2c5d120e7e08a16 @cyBerta does that make sense to you?

PR #1026, opened 2021-01-06
fix truncated empty state lables 

### LINKED ISSUE #1009: sometimes truncated "empty" hints
the "no documents" hint may be truncated, depending on screen size and language, here is an example with german on an iphone7/se2:

we should also check if that can happen for the other hints (i am pretty sure, it can :)

cyBerta:
fixes #1009


deltachat-ios/View/FlexLabel.swift
```diff
@@ -1,84 +0,0 @@
...
-        addSubview(label)
-        label.translatesAutoresizingMaskIntoConstraints = false
-        label.centerYAnchor.constraint(equalTo: centerYAnchor, constant: 0).isActive = true
-        label.centerXAnchor.constraint(equalTo: centerXAnchor, constant: 0).isActive = true
-        label.topAnchor.constraint(equalTo: layoutMarginsGuide.topAnchor).isActive = true
-        label.bottomAnchor.constraint(equalTo: layoutMarginsGuide.bottomAnchor).isActive = true
-        label.widthAnchor.constraint(lessThanOrEqualTo: widthAnchor, multiplier: 0.95).isActive = true
-    }
-
-    ///FIXME: - replace this implementation, it's cutting off long texts, check PaddingTextView
```
cyBerta, line 59: there was already a fixme, to replace the old implementation...

r10s (commented):
lgtm

r10s (approved):

PR #1027, opened 2021-01-06
Adapt gallery image preview

### LINKED ISSUE #1025: weird enlarging of image in gallery
for the new gallery-context-menu, at https://github.com/deltachat/deltachat-ios/pull/985#issuecomment-726792484 i suggested that a tap inside the preview-image of the context-menu should directly enlarges the image; similar to the ios gallery.

some changes were done, however, the current state is not really nice :)

a tap on the preview-image first closes the preview with some animation and then, with another animation, the fullscreen-view is opened. this looks weird and is imho unexpected to the user. normally, animations should help to follow what is going on, this is not the case here.

so, we should either:
1. get rid of the useless stopover - the mentioned ios gallery does not do this stopover as well. please heave a look at the system gallery or on whatsapp, EDIT: i would not do 1. for now, see reasoning below
2. or close the context menu on a tap as before - that would be better as the current state as well. telegram does the same, just closes the context-menu

for 2., that is probably just some code-removal. so, i would be fine with 2. as well :)

cyBerta:
closes #1025

* for now we don't open the preview on tap on the image in the context menu
* however the transition animation looks now like a zoom-in/zoom-out of the image, the zoom effect starts from the original position of the selected cell in the grid

### DIFF
--- a/deltachat-ios/Controller/GalleryViewController.swift
+++ b/deltachat-ios/Controller/GalleryViewController.swift
@@ -254,13 +254,6 @@ extension GalleryViewController: UICollectionViewDataSource, UICollectionViewDel
             }
         )
     }
-
-    @available(iOS 13, *)
-    func collectionView(_ collectionView: UICollectionView, willEndContextMenuInteraction configuration: UIContextMenuConfiguration, animator: UIContextMenuInteractionAnimating?) {
-        if let msgId = (animator?.previewViewController as? ContextMenuController)?.msg.id {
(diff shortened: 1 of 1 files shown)


r10s:
> however the transition animation looks now like a zoom-in/zoom-out of the image, the zoom effect starts from the original position of the selected cell in the grid

i'd say that's fine and expected. it is the same as for the chat-preview (and also the same as eg. in the whatsapp chat-preview)

r10s (approved):
lgtm

PR #1028, opened 2021-01-06
don't show sender name in a quoted chat message 

### LINKED ISSUE #1012: quoting forwarded message no longer look as beeing forwarded
see https://github.com/deltachat/deltachat-android/issues/1743 fro details, reasoning and solution.

also the vertical bar should be displayed in unknown-sender-grey if the sender is known (either original message not available or forwarded message)

cyBerta:
... if quoted message was forwarded
* both for a sent message and a drafted message

closes #1012


r10s (approved):
lgtm

PR #1031, opened 2021-01-07
fix rebase

cyBerta:
* readd QLPreviewControllerDelegate method to GalleryViewController
* part of #1027


r10s (approved):
thanks for fixing my broken rebase 🙏

PR #1033, opened 2021-01-14
remove unused isIoRunning() api

r10s:
the api is about to be removed from core as well,
see https://github.com/deltachat/deltachat-core-rust/pull/2139


cyBerta (approved):
LGTM

PR #1035, opened 2021-01-18
vertically center message cells

cyBerta:
* improves the UI when a cell is highlighted in the context menu


r10s (approved):
yeah, MUCH better :)

PR #1036, opened 2021-01-19
Multi select

### LINKED ISSUE #1024: Add multi-select in chats
* add an option in the context menu to enable multi-select mode
* when multi-select mode enabled, replace messageInputBar by a control containing a 'delete' button, in a further iteration we'll add 'forward there as well
* add an option to cancel multi-selection: @r10s I would propose to add another 'cancel' button next to the delete button, what do you think?
* highlight selected messages, but for now there's no need to implement fancy UI to indicate the selection
* delete multiple messages on tap on 'delete'

cyBerta:
closes #1024 
implements multi deletion and multi forwarding


r10s:
i know, this is WIP, however: with the pr in the current form, opening PDF stop working, maybe also other attachments. images do work. just in case you did not notice that yet :)

r10s:
things that came up in review a/v
- trash bin should be system-icon, for ios <13, we just use the string "Delete" instead of the icon
- order of edit-more-menu-bar: "Cancel", Forward, Delete. "Cancel" should just be a text string
- in lite-mode, "delete" and "show more" have wrong colors, dark-mode is fine
- "Select more" should be the most-bottom item, there should be a divider between the other items and "Show more"
- long-press in select-more-mode should be disabled

cyBerta:
+ divider at the top

r10s:
@cyBerta i added the string `select_more` to master - so, if you rebase the pr you can directly adapt to it and drop it from deltachat-ios/en.lproj/Localizable.strings from this pr.

r10s:
looks great meanwhile!

just one more tiny thing: can we have the forward icon a bit smaller in the edit-more-menu-bar? looks bold like an hammer currently :) maybe there is also a stock-icon that can be used? if not, maybe just make it a but smaller.



_the one between "cancel" and the trashbin is a bit bold imho_

cyBerta:
@r10s I've reduced the size of the forward icon and rebased the branch onto master .

r10s (approved):
great, looks very good to me now 👍

r10s:
\o/

PR #1043, opened 2021-01-25
fix swipe-to-reply icon visibility issue on iOS 11 and 12

cyBerta:
I noticed that on iOS 11 and 12 the reply icon has been invisible in light mode (white icon on white background).
I tried several approaches to force iOS to show the original image, showing a black arrow, without success. On these iOS versions the images are used as templates and the original color is replaced by white. 
Instead of fighting against the intended behaviour I'm suggesting to use a system blue background color for the swipe-to-reply feature on iOS 11 and 12..
 iOS 13+ allows different colored icons, so we can keep there the black icon on white background and vice versa.

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -493,14 +493,15 @@ class ChatViewController: UITableViewController {
                                             self?.replyToMessage(at: indexPath)
                                             completionHandler(true)
                                         })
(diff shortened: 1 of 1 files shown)


r10s (approved):
great, and totally reasonable. does not even look bad imho :)

r10s:
came over https://github.com/deltachat/deltachat-ios/issues/1044 on testing ios12 emulator :)

PR #1047, opened 2021-02-01
disable swipe left-to-right in chats

### LINKED ISSUE #1044: unfunctional and unneeded swipe-delete-button on ios12
while testing https://github.com/deltachat/deltachat-ios/pull/1043, i noticed that you can   swipe-left messages on ios12  and an unfunctional, red delete-buttob appears. i think this button is just wrong here, we can maybe think to add sth. like that, however, the long-tap is currently just fine and enough.

cyBerta:
fixes #1044


r10s (approved):
lgtm

PR #1049, opened 2021-02-02
prepare 1.16.0

r10s:
just noticed, ios is the first platform where "show in chat" will be released (although it exists a bit longer on android :)

draft of the device message:

Screen Shot 2021-02-02 at 23 06 01


cyBerta (approved):
:rocket:

PR #1051, opened 2021-02-08
add mailinglist and bot-impersonation support

r10s:
- [x] show mailinglists a multi-user-chat (similar to groups)
- [x] adapt titles in chat view/profile
- [x] hide unused options from mailinglist profile (disappearing/memberlist/leave)
- [x] adapt to new `dc_decide_on_contact_request()` api
- [x] make sure `dc_contact_get_name()` is used on-purpose only, see https://github.com/deltachat/deltachat-core-rust/pull/2206
- [x] show overridden names using `dc_msg_get_override_sender_name()`, prefix overridden names by `~`

 Screen Shot 2021-02-16 at 18 50 21

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -687,6 +687,10 @@ public class DcChat {
         return type == DC_CHAT_TYPE_GROUP || type == DC_CHAT_TYPE_MAILINGLIST
     }
 
+    public var isMailinglist: Bool {
+        return Int(dc_chat_get_type(chatPointer)) == DC_CHAT_TYPE_MAILINGLIST
+    }
+
     public var isSelfTalk: Bool {
         return Int(dc_chat_is_self_talk(chatPointer)) != 0
     }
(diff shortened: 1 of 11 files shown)


r10s:
so, this is read for review now :)

deltachat-ios/Controller/BlockedContactsViewController.swift
```diff
@@ -43,7 +43,7 @@ class BlockedContactsViewController: GroupMembersViewController, GroupMemberSele
     func selected(contactId: Int, selected: Bool) {
         if !selected {
             let dcContact = DcContact(id: contactId)
-            let title = dcContact.displayName.isEmpty ? dcContact.email : dcContact.displayName
+            let title = dcContact.displayName
```
r10s, line 46: just came over that when grep'ing for displayName: when displayName is empty it is set to address by definition.

DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -832,6 +833,10 @@ public class DcMsg {
         return Int(dc_msg_get_chat_id(messagePointer))
     }
 
+    public var realChatId: Int {
+        return Int(dc_msg_get_real_chat_id(messagePointer))
```
cyBerta, line 837: that's a weird naming. But a core issue first of all.

cyBerta (approved):
nice! looks good!

r10s:
thanks for review and merge!

PR #1058, opened 2021-02-17
HTML Message view

### LINKED ISSUE #1046: add option to view full (html) message
this should be targeted after merging https://github.com/deltachat/deltachat-android/pull/1763, so that strings are available and we're already through most pitfalls :)

once merged, however,
1. add a button "Show full message" (string `show_full_message`) inside the bubble of messages when [`dc_msg_has_html()`](https://c.delta.chat/classdc__msg__t.html#aab7b069edb647e01311ee675b267d99a) is true (this can happen also for _outgoing_ messages, eg. when html-messages are forwarded or messages sent with another mua support)

2. a tap in the button shall open sth. as a `FullMsgViewController` that can display HTML, similar to the one we're already using for the `HelpViewController`. the latter has surprisingly little code, however, it is probably still reasonable to use a common base class as `WebViewViewController` (sic!) - sooner or later we will add sth. as "search", that way, all HTML-views will benefit from that (naming scheme is borrowed from android, btw, makes sense to keep that)

3. once `FullMsgViewController` is opened, configure the used [`WKWebView` to disable scripts](https://stackoverflow.com/questions/34404481/swift-wkwebview-disable-javascript) (they're enabled by default) and somehow to not load remote content (unless configured by the user differently, see below).  then, the HTML can be retrieved using [`dc_get_msg_html()`](https://c.delta.chat/classdc__context__t.html#aaedfc31d404f1adbd266cfd663ee78bc). i suggest to do this async to not block the ui.

4. in the upper-right-corner, add a button to toggle loading remote content. the button shall opens a `safeActionSheet` with the title=`load_remote_content`, text=`load_remote_content_ask` and the buttons `always`, `never`, `once` and the obligatory `cancel`. see the [android screenshot](https://github.com/deltachat/deltachat-android/pull/1763) to get an idea.
a setting as `Bool AlwaysLoadRemoteContent` should be persisted somewhere and respected on subsequent openings of the `FullMsgViewController`

related: https://github.com/deltachat/deltachat-desktop/issues/2078, https://github.com/deltachat/deltachat-android/pull/1763, https://github.com/deltachat/deltachat-core-rust/pull/2125

cyBerta:
closes #1046 

also fixed a small UI Bug with Draft preview images. Having the show complete message button left aligned the cell had a lot of empty space at the right side of the cell if the quoted message image was empty. 905f2fb7692f9ccf61f56fd74db57d8a865e24a9 fixes that.


r10s:
not sure if this is already ready-for-review, however: **great stuff!**

i did not look much at the code yet, so ui nitpicking:

- the "Show fill message" button is nice, i like it :) - if we leave it that simple, however, i would left-align it.

- the icon for "show messages" should be another icons as "sync", maybe a little image or a broken image or so, possibly sth. that is available as a stock image (i can also help on selecting - where are they browsable?)

- add a simple unicode-checkmark as ✔️ before one of "Once", "Always", "Never":
    - before "Always" if "Always" was selected before
    - before "Once" if once was tapped for this session
    - otherwise before "Never"

DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -996,6 +996,17 @@ public class DcMsg {
         return dc_msg_is_setupmessage(messagePointer) == 1
     }
 
+    public var hasHtml: Bool {
+        return dc_msg_has_html(messagePointer) == 1
+    }
+
+    public lazy var html: String = {
+        guard let cString = dc_get_msg_html(DcContext.shared.contextPointer, dc_msg_get_id(messagePointer)) else { return "" }
```
r10s, line 1004: `dc_get_msg_html()` belongs to the context object, not to the message object; if we move it there, we also do not need the deprecated DcContext.shared.contextPointer.

r10s (commented):
lgtm, apart of the tiny things mentioned.

DcCore/DcCore/DC/Wrapper.swift
cyBerta: I changed that

cyBerta:
one last thing, I not sure with: 
currently I do reset the 'load always remote content' state if I tap on 'once'. So if the state was 'always' and I tapped 'once' the next time I'll open the webview the checkmark will be at 'never'. I'm not sure it this is what we want.

r10s:
> currently I do reset the 'load always remote content' state if I tap on 'once'. So if the state was 'always' and I tapped 'once' the next time I'll open the webview the checkmark will be at 'never'.

this is exactly what we're doing on android and i think it makes much sense.

r10s (approved):
great!

plesae let me play around with "show full message" button padding before merging :)

PR #1059, opened 2021-02-18
fix send button state after video draft has been added

### LINKED ISSUE #1052: send button is disabled after selecting video
after selecting a video (not images) from the paper clip, the send button is not turning blue. only after typing (and deleting) one charakter i am able to send the video without text.

cyBerta:
fixes #1052


r10s (approved):
lgtm, not tested, however.

PR #1067, opened 2021-03-01
call stopIo() before imex()

### LINKED ISSUE #1065: Call stop_io()/start_io() before/after backup export
See https://github.com/deltachat/deltachat-core-rust/pull/2255

For Android, I already made a PR: https://github.com/deltachat/deltachat-android/pull/1822/files

For backup import it's not necessary as IO is not running before doing an import.

For key import/export, it doesn't hurt but isn't necessary.

r10s:
startIo() is already called when imex() ends, see addProgressAlertListener()

closes #1065


Hocuri (commented):
Stopping looks good, but don't you want to start IO again after imex?

Hocuri (approved):
Oops, just now read the description :)

PR #1069, opened 2021-03-02
add top padding to 'Show full message' button

r10s:
goal is to have roughly the same padding between
message and button as between two paragraphs.
(so that the button look more as belonging to the message
than to the last paragraph)

@cyBerta if that is okay, please merge this pr to message_web_view
and message_web_view to master then :)


cyBerta (approved):
lgtm

PR #1070, opened 2021-03-02
add default call iOS action if phone number was tapped

### LINKED ISSUE #1037: Cannot select highlighted phone number
Currently only the bubble is activated, but I can't directly interact with highlighted phone numbers

cyBerta:
closes #1037

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -1302,6 +1302,9 @@ extension ChatViewController: BaseMessageCellDelegate {
         if handleUIMenu() || handleSelection(indexPath: indexPath) {
             return
         }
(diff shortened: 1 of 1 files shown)


r10s:
having a number with a slash, only the part before the slash is called - whereas the number is correctly underlined, see first example:

image_2021-03-02_17-08-02

second example, plain number is okay, third one does no action at all - while also here, the underline seems to be correct.

cyBerta:
Thanks for the hint, @r10s 
We now filter non-numeric characters out, so that the generation of tel:// urls works correctly.
Tested with 
+49040123456
040/123456
12345-67

iOS detects correctly the country number code even if '+' is filtered out

r10s (approved):

PR #1072, opened 2021-03-05
mark overridden senders in quotes

r10s:
if the original author of a quote is shown
and known to be overridden,
that should be marked with the character `~`
as for the normal names,
esp. as they are look very similar otherwise,
eg. also bold, also colored.
the missing `~` looks like a bug here.

cmp. https://github.com/deltachat/deltachat-android/pull/1826


cyBerta (approved):
lgtm

PR #1073, opened 2021-03-07
show subject in html-view, if available

r10s:
after https://github.com/deltachat/deltachat-android/pull/1763 is merged, we can make the html-view even nicer :)

Screen Shot 2021-03-07 at 21 05 53


cyBerta (approved):
nice

PR #1078, opened 2021-03-08
Unsupported audio

### LINKED ISSUE #943: no error message when trying to play unsupported audio
ios cannot play eg. ogg-opus out-of-the box, however, as .ogg is defined as being an audio-message it is shown as such, including play-button etc.

there is no much we can/should do about that currently (esp. as in case of .ogg, other codecs will work), however, there should be an alert saying, the file cannot be played when the play-button is pressed. currently just nothing happens.

even if we get the codec to be supported, there will be other ones that do not work, so, this issue is not about making some codec work but just about saying that the file cannot be played.

EDIT: i added a testfile to "iOS DC" groups

cyBerta:
closes #943 
any error while trying to play an audio file will result in the following error message:

audio_error

### DIFF
--- a/deltachat-ios/Chat/AudioController.swift
+++ b/deltachat-ios/Chat/AudioController.swift
@@ -15,9 +15,15 @@ public enum PlayerState {
     case stopped
 }
 
+public protocol AudioControllerDelegate: AnyObject {
+    func onAudioPlayFailed()
+}
+
 /// The `AudioController` update UI for current audio cell that is playing a sound
 /// and also creates and manage an `AVAudioPlayer` states, play, pause and stop.
 open class AudioController: NSObject, AVAudioPlayerDelegate, AudioMessageCellDelegate {
+
+    open weak var delegate: AudioControllerDelegate?
 
     lazy var audioSession: AVAudioSession = {
         let audioSession = AVAudioSession.sharedInstance()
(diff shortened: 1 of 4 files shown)


cyBerta:
theroretically we could handle audio errors more specifically.
these are the possible errors and their descriptions:

```
/*!
    AudioFile error codes
    @abstract   These are the error codes returned from the AudioFile API.
    @constant   kAudioFileUnspecifiedError 
		An unspecified error has occurred.
    @constant   kAudioFileUnsupportedFileTypeError 
		The file type is not supported.
    @constant   kAudioFileUnsupportedDataFormatError 
		The data format is not supported by this file type.
    @constant   kAudioFileUnsupportedPropertyError 
		The property is not supported.
    @constant   kAudioFileBadPropertySizeError 
		The size of the property data was not correct.
    @constant   kAudioFilePermissionsError 
		The operation violated the file permissions. For example, trying to write to a file opened with kAudioFileReadPermission.
    @constant   kAudioFileNotOptimizedError 
		There are chunks following the audio data chunk that prevent extending the audio data chunk. 
		The file must be optimized in order to write more audio data.
    @constant   kAudioFileInvalidChunkError 
		The chunk does not exist in the file or is not supported by the file. 
    @constant   kAudioFileDoesNotAllow64BitDataSizeError 
		The a file offset was too large for the file type. AIFF and WAVE have a 32 bit file size limit. 
    @constant   kAudioFileInvalidPacketOffsetError 
		A packet offset was past the end of the file, or not at the end of the file when writing a VBR format, 
		or a corrupt packet size was read when building the packet table. 
    @constant   kAudioFileInvalidPacketDependencyError
		Either the packet dependency info that's necessary for the audio format has not been provided,
		or the provided packet dependency info indicates dependency on a packet that's unavailable.
    @constant   kAudioFileInvalidFileError 
		The file is malformed, or otherwise not a valid instance of an audio file of its type. 
    @constant   kAudioFileOperationNotSupportedError 
		The operation cannot be performed. For example, setting kAudioFilePropertyAudioDataByteCount to increase 
		the size of the audio data in a file is not a supported operation. Write the data instead.
    @constant   kAudioFileNotOpenError 
		The file is closed.
	@constant   kAudioFileEndOfFileError 
		End of file.
	@constant   kAudioFilePositionError 
		Invalid file position.
	@constant   kAudioFileFileNotFoundError 
		File not found.
 */
```

r10s (approved):

PR #1068, opened 2021-03-01
Prevent background crash

cyBerta:
fixes #1057 

fixes also a potential memory leak by removing the keyboard observer on viewDidDisappear

PR #1083, opened 2021-03-17
Chatlist update

### LINKED ISSUE #1079: Chatlist is not updated when coming back to the app
Can be easily seen when sending a message, it only gets updated if I wait for a while? and when going into a chat and coming back.

cyBerta:
fixes #1079 

and also ensures the chat list is updated every minute while the chat list is visible


r10s (approved):
lgtm, that uses the same method as introcuded in https://github.com/deltachat/deltachat-ios/pull/1068, right? so we probably do not run into the same issues :)

cyBerta:
exactly :)

PR #1086, opened 2021-03-18
Group creation issues

### LINKED ISSUE #975: group creation issues
creating a group has some minor issues, some may regarded as bugs even:

1. when using "New contact" in the process, an additional one-to-one-chat is created with the added contact, this is unexpected

2. when entering an exiting contact in the "New contact" dialog, this contact is not added as a member

3. when hitting "New contact" and the search contains an email-address,  we should use that to prefill the "New contact" dialog

4. when leaving the "New contact" dialog via "Done" button, currently, we go back to the "Add members" screen, the user has no clue if the added member is added or not. it is better to go back to the "Create group" screen where the user can directly see the added member (we should not forget to also add normally checked members)  
this is one click more when adding another member, but i think, this is more straight-forward.

nb: szenario i have in mind: 10 ppl sitting together and want a collective group; each ppl tells the group creator their email address. or the group-creator has a list of addresses that should go to a group.

cyBerta:
targets #975

not implemented:

> when leaving the "New contact" dialog via "Done" button, currently, we go back to the "Add members" screen, the user has no clue if the added member is added or not. it is better to go back to the "Create group" screen where the user can directly see the added member (we should not forget to also add normally checked members)
this is one click more when adding another member, but i think, this is more straight-forward.

The newly created member has a check mark in the member list in the AddMembersViewController. So all it's needed is to scroll in the contact list or search for the contact to check if the new created member is added. At the same time it might be unexpected that a user doesn't get back where they came from. Plus adding further contacts are done with one tap less.

### DIFF
--- a/deltachat-ios/Controller/AddGroupMembersViewController.swift
+++ b/deltachat-ios/Controller/AddGroupMembersViewController.swift
@@ -186,7 +186,7 @@ class AddGroupMembersViewController: GroupMembersViewController {
 
     private func showNewContactController() {
         let newContactController = NewContactController(dcContext: dcContext)
-        newContactController.openChatOnSave = false
+        newContactController.createChatOnSave = false
         navigationController?.pushViewController(newContactController, animated: true)
     }
 }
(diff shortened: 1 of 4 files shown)


r10s (approved):
point 4. in #975 is not targeted by this pr, however, we can also do that in a subsequent pr if we still think, it is useful (i'd say so :)

PR #1089, opened 2021-03-19
keep scroll position 

### LINKED ISSUE #1087: resume a chat with same scroll position
After coming back from the chat profile or the HTML view controller, the chat should be displayed at the scroll position it was left.

cyBerta:
when navigating from chat to html view or chat profile

closes #1087


r10s:
tested, lgtm :)

PR #1093, opened 2021-03-19
start IO on didReceiveRemoteNotification

r10s:
when didReceiveRemoteNotification, the app may be in "suspended" state, which is sort of a freezed app, with gracefully stopped IO - but still in memory, so context is up and existing.

however, to actually get messages, IO needs to be started.

successor of #889


cyBerta (approved):
+1

PR #1095, opened 2021-03-20
NotificationManager

cyBerta:
initial implementation

this PR only moves the code that generates notifications out of event.swift because the code in that file is only reponsible for emitting events that other components listen to. 

adding code to jump to the right message will be done in a subsequent PR

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -60,6 +60,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         appCoordinator = AppCoordinator(window: window, dcContext: dcContext)
         locationManager = LocationManager(context: dcContext)
         UIApplication.shared.setMinimumBackgroundFetchInterval(UIApplication.backgroundFetchIntervalMinimum)
+        notificationManager = NotificationManager()
         dcContext.maybeStartIo()
         setStockTranslations()
 
@@ -296,7 +297,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
                 self?.maybeRegisterForRemoteNotifications()
             }
         }
-        notificationManager = NotificationManager()
     }
 
     // register on apple server for receiving push notifications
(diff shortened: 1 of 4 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -295,6 +296,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
                 self?.maybeRegisterForRemoteNotifications()
             }
         }
+        notificationManager = NotificationManager()
```
r10s, line 299: `registerForNotifications()` may be called multiple times; that would overwrite existing NotificationManager objects, not sure if that is what we want to do.

r10s (commented):
i think, we should decouple the NotificationManager from registerForNotification and also created the object before we start IO - otherwise, we may miss notifications, esp. on startup.

tbh, i am still in favor of _not_ doing this observer-thingie for notifications, this just adds another indirection over apple api without any benefit (as far as i can see :)

i would still prefer having just an `NotificationManager.addNotification(chatId, msgId)` function that is called by `handleEvent()` - less code and less things that can go wrong, esp. during statup/creation. another argument for a simple function is that we would need a `removeNotifications(chatId)` anyway (there is no event for that).

(this `addNotification()`/`removeNotifications()` would be roughly the same what we are doing on android then)

but anyway, maybe the observer-thingie is fine as well, however, the decoupling from registerForNotification() seems to be needed, then this pr can be merged in to not block other things :)

deltachat-ios/AppDelegate.swift
cyBerta: fixed

cyBerta:
My long term vision for the notifications would be to move all related notification code to the notification manager, so that we have everything at one place and got clarity about where to look at if we want to change sth. to the notifications. This would mean that the notification manager implements `UNUserNotificationCenterDelegate` and `func application( _ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data)` delegates for example to a method of the notification manager. 

What I don't like about the idea of calling the nofication manager directly from events.swift is that events.swift is  part of the DcCore package whereas (IMO) the notification manager shouldn't be part of it. It is UI code that is only relevant for the main app, but not not for any extension. 
I agree that we need to care about the registration and removal of observers though. I changed the code the way you proposed to register unconditionally the notification manger once.

r10s (approved):
lgtm now, i can live with the observer-thingie for notifications, np :)

PR #1096, opened 2021-03-20
update documentation, add some notify measures to debug view

r10s:


r10s:
merging that in, just for a testflight :)

PR #1105, opened 2021-03-25
Notification jump to message

### LINKED ISSUE #721: tapping notifications do not open the correct chat

cyBerta:
* implements jumping from a notification to the message
* ensures foreground notifications (before not working at all, since the method signature was wrong), are shown in a non-intrusive way. Instead of showing an alert for each incoming message, (the last) unread message is only shown in the message center.

fixes #721

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -129,6 +129,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
     func applicationWillEnterForeground(_: UIApplication) {
         logger.info("---- foreground ----")
+        logger.debug("---- UNUserNotificationCenter.delegate:  \(UNUserNotificationCenter.current().delegate)---- ")
         appIsInForeground = true
         dcContext.maybeStartIo()
         if reachability.connection != .none {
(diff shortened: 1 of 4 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -413,23 +413,23 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
     // MARK: - Handle notification banners
 
-    private func userNotificationCenter(_: UNUserNotificationCenter, willPresent _: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
+    func userNotificationCenter(_: UNUserNotificationCenter, willPresent _: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
```
r10s, line 416: is this callback about the "foreground notifications" that are working now? what are "foreground notifications" at all?

maybe add some lines of documentation about when this is called and why we call the completion handler in the given way.

deltachat-ios/AppDelegate.swift
```diff
@@ -413,23 +413,23 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
-        completionHandler([.alert, .sound])
+        if #available(iOS 14.0, *) {
+            completionHandler([.list, .badge])
+        } else {
+            completionHandler([.badge])
+        }
     }
 
-    private func userNotificationCenter(_: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
+    func userNotificationCenter(_: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
```
r10s, line 425: also here - maybe add a line about that this function is called when the notification is tapped.

r10s (approved):
in general, this lgtm, apart form the question wrt "foreground notification", but this is more an understanding problem :)

i would add some more documentation (as we did for some other functions in AppDelegate).

another thing: there is a useless animation when tapping a notification - the chatlist is shown shortly, and then the chat appears - this is no blocker i'd say, however, if this is not fixed within this pr, please file an issue for it :)

PR #1108, opened 2021-03-25
Improve notification layout

### LINKED ISSUE #1107: improve notification layout
* notifications should show a preview image if an image is attached to the message
* notifications should show group names for group chats

cyBerta:
on top of #1095 (which needs to be reviewed and merged first)

closes #1107 

notification for a group:

Notification1

notification for a group - image message without text:
Notification2

notification for 1-1 chat from image message with text:
Notification3

### DIFF
--- a/deltachat-ios/Helper/NotificationManager.swift
+++ b/deltachat-ios/Helper/NotificationManager.swift
@@ -17,35 +17,37 @@ public class NotificationManager {
                let messageId = ui["message_id"] as? Int {
                 let chat = DcContext.shared.getChat(chatId: chatId)
                 if !UserDefaults.standard.bool(forKey: "notifications_disabled") && !chat.isMuted {
-                    let content = UNMutableNotificationContent()
-                    let msg = DcMsg(id: messageId)
-                    content.title = chat.isGroup ? chat.name : msg.getSenderName(msg.fromContact)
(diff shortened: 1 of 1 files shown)


deltachat-ios/Helper/NotificationManager.swift
```diff
@@ -19,11 +19,28 @@ public class NotificationManager {
...
+                    content.title = chat.isGroup ? chat.name : msg.getSenderName(msg.fromContact)
+                    content.body =  msg.summary(chars: 80) ?? ""
+                    content.subtitle = chat.isGroup ?  msg.getSenderName(msg.fromContact) : ""
                     content.userInfo = ui
                     content.sound = .default
 
+                    if msg.type == DC_MSG_IMAGE || msg.type == DC_MSG_GIF,
+                       let url = msg.fileURL {
+                        do {
+                            // make a copy of the file first since UNNotificationAttachment will move attached files into the attachment data store
```
r10s, line 31: does that run in the ui-thread (that may get some anr then) or in some background-thread?

not sure what `OperationQueue.main` means here suggest also here to add a documentation line atop, even if everything is already fine.

r10s (approved):
in general, this lgtm. maybe it is needed to dispatch the longer-lasting image copy to some background thread, see comment below.

nb: i rebased the pr on master.

PR #1119, opened 2021-03-31
more more to bg-thread

r10s:
precaution against eg. getFreshMessages() or taking longer.

- the IncomingMsgsObserver did already use a thread, this pr moves more to the thread (UserDefaults is thread-safe https://developer.apple.com/documentation/foundation/userdefaults)

- the MsgsNoticedObserver also spans a thread

(review with hiding whitespace)

### DIFF
--- a/deltachat-ios/Helper/NotificationManager.swift
+++ b/deltachat-ios/Helper/NotificationManager.swift
@@ -28,20 +28,20 @@ public class NotificationManager {
             forName: dcNotificationIncoming,
             object: nil, queue: OperationQueue.main
         ) { notification in
-            if let ui = notification.userInfo,
-               let chatId = ui["chat_id"] as? Int,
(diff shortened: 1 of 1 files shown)


cyBerta:
@r10s unfortunately, we're not allowed to call UIApplication.applicationIconBadgeNumber from a background thread. It produces at least runtime warnings: `UIApplication.applicationIconBadgeNumber must be used from main thread only`. I'll fix that in a subsequent PR.

r10s:
> @r10s unfortunately, we're not allowed to call UIApplication.applicationIconBadgeNumber from a background thread. It produces at least runtime warnings: UIApplication.applicationIconBadgeNumber must be used from main thread only. I'll fix that in a subsequent PR.

good catch! that explains some missing badge updates i encountered recently...

PR #1124, opened 2021-04-01
multiple notifications

### LINKED ISSUE #1106: show multiple notifications
if it's easy to implement we should allow to show more than one notification in the message center.

cyBerta:
* show multiple background notifications, threaded per chat
* remove notifications if corresponding chat was opened
closes #1106 
* removes foreground notifications feature, see https://github.com/deltachat/deltachat-ios/pull/1124/commits/ed87be56092f446b465177af351202d70b304f7e: iOS handles foreground notifications visually differently than background notifications. It puts foreground notifications above grouped background notifications as a list of single notifications (thus the foreground notification option .list). This clutters the notification area (that you can reach by pulling down from the top of the screen) extremely fast. There's no API to force those foreground notifications being grouped even though they've got a threadIdentifier. After all, grouped background notifications seem to be more valuable than foreground notifications, so we remove the latter feature in favor of getting the former done right.

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -426,6 +426,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     // MARK: - Handle notification banners
 
     // This method will be called if an incoming message was received while the app was in forground.
+    // We don't show foreground notifications in the notification center because they don't get grouped properly
     func userNotificationCenter(_: UNUserNotificationCenter, willPresent notification: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
         logger.info("forground notification")
         completionHandler([.badge])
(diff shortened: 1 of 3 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -426,20 +426,14 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     // MARK: - Handle notification banners
 
     // This method will be called if an incoming message was received while the app was in forground.
-    // On iOS 14+ we show a non-intrusive notification in the notification center, if the notification doesn't belong to a currently opened chat.
-    // On pre iOS 14 this feature isn't available so we're just triggering to show the badge at the launcher icon. This will only have an effect for unread messages > 0
     func userNotificationCenter(_: UNUserNotificationCenter, willPresent notification: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
```
r10s, line 429: maybe add some lines from the reasoning from this pr to the comment. so we do not make the same mistake again :)

cyBerta: done

r10s (approved):
lgtm, also tested that for a day -- what a huge improvement!

r10s:
two thing i noticed during testing:

- if there are **notifications for two different chats,** ChatA and ChatB:
   - tap on ChatA
   - Delta Chat opens with ChatA
   - do not leave ChatA but tap on ChatB in the notification center
   - Delta Chat opens with ChatB
   - hit the "back" button in the upper right cornder
   - -> **instead of going to the chatlist, ChatA opens,** i'd say, this is unexpected

- if there are **multiple notifications for the same chat, the last one is shown outside the notification center** - is that on-purpose? or even the normal ios behavior?<br />

however, both can maybe better be targeted in another pr, i would like to merge this one in to keep reviewing easier :)

Issue #1085, opened 2021-03-18
check for anr in chatlist


r10s:
sometimes, there are anr in charlist and chat, maybe because of incoming messages, maybe because of markseen, not sure, this does not happens always and i did not find a way to reproduce that, unfortunately.

a symptom of that is that views may be opened several times, eg. when hitting the "open chat" button during an anr several times, the chat is opened several times as well (you need to hit "back" several times to go back to the chatlist then). however, idea is to get anr fixed, not the symptom :)

r10s:
@dignifiedquire reported that the used sqlite-core-library in combination with the limited number of threads may cause problems:

due to async, sql-requests are scheduled over the threads available, the scheduler does not take care much about the calling thread - so it may happen, that a theoretically fast call takes some seconds as it is randomly in the same thread-queue as eg. some longer taking operation that is done before.

the "real" fix could be sth. as using sqlx in core, however, we could mitigate the issue by avoiding calls to core in ui-thread (where possible) and also by taking care not doing lots unneeded core calls (we had that here and there by these "computed properties" where `sth.text` does not look like an expensive call - but in reality is a long-taking core call that is evaluated over and over. also, that is easily mixed with `lazy` ... i consider  "computed properties" harmful meanwhile ;)

@dignifiedquire is that summary roughly correct? i am not sure if i got it correctly :)

r10s:
meanwhile, there is sqlx coming, so we should reconsider that once https://github.com/deltachat/deltachat-core-rust/pull/2089 is merged and/or in use on ios.

r10s:
after some days of running sqlx on my testphone, note sure, if sqlx helps on these anr, at least not alone.

first, i had the feeling, things are a bit faster, but that seemed to be a false impression, resulting just from the restart of the app.

to reproduce, it is needed that the app is running for some hours, then from inside the app, hit the home button (not: swipe away to exit) and the hit the app-icon. that results in an anr of sometimes several seconds pretty reliably (at least on my 3gig db).  
unfortunately, i do not know of an easier/faster way to get the anr.

general impression is, that i got anr only for the top-level viewcontrollers, so, qr, chatlist, top-settings. being eg. in the chat, things are smoother.

meanwhile, i dived a bit in the code:
- in appDelegate, there is a maybeNetwork that should be moved to a thread, again, that alone is not the solution
- ChatlistViewModel::refreshData is called often, maybe we should debounce that

might be, the anr is not solvable by a single code-change.

[...]

PR #1127, opened 2021-04-06
call maybeNetwork() from non-ui thread, fix background thread

### LINKED ISSUE #1128: declare background task later and redeclare them as needed
currently, we are declaring the background task already on startup, in installEventHandler() and getting a warning as follows:

```
2021-04-06 18:22:37.504670+0200 deltachat-ios[5801:4433962] [BackgroundTask] Background Task 1 ("Called by deltachat-ios, from $s13deltachat_ios11AppDelegateC19installEventHandleryyFyycfU_"), was created over 30 seconds ago. In applications running in the background, this creates a risk of termination. Remember to call UIApplication.endBackgroundTask(_:) for your task in a timely manner to avoid this.
```

so, ios, seems to use this also as a time-based-marker, might be that this was different in older os.

https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtask say, the background task can be declared at "any point", however, also speaks of "mark the beginning". i suggest to move the registering to start of  applicationDidEnterBackground.

moreover, the current logic allows only one background task at all - then the app has to be terminated and restarted to have another background task. in practise, however, apps are suspended nowadays.

r10s:
targets #1085 
closes #1128 

for review, compare commit-by-commit.

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -175,26 +175,14 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         }
     }
 
+    func applicationWillResignActive(_: UIApplication) {
+        logger.info("⬅️ applicationWillResignActive")
+        registerBackgroundTask()
+    }
+
     func applicationDidEnterBackground(_: UIApplication) {
         logger.info("⬅️ applicationDidEnterBackground")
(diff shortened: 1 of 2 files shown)


r10s:
merging that in to move forward, @cyBerta please have a look at the changes once you are back to keyboard and fine time for it :)

PR #1138, opened 2021-04-08
fix chatlist anr

### LINKED ISSUE #1085: check for anr in chatlist

r10s:
call the correct removeObserver functions:
    
- `removeObserver(observer)` only for observers with block-closure,
- `removeObserver(self)` for others.

when the wrong function is called, the observer is not removed - and the next time the view gets active, a second, third, fourth observer is added ... this results in things being executed tens of times.

this issue was introduced by recent update fixes as of https://github.com/deltachat/deltachat-ios/pull/1083.

TODO, maybe in another pr:
- check for similar issues for the other observers
- check if the remaining call to getChatlist-on-active can be improved further (partly moved to a background thread or so) (EDIT: started that at https://github.com/deltachat/deltachat-ios/pull/1140)

closes #1085

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -152,7 +152,8 @@ class ChatListController: UITableViewController {
             nc.removeObserver(msgsNoticedObserver)
         }
         // remove non-block observers
-        nc.removeObserver(self)
+        NotificationCenter.default.removeObserver(self, name: UIApplication.didBecomeActiveNotification, object: nil)
+        NotificationCenter.default.removeObserver(self, name: UIApplication.willResignActiveNotification, object: nil)
     }
     
     // MARK: - setup
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -12,9 +12,6 @@ class ChatListController: UITableViewController {
     private var msgChangedObserver: Any?
     private var msgsNoticedObserver: Any?
     private var incomingMsgObserver: Any?
-    private var viewChatObserver: Any?
-    private var foregroundObserver: Any?
-    private var backgroundObserver: Any?
 
     private weak var timer: Timer?
 
@@ -127,14 +124,16 @@ class ChatListController: UITableViewController {
             queue: nil) { [weak self] _ in
                 self?.viewModel.refreshData()
         }
-        foregroundObserver = nc.addObserver(self,
-                                            selector: #selector(applicationDidBecomeActive(_:)),
-                                            name: UIApplication.didBecomeActiveNotification,
-                                            object: nil)
-        backgroundObserver = nc.addObserver(self,
-                                            selector: #selector(applicationWillResignActive(_:)),
-                                            name: UIApplication.willResignActiveNotification,
-                                            object: nil)
+        nc.addObserver(
+            self,
+            selector: #selector(applicationDidBecomeActive(_:)),
+            name: UIApplication.didBecomeActiveNotification,
+            object: nil)
+        nc.addObserver(
+            self,
+            selector: #selector(applicationWillResignActive(_:)),
+            name: UIApplication.willResignActiveNotification,
+            object: nil)
     }
 
     override func viewDidDisappear(_ animated: Bool) {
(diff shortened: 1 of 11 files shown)


cyBerta:
@r10s thanks for pushing this topic forward!
I haven't found yet any references in apples documentation that show the distinction you're making here (or I read them but didn't understand them correctly). Do you have any links?

cyBerta:
for reference: https://developer.apple.com/documentation/foundation/notificationcenter/1413994-removeobserver

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -151,12 +151,8 @@ class ChatListController: UITableViewController {
...
             nc.removeObserver(msgsNoticedObserver)
         }
-        if let foregroundObserver = self.foregroundObserver {
-            nc.removeObserver(foregroundObserver)
-        }
-        if let backgroundObserver = self.backgroundObserver {
-            nc.removeObserver(backgroundObserver)
-        }
+        // remove non-block observers
+        nc.removeObserver(self)
```
cyBerta, line 155: we can do that in this case, but need to be careful, if not all observers are added and removed at the same step of the lifecycle. The documentation recommends therefore: 

> When removing an observer, remove it with the most specific detail possible. For example, if you used a name and object to register the observer, use removeObserver(_:name:object:) with the name and object.

cyBerta (approved):
🌈 great🌈 
I left a comment about removing observers. Maybe we can agree to use `nc.removeObserver(self)` only if it's called in dealloc?

r10s:
> Maybe we can agree to use nc.removeObserver(self) only if it's called in dealloc?

hm, but `viewDidDisappear` is not "dealloc". or what do you mean with "dealloc"?

but this add/remove observer is weird, i need a bit deeper understanding i fear :)
i just noticed that we were adding observers but not removing them - and the changed fixed things :)

cyBerta:
> > Maybe we can agree to use nc.removeObserver(self) only if it's called in dealloc?
> 
> hm, but `viewDidDisappear` is not "dealloc". or what do you mean with "dealloc"?
>

sorry I mean the method deinit() that can be called to free resources before an object gets deallocated: 
https://docs.swift.org/swift-book/LanguageGuide/Deinitialization.html

For simple classes or views we could call the more generic `nc.removeObserver(self)` in `deinit`  as it is ensured that no observer will be needed in those classes if deinit has been called. 

Wrt. UIViewControllers we might register and remove observers in different lifecycle hooks, for example some in viewWillAppear/viewWillDisappear others in viewDidAppear/viewDidDisappear. In those cases calling `nc.removeObserver(self)` cannot be done in viewWillDisappear, because observers that should be removed in viewDidDisappear would be removed too early.

So my proposal is to remove observers using a specified name if we're not already deallocating an object. See https://developer.apple.com/documentation/foundation/notificationcenter/1415360-addobserver in the Discussion section

`To unregister an observer, use removeObserver(_:) or removeObserver(_:name:object:) with the most specific detail possible. For example, if you used a name and object to register the observer, use the name and object to remove it.`

> i just noticed that we were adding observers but not removing them - and the changed fixed things :)

That is true. I approved the PR because it does the right thing. 👍

r10s:
indeed, it seems to be better to call removeObserver(self) only in deinit, when using NotificationCenter.default.removeObserver instead of nc.removeObserver, xcode also prints a warning:

Screen Shot 2021-04-10 at 23 30 00

moreover, i've found two similar issues and w save the observer-handle in `NSObjectProtocol?` instead of `Any?` now to catch these errors sooner - swift allows to assign "no return value" to `Any?` ...

@cyBerta i am merging that in to move forward, please check the code and the changes anyway after merging.

cyBerta:
@r10s very cool! that will boost up the app! Thanks for investigating this issue.

PR #1139, opened 2021-04-09
set bgIoTimestamp also on init

r10s:
the main purpose of the timestamp is to block rapidly subsequent calls
to remote- or local wakeup;
therefore set it also on init.

moreover, also on init, we're in background,
so it makes also sense from that perspective.

### DIFF
--- a/deltachat-ios/Controller/SettingsController.swift
+++ b/deltachat-ios/Controller/SettingsController.swift
@@ -488,10 +488,6 @@ internal final class SettingsViewController: UITableViewController, ProgressAler
 
         if let appDelegate = UIApplication.shared.delegate as? AppDelegate {
             info += "notify-token=\(appDelegate.notifyToken ?? "<unset>")\n"
-
-            let timestampDbl = appDelegate.bgIoTimestamp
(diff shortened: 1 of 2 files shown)


deltachat-ios/Controller/SettingsController.swift
```diff
@@ -488,10 +488,6 @@ internal final class SettingsViewController: UITableViewController, ProgressAler
 
         if let appDelegate = UIApplication.shared.delegate as? AppDelegate {
             info += "notify-token=\(appDelegate.notifyToken ?? "<unset>")\n"
-
-            let timestampDbl = appDelegate.bgIoTimestamp
-            let timestampStr = timestampDbl==0.0 ? "<unset>" : DateUtils.getExtendedRelativeTimeSpanString(timeStamp: timestampDbl)
-            info += "last-bg-io=\(timestampStr)\n"
```
r10s, line 494: as we set it also on init, i do not see any usecase of having that in the log, but lots of misinterpretion.

r10s:
merging that in to move forward ...

PR #1158, opened 2021-04-14
allow different senders in one-to-one-chats

### LINKED ISSUE #1145: add alias support
once https://github.com/deltachat/deltachat-core-rust/pull/2297 is merged, we need to check for [`dc_msg_get_override_sender_name()`](https://c.delta.chat/classdc__msg__t.html#a47c8d53848c954277926f456919c7eab) even in one-to-one-chats.

if there is an override-sender-name, _and only then_, the name prefixed by a `~` and the avatar should be shown (as for groups and mailinglists).

if there is no override-sender-name, name and avatar should stay hidden for one-to-one-chats.

see also https://github.com/deltachat/deltachat-android/pull/1840, https://github.com/deltachat/deltachat-desktop/issues/2215

r10s:
closes #1145


cyBerta (approved):
lgtm

PR #1162, opened 2021-04-17
make repo's local issue template work

### LINKED ISSUE #1161: ISSUE_TEMPLATE.md is ignored
GitHub shows in page https://github.com/deltachat/deltachat-ios/blob/master/.github/ISSUE_TEMPLATE.md  the message:

```
You are using an old version of issue templates. Please update to the new issue template workflow
```
the local template is not used, instead templates from https://github.com/deltachat/.github/blob/master/.github/ISSUE_TEMPLATE/ are used, 

solution could be to just remove the local template and use the organization-level templates but they are too generic, and looks weird that you ask users if they are using Windows, Android, etc in the iOS repo, so it makes sense to actually make the local template work

adbenitez:
close #1161

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,3 +1,11 @@
+---
+name: Bug report
+about: Report something that isn't working.
+title: ''
+assignees: ''
+
+---
+
 <!--
 This is a bug report tracker. New features are discussed in the forum: https://support.delta.chat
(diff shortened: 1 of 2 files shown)


r10s (approved):
nice, thanks for figuring that out - so there was probably a recent change for the ISSUE_TEMPLATE? or was that always wrong?

r10s:
ahh, i just read the corresponding issue, there you explained that :)

PR #1171, opened 2021-04-26
add daymarkers to chats

### LINKED ISSUE #1117: re-add daymarkers
the new-day-headlines were lost somehow :)

cyBerta:
closes #1117

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -509,11 +509,18 @@ class ChatViewController: UITableViewController {
         _ = handleUIMenu()
 
         let id = messageIds[indexPath.row]
+        if id == DC_MSG_ID_DAYMARKER, messageIds.count > indexPath.row + 1 {
+            let nextMessageId = messageIds[indexPath.row + 1]
+            let nextMessage = DcMsg(id: nextMessageId)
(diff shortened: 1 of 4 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -509,11 +509,18 @@ class ChatViewController: UITableViewController {
         _ = handleUIMenu()
 
         let id = messageIds[indexPath.row]
+        if id == DC_MSG_ID_DAYMARKER, messageIds.count > indexPath.row + 1 {
```
r10s, line 512: on errors, if, for whatever reason the daymarker is the last item, this condition will fail - and then an invalid message object will be created below with `let message = DcMsg(id: id)`.

i suggest to always add a daymarker when the first condition is true; if the second condition fails, however, fall back to a standard text as "ErrDaymarker" - usually, this should not happen anyway.

nb: since https://github.com/deltachat/deltachat-core-rust/pull/1677 we can even get the timestamp from the daymarker itself, however, that would require some larger refactoring. i'd say if we really see "ErrDaymarker", we can change that in a subsequent pr at some later point :)

r10s (approved):
lgtm, i would change the error handling, however.

deltachat-ios/Chat/ChatViewController.swift
cyBerta: I fixed that

PR #1172, opened 2021-04-26
search only verified contacts when adding new members to verified groups

### LINKED ISSUE #1103: Verified group: adding a member i can search for non verified contacts
I have a verified group, i go to the group info screen and press "Add member" then i type something in the search and also non verified contacts show up.

cyBerta:
closes #1103


r10s (approved):
lgtm

PR #1173, opened 2021-04-26
adjust quality-of-service for background task

r10s:
~~`.userInteractive`~~ `.userInitiated` should be used for background tasks affecting the UI
and may block the user.
not sure, if that makes a huge difference, ist just came over that -
and i have the feeling, html-pages are loaded faster now :)

https://developer.apple.com/documentation/dispatch/dispatchqos


cyBerta:
We can try that. I would probably use .userInitiated in most cases, because I read .userInteractive as a thread dedicated for animations. Loading data or writing to the database on that thread might block them.

cyBerta (approved):

r10s:
you're right, .userInitiated fits better. i will change that, it is still 3 "quality-classes" better than .background (which is wrong in these cases)

PR #1174, opened 2021-04-27
handle register for notifications after scan to create account

### LINKED ISSUE #1136: registerForNotifications not called after qr-account-scan
`registerForNotifications()` seems not to be called when an account was scanned from a qr code.

we should also check that for backup-import.

cyBerta:
fixes #1136


r10s (approved):
lgtm

PR #1176, opened 2021-04-27
improve location permission handling

### LINKED ISSUE #1137: location streaming messages in log although location streaming is disabled
my log shows entries as 

```
22:40:33.262 💚 DEBUG LOCATION MANAGER: didChangeAuthorization: CLAuthorizationStatus
22:40:34.184 💚 DEBUG LOCATION: didUpdateLocations
22:40:34.185 💚 DEBUG LOCATION: isBetterLocation: true
22:40:34.191 💚 DEBUG LOCATION: didUpdateLocations
22:40:34.191 💚 DEBUG LOCATION: isBetterLocation: true
22:40:34.200 💚 DEBUG LOCATION: didUpdateLocations
22:40:34.200 💚 DEBUG LOCATION: isBetterLocation: true
22:40:34.202 💚 DEBUG LOCATION: didUpdateLocations
22:40:34.202 💚 DEBUG LOCATION: isBetterLocation: true
```

although global location streaming option is disabled. i assume nothing happens at the end as `isSendingLocationsToChat` is probably false, however, we get some locations and also the system dialog for locations may pop up (of course, the latter only as i have enabled at some point before)

however, it would be better to not do anything with locations and not create the location-manager at all if location streaming is disabled.

nb: would be nice if `isSendingLocationsToChat` is shown somewhere in `get_info`, maybe from core.

cyBerta:
fixes #1137

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -1100,7 +1100,6 @@ class ChatViewController: UITableViewController {
         guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else {
             return
         }
-        self.dcContext.sendLocationsToChat(chatId: self.chatId, seconds: seconds)
         appDelegate.locationManager.shareLocation(chatId: self.chatId, duration: seconds)
     }
 
--- a/deltachat-ios/Controller/SettingsController.swift
+++ b/deltachat-ios/Controller/SettingsController.swift
@@ -396,6 +396,7 @@ internal final class SettingsViewController: UITableViewController, ProgressAler
         let title = locationStreaming ?
             "Disable on-demand location streaming" : String.localized("pref_on_demand_location_streaming")
         alert.addAction(UIAlertAction(title: title, style: .default, handler: { [weak self] _ in
+            guard let self = self else { return }
             UserDefaults.standard.set(!locationStreaming, forKey: "location_streaming")
             if !locationStreaming {
                 let alert = UIAlertController(title: "Thanks for trying out the experimental feature 🧪 \"Location streaming\"",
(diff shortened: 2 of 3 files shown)


r10s (approved):
this prevents unexpected location updates by disabling all location-requests on disabling the feature, right?  
so, if i got the problem in the past, i have to enable and disable again? that would be okay, esp. as it is experimental :) and definitely an improvement.

one question to the CLLocationManagerDelegate - that object only takes a bit of memory when not in use? can you double-check that? what we really should avoid is unexpected location-warnings popping up if the location-streaming was never enabled.

cyBerta:
> this prevents unexpected location updates by disabling all location-requests on disabling the feature, right?
> so, if i got the problem in the past, i have to enable and disable again? that would be okay, esp. as it is experimental :) and definitely an improvement.

The app does now disable all location streaming if the feature gets disabled. Besides that, location streaming gets disabled completely if the last ongoing location streaming period was expired or the user disabled the last ongoing location streaming.

cyBerta:
> what we really should avoid is unexpected location-warnings popping up if the location-streaming was never enabled.

This shouldn't happen anymore. The permission handling flow was buggy before: if the user allowed once location streaming, the app would start the location engine on each app start. This has been fixed by only starting the location streaming if the core tells the UI that there's any location streaming chat enabled: https://github.com/deltachat/deltachat-ios/pull/1176/files#diff-3acfeb56e644e8d4a825d4865e02f0df09187a6bc3fc1a58732f1a288c23bba7R96

cyBerta:
> that object only takes a bit of memory when not in use?

The location manager allocates 80 bytes, CLLocationManager another 32 bytes. So that should be fine.

r10s:
thanks for re-checking and clarifying!

PR #1177, opened 2021-04-27
on show_emails=all, open contact-request-chat

r10s:
if all emails are handled by Delta Chat,
there will typicaly be much more "contact requests" as usual.

handling these contact requests in the stacked way, one by one,
is some pain.
for a first improvement, we just open the contact-request-chat directly,
the user can pick a contact request from there.

more improvements are in disussion, see
https://github.com/deltachat/deltachat-android/issues/1896
and recent irc discussions.

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -287,7 +287,12 @@ class ChatListController: UITableViewController {
         switch cellData.type {
         case .deaddrop(let deaddropData):
             safe_assert(deaddropData.chatId == DC_CHAT_ID_DEADDROP)
-            showDeaddropRequestAlert(msgId: deaddropData.msgId)
+            if dcContext.showEmails == DC_SHOW_EMAILS_ALL {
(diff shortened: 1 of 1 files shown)


r10s:
i did the same for ios at https://github.com/deltachat/deltachat-ios/pull/1177/files

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -287,7 +287,12 @@ class ChatListController: UITableViewController {
         switch cellData.type {
         case .deaddrop(let deaddropData):
             safe_assert(deaddropData.chatId == DC_CHAT_ID_DEADDROP)
-            showDeaddropRequestAlert(msgId: deaddropData.msgId)
+            if dcContext.showEmails == DC_SHOW_EMAILS_ALL {
```
r10s, line 290: this extra-handling is probably temporary. once that is settled (see recent dc and irc discussions), we may again use the same approach for all `show_emails` setting.

cyBerta (approved):

PR #1179, opened 2021-04-28
fix app crash if saved messages chat gets deleted

### LINKED ISSUE #1142: crash when deleting save-messages as the last chat
the app crashes when saved-messages chat is the only chat and that is deleted then.

(the core adds a device message about how the saved-messages chat can be recreated, so there will be one chat again; maybe it is related to that)

cyBerta:
fixes #1142

### DIFF
--- a/deltachat-ios/ViewModel/ChatListViewModel.swift
+++ b/deltachat-ios/ViewModel/ChatListViewModel.swift
@@ -23,7 +23,7 @@ protocol ChatListViewModelProtocol: class, UISearchResultsUpdating {
     var emptySearchText: String? { get }
 
     /// returns ROW of table
-    func deleteChat(chatId: Int) -> Int
+    func deleteChat(chatId: Int) -> Int?
     func archiveChatToggle(chatId: Int)
     func pinChatToggle(chatId: Int)
     func refreshData()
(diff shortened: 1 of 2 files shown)


deltachat-ios/ViewModel/ChatListViewModel.swift
```diff
@@ -201,13 +201,18 @@ class ChatListViewModel: NSObject, ChatListViewModelProtocol {
...
+        let chat = dcContext.getChat(chatId: chatId)
         dcContext.deleteChat(chatId: chatId)
         updateChatList(notifyListener: false)
         safe_assert(indexToDelete != nil)
-        return indexToDelete ?? -1
+        // Do not return the index of the tableview cell for the selfTalk chat:
+        // Immediately after deleting the chat, a message to the device talk chat gets added.
+        // If the device talk chat was not existing it would be created and added to the table data source.
+        // That case invalidates the index.
+        return chat.isSelfTalk ? nil : indexToDelete
```
r10s, line 215: hm, pretty special knowledge :)

i think, however, we're doing far too much here. [dc_delete_chat()](https://c.delta.chat/classdc__context__t.html#ad50eed96a11b113c4080886b2b748ff3) already emits the event DC_EVENT_MSGS_CHANGED that we observe and update the chatlist - neither the updateChatList() a few lines above is needed nor the index return value.

r10s:
i did a mini-pr while testing the things i wrote above - https://github.com/deltachat/deltachat-ios/pull/1183

r10s (approved):
lgtm - for the skipped animation:

- we do not have animations or pin/unpin/archive, so that skipped animation kind of streamlines things :)  

- we get reduced complexity and removed a bug and potential subsequent bugs (if core decides not to add device message or so)

- i think, if really wanted, we can re-introduce animations in another pr, but then in a more compatible way, maybe by checking for simple changes (delete, swap) on the update-event. but that has no priority currently :)

PR #1180, opened 2021-04-28
always refresh messages after forwarding to saved messages chat

### LINKED ISSUE #1175: forwarding to "saved messages" does not show the forwarded message
when forwarding a message to "saved messages", the message is not displayed in the chat directly.

however, the message is forwarded leaving and re-entering the chat will show the chat. might be, some update-event is missing from core, android does the saving a bit differently (cmp. #1039), so that is not directly comparable.

as the "saved messages" chat is an essential ui element, i would regard this bug as a blocker.

cyBerta:
fixes #1175

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -976,6 +976,7 @@ class ChatViewController: UITableViewController {
         let chat = dcContext.getChat(chatId: self.chatId)
         if chat.isSelfTalk {
             RelayHelper.sharedInstance.forward(to: self.chatId)
+            refreshMessages()
         } else {
             confirmationAlert(title: String.localizedStringWithFormat(String.localized("ask_forward"), chat.name),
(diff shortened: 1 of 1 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -976,6 +976,7 @@ class ChatViewController: UITableViewController {
         let chat = dcContext.getChat(chatId: self.chatId)
         if chat.isSelfTalk {
             RelayHelper.sharedInstance.forward(to: self.chatId)
+            refreshMessages()
```
r10s, line 979: wondering why that is not needed below.

r10s (approved):
lgtm, anyway :)

deltachat-ios/Chat/ChatViewController.swift
cyBerta: I bet there's a event missing if dc_forward_msgs is called for the self chat

PR #1182, opened 2021-04-28
Read receipts

### LINKED ISSUE #1178: read receipts are sent when chat is opened while screen turns off
read receipt should only be sent when messages are really displayed to the user.

if the chat is open but the screen is turned off, no read receipt should be sent.

if later the screen is turned on again, and the user sees the messages, the read receipts should be sent.

i assume this is what really happen when filing https://github.com/deltachat/deltachat-ios/issues/1074

cyBerta:
fixes #1178


r10s (commented):
lgtm, also tested a bit :)

r10s (approved):

PR #1183, opened 2021-04-28
simplify chat deletion

r10s:
that should be merged into https://github.com/deltachat/deltachat-ios/pull/1179

dc_delete_chat already emits DC_EVENT_MSGS_CHANGED
that we are observing and update the chatlist on changes -
so, there is nothing special to do beside calling dc_delete_chat :)

archive/pin also might do superfluous things, i have not checked that, that can go to another pr, however.

EDIT: that may skip a remove-animation, however, as we do not have that for "archive" as well (which is probably used as many times), that'd be fine. if desired, in another pr, we can re-add animation by _first_ doing ui-cell-removal and in its closure, finally calling deleteChat.

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -458,9 +458,7 @@ class ChatListController: UITableViewController {
             return
         }
 
-        if let row = viewModel.deleteChat(chatId: chatId) {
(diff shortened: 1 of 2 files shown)


cyBerta:
This works, but removes the row animation. For this reason I left that code there.

cyBerta:
anyways, it reduces the complexity. So I think it's worth it, too.

cyBerta (approved):

PR #1185, opened 2021-04-29
Narrow chat bubbles

### LINKED ISSUE #1147: small images lead to too narrow bubbles
small images lead to too narrow bubbles. sth. was changed in this area, not sure, though.

expected behavior would be some minimal bubble size. when fixing that, we should do some research about what was fixed before and if that is still working :)

cyBerta:
fixes #1147 

sets the minimal width of images with text to 175 dp, same for landscape images without text, see the image below 

Screen Shot 2021-04-29 at 4 59 02 PM

We might increase the minimal width even a little bit more. That can be done easily by changing a constant.

### DIFF
--- a/deltachat-ios/Chat/Views/Cells/ImageTextCell.swift
+++ b/deltachat-ios/Chat/Views/Cells/ImageTextCell.swift
@@ -4,7 +4,7 @@ import DcCore
 import SDWebImage
 
 class ImageTextCell: BaseMessageCell {
-
+    let minImageWidth: CGFloat = 175
     var imageHeightConstraint: NSLayoutConstraint?
     var imageWidthConstraint: NSLayoutConstraint?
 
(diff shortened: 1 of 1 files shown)


r10s (approved):
i am wondering if it would not be better to let the text define the width of the cell; this is what we are doing on android. if not text is set, we can leave it smaller (a minimal width is still useful, but could be smaller then)

however, if that gets too complicated, we can merge that in as is, it is already an improvement and mitigates things with less effort - so maybe that's sufficient. most images are larger, so that is an cornercase anyway.

cyBerta:
I've checked file messages and file+text messages. Those message bubbles grow and shrink with the message text size.

cyBerta:
I would like to merge that in and iterate over it later on.

r10s:
> I would like to merge that in and iterate over it later on.

yeah, let's see, maybe it is sufficient and "good enough" :)

PR #1186, opened 2021-04-30
adapt chat bubble size

cyBerta:
* reduces minimal image width
* scales small images depending on message text width

Screen Shot 2021-04-30 at 1 06 05 PM
Screen Shot 2021-04-30 at 1 06 43 PM

follow up of https://github.com/deltachat/deltachat-ios/pull/1185

* personally, I like the design of https://github.com/deltachat/deltachat-ios/pull/1185 more

### DIFF
--- a/deltachat-ios/Chat/Views/Cells/ImageTextCell.swift
+++ b/deltachat-ios/Chat/Views/Cells/ImageTextCell.swift
@@ -4,7 +4,7 @@ import DcCore
 import SDWebImage
 
 class ImageTextCell: BaseMessageCell {
-    let minImageWidth: CGFloat = 175
+    let minImageWidth: CGFloat = 125
     var imageHeightConstraint: NSLayoutConstraint?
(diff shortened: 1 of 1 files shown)


r10s:
hm, not sure. i think, avoiding messages getting too high (what will happen when text is too narrow) is also very important, maybe more important than having a maybe ugly image.

however, i would expect sth. as `< bubble-width` instead of `< 125` - in the example above when the image has a width of 126, the image would again be narrow, or? btw. this is probably also true for  #1185.

cyBerta:
> however, i would expect sth. as `< bubble-width` instead of `< 125` - in the example above when the image has a width of 126,

we don't calculate the overall bubble-width, it's depending on constraints. currently there's no way to get the value for 'bubble-width' / text width before or during the cells have been calculated and drawn. 

But we could define sth.  like half of the screen instead of '< 125'

r10s (approved):
nice idea, that could work, indeed :)

PR #1187, opened 2021-04-30
vertically align text in sending status bar

### LINKED ISSUE #1148: delivery-status-line not vertically centered in its container
for images (and probably also videos, gifs, stickers etc.) sent without additional text, the delivery line is integrated in the image, which is great and saves some screen space.

to make it stand out on any background, there is some transparent gray added below the delivery-status-line.

however, the text is not centered vertically in this background:

expected behavior would be to have that vertically centered.

reported by @Simon-Laux and others :)

cyBerta:
fixes #1148

Screen Shot 2021-04-30 at 4 43 29 PM
Screen Shot 2021-04-30 at 4 43 43 PM
Screen Shot 2021-04-30 at 4 44 13 PM


deltachat-ios/Assets.xcassets/ic_lock.imageset/Contents.json
```diff
@@ -1,23 +1,23 @@
 {
   "images" : [
     {
+      "filename" : "ic_lock_1x.png",
       "idiom" : "universal",
-      "filename" : "ic_lock_lightergrey_1x.png",
```
r10s, line 5: what is that change about?

just a renaming? wondering as the file size changed.

deltachat-ios/Chat/Views/Cells/BaseMessageCell.swift
```diff
@@ -417,11 +417,19 @@ public class BaseMessageCell: UITableViewCell {
 
     private func attachPadlock(to text: NSMutableAttributedString, color: UIColor? = nil) {
         let imageAttachment = NSTextAttachment()
+        let cachedImageKey = "ic_lock\(color != nil ? "_color" : "")"
+        let cachedImage = ThumbnailCache.shared.restoreImage(key: cachedImageKey)
```
r10s, line 421: ThumbnailCache is also used for gallery items and video thumbnails - they use a url that is guaranteed not to be never ic_lock or so?

r10s (commented):
visually, this looks good to me.

what is the reasons to use an additional cache here? have we some findings that the current approach is too slow? or is this kind of best-practice? asking as caches tend to produce problems and should not be used without reason.

also, please double-check that keys for the delivery icons cannot be mixed with the other items in the cache - not that we have a user-send-image as delivery icon somehow.

if in doubt, i would do the cache things in another pr.

deltachat-ios/Assets.xcassets/ic_lock.imageset/Contents.json
cyBerta: I've changed the images. They had a lot of empty space above the key icon, so that it was annoying to align the icon correctly.

cyBerta:
> what is the reasons to use an additional cache here? have we some findings that the current approach is too slow? or is this kind of best-practice? asking as caches tend to produce problems and should not be used without reason.

I agree. The reason is that scrolling through the chat messages could be smoother. I realized that we do for each message bubble several image editing operations, like cropping, scaling or tinting. Of course, the best option would be to have better fitting image source sets as svg from where we derive all client images, but that needs some coordination with the other clients I guess. 

I will double check the performance improvement, and may revert the last commit about caching.

PR #1204, opened 2021-05-11
remove unused ChatListViewModelProtocol

r10s:
this pr removes the last unused `ViewModelProtocol`, we decided some time ago, to add complexity only when we have a use of that. as a rule of thumb, complex classes, used only once and splitting code over objects and files are questionable.

also removing `ChatListViewModel`, however, produces too much noise currently and may result in bugs, i would not go for that now. we can continue refactoring, when we are actually working on that part (we thought be would by the recent fixes, the they're related to other parts :)

overhead is already reduced by 1/3rd, so i suggest to merge that without further refactoring :)

### DIFF
--- a/deltachat-ios/ViewModel/ChatListViewModel.swift
+++ b/deltachat-ios/ViewModel/ChatListViewModel.swift
@@ -133,7 +133,6 @@ class ChatListViewModel: NSObject {
         return nil
     }
 
-    // needed to differentiate betweeen deaddrop / archive / default
     func chatIdFor(section: Int, row: Int) -> Int? {
         let cellData = cellDataFor(section: section, row: row)
         switch cellData.type {
--- a/deltachat-ios/ViewModel/ChatListViewModel.swift
+++ b/deltachat-ios/ViewModel/ChatListViewModel.swift
@@ -1,35 +1,9 @@
 import UIKit
 import DcCore
 
-// MARK: - ChatListViewModelProtocol
-protocol ChatListViewModelProtocol: class, UISearchResultsUpdating {
-
-    var onChatListUpdate: VoidFunction? { get set }
-
-    var isArchive: Bool { get }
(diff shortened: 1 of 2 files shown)


deltachat-ios/ViewModel/ChatListViewModel.swift
```diff
@@ -42,10 +16,10 @@ class ChatListViewModel: NSObject, ChatListViewModelProtocol {
         case messages
     }
 
-    var isArchive: Bool
+    private(set) public var isArchive: Bool
```
cyBerta, line 19: interesting, I haven't seen that syntax yet, a little bit confusing to read though.

is there a real reason to change the visibility here?

deltachat-ios/ViewModel/ChatListViewModel.swift
```diff
@@ -158,6 +133,7 @@ class ChatListViewModel: NSObject, ChatListViewModelProtocol {
         return nil
     }
 
+    // needed to differentiate betweeen deaddrop / archive / default
```
cyBerta, line 136: this comment is not so much adding new information (like about why) and the switch-case is quite easy to read in my opinion.

r10s: you're right, i will remove it. i copied it down without thinking much about it :)

deltachat-ios/ViewModel/ChatListViewModel.swift
r10s: i tried just to do was the protocol was doing - making `isArchive` public read and private read/write.

the syntax `private(set) public` is already used in some pods, so if you grep for it, you will find some things.

but i would also be fine in just making it `public` and not introduce a new syntax variant.

cyBerta (approved):

PR #1205, opened 2021-05-11
show audio file duration in audio message bubbles

### LINKED ISSUE #1201: total length of voice messages not displayed before playback
after a voice message is received, usually, the total time should be displayed in the message bubble before playback. currently, always `0:00` is shown there.

when playback is started, the current playing position replaces the total time, this is correct (and this is also what whatsapp does and delta chat android).

so, no need to add an additional time display; just show the total time when playback is not started.

cyBerta:
fixes #1201

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -969,8 +969,8 @@ public class DcMsg {
         return CGFloat(dc_msg_get_width(messagePointer))
     }
     
-    public var duration: Double {
-        return Double(dc_msg_get_duration(messagePointer))
+    public var duration: Int {
+        return Int(dc_msg_get_duration(messagePointer))
     }
 
     public func setLateFilingMediaSize(width: CGFloat, height: CGFloat, duration: Int) {
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -968,6 +968,10 @@ public class DcMsg {
     public var messageWidth: CGFloat {
         return CGFloat(dc_msg_get_width(messagePointer))
     }
+    
+    public var duration: Double {
+        return Double(dc_msg_get_duration(messagePointer))
+    }
 
     public func setLateFilingMediaSize(width: CGFloat, height: CGFloat, duration: Int) {
         dc_msg_latefiling_mediasize(messagePointer, Int32(width), Int32(height), Int32(duration))
@@ -1009,6 +1013,10 @@ public class DcMsg {
 
     public func setDimension(width: CGFloat, height: CGFloat) {
         dc_msg_set_dimension(messagePointer, Int32(width), Int32(height))
+    }
+    
+    public func setDuration(_ duration: Double) {
+        dc_msg_set_duration(messagePointer, Int32(duration))
     }
 
     public var filesize: Int {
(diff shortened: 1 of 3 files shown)


deltachat-ios/Chat/AudioController.swift
```diff
@@ -85,6 +85,34 @@ open class AudioController: NSObject, AVAudioPlayerDelegate, AudioMessageCellDel
...
     }
+    
+    public func getAudioDuration(messageId: Int, successHandler: @escaping (Int, Double) -> Void) {
+        let message = DcMsg(id: messageId)
+        if playingMessage?.id == messageId {
+            // irgnore messages that are currently playing or recently paused
+            return
+        }
+        
+        if let fileURL = message.fileURL {
```
r10s, line 96: before trying to get the duration from the file (which may be expensive), we should check [`dc_msg_get_duration()`](https://c.delta.chat/classdc__msg__t.html#aaedb426314f3d701d606ba3daa24fc1a) that _may_ also return the duration. 

only if that return 0 for "unknown", we should parse the file.

both checks, however, should be async inside `getAudioDuration()`

r10s (approved):
lgtm, but consider checking if core already knows the duration.

deltachat-ios/Chat/AudioController.swift
cyBerta: nice, I wasn't aware of that api

cyBerta:
@r10s I've added the calls to `dc_msg_get_duration` and `dc_msg_set_duration` as you proposed

deltachat-ios/Chat/AudioController.swift
```diff
@@ -93,22 +93,29 @@ open class AudioController: NSObject, AVAudioPlayerDelegate, AudioMessageCellDel
...
+            } else if let fileURL = message.fileURL {
+                let audioAsset = AVURLAsset.init(url: fileURL, options: nil)
+                audioAsset.loadValuesAsynchronously(forKeys: ["duration"]) {
+                    var error: NSError?
+                    let status = audioAsset.statusOfValue(forKey: "duration", error: &error)
+                    switch status {
+                    case .loaded:
+                        let duration = audioAsset.duration
+                        let durationInSeconds = Double(CMTimeGetSeconds(duration))
+                        message.setDuration(1000 * durationInSeconds)
```
r10s, line 111: i assume, the goal of that line is to write the changes to the database for later, faster access?

this does not work that way, `dc_msg_set_*()` methods only change the message object, usually to send it or save it as a draft later.

to actually change the values in the database, use [`dc_msg_latefiling_mediasize()`](https://c.delta.chat/classdc__msg__t.html#a7687ff969841f3d00c3a212f9ad27861)

DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -968,6 +968,10 @@ public class DcMsg {
     public var messageWidth: CGFloat {
         return CGFloat(dc_msg_get_width(messagePointer))
     }
+    
+    public var duration: Double {
```
r10s, line 972: why not use the same type as the underlying core function here? so an Integer?

deltachat-ios/Chat/AudioController.swift
cyBerta: thanks for the hint, I've fixed that

DcCore/DcCore/DC/Wrapper.swift
cyBerta: I changed it to Int

PR #1206, opened 2021-05-11
[WIP] fix background fetch crash

### LINKED ISSUE #1202: background handler terminated by 0xdead10cc sometimes
getting "Namespace RUNNINGBOARD, Code 0xdead10cc" crashes from time to time, i cannot reproduce that, but it was reported by @dignifiedquire; some logs are in  App Store Connect, excerpt

```
Exception Type:  EXC_CRASH (SIGKILL)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Termination Reason: Namespace RUNNINGBOARD, Code 0xdead10cc
Triggered by Thread:  0
```

[full log](https://gist.github.com/r10s/9da620f127a718e346635786a625585e) - so, there seems sth. wrong with background thread handling, searching for "Namespace RUNNINGBOARD, Code 0xdead10cc" brings up some advices, need to investigate further.

EDIT: to sum up, the issue with 0xdead10cc is that a lock is held on some files (probably sql-database) after the app ~~gets~~ signals to be suspended. reason might be dispatched jobs that are still running for a little moment, we tried to mitigate that by recent commits and to exclude this cause. however, might also be that "shared" usage of the database is the cause, that would be very hard to fix, see discussions wrt share-to-delta.

stats of recent versions:

Screen Shot 2021-05-11 at 10 45 13

[compared with older stats](https://github.com/deltachat/deltachat-ios/issues/1129#issuecomment-821821375), we had 700 installations and only 1 crash/7 days (at least after removing sqlx)

ftr, 0xdead10cc was also topic in https://github.com/deltachat/deltachat-ios/issues/1057

cyBerta:
* execute completion handler sequencially after maybenetwork was called
this is a proposal for a minor refactoring and attempt to fix #1202

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -368,27 +368,23 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         }
         bgIoTimestamp = nowTimestamp
 
-        // we're in background, run IO for a little time
-        dcContext.maybeStartIo()
-
         // maybeNetwork() shall not be called in ui thread;
-        // even if we're in backround now, app may be starting just now and maybeNetwork() will block in that case
-        // (series of performFetchWithCompletionHandler/applicationWillEnterForeground are not rare)
         DispatchQueue.global(qos: .background).async { [weak self] in
+            self?.dcContext.maybeStartIo()
             self?.dcContext.maybeNetwork()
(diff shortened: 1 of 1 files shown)


r10s:
replaced by https://github.com/deltachat/deltachat-ios/pull/1208

PR #1207, opened 2021-05-11
[WIP] close database on suspend

r10s:
maybe an alternative to #1206, targets #1202; let's discuss that tomorrow in an a/v :)


r10s:
well, it is not as easy as this pr suggests:
- i got a crash releated to that pr, for whatever reason, did not dived into that further, maybe related to
- we would also have to shutdown event handlng and restart it, that has issues on its own (eg. https://github.com/deltachat/deltachat-core-rust/issues/2280)

r10s:
replaced by https://github.com/deltachat/deltachat-ios/pull/1208

PR #1210, opened 2021-05-13
hack around 0xdead10cc

r10s:
targets #1202

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -517,7 +517,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         dcContext.setStockTranslation(id: DC_STR_FORWARDED, localizationKey: "forwarded")
     }
 
-    private func appIsInForeground() -> Bool {
+    func appIsInForeground() -> Bool {
         switch UIApplication.shared.applicationState {
         case .background, .inactive:
             return false
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -202,6 +202,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             } else if app.backgroundTimeRemaining < 10 {
                 logger.info("⬅️ few background time, \(app.backgroundTimeRemaining), stopping")
                 self.dcContext.stopIo()
+
+                // to avoid 0xdead10cc exceptions, scheduled jobs need to be done before we get suspended;
+                // we increase the probabilty that this happens by waiting a moment before calling unregisterBackgroundTask()
                 DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
                     logger.info("⬅️ few background time, \(app.backgroundTimeRemaining), done")
                     self.unregisterBackgroundTask()
@@ -382,7 +385,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             if !self.appIsInForeground() {
                 self.dcContext.stopIo()
             }
-            completionHandler(.newData)
+
+            // to avoid 0xdead10cc exceptions, scheduled jobs need to be done before we get suspended;
+            // we increase the probabilty that this happens by waiting a moment before calling completionHandler()
+            DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
+                logger.info("⬅️ fetch done")
+                completionHandler(.newData)
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


r10s:
hm, that does not work as waitForExistingJobsDispatched is itself called from main, so, mainDoneSoFar never gets true. needs some more love ...

cyBerta:
Why don't you move the completion handler itself to the background queue? 

something like:

```
DispatchQueue.global(qos: .background).async {
            completionHandler(.newData)
}
```

deltachat-ios/AppDelegate.swift
```diff
@@ -382,7 +388,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             if !self.appIsInForeground() {
                 self.dcContext.stopIo()
             }
-            completionHandler(.newData)
+
+            // to avoid 0xdead10cc exceptions, scheduled jobs need to be done before we get suspended;
+            // we increase the probabilty that this happens by adding completionHandler()
+            // to the end of both dispatch queues.  
+            DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
```
cyBerta, line 395: What is the purpose of the +1 seconds here and  [here](https://github.com/deltachat/deltachat-ios/pull/1210/files#diff-fc4377fbb4c60bfc9a11f53fd448de91bbf34787141509ddb85242515fea9e23R209?)

r10s: just to give things a bit more time, to mitigate race conditions. might be that sth. gets dispatched from global to main queue, we would not catch that. 

there was a 3 second timeout before, but that appears too long.

r10s: another thing: according to [developer.apple.com](https://developer.apple.com/documentation/dispatch/dispatchqueue/2300077-global), `DispatchQueue.global(qos:)` may return different queues, so, the additional second may catch some cases where other queues are used we are not checking. 

but it might be advantageous to first do swap the two schedules: background may easily schedule to main, the other way round, probably not so often. also, the 1 second is always added _after_ background finishes and does not run concurrently.

deltachat-ios/AppDelegate.swift
```diff
@@ -382,7 +388,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             if !self.appIsInForeground() {
                 self.dcContext.stopIo()
             }
-            completionHandler(.newData)
+
+            // to avoid 0xdead10cc exceptions, scheduled jobs need to be done before we get suspended;
+            // we increase the probabilty that this happens by adding completionHandler()
+            // to the end of dispatch queues and waiting one second.
+            DispatchQueue.global(qos: .background).async {
+                DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
```
cyBerta, line 396: in the end this asyncAfter (no matter if scheduled from background or not) might be the essential change. 

[the docs](https://developer.apple.com/documentation/dispatch/dispatchqueue/2300077-global) are confusing me:

> Tasks submitted to the returned queue are scheduled concurrently with respect to one another. 

What I conclude: So even if background task A is submitted before background task B, they will run (and finish) concurrently.

I think now, moving the completion handler to the background task will not fix the issue that other background tasks might still be running.

cyBerta: Ok, I'm getting more clarity, so the _submission_ of the execution blocks are non-blocking (async) for the global queues. The _execution_ depends on `DispatchQueue.global(qos: .background).sync / DispatchQueue.global(qos: .background).async`

cyBerta: Since we always dispatch queues to be executed asynchronously (`DispatchQueue.global(qos: .background).async`) we have no guarantee that the completion handler will be executed and finish last in the background queue

cyBerta: If we would implement our own DispatchQueues, we could make use of the .barrier flag (they have no effect in global queues). That is, in a async queue, all tasks submitted before the execution block marked with the .barrier flag, would finish before that execution block is executed.

r10s: yip, the docs are confusing. as so often :)

but anyway, on av we decided to just rely on a timeout and see if that actually fixes the issue. next escalation would be to tweak DispatchQueue.global and maybe even adding a dc_suspend()

r10s:
i just played around a bit with `DispatchQueue.global.async` in the ios emulator:

- up to 64 jobs are started concurrently
- jobs are mostly started in the correct order, at least when there is little time between scheduling them
- when the last queued job is done, up to 63 other jobs may still be running.

so, an additional `DispatchQueue.global.async` around `DispatchQueue.main.asyncAfter` only makes sense if we assume there are lots of jobs queues. but then, it may make sense.

<details>
<summary>code used for testing</summary>

```
        for i in 0 ..< 100 {
            usleep(UInt32(i) * 1000)
            logger.info("😇 queuing job \(i)")
            DispatchQueue.global().async {
                logger.info("🟢 job \(i) started")
                usleep(UInt32(i) * 1000*1000)
                logger.info("🔴 job \(i) done")
            }
        }
```

</details>

cyBerta (approved):

PR #1211, opened 2021-05-13
Updates

cyBerta:
* fixes podfile structure and thus pod updating problems due to APPLICATION_EXTENSION_API_ONLY flag set in the share extension
* updates some of the pods: SDWebImage, swifty beaver (logging lib), DBDebugToolkit, swiftlint
* removes unused UICircularProgressRing dependency

I didn't remove the pods from git thuugh. There are contradicting recommendations and I think it's wise to follow that one: https://guides.cocoapods.org/using/using-cocoapods.html#should-i-ignore-the-pods-directory-in-source-control


r10s (approved):
great!

PR #1213, opened 2021-05-14
Always set timer to nil after being invalidated

cyBerta:
related to #1210 and #1202

it's a small refactoring, which still ensures that always only 1 timer is running. however this implementation invalidates always any running timer and starts a new one, whereas the old implementation skipped starting a timer, when there was already one running before. 

on stop the timer gets nil'ed now, that's probably the most important change

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -279,10 +279,13 @@ class ChatViewController: UITableViewController {
 
     private func startTimer() {
         stopTimer()
-        timer = Timer.scheduledTimer(withTimeInterval: 60, repeats: true) { [weak self] _ in
+        timer = Timer.scheduledTimer(withTimeInterval: 60, repeats: true) { _ in
             // reload table
-            DispatchQueue.main.async {
-                guard let self = self, let appDelegate = UIApplication.shared.delegate as? AppDelegate else { return }
+            DispatchQueue.main.async { [weak self] in
+                guard let self = self,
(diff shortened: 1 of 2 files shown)


r10s:
there is a merge conflict now :)

could you reproduce/say, the timer could really started multiple times by the old logic? or is this only a general precaution?

cyBerta:
no, I couldn't reproduce to get the timer started multiple times, it's more a precaution :-/

r10s:
seems as if `invalidate()` needs to be called from the same thread as it was created - maybe this is an issue?

see https://developer.apple.com/documentation/foundation/nstimer/1415405-invalidate (via https://stackoverflow.com/questions/16475686/nstimer-not-stopping-when-invalidated)

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -279,10 +279,13 @@ class ChatViewController: UITableViewController {
 
     private func startTimer() {
         stopTimer()
-        timer = Timer.scheduledTimer(withTimeInterval: 60, repeats: true) { [weak self] _ in
+        timer = Timer.scheduledTimer(withTimeInterval: 60, repeats: true) { _ in
```
r10s, line 282: sure that `[weak self]` can be removed here? as we're referring self a line below? maybe this is processed differently by swift magic, i do not know much about that, however, if in doubt, i would leave that line using `[weak self] _ in` :)

r10s (approved):
lgtm, we'll see what happens on the next testflight ...

PR #1216, opened 2021-05-17
further hacks around 0xdead10cc

cyBerta:
* register and unregister background tasks for each background fetch

related to #1202

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -16,6 +16,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     var locationManager: LocationManager!
     var notificationManager: NotificationManager!
     private var backgroundTask: UIBackgroundTaskIdentifier = .invalid
+    private var fetchBackgroundTasks = [Double: UIBackgroundTaskIdentifier]()
     var reachability = Reachability()!
     var window: UIWindow?
     var notifyToken: String?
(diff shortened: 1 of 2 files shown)


deltachat-ios/Helper/NotificationManager.swift
```diff
@@ -37,13 +39,31 @@ public class NotificationManager {
             NotificationManager.updateApplicationIconBadge(reset: false)
         }
     }
+    
+    private func unregisterNotificationBackgroundTask(msgId: Int) {
```
r10s, line 43: wondering if we can avoid these helper functions and just use a local variable.

deltachat-ios/Helper/NotificationManager.swift
```diff
@@ -55,7 +75,8 @@ public class NotificationManager {
                     if chat.isMuted {
                         return
                     }
-
+                    
+                    self?.registerNotificationBackgroundTask(msgId: messageId)
```
r10s, line 79: i think, this registering may come too late.

we should register the task before it is really started, otherwise we may get 0xdead10cc eg. in the getChat() call a line above.

deltachat-ios/AppDelegate.swift
```diff
@@ -345,6 +346,24 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         increaseDebugCounter("notify-local-wakeup")
         performFetch(completionHandler: completionHandler)
     }
+    
+    private func unregisterFetchBackgroundTask(timestamp: Double) {
```
r10s, line 350: wondering if we can avoid these helper functions and just use a local variable.

r10s (commented):
just some thoughts, i did a pr while playing around that targets these points, https://github.com/deltachat/deltachat-ios/pull/1219

r10s (approved):

PR #1217, opened 2021-05-17
Notification archived chats

### LINKED ISSUE #1215: notifications/badge not cleared if chat is archived
similar to https://github.com/deltachat/deltachat-ios/issues/1197

if you archive a chat with unread messages before opening it, the notifications are not removed from the notification center nor the app-icon-badge is updated.

both will show the invalid information until some other action is taken that forces an update.

(the core itself marks messages correctly as "noticed" on archiving, this led to even more confusion on unarchiving as there may be chat at all having unread messages - but badge shows there re some)

cyBerta:
* fixes #1215 
* and also more bugs around notifications and chats that get deleted in the ContactDetails and GroupChatDetails ViewController. 

I purposely didn't move  `NotificationManager.removeNotificationsForChat()` into DcContext.deleteChat / DcContext.archiveChat, because imo the DcCore module shouldn't handle notifications as it is sth. specific to the main app module.

### DIFF
--- a/deltachat-ios/Controller/ContactDetailViewController.swift
+++ b/deltachat-ios/Controller/ContactDetailViewController.swift
@@ -420,6 +420,7 @@ class ContactDetailViewController: UITableViewController {
             return
         }
         viewModel.context.deleteChat(chatId: viewModel.chatId)
+        NotificationManager.removeNotificationsForChat(chatId: viewModel.chatId)
 
         // just pop to viewControllers - we've in chatlist or archive then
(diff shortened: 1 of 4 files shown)


r10s:
> I purposely didn't move NotificationManager.removeNotificationsForChat() into DcContext.deleteChat / DcContext.archiveChat, because imo the DcCore module shouldn't handle notifications as it is sth. specific to the main app module.

yip, i agree. the DcCore should do the basics and should be very close to the c-ffi.

r10s (approved):

PR #1218, opened 2021-05-17
Make lint happy

cyBerta:
fixes some lint issues


r10s (approved):
great! i was always wondering where this set-before-get-order comes from :)

Issue #1157, opened 2021-04-13
fix background poll and and finish it asap


r10s:
currently, `performFetchWithCompletionHandler` and `didReceiveRemoteNotification` let run the core IO for 10 seconds and terminate then.

it would be better to terminate once a fetch on the imap folders INBOX and DeltaChat is done (there is some info-message, but no dedicated event, maybe core needs some adaptions) _AND_ there are no more pending jobs (eg. sending messages out).

that way:
- we would usually return after a second or so and would get called faster again
- if a fetch really takes 10~20 seconds, that can be completed as well (after ~20 seconds, however, better terminate IO gracefully and call the completion handler properly) (absolute max. would be 30 seconds, however, at that point we risk to be brute-force-terminated and not to be called soon again)

r10s:
some numbers:

on my iphone7, in the last 23½ h:
- 143 wakeups in total (68 remote, 75 local, in average one wakeup every 10 min)
- this should be 143 * 11 s (10 s wait, 1 s terminate) background time = **26 min**
- battery stats say 36 min in background, however, i assume, this includes the 30 s after leaving the app - so i have left the app 20 times (36 min - 26 min = 10 min -> 20 * 30 s) today, which is reasonable

so, assuming an fetch with no data is not much more than one second, the **26 min** can probably be reduced to a fragment - which, all in all, should get us a better reputation to be called more often (i do not expect to be called more than every 10 min, but there are phones with more apps where we are currently called less often)

cyBerta:
This issue should be postponed until multi-account is implemented

r10s:
just had a look, for the code, https://github.com/deltachat/deltachat-ios/pull/1221/files is still a pretty nice pattern.

PR #1220, opened 2021-05-18
stop performFetch by a function

r10s:
still targets 0xdead10cc, #1202

i did not create an observer yet as this would mean additional complexity. however, when we need it because of #1157, all these changes will already help.

we need to take care that `performFetch()` always calls the `completionHandler`, even if `performFetch()` is called concurrently. this is mainly achieved by the time-check, allowing this function only once per minute. otherwise, things would be even more complicated.

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -16,6 +16,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     var locationManager: LocationManager!
     var notificationManager: NotificationManager!
     private var backgroundTask: UIBackgroundTaskIdentifier = .invalid
+    private var fetchBackgroundTask: UIBackgroundTaskIdentifier = .invalid
+    private var fetchBackgroundTaskEnding = false
     var reachability = Reachability()!
     var window: UIWindow?
     var notifyToken: String?
@@ -360,6 +362,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         //
         // therefore, if last fetch is less than a minute ago, we skip this call;
         // this also lets the completionHandler being called earlier so that we maybe get awakened when it makes more sense.
+        // moreover, this ensures every competionHandler belongs to one backgroundTask.
         //
         // nb: calling the completion handler with .noData results in less calls overall.
         // if at some point we do per-message-push-notifications, we need to tweak this gate.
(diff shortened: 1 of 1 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -386,25 +389,41 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
-            }
 
-            // to avoid 0xdead10cc exceptions, scheduled jobs need to be done before we get suspended;
-            // we increase the probabilty that this happens by waiting a moment before calling completionHandler()
-            DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
-                logger.info("⬅️ fetch done")
-                completionHandler(.newData)
+            logger.info("⬅️ fetch done")
+            completionHandler(.newData)
+            self.unregisterFetchBackgroundTask()
```
cyBerta, line 419: maybe we should move that line before the completion handler is executed in order to be sure this is called before the suspension of the app happened.

deltachat-ios/AppDelegate.swift
```diff
@@ -372,12 +376,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         bgIoTimestamp = nowTimestamp
 
         // make sure to balance each call to `beginBackgroundTask` with `endBackgroundTask`
-        let backgroundTask = UIApplication.shared.beginBackgroundTask {
-            // TODO: currently, this should be okay, we are not using too much background time,
-            // so that handler should never be called.
-            // the plan is to listen to an event as DC_EVENT_ENTER_IDLE and stop fetch from there.
-            // if we have such an event, we could imprive this handler and fire the event.
-            logger.info("fetch background task will end soon")
+        fetchBackgroundTaskEnding = false
```
cyBerta, line 379: this is quite confusing to read. Is it possible to move that line to the end of `endFetch()`?

cyBerta:
For the naming of the background task, maybe we can find sth. more descriptive. Afterburner reminds me of other things than what is meant here 🌶️🌶️🌶️

deltachat-ios/AppDelegate.swift
r10s: i think, this is fine, the backgroundTask avoids the suspension.

r10s:
please do not merge in, i have an idea about how to do this whole stuff more elegant with less "globals", have to think about that a bit more, however :)

r10s:
i crated an alternative in https://github.com/deltachat/deltachat-ios/pull/1221

r10s:
this pr, although having more complexity, as the same issue as #1221 - it does not call endBackgroundTask from within the handler but a moment later.

r10s:
closing in favor to https://github.com/deltachat/deltachat-ios/pull/1223

PR #1221, opened 2021-05-18
stop performFetch using a semaphore

r10s:
this is an alternative to https://github.com/deltachat/deltachat-ios/pull/1220, without using lots of AppDelegate-global variables and problems resulting from that.

that would also prepare nicely for #1157 as only a signal needs to be raised (otherwise we would also get the completion handler into the AppDelegate-global.

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -388,7 +388,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
         DispatchQueue.global(qos: .background).async { [weak self] in
             logger.info("➡️ fetching")
-            _ = fetchSemaphore.wait(timeout: .now() + 1000)
+            _ = fetchSemaphore.wait(timeout: .now() + 10)
 
             DispatchQueue.main.async { [weak self] in
                 logger.info("⬅️ finishing fetch")
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -372,38 +372,45 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         bgIoTimestamp = nowTimestamp
 
         // make sure to balance each call to `beginBackgroundTask` with `endBackgroundTask`
+        let fetchSemaphore = DispatchSemaphore(value: 0)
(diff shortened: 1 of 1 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -372,38 +372,46 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
-            // TODO: currently, this should be okay, we are not using too much background time,
-            // so that handler should never be called.
-            // the plan is to listen to an event as DC_EVENT_ENTER_IDLE and stop fetch from there.
-            // if we have such an event, we could imprive this handler and fire the event.
-            logger.info("fetch background task will end soon")
+            // we do not call `endBackgroundTask` here, even if that results in a warning:
+            // if we would do, we would risk a 0xdead10cc exception as things are still running;
+            // therefore we just signal the task below to exit gracefully and call `endBackgroundTask`.
+            // in practise, this handler should not be called anyway
+            // as we take care to stay below the 30 seconds.
```
r10s, line 381: calling the handler itself would be easy, just needs a var in local namespace, declared with `var backgroundTask: UIBackgroundTaskIdentifier = .invalid` - however, it is unclear if that is finally better.

deltachat-ios/AppDelegate.swift
```diff
@@ -372,38 +372,45 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
 
         // we're in background, run IO for a little time
         dcContext.maybeStartIo()
         dcContext.maybeNetwork()
 
-        DispatchQueue.main.asyncAfter(deadline: .now() + 10) { [weak self] in
-            logger.info("⬅️ finishing fetch")
+        DispatchQueue.global(qos: .background).async { [weak self] in
+            logger.info("➡️ fetching")
+            _ = fetchSemaphore.wait(timeout: .now() + 1000)
```
cyBerta, line 391: The initializer of DispatchTime, the argument passed to the wait function, uses nano-seconds https://developer.apple.com/documentation/dispatch/dispatchtime. 
Did you test that + 1000 is right here? You're intending 10 seconds, right?

cyBerta:
nice approach 🚀

r10s:
closing in favor to https://github.com/deltachat/deltachat-ios/pull/1223, we will probably pickup the semaphore-idea when targeting #1157

PR #1225, opened 2021-05-19
Camera permission handling

### LINKED ISSUE #677: show warning if camera access was denied
Currently if we don't have camera access permission the QrCodeReader is plain white.

We should show a hint label and also a button, that redirects to device settings -> deltachat.

cyBerta:
fixes #677

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -1050,7 +1050,32 @@ class ChatViewController: UITableViewController {
     }
 
     private func showCameraViewController() {
-        mediaPicker?.showCamera()
+        if AVCaptureDevice.authorizationStatus(for: .video) == .authorized {
+            self.mediaPicker?.showCamera()
+        } else {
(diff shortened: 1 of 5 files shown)


r10s (commented):
i rebases the pr.

i did not checked the code much yet, only functionality. it works great for the qr-code-scanner and camera that is used after configuration.

however, it does not work yet for the qr-code-scanner that is used on setup - here, if you hit one time "Don't allow" maybe accidentally, your're still stuck. there should be the same hint as for the qr-code-scanner that is used after configuration.

cyBerta:
hm... I cannot reproduce the issue. the same viewcontroller is used for the QR code scanning. 
On the welcome screen I tap Scan QR code and I can see either the QR-Code scanner or the error message. It depends on whether I have enabled or disabled the camera permission switch. 
would you mind to retry with a fresh build?

r10s:
you're right, i mixed something up.

r10s (approved):

PR #1226, opened 2021-05-19
hide keyboard if user tapped on a chat item, adapt to DC Android

### LINKED ISSUE #1114: hide the keyboard in the chat list
* After searching in the chat list and tapping on a list item, the keyboard should disappear, so that the keyboard is not visible anymore if the user comes back from the chat to the chat list

cyBerta:
closes #1114


r10s (approved):
great!

PR #1227, opened 2021-05-19
Keep keyboard in chat

### LINKED ISSUE #1113: Don't hide the keyboard after sending a message
it is quite annoying, if you're chatting and sending mulitple messages with just few words right after each other. The keyboard should only hide on manual swipe down.

cyBerta:
fixes #1113 

and adds some minor clean ups in ChatViewController

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -912,12 +912,11 @@ class ChatViewController: UITableViewController {
             .onEnabled { item in
                 UIView.animate(withDuration: 0.3, animations: {
                     item.backgroundColor = DcColors.primary
-                })
-        }.onDisabled { item in
-            UIView.animate(withDuration: 0.3, animations: {
-                item.backgroundColor = DcColors.colorDisabled
-            })
-        }
+                })}
+            .onDisabled { item in
+                UIView.animate(withDuration: 0.3, animations: {
+                    item.backgroundColor = DcColors.colorDisabled
+                })}
     }
 
     @objc private func chatProfilePressed() {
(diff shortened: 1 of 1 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -21,10 +21,11 @@ class ChatViewController: UITableViewController {
...
     // The latter is cancelable and leads to viewWillAppear is called in case the gesture is cancelled
     // We need the flag to handle that special case correctly in viewWillAppear
-    var isDismissing = false
-    var isInitial = true
-    var ignoreInputBarChange = false
+    private var isDismissing = false
+    private var isInitial = true
+    private var ignoreInputBarChange = false
     private var isVisibleToUser: Bool = false
+    private var keepKeyboard: Bool = false
```
r10s, line 28: maybe add a little comment what keepKeyboard actually does.

iiuc, it is set on sending and sets the focus to the input bar on sending so that it is not hidden?

r10s (approved):
lgtm.

after sending, there is still some nervous scrolling, it always looks as if the message scrolls in two times (EDIT: i filed an issue for that at https://github.com/deltachat/deltachat-ios/issues/1229). however, as that is no deterioration, this lgtm :)

deltachat-ios/Chat/ChatViewController.swift
cyBerta: right, it indicates that the currently focussed input field should not loose and re-request it's focus. The visibility of the keyboard is always related to the focus of an editable input field in iOS. There's no other way to control the keyboard than by requesting to become the first responder or by resigning that. The evaluation if a text input field should loose it's focus can be done in the delegate method `textViewShouldEndEditing`

PR #1233, opened 2021-05-21
don't scroll twice after message has been delivered

### LINKED ISSUE #1229: "double scrolling" on sending a message
when sending a message in a chat that has already some message in (so that scrolling is always needed), the message is scrolled in _twice_ somehow, fast after each other.

this looks a bit weird and not as stable as we would like to be :)

cyBerta:
fixes #1229

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -1112,12 +1112,7 @@ class ChatViewController: UITableViewController {
 
     func updateMessage(_ messageId: Int) {
         if messageIds.firstIndex(where: { $0 == messageId }) != nil {
-            markSeenMessage(id: messageId)
-            let wasLastSectionVisible = self.isLastRowVisible()
             reloadData()
-            if wasLastSectionVisible {
-                self.scrollToBottom(animated: true)
-            }
         } else {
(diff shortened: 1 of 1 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1112,12 +1112,7 @@ class ChatViewController: UITableViewController {
 
     func updateMessage(_ messageId: Int) {
         if messageIds.firstIndex(where: { $0 == messageId }) != nil {
-            markSeenMessage(id: messageId)
```
r10s, line 1115: `markSeenMessage` is not needed because `updateMessage` is called for outgoing messages only? or is it called implicitly somewhere else? why was `markSeenMessage` there at all?

cyBerta: I think this part was really old. When we moved to the new Chat UI I've just c&p this from the messageKit version. markseen is completely wrong here I think. This if clause will be entered if the read receipt was received / the message state changes. markmessage needs to be implemented for newly received messages (among others), which is handled elsewhere.

r10s (commented):
lgtm

r10s (approved):

Issue #1055, opened 2021-02-11
Display contact status in the profile


link2xt:
See related core PR: https://github.com/deltachat/deltachat-core-rust/pull/2218

New API is `dc_contact_get_status`

r10s:
meanwhile, there are [screenshots from android](https://github.com/deltachat/deltachat-android/pull/1805) and a [mockup from desktop](https://github.com/deltachat/deltachat-desktop/pull/2126#issuecomment-793999988).

for ios, i think, 

- we should add a complete section to the contact-profile with the usual upper-case title of the string `pref_default_status_label`

- the section should be placed between the first section and second section, so between "Send message" and "Shared chats"

- the section will have only one item, the signature-text. if would be great if links in this text are clickable.

- if dc_contact_get_status() returns an empty string, the whole section can disappear. 

- wrt scrolling: it is okay if the signature-text grows in height, the most-important functions are above this section and the whole ViewController is scrollable anyway

PR #1234, opened 2021-05-24
show status/signatures in contact-profiles

r10s:
- [x] add `DcContact.status` api
- [x] basically show status
- [x] make links clickable, maybe better use ~~`UILabel`~~ `MessageLabel` inside `MultilineLabelCell` for that purpose
- [x] make phone number clickable
- [x] make email addresses clickable
- [x] let `MultilineLabelCell` grow in height as needed

  

closes https://github.com/deltachat/deltachat-ios/issues/1055

### DIFF
--- a/deltachat-ios/Controller/ContactDetailViewController.swift
+++ b/deltachat-ios/Controller/ContactDetailViewController.swift
@@ -90,6 +90,7 @@ class ContactDetailViewController: UITableViewController {
 
     private lazy var statusCell: MultilineLabelCell = {
         let cell = MultilineLabelCell()
+        cell.multilineDelegate = self
         return cell
     }()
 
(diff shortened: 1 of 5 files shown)


cyBerta:
>     * make links clickable, maybe better use `UILabel` inside `MultilineLabelCell` for that purpose

You could also use MessageLabel, it already has 'Detectors' to detect links, phone numbers etc.

r10s:
> You could also use MessageLabel, it already has 'Detectors' to detect links, phone numbers etc.

sounds good!

deltachat-ios/ViewModel/ContactDetailViewModel.swift
```diff
@@ -54,6 +56,12 @@ class ContactDetailViewModel {
         self.sharedChats = context.getChatlist(flags: 0, queryString: nil, queryId: contactId)
 
         sections.append(.chatOptions)
+
+        let dcContact = DcContact(id: contactId)
```
cyBerta, line 60: see line 32/33

deltachat-ios/Controller/ContactDetailViewController.swift
```diff
@@ -430,3 +440,20 @@ class ContactDetailViewController: UITableViewController {
         }
     }
 }
+
+extension ContactDetailViewController: MultilineLabelCellDelegate {
+    func phoneNumberTapped(number: String) {
+        let sanitizedNumber = number.filter("0123456789".contains)
```
cyBerta, line 446: what about '+' ? doesn't ios handle +1 for example correctly?

r10s: i just copy+pasted from the existing code :)

cyBerta:
Nice feature!
I have found one minor issue: 
rescaling of the fonts don't work immediatly. you have to leave the ViewController and reenter to get the correct scaling. 

see the attached image, the other cells do scale immediately which leads to such a screen:
Screen Shot 2021-05-26 at 14 52 31

r10s:
> rescaling of the fonts don't work immediatly.

@cyBerta what is needed to get that correctly?

feel free to push sth. to this pr, maybe that is easier than explaining :)

cyBerta (approved):
the font issue will be targeted in another ticket

PR #1235, opened 2021-05-25
fix more lint issues

cyBerta:
no ticket associated with this PR


cyBerta:
I'll squash both commits with the `Squash and Merge` button here. So no worries about the duplicated commit message

PR #1236, opened 2021-05-25
First unread message

### LINKED ISSUE #1189: opening chat at the first unread message
the chat should open at the first unread message, not at the end.

we changed that recently on android, https://github.com/deltachat/deltachat-android/pull/1641, and that feels much better overall.

cmp. #1032

cyBerta:
closes #1189

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -27,6 +27,7 @@ public class DcContext {
         return .dcContext
     }
 
+    // TODO: remove count and from parameters if we don't use it
     public func getMessageIds(chatId: Int, count: Int? = nil, from: Int? = nil) -> [Int] {
         let start = CFAbsoluteTimeGetCurrent()
         let cMessageIds = getChatMessages(chatId: chatId)
@@ -247,6 +248,10 @@ public class DcContext {
 
     public func getFreshMessages() -> DcArray {
         return DcArray(arrayPointer: dc_get_fresh_msgs(contextPointer))
+    }
+
+    public func getFreshMessagesCount(chatId: Int) -> Int {
+        return Int(dc_get_fresh_msg_cnt(contextPointer, UInt32(chatId)))
     }
 
     public func markSeenMessages(messageIds: [UInt32]) {
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -250,10 +250,6 @@ public class DcContext {
         return DcArray(arrayPointer: dc_get_fresh_msgs(contextPointer))
     }
 
-    public func getFreshMessagesCount(chatId: Int) -> Int {
-        return Int(dc_get_fresh_msg_cnt(contextPointer, UInt32(chatId)))
-    }
-
     public func markSeenMessages(messageIds: [UInt32]) {
         messageIds.withUnsafeBufferPointer { ptr in
             dc_markseen_msgs(contextPointer, ptr.baseAddress, Int32(ptr.count))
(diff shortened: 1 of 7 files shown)


DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -249,6 +250,10 @@ public class DcContext {
         return DcArray(arrayPointer: dc_get_fresh_msgs(contextPointer))
     }
 
+    public func getFreshMessagesCount(chatId: Int) -> Int {
```
r10s, line 253: there is already `getUnreadMessages` that does exactly the same under another name :)

so, for clarity, in this pr, i'd just this function - the things may get complicated enough :) is a subsequent pr, we can rename the function.

r10s (commented):
nice!

as the line-separator is a "whole" UIView - can we have sth. as the day-titles here? so a string saying "123 new messages" centered with transparent background?

cyBerta:
> as the line-separator is a "whole" UIView - can we have sth. as the day-titles here? so a string saying "123 new messages" centered with transparent background?

Yes we can do that. But we would have to change the behavior. Right now the thin line is only an indicator which messages the user should already have seen before they entered the chat, without doing an assumptions about the amount of unread messages. If the user sends an additional message, the thin line stays, if another message comes in, too. It currently only disappears if the user deletes a message, because that operation makes it harder to track if the thin line needs to be moved or not. It also disappears if the user re-enters the chat of course. 

If we add the information about the amount of messages we would have to remove or adapt the indicator as soon as a new message comes in or as soon as the user wrote an additional message. Moreover I like the fact that this is only a very unintrusive visual hint that doesn't take any place - a decent design :).

DcCore/DcCore/DC/Wrapper.swift
cyBerta: Ups, I haven't seen that. Thanks! I'll fix that.

r10s:
> Moreover I like the fact that this is only a very unintrusive visual hint that doesn't take any place - a decent design :)

hm, "decent design" is also true for a headline - maybe even more as no new visual element is introduced. that also has the advantage, that no questions pop up as it is self-explaining; also that is already used on android, so, we can say, that this actually works ux-wise.

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -537,6 +537,12 @@ class ChatViewController: UITableViewController {
                 cell.update(text: "ErrDaymarker")
             }
             return cell
+        } else if id == DC_MSG_ID_MARKER1 {
+            // unread messages marker
+            let cell = tableView.dequeueReusableCell(withIdentifier: "info", for: indexPath) as? InfoMessageCell ?? InfoMessageCell()
+            let freshMsgsCount = self.messageIds.count - (indexPath.row + 1)
```
r10s, line 543: +1 for calculating the number of unread messages that way - so the headline can never be wrong :)

r10s (commented):
great pr, pretty clear and straight forward. the new-message-headline looks great!

i played around with it a bit:

- the new-message-headline stays when new messages come in, as the counter is also updated, this makes a lot of sense. great

- for outgoing messages, however, currently the new-message-headline also stays, probably because messageIds array is not updated. i think, here, it should be removed, esp. as the outgoing message is counted as "new" as well.  
(maybe the easiest and most-performant way to fix that is to just remove DC_MSG_ID_MARKER1 from messagesIds manually - maybe there is even a swift-array-function for that, otherwise, maybe create sth. as removeValueFromArray working on an [Int] array)

cyBerta:
>     * for outgoing messages, however, currently the new-message-headline also stays, probably because messageIds array is not updated. i think, here, it should be removed, esp. as the outgoing message is counted as "new" as well.

I've adapted the PR accordingly

PR #1240, opened 2021-05-27
Stickers

### LINKED ISSUE #1146: add support for webp, stickers
- webp stickers are already sent around here and there, it seems as if SDWebImage can support that by a plugin, so we should enable that at some point.
cc @adbenitez - or did i get sth. wrong?

- when on that, we should also show the stickers as such, so transparent, without surrounding bubble, currently stickers are displayed as images.

- the emoji keyboard can send stickers, we should consider to send them as such, cmp. #1170

cyBerta:
closes #1146


cyBerta:
@r10s for now all images pasted into the chat input bar hare handled as stickers. (You can verify by copying an image from the photos app to the clipboard and then paste them into the chat input bar). There's no way to determine exactly from which source a pasted image comes from, so that we cannot distinguish between memojis pasted from the keyboard into the pasteboard and other pasted images as describted above. The only thing we still can do is to check if an image has an alpha channel and allow only those to be sent as stickers, whereas other pasted images go into the preview area and will be handled as image messages. What do you think about that?

r10s:
i think, we can give the approach "image in input bar == sticker" a try. i do not think, many ppl will use that to share images from gallery or so.

cyBerta:
> i think, we can give the approach "image in input bar == sticker" a try. i do not think, many ppl will use that to share images from gallery or so.

true that is a quite hidden feature

r10s:
waiting for code review until "add SDWebImageWebPCoder pod" is removed from pr.

already now:

- great stuff, i build + tested + played around a bit - what a fun :)

- the stickers are a bit too large, we should maybe limit them to "half width of smaller screen side"

for comparison: from left to right, stickers on delta/ios, delta/android, whatsapp/ios:

cyBerta:
> om left to right, stickers on delta/ios, delta/android, whatsapp/ios:

whatsapp seems to show the wrong aspect ratio 😲 . Look at that strong neck!

cyBerta:
I'm closing this in favour of #1245

PR #1242, opened 2021-05-28
unify visiual feedback

r10s:
this pr adds visual feedback to all selectable cells in the UITableViewControllers.

before, it was a bit random if and which feedback was given.

see the separate commits for details.


cyBerta:
I haven't tested yet, but codewise it looks fine

cyBerta (approved):
great 🚀

PR #1245, opened 2021-06-01
Stickers merge

cyBerta:
Rebasing was difficult, so I went for merging master into the stickers branch. We now have one additional merge commit, but the diff is now easy to read at least.


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1643,3 +1656,10 @@ extension ChatViewController: UITextViewDelegate {
         return true
     }
 }
+
+// MARK: - InputBarImagePasteDelegate
```
cyBerta, line 1660: TODO: fix the comment

deltachat-ios/Model/GalleryItem.swift
```diff
@@ -43,17 +43,21 @@ class GalleryItem: ContextMenuItem {
         }
         switch viewtype {
         case .image:
-            thumbnailImage = msg.image
+            if url.pathExtension == "webp" {
+                loadAsyncImageThumbnail(from: url)
+            } else {
+                thumbnailImage = msg.image
```
cyBerta, line 49: TODO: check if we can just always call loadAsyncImageThumbnail

deltachat-ios/Model/GalleryItem.swift
```diff
@@ -43,17 +43,21 @@ class GalleryItem: ContextMenuItem {
...
+            loadAsyncImageThumbnail(from: url)
         default:
             safe_fatalError("unsupported viewtype - viewtype \(viewtype) not supported.")
         }
     }
 
-    private func loadGifThumbnail(from url: URL) {
+    private func loadAsyncImageThumbnail(from url: URL) {
         DispatchQueue.global(qos: .userInteractive).async {
             guard let imageData = try? Data(contentsOf: url) else {
```
cyBerta, line 62: we can probably shorten that: `SDAnimatedImage(contentsOfFile: url.path)`

cyBerta: I've created a separate PR to tackle the issues in GalleryItem -> #1246

cyBerta:
@r10s I think I will have to create a separate StickerCell, if we want to diverge from the aspect ratio rules made for videos, images and gifs. Otherwise things will get too messy.

r10s:
i see - and you are suggesting to do that in a subsequent pr? would be fine for me, things are a huge improvement already now :)

cyBerta:
@r10s I will do another PR that moves the sticker code into a separate sticker cell class.

r10s:
> @r10s I will do another PR that moves the sticker code into a separate sticker cell class.

to me, the current implementation looks okay.

r10s (approved):
lgtm, i am not-so-good with these constraint thingies, however, on testing, thinks look fine :)

so great thing - ios can send stickers now \o/

PR #1246, opened 2021-06-01
Async image loading

cyBerta:
on top of #1245 (needs to be merged first, before this one can be reviewed)

Calling `DcMsg.image` is blocking and should be generally avoided whenever possible. 
This PR ensures that image loading in the gallery is always async.
for $reasons, SDAnimatedImage doesn't load all jpegs correctly, so that I had to implement a separate method to load pngs and jpegs in a UIImage


r10s (approved):
lgtm, during testing, however, i got a crash in larger galleries. however, as that happens with and without this pr, it is probably unrelated and i filed [another issue](https://github.com/deltachat/deltachat-ios/issues/1248) for that.

PR #1247, opened 2021-06-01
keep sticker size on rotation

r10s:
this uses the same sticker dimensions in portait and landscape (maybe also a thing to consider for other images, however, this is out of scope :)


cyBerta:
nice idea

cyBerta (commented):
+1

PR #1251, opened 2021-06-02
Fix share to dc

### LINKED ISSUE #1249: share-to-delta from system-gallery stopped working
probably a regression from the recent changes wrt stickers :)

cyBerta:
I basically stubled into the same issue as here: https://github.com/deltachat/deltachat-ios/pull/1246#issue-659165534

closes #1249

### DIFF
--- a/deltachat-ios/Helper/ImageFormat.swift
+++ b/deltachat-ios/Helper/ImageFormat.swift
@@ -22,8 +22,7 @@ extension ImageFormat {
 
             if let dataString = String(data: subdata, encoding: .ascii),
                 dataString.hasPrefix("RIFF"),
-                dataString.hasSuffix("WEBP")
-            {
+                dataString.hasSuffix("WEBP") {
                 return .webp
             }
 
@@ -31,14 +30,12 @@ extension ImageFormat {
             let subdata = data[8...11]
 
             if let dataString = String(data: subdata, encoding: .ascii),
-                Set(["heic", "heix", "hevc", "hevx"]).contains(dataString)
-            {
+                Set(["heic", "heix", "hevc", "hevx"]).contains(dataString) {
                 return .heic
             }
 
         case 0x42 where data.count >= 2 :
-            let secondByte = data[1]
-            if secondByte == 0x4D {
+            if data[1] == 0x4D {
                 return .bmp
             }
 
@@ -46,10 +43,6 @@ extension ImageFormat {
             break
         }
         return .unknown
-    }
-
-    var contentType: String {
-        return "image/\(rawValue)"
     }
 
     static func loadImageFrom(data: Data) -> UIImage? {
--- a/deltachat-ios/Helper/ImageFormat.swift
+++ b/deltachat-ios/Helper/ImageFormat.swift
@@ -7,6 +7,8 @@ enum ImageFormat: String {
 }
 
 extension ImageFormat {
+
+    // magic bytes can be found here: https://en.wikipedia.org/wiki/List_of_file_signatures
     static func get(from data: Data) -> ImageFormat {
         switch data[0] {
         case 0x89:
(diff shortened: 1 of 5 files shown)


deltachat-ios/Helper/ImageFormat.swift
```diff
@@ -0,0 +1,64 @@
...
+
+        case 0x42 where data.count >= 2 :
+            if data[1] == 0x4D {
+                return .bmp
+            }
+
+        default:
+            break
+        }
+        return .unknown
```
r10s, line 45: where does this list come from?
maybe add a link to the source (if any).

deltachat-ios/Helper/ImageFormat.swift
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    static func loadImageFrom(data: Data) -> UIImage? {
+        switch ImageFormat.get(from: data) {
+        case .gif, .png, .webp, .heic:
```
r10s, line 50: `.heic` can be animated as well? because of the "live-image"?

r10s (approved):
lgtm.

it happend once to me that that sharing did not work at the first place: the "one moment..." dialog was displayed but vanished too fast without an error - and no image was finally shared. but i could not reproduce it - so that might be another bug and should not block this pr from merging.

r10s:
when looking at `SDAnimatedImage` source, it says:

"The image class which supports animating on `SDAnimatedImageView`. You can also use it on normal UIImageView/NSImageView."

so, maybe we can just always use `SDAnimatedImageView` unconditionally - or is there a reasonable speed advantage in using `UImage` for jpg?

deltachat-ios/Helper/ImageFormat.swift
cyBerta: yes: https://nokiatech.github.io/heif/comparison.html

deltachat-ios/Helper/ImageFormat.swift
cyBerta: https://en.wikipedia.org/wiki/List_of_file_signatures

cyBerta: I'll can add that as a comment as well

cyBerta:
> when looking at `SDAnimatedImage` source, it says:
> 
> "The image class which supports animating on `SDAnimatedImageView`. You can also use it on normal UIImageView/NSImageView."
> 
> so, maybe we can just always use `SDAnimatedImageView` unconditionally - or is there a reasonable speed advantage in using `UImage` for jpg?

Yes, I tried to use SDAnimatedImage on UIImage, but unfortunaltely if you try to parse Data from a jpeg and try to show it in SDAnimatedImageView or UIImageView, it didn't work. It let to the regression that has been fixed with this PR. 
I bumped into the same problem when working on async loading of images in GalleryItem so that I had to distinguish between different formats.

cyBerta:
> "The image class which supports animating on `SDAnimatedImageView`. You can also use it on normal UIImageView/NSImageView."

Maybe this is a bug in SDAnimatedImage

deltachat-ios/Helper/ImageFormat.swift
r10s: yip, i would add that link as a comment.

r10s:
k, then the distinction between `SDAnimatedImage` and `UIImage` makes sense, at least for now. however, maybe also add your findings to a comment atop of `loadImageFrom()`, saying, that it theoretically should work but it practically does not.

deltachat-ios/Helper/ImageFormat.swift
cyBerta: done

r10s (approved):
great!

PR #1252, opened 2021-06-03
Save animated images

cyBerta:
on top of #1251 (needs to be reviewed and merged first)

* full support of apngs now
* let new ImageFormat class handle loading and saving images with the correct formatting

### DIFF
--- a/deltachat-ios/Helper/ImageFormat.swift
+++ b/deltachat-ios/Helper/ImageFormat.swift
@@ -1,6 +1,7 @@
 import Foundation
 import UIKit
 import SDWebImage
+import DcCore
 
 enum ImageFormat: String {
     case png, jpg, gif, tiff, webp, heic, bmp, unknown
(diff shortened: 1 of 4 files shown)


deltachat-ios/Helper/ImageFormat.swift
```diff
@@ -67,4 +68,35 @@ extension ImageFormat {
...
+            if format != .unknown {
+                return ImageFormat.saveImage(data: data, suffix: format.rawValue)
+            }
+        }
+        let suffix = image.isTransparent() ? "png" : "jpg"
+        guard let data = image.isTransparent() ? image.pngData() : image.jpegData(compressionQuality: 1.0) else {
+            return nil
+        }
+
+        return saveImage(data: data, suffix: suffix)
```
r10s, line 84: i suggest to use either `saveImage()` or `ImageFormat.saveImage()` at both places (here and at line 76), but not to mix that.

deltachat-ios/Helper/ImageFormat.swift
```diff
@@ -67,4 +68,35 @@ extension ImageFormat {
...
+    }
+
+    public static func saveImage(data: Data, suffix: String) -> String? {
```
r10s, line 87: so, data contains the already encoded image data, not an rgb-bitarray? and image-(re)encoding does not take place in this function? maybe add a comment on that and/or name data to encoded_data.

deltachat-ios/Helper/ImageFormat.swift
```diff
@@ -67,4 +68,35 @@ extension ImageFormat {
         return loadImageFrom(data: imageData)
     }
 
+    public static func saveImage(image: UIImage) -> String? {
+        if image.sd_isAnimated,
+           let data = image.sd_imageData() {
```
r10s, line 73: does that result in a recoding?

that may be a bit unfortunately wrt stickers (as it would make showing every received sticker in a selector harder as they the same stickers may have different binaries)

r10s: answering myself: for animated images, there is "No more extra encoding process." says comment atop of sd_imageData.

however, for not-animated stickers, recoding is still always done (that are handled below).

PR #1257, opened 2021-06-08
tap on quotes that belong to another chat

### LINKED ISSUE #1250: tapping quotes not working if the original message is in another chat
it may happen for several reasons, that a quote refers a message in _another_ chat, eg. by android's "reply privately" feature.

a tap on these quotes should open the correct chat at the correct position, however, currently, this does not work - tapping quotes always stay in the same chat.

cyBerta:
open another chat on tapping on a quote if the quote doesn't belong to the current chat

closes #1250


r10s (approved):
nice, looks pretty straightforward!

PR #1258, opened 2021-06-08
block sending share messages until all attachments have been processed

### LINKED ISSUE #1253: share-to-delta aborted when tapping too fast
i could reproduce the bug happening at https://github.com/deltachat/deltachat-ios/pull/1251#pullrequestreview-675255093 now:

- open gallery and select an image to share

- select share-to-delta

- share-to-delta is _aborted_ when tapping the "send" button too fast. this is possible eg. by typing sth. in, so that the button is enabled (otherwise, the button is enabled when the image preview maybe before the image is ready)

_aborted_ means, the image is not shared, no error is shown. also, the system's share-to-selection dialog is not closed as usual.

a video demonstrating the issue:

https://user-images.githubusercontent.com/9800740/120701582-4cb43280-c4b3-11eb-873a-0c4336811f55.mp4

if sharing would have succeeded, the "One moment..." would appear a bit longer and the share-dialog would have been closed. you would be you have to be fast during the whole process :)

cyBerta:
fixes #1253

### DIFF
--- a/DcShare/Controller/ShareViewController.swift
+++ b/DcShare/Controller/ShareViewController.swift
@@ -38,6 +38,7 @@ class ShareViewController: SLComposeServiceViewController {
     let dbHelper = DatabaseHelper()
     var shareAttachment: ShareAttachment?
     var isAccountConfigured: Bool = true
+    var isLoading: Bool = true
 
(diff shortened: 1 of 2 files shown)


r10s (approved):
lgtm

PR #1259, opened 2021-06-08
early return on keyboardWillChangeFrameNotification 

### LINKED ISSUE #1241: Double Scrolling PT.2
when a message has been sent and the user taps immediately afterwards in the chat on an empty space (background), the scroll down animation repeats.

cyBerta:
... if keyboard size won't change

fixes #1241


r10s (approved):
if that helps, that lgtm :)

PR #1262, opened 2021-06-08
Fix chat cell layout

### LINKED ISSUE #1261: layout trouble with small images
when sending small square images the chat bubble layout is broken. It was probably already there for small gifs in the past but have been worsened with https://github.com/deltachat/deltachat-ios/pull/1245, specifically with async loading of all images https://github.com/deltachat/deltachat-ios/pull/1245/commits/e247eae8fb8c9b8ec0b3bdfec30ae0f6308ccf43

I don't like so much the Android layouting for small images, they get scaled up to a fixed square size and thus appear blurry. On the other hand implementing that is an easy way to fix the layout constraints for those small images.

Example image: https://en.wikipedia.org/wiki/APNG#/media/File:Animated_PNG_example_bouncing_beach_ball.png

cyBerta:
fixes #1261 without scaling small images up to the maximum square size

also avoids stickers from being cut-off

### DIFF
--- a/deltachat-ios/Chat/Views/Cells/ImageTextCell.swift
+++ b/deltachat-ios/Chat/Views/Cells/ImageTextCell.swift
@@ -110,6 +110,7 @@ class ImageTextCell: BaseMessageCell {
 
         self.imageHeightConstraint?.isActive = false
         self.imageWidthConstraint?.isActive = false
+        self.contentImageView.contentMode = .scaleAspectFit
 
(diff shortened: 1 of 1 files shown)


r10s (approved):
i am personally more in favor of scaling up, at least in "reasonable limits" (so: scaling 300px -> 400px might be fine while 16px -> 400 px is probably not), esp. on high-res devices, things may look too small overall.

but we can give that one a try and see how it looks like overall.

PR #1263, opened 2021-06-09
hide signature in "Saved messages"

r10s:
"Saved messages" would show the "Self" signature, while that may be okay if the chat would be named "Self chat" or so, for "Saved messages", this is a bit weird.

if we want a preview, we should add that somewhere else, it is not expected here.

also, this increases between-app-consistence: Android does not show the signature for "Saved messages" as well.


cyBerta (approved):
lgtm

PR #1268, opened 2021-06-11
refactor wrapper

### LINKED ISSUE #1264: Get rid of DcContext.shared
as part of AEAP we want to remove the global instance of DcContext

cyBerta:
* get rid of DcContext.share
* inject dcContext everyw…here instead
* changes the way DcMsg and DcContact are created

closes #1264

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -436,32 +436,6 @@ public class DcContext {
         return DcProvider(dcProviderPointer)
     }
 
-    public func previousMediaURLs(messageId: Int, messageType: Int) -> [URL] {
-            var urls: [URL] = []
-            var prev: Int = Int(dc_get_next_media(contextPointer, UInt32(messageId), -1, Int32(messageType), 0, 0))
-            while prev != 0 {
-                let prevMessage = getMessage(id: prev)
-                if let url = prevMessage.fileURL {
-                    urls.insert(url, at: 0)
-                }
-                prev = Int(dc_get_next_media(contextPointer, UInt32(prevMessage.id), -1, Int32(prevMessage.type), 0, 0))
-            }
-            return urls
-        }
-
-        public func nextMediaURLs(messageId: Int, messageType: Int) -> [URL] {
-            var urls: [URL] = []
-            var next: Int = Int(dc_get_next_media(contextPointer, UInt32(messageId), 1, Int32(messageType), 0, 0))
-            while next != 0 {
-                let nextMessage = getMessage(id: next)
-                if let url = nextMessage.fileURL {
-                    urls.append(url)
-                }
-                next = Int(dc_get_next_media(contextPointer, UInt32(nextMessage.id), 1, Int32(nextMessage.type), 0, 0))
-            }
-            return urls
-        }
-
     public func imex(what: Int32, directory: String) {
         dc_imex(contextPointer, what, directory, nil)
     }
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -93,10 +93,6 @@ public class DcContext {
         dc_block_contact(contextPointer, UInt32(id), 0)
     }
 
-    public func isFromCurrentSender(message: DcMsg) -> Bool {
-        return message.fromContactId == getContact(id: Int(DC_CONTACT_ID_SELF)).id
-    }
-
     public func getBlockedContacts() -> [Int] {
         let cBlockedContacts = dc_get_blocked_contacts(contextPointer)
         return DcUtils.copyAndFreeArray(inputArray: cBlockedContacts)
@@ -875,6 +871,10 @@ public class DcMsg {
 
     public var fromContactId: Int {
         return Int(dc_msg_get_from_id(messagePointer))
+    }
+
+    public var isFromCurrentSender: Bool {
+        return fromContactId == DC_CONTACT_ID_SELF
     }
 
     public var chatId: Int {
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -6,16 +6,16 @@ public class DcContext {
 
     /// TODO: THIS global instance should be replaced in the future, for example for a multi-account scenario,
     /// where we want to have more than one DcContext.
-    static let dcContext: DcContext = DcContext()
+    //static let dcContext: DcContext = DcContext()
     public var logger: Logger?
     var contextPointer: OpaquePointer?
     public var lastErrorString: String?
     public var lastWarningString: String = "" // temporary thing to get a grip on some weird errors
     public var maxConfigureProgress: Int = 0 // temporary thing to get a grip on some weird errors
 
-    private init() {
+    public init() {
     }
-
+    
     deinit {
         if contextPointer == nil { return } // avoid a warning about a "careless call"
         dc_context_unref(contextPointer)
@@ -23,8 +23,22 @@ public class DcContext {
     }
 
     /// Injection of DcContext is preferred over the usage of the shared variable
-    public static var shared: DcContext {
+/*    public static var shared: DcContext {
         return .dcContext
+    }*/
+
+    public func newMessage(viewType: Int32) -> DcMsg {
+        let messagePointer = dc_msg_new(contextPointer, viewType)
+        return DcMsg(pointer: messagePointer)
+    }
+
+    public func getMessage(id: Int) -> DcMsg {
+        let messagePointer = dc_get_msg(contextPointer, UInt32(id))
+        return DcMsg(pointer: messagePointer)
+    }
+
+    public func sendMessage(chatId: Int, message: DcMsg) {
+        dc_send_msg(contextPointer, UInt32(chatId), message.messagePointer)
     }
 
     // TODO: remove count and from parameters if we don't use it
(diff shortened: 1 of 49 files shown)


DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -63,6 +77,27 @@ public class DcContext {
...
+    public func unblockContact(id: Int) {
+        dc_block_contact(contextPointer, UInt32(id), 0)
+    }
+
+/*    public func getFromContact(messageId: Int): DcContact? {
+        DcContact(id: fromContactId)
+    }() */
+
+    public func isFromCurrentSender(message: DcMsg) -> Bool {
+        return message.fromContactId == getContact(id: Int(DC_CONTACT_ID_SELF)).id
```
r10s, line 98: as `getContact(id: Int(DC_CONTACT_ID_SELF)).id` is always just  DC_CONTACT_ID_SELF (that was wrong before), there is no need to move that to `DcMsg` to `DcContext`.

nb: on android, we call this function `isOutgoing()` which is a better name, however, to keep diff small, we should not rename the function.

DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -837,12 +848,8 @@ public class DcMsg {
         return Int(dc_msg_get_from_id(messagePointer))
     }
 
-    public lazy var fromContact: DcContact = {
-        DcContact(id: fromContactId)
-    }()
-
     public var isFromCurrentSender: Bool {
-        return fromContact.id == DcContact(id: Int(DC_CONTACT_ID_SELF)).id
+        return fromContactId == DC_CONTACT_ID_SELF
```
r10s, line 852: that will indeed save two db-calls per call to `isFromCurrentSender` - which called quite often! might be an even noticeable speed-improvement on scrolling through a chat :)

DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -837,12 +848,8 @@ public class DcMsg {
         return Int(dc_msg_get_from_id(messagePointer))
     }
 
-    public lazy var fromContact: DcContact = {
-        DcContact(id: fromContactId)
```
r10s, line 841: good to get rid of that:
- it is an expensive call that does not look as such
- it was never updated due to `lazy`

DcCore/DcCore/Views/InitialsBadge.swift
```diff
@@ -112,7 +112,7 @@ public class InitialsBadge: UIView {
         return !label.isHidden
     }
 
-    public func setColor(_ color: UIColor) {
+    public func setColor(_ color: UIColor?) {
```
r10s, line 115: color does not need to to optional, i have already changed this, no further action needed :)

deltachat-ios/ViewModel/ContactCellViewModel.swift
```diff
@@ -17,7 +18,7 @@ enum CellModel {
 }
 
 struct ContactCellData {
-    let contactId: Int
+    let contact: DcContact
```
r10s, line 21: i was first facing this change very skeptical as that may be easily result in a performance degradation due to additional database calls.

however, for both places, this struct is created, DcContact is created anyway, so it is okay from that point of view.

memory might still be an issue - as well as missing updates (contact is not updated while contactId forces a reload), i did not checked these points in detail yet.

DcShare/Controller/ShareViewController.swift
```diff
@@ -31,11 +31,14 @@ class ShareViewController: SLComposeServiceViewController {
         }
     }
 
+    lazy var dbHelper: DatabaseHelper = {
+       return DatabaseHelper(dcContext: dcContext)
+    }()
+
     let logger = SimpleLogger()
-    let dcContext = DcContext.shared
+    let dcContext: DcContext = DcContext()
```
r10s, line 39: dcContext is now unusable until presentationAnimationDidFinish() is processed, just saying, probably okay.

deltachat-ios/Chat/Views/Cells/BaseMessageCell.swift
```diff
@@ -272,7 +272,8 @@ public class BaseMessageCell: UITableViewCell {
     }
 
     // update classes inheriting BaseMessageCell first before calling super.update(...)
-    func update(msg: DcMsg, messageStyle: UIRectCorner, showAvatar: Bool, showName: Bool) {
+    func update(dcContext: DcContext, msg: DcMsg, messageStyle: UIRectCorner, showAvatar: Bool, showName: Bool) {
+        let fromContact = dcContext.getContact(id: msg.fromContactId)
```
r10s, line 276: nice! that saves on database access per bubble!

deltachat-ios/Controller/ContactDetailViewController.swift
```diff
@@ -33,8 +33,8 @@ class ContactDetailViewController: UITableViewController {
 
     private lazy var blockContactCell: ActionCell = {
         let cell = ActionCell()
-        cell.actionTitle = viewModel.contact.isBlocked ? String.localized("menu_unblock_contact") : String.localized("menu_block_contact")
-        cell.actionColor = viewModel.contact.isBlocked ? SystemColor.blue.uiColor : UIColor.red
+        cell.actionTitle = (viewModel.contact.isBlocked ?? false) ? String.localized("menu_unblock_contact") : String.localized("menu_block_contact")
+        cell.actionColor = (viewModel.contact.isBlocked ?? false) ? SystemColor.blue.uiColor : UIColor.red
```
r10s, line 37: `?? false` is not needed by recent commits. i will add a commit, i just tested that anyway.

deltachat-ios/Model/GalleryItem.swift
```diff
@@ -44,8 +44,8 @@ class GalleryItem: ContextMenuItem {
         }
     }
 
-    init(msgId: Int) {
-        self.msg = DcMsg(id: msgId)
+    init(msg: DcMsg) {
```
r10s, line 47: i would have passed context+msgId to the constructor, that way one would stay more flexible and also saves some lines of code on construction.

but this is not a blocker, it is also okay this way, we can change it if it turns out that we need the context.

deltachat-ios/Controller/GalleryViewController.swift
```diff
@@ -183,12 +184,14 @@ extension GalleryViewController: UICollectionViewDataSource, UICollectionViewDel
...
+            galleryCell.update(item: item)
         } else {
-            let galleryItem = GalleryItem(msgId: msgId)
+            let message = dcContext.getMessage(id: msgId)
+            let galleryItem = GalleryItem(msg: message)
             items[indexPath.row] = galleryItem
             item = galleryItem
+            galleryCell.update(item: item)
         }
-        galleryCell.update(item: item)
```
r10s, line 191: why is `galleryCell.update(item: item)` no longer outfactored?

deltachat-ios/Helper/NotificationManager.swift
```diff
@@ -51,21 +53,25 @@ public class NotificationManager {
                 logger.info("notification background task will end soon")
             }
 
-            DispatchQueue.global(qos: .background).async {
+            DispatchQueue.global(qos: .background).async { [weak self] in
```
r10s, line 56: great you take care of this `[weak self]` stuff!

deltachat-ios/ViewModel/ContactCellViewModel.swift
```diff
@@ -32,11 +33,13 @@ struct DeaddropCellData {
     let chatId: Int
     let msgId: Int
     let summary: DcLot
+    let deaddropContact: DcContact
```
r10s, line 36: why add and cache that here?

i'd prefer a more 1:1 adaption as in the rest of this pr.

r10s (commented):
generally, this lgtm, i did some minor comments and commits, please pull.

i am a bit skeptical wrt introducing `ContactCellData.contact` and `DeaddropCellData.deaddropContact` - this may have side effects eg. because of additional memory usage, non-updated objects, too many loaded objects while adding few benefits. maybe everything is fine, however, i cannot oversee that atm.

so - maybe better keep this pr stupid and leave ids as ids (as at other places in this pr) and do possible optimizations in another pr.

but i will also sleep over that,  a but tiered but the huge review :)

r10s (approved):
great job, @cyBerta!

in the last two commits, i reverted the additional caches, mainly to avoid possible troubles, eg. one is referred in an update() routine. if they are really needed, we should add them in another pr.

deltachat-ios/ViewModel/ContactCellViewModel.swift
r10s: i added a commit that discards the additional caching, just to avoid introducing new issues as side effects here. the code is also closer to the old one and easier to review.

deltachat-ios/ViewModel/ContactCellViewModel.swift
r10s: i added a commit that discards the additional caching, just to avoid introducing new issues as side effects here. the code is also closer to the old one and easier to review.

DcShare/Controller/ShareViewController.swift
cyBerta: yes it is the main entry point for the extension

deltachat-ios/Controller/GalleryViewController.swift
cyBerta: that's not how it should be. I'll correct it

PR #1269, opened 2021-06-14
fixing sharing text

### LINKED ISSUE #1266: send-button disabled on "share text"
when selecting some text eg. in Safari and selecting "Share..." from Safari's context menu, the share-extensions are listed.

when selecting "Delta Chat" there, the "Send" button is initially disabled - only when you add/remove a character, the send button is selectable. that worked before the [most recent changes wrt share](https://github.com/deltachat/deltachat-ios/issues?q=is%3Aclosed+label%3A%22share+extension%22).

cyBerta:
set isLoadingFlag only if share items are really existing
validateContent() on loading finished

fixes #1266


r10s (approved):
lgtm, tested, sharing text works now, also sharing images is still working.

curious what breaks next here, this send-button is a beast :)

PR #1212, opened 2021-05-14
Show in chat

cyBerta:
fixes #1150

PR #1270, opened 2021-06-14
Fix show in chat

### LINKED ISSUE #1267: "Show in chat" may open wrong chat
"Profile / Images / Show in chat" and "Profile / Documents / Show in chat" seem to pass the message-id just to the underlying chat view controller - that is correct, when the profile belongs to that chat.

however, when opening a contact profile from a group profile, that assumption is wrong (the profile belongs to the contact while the underlying chat view controller belongs to the group.

to reproduce:
1. open a group
2. open group profile
3. tap a contact in the memeberlist to go to the contact profile
4. open contact profile gallery
5. tap "Show in chat" on some image

that image is now passed to the chat view controller of step 1. - which, does not have the image nor a message with the requested id.

might be i've seen this issue when reporting https://github.com/deltachat/deltachat-ios/issues/1150 at some point, maybe have also have a closer look at the fix belonging to #1150 what that is actually doing and if that still makes sense.

cyBerta:
fixes #1267

https://github.com/deltachat/deltachat-ios/pull/1212 is still valid

### DIFF
--- a/deltachat-ios/Controller/DocumentGalleryController.swift
+++ b/deltachat-ios/Controller/DocumentGalleryController.swift
@@ -169,18 +169,10 @@ extension DocumentGalleryController {
 
     func redirectToMessage(of indexPath: IndexPath) {
         let msgId = fileMessageIds[indexPath.row]
+        let chatId = dcContext.getMessage(id: msgId).chatId
 
-        guard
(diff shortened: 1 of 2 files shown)


r10s (approved):
lgtm, much simplified code, great!

PR #1272, opened 2021-06-14
DcAccounts

### LINKED ISSUE #1265: DcAccounts wrapper
* we should start using the dc_accounts api as first step to prepare multi-account
* therefore we need to implement the wrapper for dc accounts
* we need to do kind of a migration with existing databases by using the import function (`migrate_account`)
* before we start implementing multi-account we can make sure the dc_accounts api works correctly in a single-account-scenario

cyBerta:
closes #1265

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -30,6 +30,10 @@ public class DcAccounts {
         return Int(dc_accounts_migrate_account(accountsPointer, dbLocation))
     }
 
+    public func getAll() -> [Int] {
+        let cAccounts = dc_accounts_get_all(accountsPointer)
+        return DcUtils.copyAndFreeArray(inputArray: cAccounts)
+    }
 }
 
 public class DcContext {
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -36,7 +36,7 @@ public class DcAccounts {
         return DcUtils.copyAndFreeArray(inputArray: cAccounts)
     }
 
-    public func get() -> DcContext {
+    public func getSelected() -> DcContext {
         let cPtr = dc_accounts_get_selected_account(accountsPointer)
         return DcContext(contextPointer: cPtr)
     }
@@ -47,7 +47,7 @@ public class DcAccounts {
     }
 
     public func maybeStartIo() {
-        if get().isConfigured() {
+        if getSelected().isConfigured() {
             dc_accounts_start_io(accountsPointer)
         }
     }
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -30,10 +30,56 @@ public class DcAccounts {
         return Int(dc_accounts_migrate_account(accountsPointer, dbLocation))
     }
 
+    public func addAccount() -> Int {
+        return Int(dc_accounts_add_account(accountsPointer))
+    }
+
+    public func getAccount(id: Int) -> DcContext {
+        let contextPointer = dc_accounts_get_account(accountsPointer, UInt32(id))
+        return DcContext(contextPointer: contextPointer)
+    }
+
     public func getAll() -> [Int] {
         let cAccounts = dc_accounts_get_all(accountsPointer)
         return DcUtils.copyAndFreeArray(inputArray: cAccounts)
     }
+
+    public func maybeNetwork() {
+        dc_accounts_maybe_network(accountsPointer)
+    }
+
+    public func maybeStartIO() {
+        if getSelectedAccount().isConfigured() {
+            dc_accounts_start_io(accountsPointer)
+        }
+    }
+
+    public func stopIO() {
+        dc_accounts_stop_io(accountsPointer)
+    }
+
+    public func selectAccount(id: Int) -> Bool {
+        return dc_accounts_select_account(accountsPointer, UInt32(id)) == 1
+    }
+
+    public func getSelectedAccount() -> DcContext {
+        let cPtr = dc_accounts_get_selected_account(accountsPointer)
+        return DcContext(contextPointer: cPtr)
+    }
+
+    public func removeAccount(id: Int) -> Bool {
+        return dc_accounts_remove_account(accountsPointer, UInt32(id)) == 1
+    }
+
+    public func importAccount(filePath: String) -> Int {
+        return Int(dc_accounts_import_account(accountsPointer, filePath))
+    }
+
+    public func getEventEmitter() -> DcAccountsEventEmitter {
+        let eventEmitterPointer = dc_accounts_get_event_emitter(accountsPointer)
+        return DcAccountsEventEmitter(eventEmitterPointer: eventEmitterPointer)
+    }
+
 }
 
 public class DcContext {
@@ -45,6 +91,10 @@ public class DcContext {
     public var maxConfigureProgress: Int = 0 // temporary thing to get a grip on some weird errors
 
     public init() {
+    }
+
+    public init(contextPointer: OpaquePointer?) {
+        self.contextPointer = contextPointer
     }
     
     deinit {
@@ -627,6 +677,25 @@ public class DcContext {
     // do not use. use DcContext::isConfigured() instead
     public var configured: Bool {
         return getConfigBool("configured")
+    }
+}
+
+
+public class DcAccountsEventEmitter {
+    private var eventEmitterPointer: OpaquePointer?
+
+    // takes ownership of specified pointer
+    public init(eventEmitterPointer: OpaquePointer?) {
+        self.eventEmitterPointer = eventEmitterPointer
+    }
+
+    public func getNextEvent() -> DcEvent? {
+        guard let eventPointer = dc_accounts_get_next_event(eventEmitterPointer) else { return nil }
+        return DcEvent(eventPointer: eventPointer)
+    }
+
+    deinit {
+        dc_accounts_event_emitter_unref(eventEmitterPointer)
     }
 }
 
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -316,20 +316,6 @@ public class DcContext {
         return "ErrGetContactEncrInfo"
     }
 
-    public func openDatabase(dbFile: String) {
-        var version = ""
-        if let appVersion = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String {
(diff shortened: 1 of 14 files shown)


DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -596,6 +680,25 @@ public class DcContext {
...
+        guard let eventPointer = dc_accounts_get_next_event(eventEmitterPointer) else { return nil }
+        return DcEvent(eventPointer: eventPointer)
+    }
+
+    deinit {
+        dc_accounts_event_emitter_unref(eventEmitterPointer)
+    }
+}
+
 public class DcEventEmitter {
```
r10s, line 702: the old DcEventEmitter could probably be removed.

DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -2,6 +2,86 @@ import Foundation
...
+    var accountsPointer: OpaquePointer?
+
+    public init() {
+        var version = ""
+        if let appVersion = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String {
+            version += " " + appVersion
+        }
+
+        if let sharedDbLocation = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: applicationGroupIdentifier) {
+            accountsPointer = dc_accounts_new("iOS\(version)", sharedDbLocation.path)
```
r10s, line 19: the directory we're passing here is not empty, there may be the old messenger.db and messenger.db-blobs in.

to avoid problems, i would use a subdirectory as "accounts". core will create the subdirectory if it does not exist (as well as all the needed subdirectories for the accounts).

DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -2,6 +2,95 @@ import Foundation
...
+        let contextPointer = dc_accounts_get_account(accountsPointer, UInt32(id))
+        return DcContext(contextPointer: contextPointer)
+    }
+
+    public func getAll() -> [Int] {
+        let cAccounts = dc_accounts_get_all(accountsPointer)
+        return DcUtils.copyAndFreeArray(inputArray: cAccounts)
+    }
+
+    public func get() -> DcContext {
```
r10s, line 39: let's stay closer to the ffi and call this function `getSelected()`

deltachat-ios/AppDelegate.swift
```diff
@@ -445,24 +450,54 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
+        locationManager.reloadDcContext()
+        notificationManager.reloadDcContext()
+        if dcAccounts.get().isConfigured() {
+            appCoordinator.presentTabBarController()
+        } else {
+            appCoordinator.presentWelcomeController()
+        }
     }
 
+    func openDatabase() {
```
r10s, line 483: i suggest to take the chance to strike these tiny one-line-helper. the code is easier to read if you see eg. `dcAccounts.openDatabase()` directly.

deltachat-ios/AppDelegate.swift
```diff
@@ -445,24 +450,54 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
+
+        // first database migration to shared container, can be removed at some point
+        // implemented in April 2020 (https://github.com/deltachat/deltachat-ios/pull/612)
+        let databaseHelper = DatabaseHelper(dcLogger: dcContext.logger)
+        guard let databaseLocation = databaseHelper.updateDatabaseLocation() else {
             fatalError("Database could not be opened")
         }
-        logger.info("open: \(databaseLocation)")
         dcContext.openDatabase(dbFile: databaseLocation)
+        if dcContext.isConfigured() {
```
r10s, line 463: i think, it is not needed to try to open the database and check for configured, esp. as this would also happen when the file is no longer existant. i would just check for existence of the old database file and pass it to migrateAccount() then.

deltachat-ios/AppDelegate.swift
```diff
@@ -445,24 +450,54 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
+            dcContext.closeDatabase()
+            if dcAccounts.migrateAccount(dbLocation: databaseLocation) == 0 {
```
r10s, line 465: i assume that will result in accounts-manager having two accounts. as that is a bit unexpected at that point, maybe we should change core to not create a default account. in this case, if there is no account to migrate, we would just say add-account.

cc @link2xt

cyBerta: from my point of view, we don't need the default account context. We can add one manually (programmatically), when we need it.

link2xt: This is fixed in https://github.com/deltachat/deltachat-core-rust/pull/2500

r10s: i pulled current core and rebased this pr.

DcShare/Controller/ShareViewController.swift
```diff
@@ -77,27 +77,25 @@ class ShareViewController: SLComposeServiceViewController {
...
-                }
+            if selectedChatId == nil {
+                selectedChatId = dcContext.getChatIdByContactId(contactId: Int(DC_CONTACT_ID_SELF))
+            }
+            if let chatId = selectedChatId {
+                selectedChat = dcContext.getChat(chatId: chatId)
+            }
+            DispatchQueue.global(qos: .userInitiated).async { [weak self] in
+                guard let self = self else { return }
+                self.shareAttachment = ShareAttachment(dcContext: self.dcAccounts.getSelected(), inputItems: self.extensionContext?.inputItems, delegate: self)
```
r10s, line 98: why use `self.dcAccounts.getSelected()` if `self.dcContext` is available as well.

unless there is a good reason, maybe better use either the one or the other throughout the file. getSelected() has some risk to be changed, so maybe better use `self.dcContext`.

deltachat-ios/AppDelegate.swift
```diff
@@ -55,7 +55,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         let console = ConsoleDestination()
         console.format = "$DHH:mm:ss.SSS$d $C$L$c $M" // see https://docs.swiftybeaver.com/article/20-custom-format
         logger.addDestination(console)
-        dcContext.logger = DcLogger()
+
+        dcAccounts.openDatabase()
+        migrateToDcAccounts()
+        if dcAccounts.getAll().isEmpty {
+            dcAccounts.add()
```
r10s, line 62: results in linter warning `Result of call to 'add()' is unused`. however, if that fails we cannot do much but thowing a fatal error to the user as needed at https://github.com/deltachat/deltachat-ios/pull/1272/files?diff=unified&w=1#diff-fc4377fbb4c60bfc9a11f53fd448de91bbf34787141509ddb85242515fea9e23R469 - do we have a function to show such an fatal error?

deltachat-ios/AppDelegate.swift
```diff
@@ -445,24 +450,47 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
+
+        if databaseHelper.unmanagedDatabaseLocation == nil {
+            return
+        }
+
+        if dcAccounts.migrate(dbLocation: databaseLocation) == 0 {
+            fatalError("Account could not be migrated")
+            // TODO: show error message in UI
+        }
+        databaseHelper.clearUnmanagedAccountData()
```
r10s, line 471: `dcAccounts.migrate()` has cleared/moved the unmanaged account data, this function should not be needed.

deltachat-ios/AppDelegate.swift
```diff
@@ -445,24 +450,47 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
 
-    func openDatabase() {
-        guard let databaseLocation = DatabaseHelper(dcContext: dcContext).updateDatabaseLocation() else {
+    func migrateToDcAccounts() {
+        let dcContext = DcContext()
+
+        // first database migration to shared container, can be removed at some point
+        // implemented in April 2020 (https://github.com/deltachat/deltachat-ios/pull/612)
+        let databaseHelper = DatabaseHelper(dcLogger: dcContext.logger)
+        guard let databaseLocation = databaseHelper.updateDatabaseLocation() else {
```
r10s, line 459: is that call really needed? even if `localDbFile` still exists and is returned as `unmanagedDatabaseLocation` below, migrate() should be able to move it.

r10s (commented):
already some comments ...

deltachat-ios/Helper/RelayHelper.swift
```diff
@@ -30,7 +30,7 @@ class RelayHelper {
 
     func forward(to chat: Int) {
         if let messageIds = self.messageIds {
-            RelayHelper.dcContext?.forwardMessages(with: messageIds, to: chat)
+            RelayHelper.dcAccounts?.getSelected().forwardMessages(with: messageIds, to: chat)
```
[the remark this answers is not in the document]
cyBerta: I think we should reset the messageIds when switching the account. That way we don't need to compare accountIds on forwarding.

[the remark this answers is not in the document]
cyBerta: But it's a good point you made

deltachat-ios/AppDelegate.swift
cyBerta: No, we don't have such a thing. Normally we're throwing fatal errors (like exceptions in Java world). So the app would just close and we would see the crashes in the reports.

deltachat-ios/AppDelegate.swift
cyBerta: cool, good to know :)

deltachat-ios/AppDelegate.swift
cyBerta: we could do sth. like
```
if dcAccounts.getAll().isEmpty, dcAccounts.add() == 0 {
      fatalError("Could not initialize a new account.")
}
```

deltachat-ios/AppDelegate.swift
cyBerta: True!

deltachat-ios/Helper/RelayHelper.swift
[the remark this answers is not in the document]
cyBerta: I did both now: DcContext is passed into RelayHelper and forwarding gets cancelled if the user switches the Account in between

r10s (approved):
yeah, i have no more comments so far :)

i also tested migrate in simulator (run once on master, then this pr), things seem to run as expected :)

PR #1279, opened 2021-06-22
ignore DC_MSG_ID_MARKER1 message 

### LINKED ISSUE #1276: date label stuck in the 70's
@dignifiedquire reported a bug where the date info message in a chat has an incorrect value. 

The reason for that bug was that the next message didn't have a proper sent date. It was another info message, the new messages label.

cyBerta:
... when picking up the date for the date info message

fixes #1276

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -529,12 +529,22 @@ class ChatViewController: UITableViewController {
         let id = messageIds[indexPath.row]
         if id == DC_MSG_ID_DAYMARKER {
             let cell = tableView.dequeueReusableCell(withIdentifier: "info", for: indexPath) as? InfoMessageCell ?? InfoMessageCell()
-            if messageIds.count > indexPath.row + 1 {
-                let nextMessageId = messageIds[indexPath.row + 1]
-                let nextMessage = dcContext.getMessage(id: nextMessageId)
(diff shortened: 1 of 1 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -529,12 +529,22 @@ class ChatViewController: UITableViewController {
...
+                    if nextMessageId != DC_MSG_ID_MARKER1 {
+                        let nextMessage = dcContext.getMessage(id: nextMessageId)
+                        cell.update(text: DateUtils.getDateString(date: nextMessage.sentDate))
+                        stop = true
+                    }
+                    n += 1
+                } else {
+                    cell.update(text: "ErrDaymarker")
+                    stop = true
+                }
```
r10s, line 547: i am a bit sceptical using a loop here, that may easily go wrong and we would have a deadlock (off by one? end conditon never set? things may change in the future and result in errors?) 

as we do accept everything but DC_MSG_ID_MARKER1 for  the date, and there is only one DC_MSG_ID_MARKER1 by definition, however, a loop seems not to be needed.

i would suggest to sth. as https://github.com/deltachat/deltachat-ios/pull/1280 therefore.

cyBerta: true, I was not sure about info messages in first place, but I left them out in the end

PR #1281, opened 2021-06-22
don't force max message input bar height for iPads in landscape mode

### LINKED ISSUE #1271: message input bar iPad layout broken
@Simon-Laux reported that the height of the messageInputBar on iPads is sometimes not correct (too high). 
leaving the chat view and returning 'fixes' the issue. 

image_2021-06-13_12-17-31

cyBerta:
fixes #1271


r10s (approved):
let's try it :)

PR #1282, opened 2021-06-23
video invite

### LINKED ISSUE #1273: Video Chat invitation
- [x] render video chat invitation correctly (same as on Android)+ (message type is [`DC_MSG_VIDEOCHAT_INVITATION`](https://c.delta.chat/group__DC__MSG.html#ga4289e30c154f83ee76c56bb1fcd8fe85))
- [x] add video chat provider settings to settings menu (setting name is [`webrtc_instance`](https://c.delta.chat/classdc__context__t.html#aff3b894f6cfca46cab5248fdffdf083d))
- [x] react to [`DC_QR_WEBRTC_INSTANCE`](https://c.delta.chat/classdc__context__t.html#a34a865a52127ed2cc8c2f016f085086c) on scanning a qr-code (using the existing scanner) and call  [`dc_set_config_from_qr()`](https://c.delta.chat/classdc__context__t.html#ac95d7275f01862fba70169b8cbb937c0)
- [x] add entry "Video chat" to attach menu to allow sending video chat invitations via [`dc_send_videochat_invitation()`](https://c.delta.chat/classdc__context__t.html#aa2995e39153c5007485541da6b5b0f1a)

cyBerta:
closes #1273


deltachat-ios.xcodeproj/project.pbxproj
```diff
@@ -24,6 +25,8 @@
 		302B84C6239676F0001C261F /* AvatarHelper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 30AC265E237F1807002A943F /* AvatarHelper.swift */; };
 		302B84C72396770B001C261F /* RelayHelper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 302B84C42396627F001C261F /* RelayHelper.swift */; };
 		302B84CE2397F6CD001C261F /* URL+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 302B84CD2397F6CD001C261F /* URL+Extension.swift */; };
+		302D5450268B6B2300A8B271 /* MessageUtils.swift in Sources */ = {isa = PBXBuildFile; fileRef = 302D544F268B6B2300A8B271 /* MessageUtils.swift */; };
+		302D5454268B84CB00A8B271 /* SettingsVideoChatInstanceController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 302D5453268B84CB00A8B271 /* SettingsVideoChatInstanceController.swift */; };
```
r10s, line 29: just a short comment, although still WIP: 

`SettingsVideoChatInstanceController.swift `  is a `ViewController`, right?

so we should add that to the name; even if some files are not named that way, we should try to get some consistency over time.

as other settings view controller start with settings, maybe `SettingsVideoChatViewController.swift`? that would also be fine if some future settings won't affect the instance only - and is even a bit shorter.

deltachat-ios/Chat/Views/Cells/VideoInviteCell.swift
```diff
@@ -0,0 +1,142 @@
...
+            bottomLabel.constraintAlignTrailingTo(messageBackgroundContainer),
+            bottomLabel.constraintToBottomOf(messageBackgroundContainer, paddingTop: 8),
+            bottomLabel.constraintAlignLeadingMaxTo(contentView, paddingLeading: UIDevice.current.userInterfaceIdiom == .pad ? 150 : 50),
+            bottomLabel.constraintAlignBottomTo(contentView, paddingBottom: 12)
+        ])
+        selectionStyle = .none
+    }
+
+    func update(dcContext: DcContext, msg: DcMsg) {
+        messageLabel.text = msg.text
```
r10s, line 115: the cell should not contain the text of the message but a dedicated string:

- for outgoing invites, use string `videochat_you_invited_hint`
- for incoming invited, use `videochat_contact_invited_hint` with the placeholder replaced by the contacts display name

deltachat-ios/Chat/Views/Cells/VideoInviteCell.swift
```diff
@@ -0,0 +1,142 @@
...
+    }()
+
+    lazy var openLabel: UILabel = {
+        let label = UILabel()
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.numberOfLines = 0
+        label.lineBreakMode = .byWordWrapping
+        label.textAlignment = .center
+        label.font = UIFont.preferredFont(for: .body, weight: .bold)
+        label.text = String.localized("open")
```
r10s, line 51: the label should be the string `videochat_tap_to_open` for outgoing invites and `videochat_tap_to_join` for incoming invites. i suggest to set that together where `messageLabel.text` is set.

deltachat-ios/Chat/Views/Cells/VideoInviteCell.swift
```diff
@@ -0,0 +1,142 @@
...
+
+    lazy var avatarView: InitialsBadge = {
+        let view = InitialsBadge(size: 28)
+        view.setColor(DcColors.systemMessageBackgroundColor)
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.setContentHuggingPriority(.defaultHigh, for: .horizontal)
+        return view
+    }()
+
+    lazy var videoIcon: InitialsBadge = {
```
r10s, line 24: i would like to have the video icon a tiny bit smaller (circle should stay as is) and closer to the avatar icon, cmp android.

maybe that already makes the circle more visible, otherwise, we should add a bit more contrast here (is that already the same shading as used eg. for date labels and info messages?)

both are no blockers, however.

r10s (approved):
tested a bit, and skimmed over code :)

r10s:
wow, that: https://github.com/deltachat/deltachat-ios/pull/1282/commits/5131c0844a9e93bc92be3cc7c08f42ed0ecd7f54 is a lot of stuff for just hiding the reply option ... it needs to be added and then filtered out? that's a lot of complicated code, looks etc. also an additional db access. wondering if it isn't better to just keep it.

desktop allows to reply to video chat invitations, so, might be that is not decided yet at all :) so maybe stash the last commit somewhere and remove it from this pr for now?

some other thing: 

-  "Attach / video chat" should be hidden as long as no instance is set.
- for the placeholder at "Settings / Video chat instance / Video chat instance: " i would use the string "URL"

cyBerta:
> that's a lot of complicated code, looks etc. also an additional db access.

performance wise it's not a big deal. the code is only executed on longpress on a cell, not while scrolling for example.

anyways, I'm going to remove it

deltachat-ios/Chat/Views/Cells/VideoInviteCell.swift
cyBerta: ok, I did both, reduced the size of the video Icon and darkened a little bit the background color for system messages

cyBerta:
>     * for the placeholder at "Settings / Video chat instance / Video chat instance: " i would use the string "URL"

this should be fixed on Android, too. There also `R.string.videochat_instance_placeholder` is used as the placeholder.

cyBerta:
>     * for the placeholder at "Settings / Video chat instance / Video chat instance: " i would use the string "URL"

I will leave it as it is for now. In contranst to "Video chat instance" there's no translations for "URL" and there might be a more adequate word for it in different languages.

PR #846, opened 2020-07-14
[wip] switch accounts

r10s:
once done, this pr will add multi-account support to deltachat-ios as described in #829

PR #1286, opened 2021-06-25
add basic switch-account controls

### LINKED ISSUE #829: add multi-account support
add a basic multi-account support as on android and desktop (only one active account at a time) using [new dc_accounts_t core functionality](https://c.delta.chat/classdc__accounts__t.html) that may need tweaking before.

- at "settings" there should be a "switch account" button at the end of the screen

- this button should allow account switching/creation/deletion

- the existing "delete account" at "settings / your profile info / password and account" can be removed

- scanning a create-account qr-code shall offer to create a new account directly (as on android)

- notifications  and donateSendMessageIntent() need to be adapted as chatId is ambiguous on multi-account; a simple approach would be to remove all notifications/donations on account switching

### LINKED ISSUE #1288: logging not working in multi-account, too many logger objects
logging does not really work with multi-account as the logger object is bound to DcContext - which is not always done when calling get(), getSelected() etc.

however, that also does not seems to be needed, we have a global logger object in created in AppDelegate - why not just use that?

as a 3rd logging thing, there is SimpleLogger, wondering if that is really needed.

might be that this maybe unneeded mess is also responsible for #1131 somehow.

r10s:
this pulls in the basic ui from #846 and closes #829 and closes #1288

successor of #1268, #1272

- [x] add ui
- [x] "add account"
- [x] "switch account"
- [x] "delete account" (if possible, that could delete the selected account; a list as on android is not-so-nice)
- [x] "welcome screen" needs a "cancel" button if there are other accounts available, on "cancel" the unconfigured account can be deleted
- [x] remove old "delete account" option
- [x] check that "import" works as expected ([`dc_accounts_import_account()`](https://c.delta.chat/classdc__accounts__t.html#ab554e69f329dd19b099a9bc425db925f) creates a new account) - EDIT: we keep on using [`dc_imex()`](https://c.delta.chat/classdc__context__t.html#ab04a07bb49363824f6fe3b03e6aaaca7), so things should be fine. wondering if `dc_accounts_import_account()` is needed at all, cmp https://github.com/deltachat/deltachat-core-rust/pull/2521
- [x] check location streaming, old delete account called disableLocationStreamingInAllChats on remove - what is needed now? however, as being experimental anyway, this is maybe sth. for another pr.

**there is a bug left:** subsequent account configuration does not end (account is okay after restart), however, that needs to be fixed in core, see https://github.com/deltachat/deltachat-core-rust/issues/2522

### DIFF
--- a/deltachat-ios/Controller/SettingsController.swift
+++ b/deltachat-ios/Controller/SettingsController.swift
@@ -469,6 +469,7 @@ internal final class SettingsViewController: UITableViewController, ProgressAler
                     message: String.localized("forget_login_confirmation_desktop"), preferredStyle: .alert)
                 confirm2.addAction(UIAlertAction(title: String.localized("delete"), style: .destructive, handler: { [weak self] _ in
                     guard let self = self else { return }
+                    appDelegate.locationManager.disableLocationStreamingInAllChats()
                     _ = self.dcAccounts.remove(id: selectedAccountId)
                     if self.dcAccounts.getAll().isEmpty {
                         _ = self.dcAccounts.add()
--- a/deltachat-ios/Controller/SettingsController.swift
+++ b/deltachat-ios/Controller/SettingsController.swift
@@ -438,38 +438,52 @@ internal final class SettingsViewController: UITableViewController, ProgressAler
 
     private func showSwitchAccountMenu() {
         let accountIds = dcAccounts.getAll()
-        let selectedAccount = dcAccounts.getSelected()
+        let selectedAccountId = dcAccounts.getSelected().id
+        guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { return }
 
         // switch account
-        let alert = UIAlertController(title: String.localized("switch_account"), message: nil, preferredStyle: .safeActionSheet)
+        let menu = UIAlertController(title: String.localized("switch_account"), message: nil, preferredStyle: .safeActionSheet)
         for accountId in accountIds {
             let account = dcAccounts.get(id: accountId)
-            var title = account.addr ?? ""
-            if let displayname = account.displayname {
-                title = "\(displayname) (\(title))"
-            }
-            title += account.configured ? "" : " (not configured)"
-            title = (selectedAccount.id==accountId ? "✔︎ " : "") + title
-            alert.addAction(UIAlertAction(title: title, style: .default, handler: { [weak self] _ in
-                guard let self = self else { return }
-                if self.dcAccounts.select(id: accountId) {
-                    // TODO: go to chatlist with new account
-                }
+            var title = account.displaynameAndAddr
+            title = (selectedAccountId==accountId ? "✔︎ " : "") + title
+            menu.addAction(UIAlertAction(title: title, style: .default, handler: { [weak self] _ in
+                _ = self?.dcAccounts.select(id: accountId)
+                appDelegate.reloadDcContext()
             }))
         }
 
         // add account
-        alert.addAction(UIAlertAction(title: String.localized("add_account"), style: .default, handler: { [weak self] _ in
-            // TODO
+        menu.addAction(UIAlertAction(title: String.localized("add_account"), style: .default, handler: { [weak self] _ in
+            _ = self?.dcAccounts.add()
+            appDelegate.reloadDcContext()
         }))
 
         // delete account
-        if accountIds.count > 1 {
-            alert.addAction(UIAlertAction(title: String.localized("delete_account"), style: .default, handler: nil)) // TODO
-        }
+        menu.addAction(UIAlertAction(title: String.localized("delete_account"), style: .default, handler: { [weak self] _ in
+            let confirm1 = UIAlertController(title: String.localized("delete_account_ask"), message: nil, preferredStyle: .safeActionSheet)
+            confirm1.addAction(UIAlertAction(title: String.localized("delete_account"), style: .destructive, handler: { [weak self] _ in
+                guard let self = self else { return }
+                let account = self.dcAccounts.get(id: selectedAccountId)
+                let confirm2 = UIAlertController(title: account.displaynameAndAddr,
+                    message: String.localized("forget_login_confirmation_desktop"), preferredStyle: .alert)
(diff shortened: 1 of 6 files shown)


r10s:
> fix bug: subsequent account configuration does not end (account is okay after restart)

it is still unclear to me what happens here. it seems we're not getting events during configure, therefore the progress is not updated and not finished. i suspect deltachat-ios more as deltachat-core here, however :) @link2xt on kdeltachat, that works, i assume?

cyBerta:
>     * check location streaming, old delete account called disableLocationStreamingInAllChats on remove - what is needed now? however, as being experimental anyway, this is maybe sth. for another pr

From my point of view, we need to disable location streaming in all chats of the deleted account. 
In case of switching an account we won't do anything, as an account is allowed to send and receive mails while another account is active.

cyBerta:
> > fix bug: subsequent account configuration does not end (account is okay after restart)
> 
> it is still unclear to me what happens here. it seems we're not getting events during configure, therefore the progress is not updated and not finished. i suspect deltachat-ios more as deltachat-core here, however :) @link2xt on kdeltachat, that works, i assume?

I didn't dive into this issue, but I remember this note:

> Note: Use only one event emitter per account manager. Having more than one event emitter running at the same time on the same account manager will result in events randomly delivered to the one or to the other. 

https://c.delta.chat/classdc__accounts__t.html#ad74f8ea4e3ad2cb14f459a0661238c9e

r10s:
> Note: Use only one event emitter per account manager. Having more than one event emitter running at the same time on the same account manager will result in events randomly delivered to the one or to the other.

we only have one event emitter, so that seems to be fine. the issue seems to be that the newly added account is not part of the existing event emitter, this seems to be a core issue, see https://github.com/deltachat/deltachat-core-rust/issues/2522

> From my point of view, we need to disable location streaming in all chats of the deleted account.

k, i will re-add the corresponding statement, we'll see how that works out.

deltachat-ios/Controller/WelcomeViewController.swift
```diff
@@ -143,6 +155,24 @@ class WelcomeViewController: UIViewController, ProgressAlertHandler {
...
+        // (normally, both checks are not needed, however, some resilience wrt future program-flow-changes seems to be reasonable here)
+        let selectedAccount = dcAccounts.getSelected()
+        if !selectedAccount.isConfigured() {
+            _ = dcAccounts.remove(id: selectedAccount.id)
+            if self.dcAccounts.getAll().isEmpty {
+                _ = self.dcAccounts.add()
+            }
+        }
+
+        // do not just pop view controller as the stack is empty on restarts
```
cyBerta, line 173: What does this comment mean? 

`appDelegate.reloadDcContext()` is exactly doing that: popping view controllers

deltachat-ios/Controller/SettingsController.swift
```diff
@@ -420,6 +430,63 @@ internal final class SettingsViewController: UITableViewController, ProgressAler
...
+        guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { return }
+
+        // switch account
+        let menu = UIAlertController(title: String.localized("switch_account"), message: nil, preferredStyle: .safeActionSheet)
+        for accountId in accountIds {
+            let account = dcAccounts.get(id: accountId)
+            var title = account.displaynameAndAddr
+            title = (selectedAccountId==accountId ? "✔︎ " : "") + title
+            menu.addAction(UIAlertAction(title: title, style: .default, handler: { [weak self] _ in
+                _ = self?.dcAccounts.select(id: accountId)
```
cyBerta, line 451: maybe do `guard let self = self else { return }` here as well, we probably don't want to call `appDelegate.reloadDcContext()` in case self is not existing enymore.

deltachat-ios/Controller/WelcomeViewController.swift
r10s: the comment should underline that a simple `popViewController()`is not sufficient in all cases (but might work in some), and in fact `appDelegate.reloadDcContext()` does not use `popViewController()`.

but i agree, that this comment is obviously not helpful. i will remove it :)

deltachat-ios/Controller/SettingsController.swift
r10s: yip, that's better, i added a commit.

deltachat-ios/Controller/SettingsController.swift
```diff
@@ -420,6 +430,65 @@ internal final class SettingsViewController: UITableViewController, ProgressAler
...
+
+        // delete account
+        menu.addAction(UIAlertAction(title: String.localized("delete_account"), style: .default, handler: { [weak self] _ in
+            let confirm1 = UIAlertController(title: String.localized("delete_account_ask"), message: nil, preferredStyle: .safeActionSheet)
+            confirm1.addAction(UIAlertAction(title: String.localized("delete_account"), style: .destructive, handler: { [weak self] _ in
+                guard let self = self else { return }
+                let account = self.dcAccounts.get(id: selectedAccountId)
+                let confirm2 = UIAlertController(title: account.displaynameAndAddr,
+                    message: String.localized("forget_login_confirmation_desktop"), preferredStyle: .alert)
+                confirm2.addAction(UIAlertAction(title: String.localized("delete"), style: .destructive, handler: { [weak self] _ in
```
cyBerta, line 472: I think needing to tap 3 times on delete to actually delete the account is one step too much. Android doesn't have it too (it has an extra step to select the account for deletion though)

r10s: i'd say this double-confirmation is not that bad, esp. as on iphone, the "confirm" buttons are at different places. a simple confirmation as for deletion of a single chat or message seems to be a bit too less.

r10s:
> isConfigured check 

i think, it is fine that that pops up in the list, it currently mainly happens because of the core bug i'd say.

cyBerta (approved):

deltachat-ios/Controller/SettingsController.swift
dumblob: I also think on iOS devices it's better to have more confirmations than less.

PR #1287, opened 2021-06-28
share adaptions

### LINKED ISSUE #1275: share adaptions
- [x] if possible: revoke share donations on switching account -> this should avoid that shortcuts are shown for accounts that are not currently selected
- [x] same for chat and contact deletion: if possible revoke share donations so that shortcuts of deleted contacts/chats are never shown
- [ ] ~~if revocation is impossible: switch accounts on selection of a shortcut - this involves that shortcut donation id's contain context ids that we can parse the related account~~

cyBerta:
closes #1275

remove donated siri message intents:
* if account gets switched
* if a chat gets deleted
* if a contact gets deleted


cyBerta:
as mentioned on IRC I'll remove all donated siri intents as part of the multi-account migration. this way we avoid stale shortcuts that remain for a while until they vanish by newer shortcuts

PR #1293, opened 2021-07-01
update badge counter on account switch

### LINKED ISSUE #1292: badge on app-icon not reset on account switching
on switching account, the notificiations seems to be removed, however, the badge on the app-icon not reset.

cyBerta:
fixes #1292

Instead of resetting the badge to 0 we're updating to the unread messages of the newly selected account for now. This way the badge counter aligns with the sum of all unread message counters in the chat list.


Simon-Laux (approved):
LGTM

PR #1294, opened 2021-07-05
Contact Requests overhaul

cyBerta:
depending on https://github.com/deltachat/deltachat-core-rust/pull/2514 to be merged

other than that ready for review


r10s:
did not go through the code yet, but a little layout comment:

i assume contact requests are not notified? if so, the badge of contact requests should not be _red_ but in the _grey_ of muted chats.

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -266,15 +274,20 @@ class ChatViewController: UITableViewController {
         configureEmptyStateView()
 
         if !disableWriting {
-            configureMessageInputBar()
-            draft.parse(draftMsg: dcContext.getDraft(chatId: chatId))
-            messageInputBar.inputTextView.text = draft.text
-            configureDraftArea(draft: draft, animated: false)
-            editingBar.delegate = self
```
r10s, line 273: `editingBar.delegate = self`  is not needed in `configureUIForWriting()` as it was always already done in `lazy var editingBar`, right?

deltachat-ios/Controller/MailboxViewController.swift
```diff
@@ -1,68 +0,0 @@
-import UIKit
```
r10s, line 1: cool that we never renamed that to `ContactRequestsViewController` - would have been a waste of time :)

”Wisdom of a Stone" :)

r10s (approved):
lgtm, there are some minor things to be done, but that should better go to subsequent prs to have a ios-master working again with core-master.

deltachat-ios/Chat/ChatViewController.swift
cyBerta: that's right

deltachat-ios/Controller/MailboxViewController.swift
cyBerta: 🤣

PR #1298, opened 2021-07-15
force initiating preview with correct size

### LINKED ISSUE #1285: fix share layout for ipads
tested on an ipad simulator:

simulator_screenshot_5CF797B1-18DE-420E-8EFE-EB1D2F601B20

cyBerta:
fixes #1285

### DIFF
--- a/DcShare/Controller/ShareViewController.swift
+++ b/DcShare/Controller/ShareViewController.swift
@@ -47,11 +47,16 @@ class ShareViewController: SLComposeServiceViewController {
     var previewImageWidthConstraint: NSLayoutConstraint?
 
     lazy var preview: SDAnimatedImageView? = {
-        let imageView = SDAnimatedImageView(frame: .zero)
+
+        UIGraphicsBeginImageContext(CGSize(width: 96, height: 96))
+        let image = UIGraphicsGetImageFromCurrentImageContext()
+        UIGraphicsEndImageContext()
+        let imageView = SDAnimatedImageView(image: image)
         imageView.clipsToBounds = true
         imageView.shouldGroupAccessibilityChildren = true
         imageView.isAccessibilityElement = false
         imageView.contentMode = .scaleAspectFit
+        imageView.translatesAutoresizingMaskIntoConstraints = false
         previewImageHeightConstraint = imageView.constraintHeightTo(96)
         previewImageWidthConstraint = imageView.constraintWidthTo(96)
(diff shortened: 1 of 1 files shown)


r10s:
ah, so that scales up the preview window?

cyBerta:
normal images: 

Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-07-15 at 18 34 47



webp, converted as stickers, text is not allowed:

Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-07-15 at 18 34 05

the screenshots are made on an simulator, the keyboard is configured to not be shown there. 
The half-transparent background is the default background of the SLComposeServiceViewController, the standard iOS comonent for sharing.

r10s:
> webp, converted as stickers, text is not allowed:

offtopic - but why is that image in webp format? @dignifiedquire posted recently screenshots in the ios group that also arrive as stickers unexpectedly - so, maybe we should be a bit more conservative with the sticker-image-detection and use images when in doubt, esp. on sharing (keyboard is maybe differently)

cyBerta:
> > webp, converted as stickers, text is not allowed:
> 
> offtopic - but why is that image in webp format? @dignifiedquire posted recently screenshots in the ios group that also arrive as stickers unexpectedly - so, maybe we should be a bit more conservative with the sticker-image-detection and use images when in doubt, esp. on sharing (keyboard is maybe differently)

I chose a webp image from https://en.wikipedia.org/wiki/WebP. We said that we would handle shared webp images as stickers. We can change that of course.

cyBerta:
But I also wonder why @dignifiedquire's screenshot was in webp format.

r10s:
> @r10s shall we just remove the option to share webp files as stickers? Maybe it causes more confusion than helping users to play with stickers.

yip, probably makes sense, seems as if webp is used eg. for screenshots, at least sometimes.

that does not affect sharing via keyboard, right?

cyBerta:
I'm going to implement

> > @r10s shall we just remove the option to share webp files as stickers? Maybe it causes more confusion than helping users to play with stickers.
> 
> yip, probably makes sense, seems as if webp is used eg. for screenshots, at least sometimes.
> 
> that does not affect sharing via keyboard, right?

Yes it doesn't affect sharing via keyboard. I'm going remove that share-webp-as-stickers-part as  a separate PR.

r10s (approved):

PR #1300, opened 2021-07-19
Connectivity view

### LINKED ISSUE #1299: add connectivity view
- [x] in the chatlist's title bar, show the current state as returned from the new api [`dc_get_connectivity()`](https://c.delta.chat/classdc__context__t.html#a80bf9812eb5a50f8eb097cbe72df1096).  
for the state DC_CONNECTIVITY_CONNECTED we show the string "Delta Chat" as before, for the other states, show the strings `connectivity_getting_new_msgs`, `connectivity_connecting` and `connectivity_not_connected` 
we do not want to show dots or other icons currently.
the text should be updated on [`DC_EVENT_CONNECTIVITY_CHANGED`](https://c.delta.chat/group__DC__EVENT.html#ga3ced1bbde752feba46ab089611287d71).
(corresponding [android code](https://github.com/deltachat/deltachat-android/blob/b3fea81bed308275425e0136233465f589cf4fe0/src/org/thoughtcrime/securesms/connect/DcHelper.java#L79))

- [x] a tap on the chatlist's title bar text (so also on "Delta Chat") should open `ConnectivityViewConroller` (based on `WebViewViewController`) that just shows the html-code returned by [`dc_get_connectivity()`](https://c.delta.chat/classdc__context__t.html#a98d41f00d9a96d83a662f39d0fc23afb).  
the page should be updated on [`DC_EVENT_CONNECTIVITY_CHANGED`](https://c.delta.chat/group__DC__EVENT.html#ga3ced1bbde752feba46ab089611287d71).

- [x] the same `ConnectivityViewConroller` should be accessibe from the settings (entry above "help") (maybe with disclosureIndicator and textLabel from `dc_get_connectivity()`, if so, it should also be updated on `DC_EVENT_CONNECTIVITY_CHANGED`)

all that is implemented on android at https://github.com/deltachat/deltachat-android/pull/1860 (cc @Hocuri - did i missed an important point?)

cyBerta:
closes #1299

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -323,6 +323,13 @@ public class DcContext {
         return dc_get_connectivity(contextPointer)
     }
 
+    public func getConnectivityHtml() -> String {
+        guard let cString = dc_get_connectivity_html(contextPointer) else { return ""}
+        let swiftString = String(cString: cString)
+        dc_str_unref(cString)
+        return swiftString
+    }
+
     public func setStockTranslation(id: Int32, localizationKey: String) {
         dc_set_stock_translation(contextPointer, UInt32(id), String.localized(localizationKey))
     }
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -47,6 +47,10 @@ public class DcAccounts {
         dc_accounts_maybe_network(accountsPointer)
     }
 
+    public func maybeNetworkLost() {
+        dc_accounts_maybe_network_lost(accountsPointer)
+    }
+
     public func maybeStartIo() {
         if getSelected().isConfigured() {
             dc_accounts_start_io(accountsPointer)
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -319,6 +319,10 @@ public class DcContext {
         return "ErrGetContactEncrInfo"
     }
 
+    public func getConnectivity() -> Int32 {
+        return dc_get_connectivity(contextPointer)
+    }
+
     public func setStockTranslation(id: Int32, localizationKey: String) {
         dc_set_stock_translation(contextPointer, UInt32(id), String.localized(localizationKey))
     }
--- a/DcCore/DcCore/DC/events.swift
+++ b/DcCore/DcCore/DC/events.swift
@@ -227,7 +227,7 @@ public class DcEventHandler {
             }
 
         case DC_EVENT_CONNECTIVITY_CHANGED:
-            dcContext.logger?.info("network: DC_EVENT_CONNECTIVITY_CHANGED")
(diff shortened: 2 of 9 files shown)


r10s:
did a quick functional test:
- the app crashes when switching flightmode on/off - unfortunately, also the connection to a debugger is lost, so i could not easily check a log; i also failed in reproducing, but we should double-check
- the font-size in the title bar is increased, it is now too large (cmp. system "Settings" title with dc ones) - that does not only affect the chatlist but all views

cyBerta:
Thanks for having a look at the PR.

> the update of the title bar is not always happening, could not really debug that because of the crashes :) (seems as if the first update works, i have never seen subsequent update when switching flightmode several times on+off)

I think this issue is core related. I've seen the same problem and added a little bit more verbose logging with commit https://github.com/deltachat/deltachat-ios/pull/1300/commits/063402e5e1fe0e093234a5b3d83fc154d5c56ddd. 
When turning on flightmode we're getting immediately `DC_EVENT_CONNECTIVITY_CHANGED`. If I call dc_get_connectivity() directly afterwards the state is still 4000 (`DC_CONNECTIVITY_CONNECTED`).

r10s:
> I think this issue is core related.

hm, on android we do not have that issue - or are events somehow debounced there? cc @Hocuri

cyBerta:
https://github.com/deltachat/deltachat-android/pull/1860/files doesn't seem to have debouncing implemented. 
I'm wondering how if multi-account is playing a role here. We're getting the event for each account but we also process the event on each dcContext (`let dcContext = dcAccounts.get(id: event.accountId)`. On any of those dcContext the result of  `dc_get_connectivity` is correct

Hocuri:
> hm, on android we do not have that issue - or are events somehow debounced there? cc @Hocuri

No. But I suspect that the Android system somehow debounces the updating of the title bar.

Generally we can debounce connectivity events in the core, maybe Desktop will also need that, anyway

cyBerta:
I don't think this issue is related to debouncing. To me, it looks like the detection of the current connectivity doesn't work reliable on iOS.
We're having a swift pod called 'ReachabilitySwift' that does a similar job. We _could_ use that platform specific implementation instead of the core as a workaround.

cyBerta:
@r10s I couldn't reproduce the crash yet. Would it be possible to get the logs and attach them here?

r10s:
>  Would it be possible to get the logs and attach them here?

i'll have a look, however, debugger is detached by the crash, so, not sure what there is available locally.

cyBerta:
I've created a ticket here: https://github.com/deltachat/deltachat-core-rust/issues/2547

r10s (approved):
lgtm, the last commit however requires https://github.com/deltachat/deltachat-core-rust/pull/2550 to be merged.

@cyBerta i think, you should also check my latest commits :)

r10s:
> lgtm, the last commit however requires deltachat/deltachat-core-rust#2550 to be merged.

that is merged now and i rebased the pr.

r10s:
\o/

PR #1304, opened 2021-07-21
fix scrolling of the chat when the inputbar size increases

cyBerta:
changing the way we scroll down in case the input bar size changes seems to resolve the sudden (and unwanted) scroll animation

the tested scenario: write a message, send it, write a second one with at least two lines of text. Without this PR, the scrolling animation  showed up, now it should not anymore. 

@r10s could you please check if you can still reproduce this particular scroll bug with this PR?


r10s (approved):
i think, we should merge it in, it does not happen all the time, but during usage. we will know after some hours of using if things are finally better :)

PR #1307, opened 2021-07-26
Scroll down button

### LINKED ISSUE #1032: Add button to scroll to the bottom of the chat
Almost all other messengers have this, and I found myself wishing for this recently again. 

- Hidden by default.
- When scrolling up, such that the last message is not visible anymore, show a button at the right bottom corner, above the input field, that when tapped scrolls to the bottom of the chat again.

cyBerta:
closes #1032


deltachat-ios/Assets.xcassets/ic_scrolldown.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
+{
+  "images" : [
```
r10s, line 2: i suggest to add suffixes to the filenames matching their resolution, as `_x1.png`, `_x2.png` , `_x3.png` - see the other assets to get an idea.

r10s (commented):
apart from the comment below, that lgtm. did not test it yet (in queue :)

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -463,7 +464,7 @@ class ChatViewController: UITableViewController {
         nc.removeObserver(self, name: UIApplication.didBecomeActiveNotification, object: nil)
         nc.removeObserver(self, name: UIApplication.willResignActiveNotification, object: nil)
         audioController.stopAnyOngoingPlaying()
-
+
```
r10s, line 467: whitespace error, for whatever reason, that prevents rebasing, so better fix that :)

r10s (approved):
just tested that - works great and looks great, tested in light and in darkmode :)

i suggest to target the two minor things, and then merge that in :)

PR #1308, opened 2021-07-26
Contact request tweaks

### LINKED ISSUE #1289: adapt to upcoming contact-request api
see https://github.com/deltachat/deltachat-android/issues/1963 and https://github.com/deltachat/deltachat-core-rust/pull/2514 for details, this issue is mainly to remind about the changes needed.

### LINKED ISSUE #1306: contact request tweaks
there are some more tweaks needed, from https://github.com/deltachat/deltachat-ios/pull/1294#issuecomment-886063391 :

- i assume contact requests are not notified? if so, the badge of contact requests should not be _red_ but in the _grey_ of muted chats.
EDIT: or do not show them at all as on android?, cmp https://github.com/deltachat/deltachat-android/issues/1990
EDIT II: i tend to just leave the badge out completely for contact requests. that makes things less noisy, which is a good thing here, i'd say. and we're on pair with android :)

- switch the order of the buttons to "Block - Accept" (Accept should be right, also, same as on android)

- for groups, the label should read "Delete" and not "Block" (currently core cannot block groups, if that is added at some point, we can modify the label again)

- "Delete" and "Block" should be shown in red (even if partly recoverable)

- on android "Delete" shows a confirmation dialog - probably makes some sense

- the "Contact Request" label should just read "Request"

all strings are available on transifex now, i just pulled the translations.

cyBerta:
closes #1306 
closes #1289

successor of #1294

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -1053,6 +1053,18 @@ class ChatViewController: UITableViewController {
             }))
         present(alert, animated: true, completion: nil)
     }
+    
+    private func askToDeleteChat() {
+        let title = String.localized(stringID: "ask_delete_chat", count: 1)
+        confirmationAlert(title: title, actionTitle: String.localized("delete"), actionStyle: .destructive,
+                          actionHandler: { [weak self] _ in
+                            guard let self = self else { return }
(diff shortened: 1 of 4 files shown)


r10s (approved):
lgtm

PR #1309, opened 2021-07-27
Rename back btn

### LINKED ISSUE #1305: rename Back button
In chat profile we want to have 'Chat' instead of 'Back' as string on the back button

cyBerta:
closes #1305


r10s (approved):
lgtm, nice! depending on the width of the screen, "<", "< Back" or "< Settings" is shown. i was condused by the "Back" 2y ago, however, it is just nice ios, this is what happens in all apps :)

i rebased the pr and added a minor commit.

PR #1313, opened 2021-07-28
Fix account setup

### LINKED ISSUE #1290: scan account broken
on new installation:

- after the question "allow delta chat to access camera for qr code scan", a white screen is shown, waiting a bit let the app crash (not sure about the latter)

- on restart, camera works, however, the progress dialog is stuck at "one moment ..."

the bug appears on [current master](https://github.com/deltachat/deltachat-ios/commit/57dd4093297fe7ff258424f867b49e827e91d45c) as well as in https://github.com/deltachat/deltachat-ios/pull/1286, however, the latter seems to be unrelated to https://github.com/deltachat/deltachat-core-rust/issues/2522 as due to restart, the events should arrive.

the bug does not exist in current app store release (1.20.5), so it is introduced recently.

might be advantageous to target this issue after fixing https://github.com/deltachat/deltachat-core-rust/issues/2522 as it may be hard to distinguish these bugs.

cyBerta:
closes #1290

### DIFF
--- a/deltachat-ios/Controller/AccountSetupController.swift
+++ b/deltachat-ios/Controller/AccountSetupController.swift
@@ -677,8 +677,10 @@ class AccountSetupController: UITableViewController, ProgressAlertHandler {
             notification in
             if let ui = notification.userInfo {
                 if ui["error"] as! Bool {
+                    self.dcAccounts.maybeStartIo()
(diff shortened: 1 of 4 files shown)


r10s (approved):
code looks good - and i tested a qr-code account and a manual setup - both work fine :)

thanks a lot for figuring that out!

PR #1315, opened 2021-07-28
fix inputbarAccessoryView background color

### LINKED ISSUE #1312: new request buttons look weird in dark mode
in dark mode, the new request buttons have a weird white border:

unless you switch github to dark mode, it is not so good to see on the screenshot. the border is top, right and buttom, _not_ on the left. might be, the border is present also in lite mode, however, invisible there.

i think, the border can be removed completely; there is already the thin grey line, that should be sufficient.

cyBerta:
fixes #1312 🐷 🚓


r10s (approved):
lgtm - for the emojis in the pr description, i have some assumptions, but not sure, however ;)

PR #1318, opened 2021-07-28
fix flickering background when entering the connectiviy view

### LINKED ISSUE #1317: flickering connection view background in dark mode
when having dark mode enabled and entering the connection view, you can see it has initially a white background and turns black afterwards. 
I think the default background color of the WKWebview should be adapted to the dark mode, that should mitigate the effect.

cyBerta:
fixes #1317 

fixes the same bug for the help pages

the fix not implemented in the root class (WebViewViewController) because FullMessageViewController doesn't adapt font colors in dark mode, thus we keep the webview background color there alsways white.


r10s (approved):
nice!

great to take care of such details!

PR #1319, opened 2021-07-28
prepare 1.21.0 testflight ✈️

r10s:
i submitted to testflight from this pr already; will take 2 days or so anyway.


cyBerta (approved):
wow, that's a lot :partying_face:

PR #1328, opened 2021-08-04
don't add message on DC_EVENT_MSGS_CHANGED if message is a draft

### LINKED ISSUE #1327: draft becomes part of chat visually
- type a draft
- stay in chat
- leave the app and restart it (eg. tap home system key and tap the icon again)

-> the draft is now shown as a weird message:

cyBerta:
fixes #1327

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -1266,7 +1266,8 @@ class ChatViewController: UITableViewController {
         } else {
             // new outgoing message
             let msg = dcContext.getMessage(id: messageId)
-            if msg.chatId == chatId {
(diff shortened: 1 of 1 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1266,7 +1266,8 @@ class ChatViewController: UITableViewController {
         } else {
             // new outgoing message
```
r10s, line 1267: i think, this else-branch is not only about "new outgoing message" but also about incoming messages, not in the chatlist for whatever reason, so, it still may be wrong to insert that message to the end.

however, this is another issue, so, we can leave that as is for now.

r10s (approved):

PR #1332, opened 2021-08-05
[postponed] tappable openpgp4fpr codes

### LINKED ISSUE #1331: send group invites to other groups
in addition to allow adding someone to a group manually or by showing a qr code, it would be nice to send an invite to someone who can then decide whether to join or not at some later point.

we already have the `OPENPGP4FPR:` protocol working on all systems, if you tap the link from eg, the browser, Delta Chat is started and the secure-join is stated.

also, you can already copy the `OPENPGP4FPR:` links to clipboard on desktop. 

what's missing:

- add a "copy" button to the "show qr code" page, the upper right corner is unused, so, just put it there
- make `OPENPGP4FPR:` links clickable and working.

once that is done, issues with the missing parts should be filed for android/ios.

r10s:
closes #1331, however, should be tested before merging and maybe also being merged after 1.22 is released (changing basic regex just before a release is maybe not advisable.

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -135,12 +135,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
         switch url.scheme?.lowercased() {
         case "openpgp4fpr":
-            // Hack to format url properly
-            let urlString = url.absoluteString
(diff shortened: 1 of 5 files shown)


r10s:
postponed as it should be figured out before if that may weak the two-factor component in the qr-codes.

PR #1338, opened 2021-08-20
add 'connectivity changed' observers in time

### LINKED ISSUE #1337: connectivity title is stuck sometimes
sometimes, the title shows "Connecting..." or even "Not connected" while the connection is actually okay.

i just had a look: the reason is that we update the title before we add the DC_EVENT_CONNECTIVITY_CHANGED observer - that way, we may miss an event.

same is true for the connectivity-html-view and for the connectivity-settings-label - in all cases the event handler must be installed _before_ the state is set to a control.

r10s:
'connectivity changed' observers must be added before
the UI is inititalized - otherwise,
we may miss an event and UI is stuck at a wrong state.

in general, that means to move adding observers to viewWillAppear -
which seems also be more balanced as they're removed on viewDidDisappear -
so both in situations the view is not visible.

@cyBerta please have a closer look at this pr after your holidays - i merge that in to move forward, however.

closes #1337 (magic number!)

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -93,6 +93,15 @@ class ChatListController: UITableViewController {
 
         // create view
         navigationItem.titleView = titleView
+
+        // set connectivity changed observer before we acutally init the title,
+        // otherwise, we may miss events and the title is not correct.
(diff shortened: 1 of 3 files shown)


cyBerta (commented):
lgtm

r10s:
thanks!

Issue #1330, opened 2021-08-05
sharing panoramas not working


r10s:
sharing a panorama to delta chat does not work - the panorama is not shared, and the share-to dialog does not disappear (tested 2021 on iphone7 with ios 14.7.1) or the share-to-dialog appears but sharing fails then (tested 2023 on iphone13 with ios 16.1.2). sharing normal photos to delta chat works as expected.

r10s:
edit: same is true for videos shared from the gallery to delta

cyBerta:
the debugger shows 
`findWriterForTypeAndAlternateType:119: unsupported file format 'public.heic'`

the exception thrown is: `worker thread for component 0 (36): EXC_RESOURCE RESOURCE_TYPE_MEMORY (limit=120 MB, unused=0x0)`

the exception indicates that too much memory is used, probably in the method findWriterForTypeAndAlternateType (core?).
The exception occurs after calling `dcContext.sendMsgSync`. I checked the memory usage and indeed the memory usage raises from ~70MB to +120 MB. I will try to release as much memory as possible on the UI side, but it might be necessary to work on the core side as well.

cyBerta:
The same spike in memory consumption  (only for processing the image ~250 mb) appears if I send a panorama image by attaching them in a chat. The only difference is that there's no hard memory limitation of 120 mb in our main app. That's why sending panorama images from the app works while sending them from the extension doesn't. 

I have 3 proposals how to improve the current situation:
1. We reduce the memory consumption for showing the thumbnail and creating DcMsgs.
2. In the core the memory consumption should be optimized wrt. image processing if possible
3. if an image exceeds a certain size we could try to open the main app to proceed with sending

r10s:
but there should be at least a correct error message - that is where i would begin with.

cyBerta:
> but there should be at least a correct error message - that is where i would begin with.

it's impossible to do any error handling after `EXC_RESOURCE RESOURCE_TYPE_MEMORY` has been thrown. The only thing we can do is to check the size of an image and estimate the used size in the core and show an error on these estimated restrictions.

r10s:
> a jpg of 9.5mb already consumes 285mb internal memory while dcContext.sendMsgSync is called.

can you tell the pixel dimensions of that image? 250mb for recoding in core is a lot, but may still be reasonable somehow.

what about the .heic format - is that an issue as well?

cyBerta:
the example image has the dimensions 8390 × 3780 
[...]

PR #1350, opened 2021-09-13
reduce memory consumption of share extension 

cyBerta:
by scaling down image thumbnail

fixes point one of https://github.com/deltachat/deltachat-ios/issues/1330#issuecomment-918108809

### DIFF
--- a/DcShare/Helper/ShareAttachment.swift
+++ b/DcShare/Helper/ShareAttachment.swift
@@ -120,7 +120,7 @@ class ShareAttachment {
                 self.messages.append(msg)
                 self.delegate?.onAttachmentChanged()
                 if self.imageThumbnail == nil {
-                    self.imageThumbnail = result
+                    self.imageThumbnail = result.scaleDownImage(toMax: 96)
                     self.delegate?.onThumbnailChanged()
                 }
             }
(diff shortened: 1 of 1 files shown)


DcShare/Helper/ShareAttachment.swift
```diff
@@ -120,7 +120,7 @@ class ShareAttachment {
                 self.messages.append(msg)
                 self.delegate?.onAttachmentChanged()
                 if self.imageThumbnail == nil {
-                    self.imageThumbnail = result
+                    self.imageThumbnail = result.scaleDownImage(toMax: 96)
```
r10s, line 123: what exactly does this `96` mean? max width/height in pixel? would be nice to have some comment at the corresponding function (not really an issue of this pr).

r10s (approved):

r10s:
do you know how much memory this frees up?

i did not manage to share a panorama still.

DcShare/Helper/ShareAttachment.swift
cyBerta: I replaced the hard coded value by `self.thumbnailSize` which also helps to understand where the value comes from.

cyBerta:
> do you know how much memory this frees up?
> 
> i did not manage to share a panorama still.

It reduces the amount of mermory invovled for the 9.5mb image from ~70 mb to ~14 mb. However, the image processing in the core still consumes too much memory.

PR #1351, opened 2021-09-13
persist last selected account

### LINKED ISSUE #1342: cancelling account selection may select wrong account
- assume you have already two configured accounts, AccountA and AccountB
- select AccountB
- tap "add account" create a 3rd account
- tap "cancel"

=> now, AccountA is selected instead of the previously selected AccountB

android has a similar issue: https://github.com/deltachat/deltachat-android/issues/2006

cyBerta:
return to last selected account 
- if current account was deleted or 
- if adding a new one was cancelled

closes #1342


r10s:
i think, this pr does a bit too much by trying to track the last selected account in all situations, i would no try to do that, also as it may happen at some point, that core or notification change selection, that would make things even harder.

instead i would really focus on the one situation where the issue can happen - adding accounts. so: remember selection before adding - and using that if, and only if, creation is cancelled.

r10s (approved):

Issue #1348, opened 2021-09-08
crash on deleting a chat from the search result


r10s:
- search for something
- open a chat from the search result
- delete that chat via "Profile / Delete Chat"

=> crash

cyBerta:
The same bug appears if I delete a searched message and return to the chat list.

PR #1352, opened 2021-09-13
update search result on messages changed event

cyBerta:
fixes https://github.com/deltachat/deltachat-ios/issues/1348

### DIFF
--- a/deltachat-ios/ViewModel/ChatListViewModel.swift
+++ b/deltachat-ios/ViewModel/ChatListViewModel.swift
@@ -355,7 +355,6 @@ extension ChatListViewModel: UISearchResultsUpdating {
         } else {
             inBgSearch = true
             DispatchQueue.global(qos: .userInteractive).async { [weak self] in
-                usleep(100000)
                 self?.needsAnotherBgSearch = false
                 self?.filterContentForSearchText(self?.searchText ?? "")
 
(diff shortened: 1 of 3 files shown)


deltachat-ios/ViewModel/ChatListViewModel.swift
```diff
@@ -342,7 +355,6 @@ extension ChatListViewModel: UISearchResultsUpdating {
         } else {
             inBgSearch = true
             DispatchQueue.global(qos: .userInteractive).async { [weak self] in
-                usleep(100000)
```
r10s, line 345: not sure if you have tested that with a 3gb database :)

on larger databases, a search easily takes some 100ms - this initital delay is to let users type some more characters, resulting in overall faster and responsive results.

so, i would leave that out of this pr, i think, it is still fast enough.

r10s (changes requested):
i would leave the debouncing delays as is.

an issue, that came up during testing: when a chat is deleted from the search result, the "empty cell" that is shown for a fraction of a second (which is fine so far), got a random avatar - it would be better to just show no avatar. but we can also treat that as an another issue in case it is more than a line :)

deltachat-ios/ViewModel/ChatListViewModel.swift
cyBerta: ok, then I'll undo this change

cyBerta: Typing multiple letters in 100ms is only sth. for quite advanced users. I wonder if we should increase the initial delay up to 250 ms.  However this might be an option for another PR and can be discussed separately.

r10s (approved):

PR #1354, opened 2021-09-14
fix app badge counter taking archived messages into account 

### LINKED ISSUE #1347: stucked app-badge-counter when archiving a chat with fresh messages
- receive a message, app-badge-counter and chat-badge-counter each shows "1"
- archive the chat with the message without opening

=> chat-badge-counter is removed, however, app-badge-counter stucks at "1".

the "1" is even not removed when the app is restarted.

might be a core issue, that has to be figured out, maybe by just checking if android has the same bug (however, not all android-ui have app-badge-counter)

cyBerta:
fixes #1347

### DIFF
--- a/deltachat-ios/ViewModel/ChatListViewModel.swift
+++ b/deltachat-ios/ViewModel/ChatListViewModel.swift
@@ -184,6 +184,7 @@ class ChatListViewModel: NSObject {
         let chat = dcContext.getChat(chatId: chatId)
         let isArchivedBefore = chat.isArchived
         if !isArchivedBefore {
+            dcContext.marknoticedChat(chatId: chatId)
             NotificationManager.removeNotificationsForChat(dcContext: dcContext, chatId: chatId)
         }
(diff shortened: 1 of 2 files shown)


deltachat-ios/ViewModel/ChatListViewModel.swift
```diff
@@ -184,6 +184,7 @@ class ChatListViewModel: NSObject {
         let chat = dcContext.getChat(chatId: chatId)
         let isArchivedBefore = chat.isArchived
         if !isArchivedBefore {
+            dcContext.marknoticedChat(chatId: chatId)
             NotificationManager.removeNotificationsForChat(dcContext: dcContext, chatId: chatId)
         }
         dcContext.archiveChat(chatId: chatId, archive: !isArchivedBefore)
```
r10s, line 190: `archiveChat()` already makes things as being notices. so, the issue is more that removeNotificationsForChat() is called _before_ `archiceChat()`, i am on that anyway, i will do a commit.

r10s (approved):

PR #1356, opened 2021-09-20
run camera permission completion handler code on UI thread

### LINKED ISSUE #1333: blank-screen instead of qr-scan-activity on startup
1. delete  Delta Chat to make sure, all permissions are reset
2. install Delta Chat newly
3. select "Scan QR Code" from welcome screen
4. you will be asked for camera permission
5. select "OK"
6. instead of the "scan activity", a black screen is shown (if you go back and select "Scan QR Code"  again, things will work)

(if seems that re-installation is really needed - just retracting "Camera" permission will show an "Open Settings" button instead of "OK" - and back in Delta Chat you're already in welcome screen and things will work implicitly)

tbh, i thought we had fixed that issue at some point already, however, 1.20 from the app store has the same issue, so it is no direct regression.

cyBerta:
fixes #1333


r10s (approved):
lgtm

PR #1357, opened 2021-09-20
don't show compact (image) message view if message has html

### LINKED ISSUE #1335: "Show full message" button overlaps quote if there is no text
if there is no text but only a quote and a "Show full message" button, quote and button overlap:

the second message shows the issue - the first one was sent before https://github.com/deltachat/deltachat-core-rust/issues/2475 was fixed - which results in a missing "Show full message" button, which is even worse :)

these kind of message arrive eg. when sending a webpage via apple mail.

maybe core could also be improved in picking up parts, however, the button should be rendered correctly anyway.

cyBerta:
fixes #1335

### DIFF
--- a/deltachat-ios/Chat/Views/Cells/ImageTextCell.swift
+++ b/deltachat-ios/Chat/Views/Cells/ImageTextCell.swift
@@ -44,7 +44,7 @@ class ImageTextCell: BaseMessageCell {
 
     override func update(dcContext: DcContext, msg: DcMsg, messageStyle: UIRectCorner, showAvatar: Bool, showName: Bool) {
         messageLabel.text = msg.text
-        bottomCompactView = msg.type != DC_MSG_STICKER && msg.text?.isEmpty ?? true
+        bottomCompactView = msg.type != DC_MSG_STICKER && !msg.hasHtml && msg.text?.isEmpty ?? true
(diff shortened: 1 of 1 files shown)


deltachat-ios/Chat/Views/Cells/ImageTextCell.swift
```diff
@@ -44,7 +44,7 @@ class ImageTextCell: BaseMessageCell {
 
     override func update(dcContext: DcContext, msg: DcMsg, messageStyle: UIRectCorner, showAvatar: Bool, showName: Bool) {
         messageLabel.text = msg.text
-        bottomCompactView = msg.type != DC_MSG_STICKER && msg.text?.isEmpty ?? true
+        bottomCompactView = msg.type != DC_MSG_STICKER && !msg.hasHtml && msg.text?.isEmpty ?? true
```
r10s, line 47: in other words, the "Full View" is shown when the type is a sticker or some text is set?

i find that hard to read :)

r10s (approved):

deltachat-ios/Chat/Views/Cells/ImageTextCell.swift
cyBerta: You're right about both. I added a second commit that improves the readability.

PR #1358, opened 2021-09-20
Message input bar overlapping

### LINKED ISSUE #1349: message-input-bar overlaps bubbles, unneccessary scrolling
reported by @paulaluap: the following needs an **iphone x** or a newer device with space below input bar:

- open a chat
- open profile
- go back

-> the input is scrolled down a line or so and overlaps the bubble. i bet, that is exactly the amount of space below the input bar (that is mostly free due to the rounded corners:

EDIT: i just tested, this is repoducible eg. in the iphone12mini simulator.

also, there is just unneeded scrolling. i am wondering why we scroll at all (we might have discussed that, but i forgot :) and not just pop the profile.

this unneeded scrolling is visible on all devices: things move in from the top.

cyBerta:
fixes #1349

todos: 
- [x] fix scroll animation when returning from chat profile to the chat view.
- [x] fix landscape scrolling bug 
- [x] fix landscape message input bar overlapping messages

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -338,9 +338,15 @@ class ChatViewController: UITableViewController {
         }
         if !isDismissing {
             self.tableView.becomeFirstResponder()
+            var bottomInsets = self.messageInputBar.intrinsicContentSize.height + self.messageInputBar.keyboardHeight
+            if UIApplication.shared.statusBarOrientation.isLandscape,
+               let root = UIApplication.shared.keyWindow?.rootViewController {
+                // in landscape we need to take safeAreaInsets into account, in portrait they're already part of the keyboard height
+                bottomInsets += root.view.safeAreaInsets.bottom
+            }
             self.tableView.contentInset = UIEdgeInsets(top: self.getTopInsetHeight(),
                                                        left: 0,
-                                                       bottom: self.messageInputBar.intrinsicContentSize.height + self.messageInputBar.keyboardHeight,
+                                                       bottom: bottomInsets,
                                                        right: 0)
 
             if let msgId = self.highlightedMsg, self.messageIds.firstIndex(of: msgId) != nil {
@@ -1657,9 +1663,15 @@ extension ChatViewController: InputBarAccessoryViewDelegate {
         if isDismissing {
             return
         }
+        var bottomInsets = size.height + messageInputBar.keyboardHeight
+        if UIApplication.shared.statusBarOrientation.isLandscape,
+           let root = UIApplication.shared.keyWindow?.rootViewController {
+            // in landscape we need to take safeAreaInsets into account, in portrait they're already part of the keyboard height
+            bottomInsets += root.view.safeAreaInsets.bottom
+        }
         self.tableView.contentInset = UIEdgeInsets(top: self.getTopInsetHeight(),
                                                    left: 0,
-                                                   bottom: size.height + messageInputBar.keyboardHeight,
+                                                   bottom: bottomInsets,
                                                    right: 0)
         if isLastRowVisible() && !tableView.isDragging && !tableView.isDecelerating  && highlightedMsg == nil && !ignoreInputBarChange {
             scrollToBottom()
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -291,7 +291,11 @@ class ChatViewController: UITableViewController {
     }
 
     private func getTopInsetHeight() -> CGFloat {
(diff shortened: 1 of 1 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -358,9 +368,15 @@ class ChatViewController: UITableViewController {
...
                     }
                 }, completion: { [weak self] finished in
+                    guard let self = self else { return }
                     if finished {
-                        self?.isInitial = false
-                        self?.ignoreInputBarChange = false
+                        if !self.isInitial {
+                            // relaod messages since this VC was in background in the navigation stack,
+                            // all message observers have been inactive in the meanwhile
+                            self.loadMessages()
```
r10s, line 376: in general, i am wondering why we cannot leave leave observers active somehow and do this `loadMessages()` only when there are really changes, maybe just by a single flag.

if feels so much faster when this call is skipped!

maybe this is sth. for another pr, however, the loadMessages() call was there before as well.

r10s (approved):

r10s:
i was a bit fast with approving, there are still related issues:

the weird scrolling mentioned in #1349 is still there when a chat is empty: create a new empty group and open it from the chatlist.

also, i have the feeling, the ui is less responsive when entering a chat, however, i have to double-check that (iphone7 with 3gb database here)

cyBerta:
> i was a bit fast with approving, there are still related issues:
> 
> the weird scrolling mentioned in #1349 is still there when a chat is empty: create a new empty group and open it from the chatlist.
> 
> also, i have the feeling, the ui is less responsive when entering a chat, however, i have to double-check that (iphone7 with 3gb database here)

please retry now.

I cannot reproduce the weird scrolling bug anymore with my latest change. Also I found a way to get noticed if the view controller gets added to and removed from the navigation stack, so that we can keep the observers being registered even if we add other VC's to the stack and remove them correctly when we remove the chat view controller. We load the messages only once now in viewDidLoad.

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1781,3 +1799,16 @@ extension ChatViewController: ChatInputTextViewPasteDelegate {
...
+
+// MARK: - UINavigationControllerDelegate
+extension ChatViewController: UINavigationControllerDelegate {
+    override func willMove(toParent parent: UIViewController?) {
+        if parent == nil {
+            //logger.debug("chat observer: remove")
+            removeObservers()
+        } else {
+            //logger.debug("chat observer: setup")
+            setupObservers()
```
r10s, line 1811: so, the trick here is that the observer stays active all time time the chat is present?

but why not just do that in init()/deinit() then?

in case not every willMove(parent) is paired with a willMove(nil) we will have the too-many-observers issue again.

r10s (commented):
it works as expected, but the pr as such confuses me a bit. also because some mostly unchanged methods are moved.

what exactly fixes the scrolling issue?

maybe let's do an av on that tomorrow :)

deltachat-ios/Chat/ChatViewController.swift
cyBerta: because deinit is not necessarily called immediately. This can lead to a situation when multiple observers are still registered, so that methods are called twice

cyBerta: > so, the trick here is that the observer stays active all time time the chat is present?

present as in present in the navigation stack. willMove is always called if the VC is added or removed from the navigation stack, no matter if the VC is visible to the user or somewhere hidden because other VCs are on top of it.

cyBerta: > in case not every willMove(parent) is paired with a willMove(nil) we will have the too-many-observers issue again.

from what I understood willMove is always paired

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -382,6 +394,34 @@ class ChatViewController: UITableViewController {
...
+    override func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
+        isDismissing = false
+        ignoreInputBarChange = true
+        AppStateRestorer.shared.resetLastActiveChat()
+        handleUserVisibility(isVisible: false)
+        audioController.stopAnyOngoingPlaying()
+    }
+
+    private func setupObservers() {
```
cyBerta, line 424: I moved setting up the observers into an own method, just like it is done for the removal [here](https://github.com/deltachat/deltachat-ios/pull/1358/files#diff-bf4dd3d619f49250044e05ac5777f82c9b4029e49d8d175f6e2bbfefc1fe4685R485)

cyBerta: the diff reads more difficult indeed, but the plain code is easier to follow now

r10s (approved):

PR #1364, opened 2021-09-22
Update chat profile on dc core events

### LINKED ISSUE #1360: Contact and group chat profile not updated on messages changed
currently the UI seems not to be updated if the user has opened the chat profile and new messages/events arrived.
parts that should be updated:
- [x] files and gallery counters
- [x] shared chats
- [x] double check disappearing messages settings

cyBerta:
closes #1360

### DIFF
--- a/deltachat-ios/Controller/ContactDetailViewController.swift
+++ b/deltachat-ios/Controller/ContactDetailViewController.swift
@@ -95,6 +95,9 @@ class ContactDetailViewController: UITableViewController {
         return cell
     }()
 
+    private var incomingMsgsObserver: NSObjectProtocol?
+    private var contactChangedObserver: NSObjectProtocol?
+
     init(dcContext: DcContext, contactId: Int) {
         self.viewModel = ContactDetailViewModel(dcContext: dcContext, contactId: contactId)
         super.init(style: .grouped)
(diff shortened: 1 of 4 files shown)


deltachat-ios/Controller/GroupChatDetailViewController.swift
```diff
@@ -209,15 +213,71 @@ class GroupChatDetailViewController: UIViewController {
         tableView.reloadData() // to display updates
         editBarButtonItem.isEnabled = currentUser != nil
         updateHeader()
-        updateCellValues()
+        updateMediaCellValues()
+        updateEphemeralTimerCellValue()
+        setupObservers()
```
r10s, line 218: same here, i would add the observers before initializing ui.

deltachat-ios/Controller/ContactDetailViewController.swift
```diff
@@ -122,9 +125,15 @@ class ContactDetailViewController: UITableViewController {
         super.viewWillAppear(animated)
         updateHeader() // maybe contact name has been edited
         updateCellValues()
+        setupObservers()
```
r10s, line 128: in general, to not miss an update, the observer needs to be added beforehand - otherwise, an event may arrive between ui update and overserver adding and ui will show the wrong values.

might be a _very_ minor thing here only, however, these patterns tend to be copy+pasted, cmp. https://github.com/deltachat/deltachat-ios/pull/1338, therefore i suggest to to that right ;)

r10s (approved):
some nitpicking wrt observer-add order, please check if we can add the observers a bit before (without crashing :) 

apart from that lgtm and works like a charm!

very nice and good improvement!

PR #1373, opened 2021-09-29
group 'experimental features' in separate dialog

r10s:
this moves 'experimental features' into a separate dialog
and avoids mixing them with normal supported features
(before, normal and experimental features
were just shown beside each other without visible difference)

for the future, this also allows us to add more
experimental things without blowing up the 'advanced' dialog.

hierachy-wise, this is the same as on android now, so we can say
"enable experiment xy at 'Settings / Advanced / Experimental Features".

(that missing grouping annoys me since a long time, however i always though, we will get rid of "location streaming" experiment soon :)


r10s:
merging that in, i think, please have a quick look afterwards, @cyBerta :)

PR #1377, opened 2021-10-02
broadcasts

### LINKED ISSUE #1375: add broadcasts
- add an experimental feature "Broadcast Lists", similar to "Location Streaming" [[text suggestion]](https://github.com/deltachat/deltachat-android/pull/2060/files#diff-bcffc9d0c459e82b154f67f9fc6f68b89885deac10ebcfd7fb796bc27620ad40R113)

if that is enabled:

- add the option "New Broadcast List" (string `new_broadcast_list`) below "New Verified Group" to the "New Chat" screen

- for creation, just use the same view as for normal groups, however, hide unneeded parts (name, avatar, qr invite)

- change titles/buttons to `new_broadcast_list`,  `add_recipients` etc.

- creation is done by calling `dc_create_broadcast_list()`

- check that the new chat type DC_CHAT_TYPE_BROADCAST is used accordingly, in most cases, that should behave same as for groups, cmp. android pr linked below

- in the profile for broadcasts, also hide uneeded things (cmp. android) and use the "recipients" wording.

all strings needed are already on master, see https://github.com/deltachat/deltachat-android/pull/2060 for open questions and screenshots.

r10s:
most things are done, the edit-name dialog needs some more love and recipients cannot be deleted yet from profile.

closes #1375

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -256,6 +256,10 @@ public class DcContext {
         return Int(dc_create_group_chat(contextPointer, verified ? 1 : 0, name))
     }
 
+    public func createBroadcastList() -> Int {
+        return Int(dc_create_broadcast_list(contextPointer))
+    }
+
     public func addContactToChat(chatId: Int, contactId: Int) -> Bool {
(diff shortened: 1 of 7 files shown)


cyBerta (approved):
wow, great!

PR #1381, opened 2021-10-29
smarter delete with fewer noise on the screen

r10s:
instead of flashing up the just deleted chat in a maybe weird state, go to the chatlist directly.

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -840,7 +840,7 @@ class ChatViewController: UITableViewController {
     private func updateTitle(chat: DcChat) {
         let titleView =  ChatTitleView()
 
-        var subtitle = "ErrSubtitle"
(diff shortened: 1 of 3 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -840,7 +840,7 @@ class ChatViewController: UITableViewController {
     private func updateTitle(chat: DcChat) {
         let titleView =  ChatTitleView()
 
-        var subtitle = "ErrSubtitle"
+        var subtitle = ""
```
r10s, line 843: this `ErrSubtitle` was visible before - as the "just deleted ChatViewController" is skipped on going back now, `ErrSubtitle` should no longer be shown.

however, there is also not much in it, the chat title is blank as well - and if the chat flashes up shortly for whatever reason, there is no need to show a weird string.

PR #1382, opened 2021-10-30
Update korean translation

hellojaccc:
Updated korean translation. 

It is missing many strings compared to en.lproj, how can i fix that?

### DIFF
--- a/deltachat-ios/ko.lproj/Localizable.strings
+++ b/deltachat-ios/ko.lproj/Localizable.strings
@@ -33,7 +33,7 @@
 "never" = "안함";
 "one_moment" = "잠시만 기다려 주세요…";
 "done" = "완료";
-"undo" = "실행취소";
+"undo" = "실행 취소";
 // Translators: used eg. for the next view, could also be "continue" or so. as used in iOS headers, the string should be as short as possible, though.
 "next" = "확인";
 "error" = "오류";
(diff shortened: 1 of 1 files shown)


r10s:
great that you want to take care of korean!

we do translations via transifex, https://www.transifex.com/delta-chat/public/ , you can log in there with your github handle.

once done, please tell us your transifex name so we can assign you as a translator for korean. alternatively, you can also request that on transifex and we will get a notification then, however, even then, it is saver to ping us here :)

r10s:
nb: we are using one shared translations repo for ios/android/desktop - so, also android/desktop will benefit from the effort you will do for ios :)

hellojaccc:
Hi, i requested to join team. @hellojaccc is my transifex nickname.

r10s:
> Hi, i requested to join team. @hellojaccc is my transifex nickname.

@hellojaccc thanks a lot! i just approved your requests and also gave you the rights to change existing translations.

i think, we can close this pr then, i would be great if you can apply these few changes from this pr also on transifex. we pull all translations from transifex regulary, they will find their way to ios/android/desktop then.

hellojaccc:
Thank you i appreciate it! I'll edit strings.

r10s:
> Thank you i appreciate it! I'll edit strings. 

it is a good timing, btw, i think somewhen ~next week we do new releases, at least testflight. they will include korean updates then :)

PR #1389, opened 2021-11-12
[wip] show unread-message-badge for chat requests

r10s:
chat requests are notified since core61
and may also be added to the app-icon-unread-badge by the system therefore.

to not confuse users,
which chat is unread when tapping the app icon with the unread-badge,
we should show the unread-message-badge also for chat requests.

previously it was hidden with the reasoning
that chat requests are not notified,
https://github.com/deltachat/deltachat-ios/issues/1306
this has changed since then, by strong user requests :)
cmp https://github.com/deltachat/deltachat-core-rust/pull/2690

### DIFF
--- a/deltachat-ios/View/ContactCell.swift
+++ b/deltachat-ios/View/ContactCell.swift
@@ -233,7 +233,7 @@ class ContactCell: UITableViewCell {
     func setStatusIndicators(unreadCount: Int, status: Int, visibility: Int32, isLocationStreaming: Bool, isMuted: Bool, isContactRequest: Bool) {
         if isLargeText {
             unreadMessageCounter.setCount(unreadCount)
-            unreadMessageCounter.isHidden = unreadCount == 0 || isContactRequest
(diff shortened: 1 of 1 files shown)


r10s:
k, the above is not right. 

i came over that, when @hpk42 pointed out his badge counters on android are wrong - and i rembered there was _sth_ on ios as well.

however, we create the badge counter for the app icon using [`get_fresh_msgs()`](https://c.delta.chat/classdc__context__t.html#a6d47f15d87049f2afa60e059f705c1c5) - and that explicitly excludes chat requests - one may argue that this is not correct, however, this is another point. ios ui is correct and in sync with core _without_ this pr.

i will double-check android now :)

r10s:
> i will double-check android now :)

commenting here, although in ios repo: if the notification-dot is used, which seems to be the default, things seems to be okay there as well, however, if systems just count the existant notificartions, things may be in sync only when counting the "request" badges as "+1". however, all that seems not to be worth an issue currently.

PR #1391, opened 2021-11-16
fix auto-layout of ImageTextCell

cyBerta:
* fixes missing image messages on iOS 15


r10s (approved):
i do not have an iOS 15 emulator currently, i probably need to upgrade soon.

but i assume you have tested that and that make a difference :)

PR #1392, opened 2021-11-16
make copying videochat invitations useful

r10s:
when doing 'Copy to clipboard' for a videochat-invitation,
use the text text that includes the URL to clipboard;
this is also what android/desktop are doing.

before, only the text "Videotext Invitation" was copied,
which is quote pointless :)


cyBerta (approved):

r10s:
somehow github merge button did merge (the commits are on master now) but then github broke its ui, closing this manually therefore :)

PR #1393, opened 2021-11-16
Update Localizable.strings

snan:
Change from noun form to verb form "Arkivera"

### DIFF
--- a/deltachat-ios/sv.lproj/Localizable.strings
+++ b/deltachat-ios/sv.lproj/Localizable.strings
@@ -27,7 +27,7 @@
 "back" = "Tillbaka";
 "close" = "Stäng";
 "forward" = "Vidarebefordra";
-"archive" = "Arkiv";
+"archive" = "Arkivera";
 "unarchive" = "Avarkivera";
 "mute" = "Tyst";
 "ephemeral_messages" = "Försvinnande meddelanden";


cyBerta:
Thanks for your contribution, @snan!
Genreally we use [transifex](https://www.transifex.com/delta-chat/delta-chat-app/) for our translation process. Normally, it is more likely that changes are reviewed there by people who also speak the same language. If you have time for it, it would be great if you could add your change on transifex. Otherwise hopefully @ralphtheninja can help out here. We'll transfer your change to transifex then (otherwise it would be overwritten next time we pull translations from tansifex).

snan:
OK. I submitted a join request on that website.

r10s:
@snan great, i approved!

for the string in question: to be clear, that needs to be "to archive", not "the archive", i also [clarified that in the string sources](https://github.com/deltachat/deltachat-android/commit/a9764e89edb5f43f8369ce4f0a83f55e7b3c6de5#diff-ac392d3094c70ee4fda24e1a3a80f912c9f14d01fbf4ed7a774b8ba09c301188R34) (android/ios/desktop all use the same translations) and also pushed that to transifex.

r10s:
i just pulled the strings as we are probably doing a new release soon.

ftr: the autogenerated files with the changes are at https://github.com/deltachat/deltachat-ios/commit/9ee02e23b18de81bf94c851e6e1f0d51f9a2106e#diff-d69dc2bb9500f86057f63d7dd1c06102a50ee1655c52da641c132003537e72eeL30

snan:
> for the string in question: to be clear, that needs to be "to archive", not "the archive"

Exactly♥! I changed it from indeterminate noun "[an] archive" to imperative verb "[please, do] archive".

Weirdly enough, the matching "unarchive" was already grammatically correct ("Avarkivera"). Semantically an unusual prepositional aggluttinization compared to something like "Återställ från arkivering" (which means "Restore from having been archived") which might be too lon

PR #1395, opened 2021-11-17
add and remove observers in willMove

### LINKED ISSUE #1374: speed up chatlist
with https://github.com/deltachat/deltachat-ios/pull/1358#issuecomment-928070944 we keep observers running in the chat view even when the chat is not on top (as on android) - this has the huge advantage, that the view is kept up to date when eg. the profile is opened and the back-button or -gesture is amazingly fast.

this saves the load-db call when the chatlist comes to view (only needed once on program start)

we should do the same for the chatlist, that should give the app a real speed boost :)

as the chatlist updates are even debounced, this change should be even more safe than for the chat view (which, however, is less often and shorted covered by another view).

cyBerta:
closes #1374

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -88,10 +88,6 @@ class ChatListController: UITableViewController {
         msg.text = String.localized("update_1_24_ios") + " https://delta.chat/en/blog"
         dcContext.addDeviceMessage(label: "update_1_24a_ios", msg: msg)
 
-        // create view
-        navigationItem.titleView = titleView
-        updateTitle()
-
         viewModel.refreshData()
     }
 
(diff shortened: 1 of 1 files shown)


r10s (commented):
there is still `getChatlist()` call when getting back from chat to chatlist, so there is still the same lag (see the log when going back to the chatlist).
```
23:01:24.394 💙 INFO event: src/chat.rs:2957: device-message update_1_24a_ios already added
23:01:25.271 💙 INFO ⏰ getChatlist: 0.8767440319061279 s
```
 i think, the gist of #1374 is to remove the need of an update when going back.

r10s (approved):
that is impressively faster now!

r10s:
there is some bug with the "Chats" title now - after entering+leaving chats+profile and going back some times (playing around ;) the "Chats" title gets un-centered - every other returning moves the title some further pixels to the left, very funny :)

r10s (approved):
seems to work now :)

PR #1396, opened 2021-11-18
implement SVG QR code loading

cyBerta:
implement SVG QR code loading

comp. https://github.com/deltachat/deltachat-desktop/pull/2420, https://github.com/deltachat/deltachat-core-rust/pull/2815,
https://github.com/deltachat/deltachat-android/pull/2138 

not implemented at all yet on iOS: QR code sharing


r10s:
- on rendering the fallback-avatar the letter is still misaligned
- the font is not bold (nowhere, neither in the text nor in the avatar)

the font-weight issue is not so bad as it looks still okay and not wrong.

@cyBerta any idea about the letter misalignment?

r10s (approved):

PR #1399, opened 2021-11-22
Preview svg images

cyBerta:
just as a proove of concept for now. 

Already without that PR, if you tap on a svg attachment the image is shown correctly on a current iOS. This PR just displays them as a preview directly within the message bubble, just as it is done for pdfs.

ps: built on top of #1396


r10s:
> Already without that PR, if you tap on a svg attachment the image is shown correctly on a current iOS.

that's already fine in many cases, or?

i fear, that if we merge that in, we get issues about svg not working - and we have to deal with such errors then.

and, after all experience with svg, it is pretty expectable that this happens.

so, tbh, i would not add more dependencies (without need) to svg for now.

cyBerta:
Let's close this PR.

PR #1401, opened 2021-11-22
Background image

### LINKED ISSUE #127: background image
it should be possible to select a background image as in android.

before this can be implemented, however, the message status (time etc.) should go inside the bubble as it is otherwise hard wrt contrast and readability)

cyBerta:
closes #127 (historical!)


r10s:
wow, this is huge step forward!

i know, this is still wip, so not sure if the following comments are _that_ helpful or already in some queues - however, some early comments:

- the incomings bubbles should be white, i could not wait and figured out the corresponding line, it is: `public static let messageSecondaryColor = UIColor.themeColor(light: UIColor.init(hexString: "ebebed")` in `DcColors.swift` - just use `"ffffff"`

- date and status headlines should be adapted as well, i suggest to use the same transparency-color as on android, this is battle tested on many backgrounds

- when switching from _portrait_ to _landscape_ there is an ugly flash on my device - the other way round, things are smart

- the input bar was also not transparent before, right?

cyBerta:
There's still the issue with the flash when switching between protrait and landscape. 
It's an issue with the input bar and I would like split that work. 
Other than that, this pr is ready for review @r10s

r10s:
another test:

- [ ] the **title is missing**
- [ ] is suggest to use the **same transparency, height and font** for the "Use Default Image" and "Select From Gallery" buttons as used for the "Accept"/"Block" contact request buttons
- [ ] a once selected image cannot be "overwritten" on my device - i am failing to use another background once i had selected one

cyBerta:
> * a once selected image **cannot be changed again** on my device - i am failing to use another background once i had selected one.

@r10s Can you describe more in detail how this fails? On my device / emulator it just works to select different backgrounds.

r10s:
> @r10s Can you describe more in detail how this fails? On my device / emulator it just works to select different backgrounds.

i can select an image from gallery, it is shown in the settings dialog, however, not taken over to the chat background. some previously set picture is "stuck" there, i cannot get rid of that even when selecting "default background". when going back to the background-settings, not the recently selected background is shown but the "sticky" one.  
this happens on my "real" device, might be related to some early beta version i tried out :)

cyBerta:
> i can select an image from gallery, it is shown in the settings dialog, however, not taken over to the chat background. some previously set picture is "stuck" there, i cannot get rid of that even when selecting "default background". when going back to the background-settings, not the recently selected background is shown but the "sticky" one.
> this happens on my "real" device, might be related to some early beta version i tried out :)

Might also be that overwriting files do work differently on different iOS versions? We use always the same filename for the background image, so that we (theoretically) ensure, to delete the old one once we select a new background image.

r10s:
> It would be cool, if you could retest with the latest commit at some point.

works for me now again!

r10s:
another thing during using:

- [ ] the "chat is empty" hint does not look nice now ...  
  
... the text should be white as for the other system messages  - and as on android.

- [ ] please make the corners of the system messages a bit "less round" (about half the radius)

- [ ] please make the font for the date-headlines bold (the other system messages should stay normal)

i did not check the transparent color in details, however, at a first glance, this looks good - it is the same color as on android, right.

r10s:
i again have the issue that some image is "stuck", same as described above. happens after playing around a bit, selecting different images and leaving/restarting the app.

cyBerta:
@r10s, I think it was a caching issue. Can you please retry? 

I also checked if large images are scaled down - yes they are.

r10s:
i am getting weird crashes meanwhile when just trying to open an chat, with or without the latest commit. the crashes are not preset on current master.

i cannot make much sense of that, it always crashes at the `guard` line - maybe some memory is corrupted. i cannot reproduce that in the emulator, and i assume it is just my device, maybe releated to earlier things i did and maybe not worth following. but might be you have an idea, therefore i post :)

Screen Shot 2021-12-13 at 12 26 42

cyBerta:
The text "unable to read data", is this really a message text?

r10s (approved):

Issue #1414, opened 2021-12-08
speed up opening chat


r10s:
with recent commits, we speed up going back to the chatlist (#1395) and going back to the chat (#1358).

with these great changes, other slow things are more visible now (before, most things were slow :)

- opening a chat, even an empty (!) chat takes at least half a second on an iphone7 - there is no reason for that, esp. an empty chat should be opened instantly.  
but even for chat with thousands of messages, the bottleneck is probably not the core: we log these timings in `Wrapper.swift` at `logger?.info("⏰ getMessageIds: \(diff) s")` - that takes noticable time, however, it seems to be always a small fraction of overall timing.

i suggest to add temporary more ⏰-logs and see where really the bottleneck is - might be, we cannot do much or not without reasonable effort, however, that would be good to know.

to make it clear: opening a chat is not super-slow, the mentioned iphone7 is quite old. however, if we can get better here, that would still be nice. 

once this is done, we can file an issue to check how the contact-profile opening can be speed up (however, that is even less urgent :).

cyBerta:
example output 

13:56:16.203 💙 INFO ⏰ ChatViewController init diff: 8.094310760498047e-05 s
13:56:16.213 💙 INFO ⏰ ChatViewController loadView diff: 0.0014230012893676758 s
13:56:16.388 💙 INFO ⏰ getMessageIds: 0.14438104629516602 s
13:56:16.584 💙 INFO ⏰ ChatViewController viewDidLoad diff: 0.3716210126876831 s
13:56:16.619 💙 INFO ⏰ ChatViewController viewWillAppear diff: 0.00855398178100586 s
13:56:17.727 💙 INFO ⏰ ChatViewController viewDidAppear diff: 0.001649022102355957 s

Some results: 
Appearantly viewDidLoad takes most of the time. 1/3 - 2/5 of the time in viewDidLoad is spend by the call to core's getMessageIds.
Even though viewDidAppear finishes over a second after viewDidAppear (see timestamps), the time to run the code within viewDidAppear takes much less. The measurements always start **AFTER** the call to the corresponding super method, so that we measure only our own code.
The views already appear after viewWillAppear finished, so the delay in viewDidAppear seems not to have an visual effect.

r10s:
did you also measure for an totally empty view? maybe start with that before trying to tweak tableview.

cyBerta:
yes I did, I can post an example of that, too

cyBerta:
Some more insights:

[...]

PR #1420, opened 2021-12-09
enable speed optimisation also for debug builds

r10s:
this avoids weird surprises, makes profiling possible and allows to get a 'feeling' about what is really slow.

if there are issues in the future, we can disable it again, however, i do not have the feeling build times are slower - might be, however, debugging gets issues (this was sometimes true in c in the old days ... maybe this is no issue at all nowadays)

targets #1414


cyBerta (approved):
👍

PR #1421, opened 2021-12-11
show chat title in delete confirmation dialog

r10s:
esp. when deleting from the chatlist with a swipe,
there is some chance that one accidentally swipes the wrong chat -
and deletes the wrong chat then.


r10s:
hm, i just noticed that desktop is using `ask_delete_named_chat` meanwhile, maybe this is the better alternative as this contains the chat name directly in the question _(Are you sure you want to delete "Chat name"?)_

cyBerta (approved):

PR #1424, opened 2021-12-13
Fix drafts on notification tap

### LINKED ISSUE #1408: Clicking a notification banner removes your unsent draft (if any) in the same chat thread
I have "Persistent banners" notifications on, got two emails from the same person, clicked the first, started writing a reply, needed to look something up, clicked the other banner to get rid of it, but it refreshed the chat with her and removed my unsent draft, I had to start over writing it.

cyBerta:
In case the user has enabled persitent notifications (disabled by default), drafts of an already opened chat have been withdrawn if the user tapped on a delivered notification (delivered to the UI, a state, where we cannot cancel notifications programatically anymore) for the same chat, see #1408

This fix ensures that drafts are always saved before a chat gets removed from the UIViewController stack. It also ensures that no action is made when the user tapped on a notification for an already opened chat. 

If a user has temporary notifications enabled instead of persistent notifications, there will never be the situation, where a notification appears for an already opened chat. 

fixes #1408


r10s (approved):
lgtm

PR #1425, opened 2021-12-14
add search and mute button in ContactDetailHeader

### LINKED ISSUE #1398: add some profile buttons
the "top buttons" in the profile get a bit  full meanwhile, moving member list and other things down, and there will be probable even more in the future.

i suggest to introduce a row of small, simple buttons right of the chat name:

this is rougly also what whatsapp and telegram are doing.

for now, we can move "search" and "mute" there - these icons are pretty self-explaining and, in case of mute, also already in use.

~~(we should maybe also switch from "loudspeaker" to "bell" icon for muting notifications, that seems to be more in use meanwhile)~~ EDIT: see https://github.com/deltachat/deltachat-ios/pull/1425#issuecomment-998669837 , sorry for the noise

cyBerta:
closes #1398


r10s:
great, that looks great already!

for the new notification icon: thinking that over, it is probably better to stay with the old loudspeaker symbol for now - otherwise we have to change all other placed as well - on ios, however, at best, including android and desktop. so, that is probably a bit too much effort for now.  
i also checked: whatsapp also still uses the loudspeaker, so it is not _that_ unknown.

cyBerta:
I've adapted the PR accordingly. 
I'm not 100% sure about when to show which icon on the button. For now, we're showing the current mute state on the button, not the one that an action will lead to. Button with text however show the action, not the current state.

r10s:
> I'm not 100% sure about when to show which icon on the button. For now, we're showing the current mute state on the button

i think, this is okay.

i just checked: this is what whatsapp and telegram are doing as well, so that seems to be fine,

r10s (approved):

PR #1428, opened 2021-12-15
hack notification status into Connectivity view

### LINKED ISSUE #1412: show notification state in connectivity view
as there are still always issues with notifications not reaching Delta Chat in background, we should make the current state more visible, also as this may form the expectations of the user somehow if we say that "every 20 minutes == everything's okay" :) (we also had the situation where users were wondering about missing notifications - and at the end, it was just disabled...)

- i suggest to add a single line to the Connectivity View to the section "Incoming Message"

- for adding, i would just hack that to the current view for now, replacing that first `</ul>` by `<li>new line's content</li></ul>`

- the new line should show the red/yellow/green icons and a single line of text, preferably a few words only, eg.
    - 🟢 **Notifications:** Working, every 20 minutes
    - 🟡 **Notifications:** Delayed, every 45 minutes
    - 🟡 **Notifications:** Disabled in the system settings
    - 🟡 **Notifications:** Disabled in the Delta Chat settings
    - 🔴 **Notifications:** Not working, only 1 times a day, last at 12:34
    - 🔴 **Notifications:** Could not get server token  

- i suggest to use a sliding window of 24 or 48 hours to calculate the frequency; technically, i'd save that in the user-settings as a single line of text: on receiving a wakeup from the system, add the date of the notification and throw away dates that are too old.

- i would not differ between local-receive and remote wakeup in the calculations, this is hard to explain and does not really make a difference to the user.

nb: we already track some notification state, however, that is hidden at "Settings / Advanced / View Log / Custom Variables" - and way too complicated:

however, for debugging, that is still quite useful and i would not touch collecting these variables for now. they may also be useful to check if the new "notification state" is reasonable.

cyBerta:
closes #1412

### DIFF
--- a/deltachat-ios/Controller/ConnectivityViewController.swift
+++ b/deltachat-ios/Controller/ConnectivityViewController.swift
@@ -44,7 +44,6 @@ class ConnectivityViewController: WebViewViewController {
     // this method needs to be run from a background thread
     private func getNotificationStatus() -> String {
         let title = " <b>" + String.localized("pref_notifications") + ":</b> "
-        let timestamps = UserDefaults.standard.array(forKey: Constants.Keys.notificationTimestamps) as? [Double]
         let notificationsEnabledInDC = !UserDefaults.standard.bool(forKey: "notifications_disabled")
         var notificationsEnabledInSystem = false
         let semaphore = DispatchSemaphore(value: 0)
@@ -72,6 +71,7 @@ class ConnectivityViewController: WebViewViewController {
                 .appending(String.localized("notifications_disabled"))
         }
 
+        let timestamps = UserDefaults.standard.array(forKey: Constants.Keys.notificationTimestamps) as? [Double]
         guard let timestamps = timestamps else {
             return "<span class=\"yellow dot\"></span>"
                 .appending(title)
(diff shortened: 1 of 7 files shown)


r10s:
@cyBerta i rebased this pr and will probably push some more commits and refinements - i assume, you do not have things pending?

cyBerta:
No, nothing pending

deltachat-ios/Controller/ConnectivityViewController.swift
```diff
@@ -41,10 +41,88 @@ class ConnectivityViewController: WebViewViewController {
...
+            return "<span class=\"yellow dot\"></span>"
+                .appending(title)
+                .appending(String.localized("no_data"))
+        }
+
+        if timestamps.isEmpty || timestamps.count == 1 {
+            // FIXME: for timestamp == 1, that is just okay if the timestamp is not too old
+            return "<span class=\"red dot\"></span>"
+                .appending(title)
+                .appending(String.localized("notifications_not_working"))
```
r10s, line 85: this has the issue, that if there is the first wakeup coming, this results in "no working" display.

with the second one, things are fine.

i think, maybe we should just calculate the average by `(now - oldest) / count` - that would require only one data point to work.

r10s:
there are some open questions wrt layout and workding, however, it is good to get that in now, to start collecting stats.

r10s (approved):

Issue #1011, opened 2020-11-24
content sometimes hidden behind input-bar


r10s:
in the chat-view, the chat-content is sometimes hidden behind input-bar.

to reproduce, do the "back" gesture, swipe from left to right and stop and cancel at 1/4th.

<img width=320 src=https://user-images.githubusercontent.com/9800740/100149472-85d56f80-2e9e-11eb-9723-13d532053380.png>

dignifiedquire:
still happening on current testflight

PR #1435, opened 2021-12-30
Inputbar accessory view refactoring

### LINKED ISSUE #1413: Weird animation when switching between protrait and landscape
when switching between protrait and landscape 
* in the chat view
* on phone layout
* if the keyboard is not visible in portrait: 

- the accessibityInputBar takes all available space for a part of a second and resizes back to the intended size.

another, somehow related bug:
in chat view, on phone layout, if keyboard is visible in portrait, the input  bar takes the maximum allowed size (about 8 lines) and you switch to landscape and back to portait, the height the text field is wrong (1 line instead of about 8).

cyBerta:
fixes #1413

ok, it does a little bit more.
we now integrate inputBarAccessoryView's source code files directly instead of depending on upstreams pod. This way we can get rid of some subclassing hacks in order to tweak the input bar. One example tweak is the inputbar shown in landscape mode on iPhones: it should occupy all visible space. As described in #1413 the implementation was buggy though. 

Moreover, DeltaChat's ChatViewController doesn't manually calculate top and bottom insets anymore to correctly position the message bubbles when the keyboard appears or disappears. This was done before to fix https://github.com/deltachat/deltachat-ios/issues/1011 but increased the complexity.

This refactoring also fixes chat message scrolling glitches, if the text input bar increases because of growing text input. 

the source code is forked from 5.1.0, which is why we effectivly updated InputBarAccessoryView from 4.3.1


r10s:
switching over to input-only mode is now done even when the input field is not focused, landscape now _always_ looks like that:

IMG_3448

before, this input-only mode was only used when the input field is focused.

i find that unexpected, none of the other messengers i tried behave like that. i'd prefer the old view that contains all controls as usual. iirc we also switched away from these views on android (esp. as screens on phones get larger and larger)

so, wondering, if we need that special "input-only" view at all.

cyBerta:
> landscape now always looks like that

that shouldn't be the case with the last commit. What device did you use to test? maybe the condition in 0f721e44f60602a833ba842a318c80249c347c41 is for some reason not fulfilled.

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -307,6 +309,22 @@ class ChatViewController: UITableViewController {
         navigationItem.backButtonTitle = String.localized("chat")
         definesPresentationContext = true
 
+        // Binding to the tableView will enabled interactive dismissal
+        keyboardManager.bind(to: tableView)
+
+        // Add some extra handling to manage content inset
```
cyBerta, line 315: This comment is wrong and should be removed

cyBerta:
> iirc we also switched away from these views on android (esp. as screens on phones get larger and larger)

I created a  bug ticket for android: https://github.com/deltachat/deltachat-android/issues/2195 
It seems that we need at least to pay some attention to smaller screens in landscape to avoid these kind of issues.

r10s:
k, let's merge that in and fix remaining issues in new pr.

r10s (approved):

PR #1436, opened 2022-01-02
add zh-Hans

dawei101:


cyBerta:
Thank you so much for your contribution. Our tranlations process starts normally on [transifex](https://www.transifex.com/delta-chat/delta-chat-app/translate/). We maintain one source for android, desktop, and iOS and use scripts to convert the source files for each platform. if we pull translations from transifex, any changes that are not added on that translation platform will be lost. So, do you mind to copy the strings to transifex?
That would be a great help for us!

dawei101:
> Thank you so much for your contribution. Our tranlations process starts normally on [transifex](https://www.transifex.com/delta-chat/delta-chat-app/translate/). We maintain one source for android, desktop, and iOS and use scripts to convert the source files for each platform. if we pull translations from transifex, any changes that are not added on that translation platform will be lost. So, do you mind to copy the strings to transifex? That would be a great help for us!

Ok, that's so nice. Does it need me to operate it , or you can copy it?
Thanks.

r10s:
what is `zh-Hans` on transifex then? just `zh`?

currently, we have:
- `zh_CN` that maps to `zh-Hant` for ios, english name "Chinese (China)", local name 简体中文
- `zh_TW` that maps to `zh-Hant-TW` for ios, english name "Chinese (Taiwan)", local name 繁體中文

we want to add
- `?1` that maps to `zh-Hant` for ios, english name `?2`, local name `?3`

@dawei101 if you can tell `?1`, `?2` and `?3`, i can at least create the language on transifex and maybe also see if i can copy the data there. for future additions, however, it would be better, if that directly happens on transifex. if you tell me your transifex handle (you can log in with github), i will add you to `zh-Hans` then.

r10s:
fixed a typo ...

r10s:
> what is zh-Hans on transifex then? just zh?

probably not, 
according to https://www.localeplanet.com/icu/zh-Hans/index.html ,  `zh` is not unique. but transifex has `zh-Hans` aka "Chinese Simplified", that might be correct ...

dawei101:
zh-Hans, the 's' means Simplified. zh-Hans is used in China mainland. 

zh-Hant's 's' means Traditional.  Taiwan and Hongkong use zh-Hant. 

zh_CN should map to zh-Hans
zh_TW should map to zh-Hant-Tw with high prority, then zh-Hant.

Does this make sence?

r10s:
superseded by https://github.com/deltachat/deltachat-ios/pull/1493

PR #1438, opened 2022-01-10
encrypted account databases

cyBerta:
iOS counterpart of https://github.com/deltachat/deltachat-android/pull/2186


r10s:
the following things popped up during testing:
 
- import&encrypt does not work, instead i got the following error:   
    
  not sure if that is related to core or to ios implementation. i was using the current core-master and this pr rebased to ios-master.
  to reproduce, in the simulator: on a fresh installation:  
  login unencrypted -> export backup -> delete account -> new account -> login encrypted -> import backup  

- the same happend when logging in encrypted at the first place.

- when logging in both times unencrypted, export/import roundtrip works

- the "three dots" menu item is "android style", we should use the "ios style" here, so an horizontal ellipsis in a circle, cmp. eg. the notes app:  
  IMG_3449  
of course in our blue :) (and we should use that also for the "sync" button in the html-view, but that is another issue :)

cyBerta:
@r10s you should try this pr with that PR with https://github.com/deltachat/deltachat-core-rust/pull/2980

r10s:
> @r10s you should try this pr with that PR: deltachat/deltachat-core-rust#2980

great! with that PR it works!

cyBerta:
@r10s actually deltachat/deltachat-core-rust#2980 should be merged first, before we merge this one

DcShare/DcShare.entitlements
```diff
@@ -6,5 +6,9 @@
 	<array>
 		<string>group.chat.delta.ios</string>
 	</array>
+	<key>keychain-access-groups</key>
+	<array>
+		<string>$(AppIdentifierPrefix)group.chat.delta.ios</string>
```
r10s, line 11: is `$(AppIdentifierPrefix)` equal to `8Y86453UA8` from above?

deltachat-ios/AppDelegate.swift
```diff
@@ -61,9 +61,35 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         dcAccounts.logger = DcLogger()
         dcAccounts.openDatabase()
         migrateToDcAccounts()
+
+        if let sharedUserDefaults = UserDefaults.shared, !sharedUserDefaults.bool(forKey: UserDefaults.hasSavedKeyToKeychain) {
```
r10s, line 65: so, `SecItem`s survives reinstall while `UserDefaults` does not? maybe add a more explicit note for that.

DcCore/DcCore/Helper/KeychainManager.swift
```diff
@@ -0,0 +1,83 @@
...
+public class KeychainManager {
+    private typealias KcM = KeychainManager
+    private static let teamId = "8Y86453UA8"
+    private static let sharedKeychainGroup = "\(KcM.teamId).group.chat.delta.ios"
+
+    public static func getAccountSecret(accountID: Int) throws -> String {
+        do {
+            return try queryAccountSecret(id: accountID)
+        } catch KeychainError.noPassword {
+            return try addAccountSecret(id: accountID)
```
r10s, line 18: this looks a bit dangerous to me:

if `queryAccountSecret()` fails temporary for some reasons (so eg. errSecSuccess), the existing secret is overwritten and the database will get unusable.

so, we should differ carefully between "Ok, but nothing found and "Error", maybe we should change the return value for `queryAccountSecret()` for that.

DcCore/DcCore/Helper/KeychainManager.swift
```diff
@@ -0,0 +1,83 @@
...
+        let query: [String: Any] = [kSecClass as String: kSecClassGenericPassword,
+                                    kSecAttrAccount as String: "\(id)",
+                                    kSecMatchLimit as String: kSecMatchLimitOne,
+                                    kSecAttrAccessGroup as String: KcM.sharedKeychainGroup as AnyObject,
+                                    kSecReturnAttributes as String: true,
+                                    kSecReturnData as String: true]
+        var item: CFTypeRef?
+        let status = SecItemCopyMatching(query as CFDictionary, &item)
+        guard status != errSecItemNotFound else {
+            throw KeychainError.noPassword
```
r10s, line 67: maybe we should return `null` if the account cannot be found instead of returning an error. that would make calling this function easier, see above.

deltachat-ios/AppDelegate.swift
```diff
@@ -61,9 +61,35 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
+                logger.warning("Failed to delete DB secrets")
+            }
+        }
+
+        let accountIds = dcAccounts.getAll()
+        for accountId in accountIds {
+            let dcContext = dcAccounts.get(id: accountId)
+            if !dcContext.isOpen() {
+                do {
+                    let secret = try KeychainManager.getAccountSecret(accountID: accountId)
```
r10s, line 77: `getAccountSecret()` creates/overwrites the secret if it is not there - but can that happen? maybe `queryAccountSecret()` is the safer choice here.

alternatively, it is sufficient to make `getAccountSecret()` safer, see other comment.

deltachat-ios/Controller/WelcomeViewController.swift
```diff
@@ -17,14 +17,7 @@ class WelcomeViewController: UIViewController, ProgressAlertHandler {
...
             guard let self = self else { return }
-            let accountSetupController = AccountSetupController(dcAccounts: self.dcAccounts, editView: false)
-            accountSetupController.onLoginSuccess = {
-                [weak self] in
-                if let appDelegate = UIApplication.shared.delegate as? AppDelegate {
-                    appDelegate.reloadDcContext()
-                }
-            }
-            self.navigationController?.pushViewController(accountSetupController, animated: true)
+            self.showAccountSetupController()
```
r10s, line 20: nice cleanup!

deltachat-ios/Controller/WelcomeViewController.swift
```diff
@@ -128,6 +135,7 @@ class WelcomeViewController: UIViewController, ProgressAlertHandler {
         if dcAccounts.getSelected().isConfigured() {
             UserDefaults.standard.setValue(dcAccounts.getSelected().id, forKey: Constants.Keys.lastSelectedAccountKey)
 
+            //FIXME: what do we want to do with QR-Code created accounts? For now: adding an unencrypted account
```
r10s, line 138: it is not unlikely, that, if things work out, encrypted accounts just get the default - then also QR-Code created accounts will get encrypted.

r10s (commented):
a handful comments, but overall, this already lgtm :)

DcShare/DcShare.entitlements
cyBerta: yes

deltachat-ios/AppDelegate.swift
cyBerta: that is the case, I'll add an comment

DcCore/DcCore/Helper/KeychainManager.swift
cyBerta: we only call `addAccountSecret()` if there's no password set at all. For all other possible error cases, the error is passed to the calling method. So no existing password will be ever overwritten.

DcCore/DcCore/Helper/KeychainManager.swift
cyBerta: Generally, the idea was taken from: https://developer.apple.com/documentation/security/keychain_services/keychain_items/adding_a_password_to_the_keychain

We can replace the `.noPassword` error, but then we'll have to handle optionals everywhere we're calling this method instead, in order to distinct between the two general error cases (no password and every other possible error).

deltachat-ios/AppDelegate.swift
cyBerta: same here: getAccountSecret will never overwrite existing passwords

DcCore/DcCore/Helper/KeychainManager.swift
```diff
@@ -0,0 +1,83 @@
...
+import Security
+
+public enum KeychainError: Error {
+    case noPassword
+    case unhandledError(message: String, status: OSStatus)
+}
+
+public class KeychainManager {
+    private typealias KcM = KeychainManager
+    private static let teamId = "8Y86453UA8"
```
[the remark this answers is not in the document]
cyBerta: We can only access the AppIdentifierPrefix if we manually add an key value pair in Info.plist

`<key>AppIdentifierPrefix</key>
<string>$(AppIdentifierPrefix)</string>` 

During build time the teamId would be written into the plist entry and on runtime we could access the value by calling 
`let appIdentifierPrefix =
    Bundle.main.infoDictionary!["AppIdentifierPrefix"] as! String
`

https://stackoverflow.com/questions/11726672/access-app-identifier-prefix-programmatically/28714850

Thumbs up, if you prefer that. Otherwise I would add the comment instead.

DcCore/DcCore/Helper/KeychainManager.swift
cyBerta: Also here, thumbs up, if you still prefer returning `null`.

DcCore/DcCore/Helper/KeychainManager.swift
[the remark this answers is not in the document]
r10s: we can leave it as it is now, now problem. but maybe add a comment, just similar to what you've just explained, maybe including the link.

DcCore/DcCore/Helper/KeychainManager.swift
r10s: i see, i've overseen that bit. thanks for explanation!

DcCore/DcCore/Helper/KeychainManager.swift
r10s: i've overseen the explicit catch above, so, we can leave this as is, it is fine.

r10s (approved):
thanks for the additional explanations.

as core is also already merged, when the discussed comments wrt AppIdentifierPrefix/teamId are in, this lgtm and can be merged :)

r10s (approved):

PR #1440, opened 2022-01-19
webxdc

### LINKED ISSUE #1439: show webxdc messages
counterpart of https://github.com/deltachat/deltachat-android/pull/2174

* [x] create basic webxdc view, possibly derived from existing WebViewViewController, open that with `/index.html`when tapping on a webxdc message
* [x] intercept requests from webview and get files only via `dc_msg_get_webxdc_blob()`. add a special behaviour for `webxdc.js` that should implement the api as in https://github.com/deltachat/deltachat-android/blob/master/res/raw/webxdc.js (no need to be a separate file, we can use whatever is easy here)
* [x] use core for sending and receiving updates when the corresponding methods are called in `webxdc.js`
* [x] nicer in-chat-presentation with icon and app name - here, wait until that is settled on android
* [x] show app name in title bar

optional (might get into a separate PR)
* allow configuration from staging area (cmp. AttachmentManager, that requires a draft to be set much sooner)

cyBerta:
closes #1439

### DIFF
--- a/deltachat-ios/Chat/Views/WebxdcPreview.swift
+++ b/deltachat-ios/Chat/Views/WebxdcPreview.swift
@@ -3,12 +3,12 @@ import DcCore
 
 public class WebxdcPreview: UIView {
     
-    private var imageWidthConstraint: NSLayoutConstraint?
-    private var imageHeightConstraint: NSLayoutConstraint?
-    
     lazy var imagePreview: UIImageView = {
         let view = UIImageView()
+        view.contentMode = .left
         view.translatesAutoresizingMaskIntoConstraints = false
+        view.layer.cornerRadius = 8
+        view.clipsToBounds = true
         return view
     }()
     
@@ -34,6 +34,7 @@ public class WebxdcPreview: UIView {
     lazy var containerStackView: UIStackView = {
         let view = UIStackView(arrangedSubviews: [imagePreview, titleView, subtitleView])
         view.axis = .vertical
+        view.alignment = .leading
         view.spacing = 6
         view.translatesAutoresizingMaskIntoConstraints = false
         return view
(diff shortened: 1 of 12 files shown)


deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -0,0 +1,263 @@
...
+          // we're passing the updates directly to this js function
+          window.__webxdcUpdate = (updateString) => {
+            var updates = JSON.parse(updateString);
+            if (updates.length === 1) {
+              update_listener(updates[0]);
+            }
+          };
+
+          return {
+            selfAddr: "\(dcContext.addr ?? "unknown")",
```
Simon-Laux, line 48: please escape user input, for desktop I used base64 encoding/decoding

deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -0,0 +1,263 @@
...
+
+            selfName: "\(dcContext.displayname ?? "unknown")",
```
Simon-Laux, line 50: same here, please escape the user input, though here it is even more important then on the email

deltachat-ios/Chat/Views/WebxdcPreview.swift
```diff
@@ -0,0 +1,85 @@
...
+    
+    private func setupSubviews() {
+        addSubview(containerStackView)
+        containerStackView.fillSuperview()
+    }
+    
+    public func configure(message: DcMsg) {
+        imagePreview.image = message.getWebxdcIcon()?.sd_resizedImage(with: CGSize(width: 175, height: 175), scaleMode: .aspectFill)
+        titleView.text = message.getWebxdcName()
+        subtitleView.text = message.getWebxdcSummary()
```
r10s, line 65: this is pretty inefficient as each line results in db-access and unzipping.

i'd just call `info = message.getWebxdcInfoDict()` once and check for `info["name"]`, `info["icon"]`  etc. then.

at least `getWebxdcIcon()` and `getWebxdcSummary()` can be deleted then, probably also`getWebxdcName()` if we add and additional line to  `WebxdcViewController.viewDidLoad()`.

that would be 3x faster and less code already now - and in the future, there may be even more things to check. having an explicit getter seem to be too much overhead unless the thing is really used often, and not only one or two times.

r10s (commented):
Triska and Othello show some weird text instead or working for me, poll app works for me, not sure what exactly is going on there, maybe there is still sth. wrong with getting blobs.

there are also some tapping features missing, but to keep review simple, that should be done when this pr is merged; i filed an issue for that at https://github.com/deltachat/deltachat-ios/issues/1466

apart from that, lgtm.

deltachat-ios/Chat/Views/WebxdcPreview.swift
cyBerta: the efficiency gain is probably neglectable, plus the code will  be less readable. I revert the getters anyways.

r10s (approved):

Issue #1446, opened 2022-01-25
fix remainging issues for notification stats


r10s:
- wrong error for timstamps.count == 1
- improve wordings, pattern "Notifications: [Delayed,|Not working,] Last check: 12:34, In Average every 12 Minutes"
cmp https://github.com/deltachat/deltachat-ios/pull/1428

PR #1447, opened 2022-01-25
fix notification stats

r10s:
targets https://github.com/deltachat/deltachat-ios/issues/1446

### DIFF
--- a/deltachat-ios/Controller/ConnectivityViewController.swift
+++ b/deltachat-ios/Controller/ConnectivityViewController.swift
@@ -80,9 +80,11 @@ class ConnectivityViewController: WebViewViewController {
 
         var averageDelta: Double = 0
         if timestamps.isEmpty {
+            // this should not happen:
+            // the array should not be empty as old notifications are only removed if a new one is added
             return "<span class=\"red dot\"></span>"
                 .appending(title)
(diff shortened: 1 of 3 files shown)


deltachat-ios/Controller/ConnectivityViewController.swift
```diff
@@ -78,20 +78,19 @@ class ConnectivityViewController: WebViewViewController {
...
-        if timestamps.isEmpty || timestamps.count == 1 {
-            // FIXME: for timestamp == 1, that is just okay if the timestamp is not too old
+        var averageDelta: Double = 0
+        if timestamps.isEmpty {
+            // this should not happen:
+            // the array should not be empty as old notifications are only removed if a new one is added
             return "<span class=\"red dot\"></span>"
                 .appending(title)
-                .appending(String.localized("notifications_not_working"))
+                .appending("Bad Data")
```
cyBerta, line 87: shall we translate that?

cyBerta (approved):

deltachat-ios/Controller/ConnectivityViewController.swift
r10s: it is an error message for a case that should not appear at all, i do not think, it make much sense to have that translated.

Issue #1442, opened 2022-01-20
"pending" icon gone


r10s:
the "pending" icon that is usually before a message gets its "tick" is gone - instead, only an empty space is shown. iirc, there was an hourglass before.

moreover, this space does not have the dimensions of the following "tick", which leads to additional flickering.

for testing, eg. send a message while in flight mode.

<img width=320 src=https://user-images.githubusercontent.com/9800740/150345434-535c0b45-57fd-422d-9ad1-149e8a52c33e.PNG>

r10s:
closed by https://github.com/deltachat/deltachat-ios/pull/1450

PR #1450, opened 2022-01-25
fix invisible hour glass for pending messages

cyBerta:
uses now the tint color (like sent and received icon) for hour glass, similar to android's color scheme for pending messages


r10s (approved):
lgtm, did not test yet - wondering why the coreDark05 was not visible. EDIT: we could clarify that in av, it _was_ visible if you have good eyes. the issue came in when we revamped some colors wrt the new wallpaper.

nb: can you please add a `closes #1442` etc.  :)

PR #1451, opened 2022-01-25
fix chat in phone landscape mode

### LINKED ISSUE #1441: landscape mode unusable
when switching from portrait to landscape:
- even if portrait is only one line in the input field, in landscape it will grow to max. 4 lines
- after some seconds (see end of video), the input fields switches to "input only mode"

video:

https://user-images.githubusercontent.com/9800740/150339682-342a0eb1-be96-4400-8859-3bb1491f41c9.MOV

i have the feeling, we're doing too much, not to less :D

cyBerta:
remove buggy auto-upscaling in phone landscape mode if keyboard is shown

fixes #1441


r10s (approved):
i think, this is prefectly fine :)

i tested on iphone se1 and on iphone7 - in landscape, you'll see as much content as possible, including quotes, however, when more lines appear, they will move content and quote upwards, out of sight. at the end, resulting in an "input-only-view" as before - but with less code and less bugs :)

PR #1465, opened 2022-02-06
make it compile on macos 12

Simon-Laux:
for better understanding(the git diff is not easy to read for this change), I just added this to the script:
```
# the same on macOS 12, TODO: is there a less duplicate way to do this check?
if [[ "$os_ver" == 12.* ]]; then
    export LIBRARY_PATH="/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib"
fi
```﻿


cyBerta:
still not optimal, but you can improve a little bit by writing:

`[[ "$os_ver" == 11.* || "$os_ver" == 12.* ]];`

r10s:
maybe we can meanwhile just skip the os-version-check and say `export LIBRARY_PATH="/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib"` unconditionally.

that way we would not have the same issue in a year again :)

all contributing devs are at least on BigSur resp. macOS 11 - and we can say, that this is the minimum requirement (which probably really is as macOS 10 does not support recent Xcode and with older Xcode you cannot build on recent iOS)

r10s (approved):

Issue #1470, opened 2022-02-08
disable pinch-to-zoom, drag-select-text for webxdc


r10s:
pinch to zoom is more annoying than helpful in most games and apps, it should just be disabled for now; also android has it disabled.

doing a quick search, i found https://stackoverflow.com/questions/36581166/programatically-enable-disable-pinch-to-zoom-on-wkwebview-and-viewforzoominginsc , not totally sure if that's helpful.

moreover, playing eg. triska, one easily does a drag-select-text, resulting in the whole app getting a blue overlay ... also that should be checked, if it can be _easily_ disabled.

r10s:
for the text part, there seems to be `textInteractionEnabled` (came over that when checking https://github.com/deltachat/deltachat-ios/pull/1477 and found https://nemecek.be/blog/111/wkwebview-improvements-in-ios-15 )

PR #1472, opened 2022-02-09
disable zoom in wkwebview

cyBerta:
it seems only to be reliable possible to disable zoom by adding javascript as it is done in this PR. 

However I'm not sure if this is the way to go or if we should instead add this example code to webxdc developer documentation, so that it's up to the webxdc to decide if zooming is allowed or not. 

relates to #1470


r10s:
thanks for figuring that out.

however, i am a bit skeptical about messing around with tags that are usually controlled by the app. we potentially overwrite other settings by that, or?  in case there is no other solution, i'd postpone that a bit as it also may cause other things not working (that are less obvious then)

Simon-Laux:
+1 for documenting scroll behavior on https://github.com/deltachat/webxdc_docs instead and also we could add this to our examples

r10s:
i created a pr to document viewport behavior at https://github.com/deltachat/deltachat-core-rust/pull/3214 and suggest to close this pr.

PR #1475, opened 2022-02-09
fix crashes on creating blobs

### LINKED ISSUE #1473: crash when opening "Stack up"
there is a reproducible crash when opening "Stack up"; i will have a look at that.

### LINKED ISSUE #1474: weird encoding of "Othello" webxdc
for reference, i think, i have already a fix.

r10s:
do not try to create a Data object by UnsafeBufferPointer that point to already unref'd data.

otherwise, we had crashes or getting already overwritten data.

that may have been also the reason for other webxdc not working as expected.

EDIT: more concrete: the weird Triska background color (white/black padding mixed) is fixed and the older Triska that did not work are working with this pr now :)

closes #1473 
closes #1474

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -1107,14 +1107,17 @@ public class DcMsg {
         defer {
             ptrSize.deallocate()
         }
+
         guard let ccharPtr = dc_msg_get_webxdc_blob(messagePointer, filename, ptrSize) else {
             return Data()
+        }
+        defer {
+            dc_str_unref(ccharPtr)
         }
 
         let count = ptrSize.pointee
         let buffer = UnsafeBufferPointer<Int8>(start: ccharPtr, count: count)
         let data = Data(buffer: buffer)
-        dc_str_unref(ccharPtr)
         return data
     }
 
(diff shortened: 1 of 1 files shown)


DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -1113,8 +1113,9 @@ public class DcMsg {
 
         let count = ptrSize.pointee
         let buffer = UnsafeBufferPointer<Int8>(start: ccharPtr, count: count)
+        let data = Data(buffer: buffer)
         dc_str_unref(ccharPtr)
```
cyBerta, line 1117: we should probably use the keyworkd defer here instead. I think generally that would be wise to introduce when unref'ing pointers.

r10s: hm, if we introduce a `defer` block after setting `ccharPtr` that opens some questions to me - is it guaranteed that that block is not exectuted before? also, we would have two defer blocks, is that an issue?

i am not that used to `defer`, so i've just chosen the thing i know of :)

but if we can answer the question, we can change the code, np - otoh - that kind of code is very rarely touched, so i would also be fine without `defer`.

r10s: multiple defers are executed in reverse order: https://docs.swift.org/swift-book/ReferenceManual/Statements.html , so that could be answered.

if we now can know that ccharPtr is nil and not garbage in case it is allocated, most questions would be answerd. but still, not sure that function is worth the detailing.

r10s: i pushed a thing. the `nil`-or-not-`nil` question only pops up when allocation of the ptrSize fails - that's probably pretty unlikely - and i would expect `ccharPtr` to be nil in this case and that's file to be passed to `dc_str_unref()`

cyBerta (approved):

PR #1477, opened 2022-02-09
stop media playback if wbxdc is closed

cyBerta:
there is an easy to use API for iOS15+ so that the issue of endless playback can be at least partially fixed without much effort.


r10s (approved):
nice! maybe that is even "good enough" for now to close the issue.

at a first glance,  targeting older os seems to r[equire javascript injection and figuring out the concrete objects and pause them](https://nemecek.be/blog/111/wkwebview-improvements-in-ios-15), not sure if that is even doable in javascript code we do not know about.

nb: there is also a `setAllMediaPlaybackSuspended()`  - but that seems to be more to avoid future playbacks, i did not fully understand that method ...

PR #1489, opened 2022-02-17
enable local storage for webxdcs

cyBerta:
compare https://github.com/deltachat/deltachat-android/pull/2221

### DIFF
--- a/deltachat-ios/Controller/WebxdcViewController.swift
+++ b/deltachat-ios/Controller/WebxdcViewController.swift
@@ -116,6 +116,8 @@ class WebxdcViewController: WebViewViewController {
         config.mediaTypesRequiringUserActionForPlayback = []
         config.allowsInlineMediaPlayback = true
 
+        config.websiteDataStore = WKWebsiteDataStore.default()
+
         if #available(iOS 13.0, *) {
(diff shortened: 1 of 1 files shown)


cyBerta:
@r10s The local storage test webxdc proves that local storage seems to work as expected (scoped to messageId)

deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -116,6 +116,8 @@ class WebxdcViewController: WebViewViewController {
         config.mediaTypesRequiringUserActionForPlayback = []
         config.allowsInlineMediaPlayback = true
 
+        config.websiteDataStore = WKWebsiteDataStore.default()
```
r10s, line 119: i think, this is not even needed, [websiteDataStore documentation](https://developer.apple.com/documentation/webkit/wkwebviewconfiguration/1395661-websitedatastore) says:

> If you don’t assign a value to this property, the configuration object uses the default data store object to store data persistently

i also tested with the test_storage.xdc, it works with and without this pr.

so not sure if we should make that explicit.

cyBerta: Hm, ok, then let's close this PR and come back to it, if issues appear

PR #1491, opened 2022-02-20
Webxdc drafts

### LINKED ISSUE #1481: webxdc drafts
The UI for drafted webxdc's should be adapted. Right now the preview image is missing for example. 

Moreover a tap on the draft should open the WebxdcView so that the user gets the opportunity to preconfigure the webxdc before sending it. (compare Android)

cyBerta:
closes #1481


r10s (approved):
lgtm, and also works as expected after some tests :)

PR #1501, opened 2022-03-19
log file size of backup import attempt

r10s:
due to apple removed the progress bar (WTF!)
on copying files from Finder to iPhone,
it may easily happen, a backup is incomplete.

this will at least give a hint.

cmp. https://developer.apple.com/forums/thread/130051


DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -709,11 +709,6 @@ public class DcContext {
         get { return getConfigInt("show_emails") }
         set { setConfigInt("show_emails", newValue) }
     }
-
-    // do not use. use DcContext::isConfigured() instead
-    public var configured: Bool {
```
r10s, line 714: in general, btw, i meanwhile see these functions that look like simple variables as harmful:

they look too cheap, the caller will not optimize the code around them and does not know that each call results in a database query, maybe even a complicated one.

PR #1514, opened 2022-03-25
reduce number of initial calls to DcContext getChatlist()

### LINKED ISSUE #1499: remove superfluous call to dc_get_chatlist()
on app start, there are two calls to `dc_get_chatlist()`:

- one in `ChatListViewModel::init()` -> `ChatListViewModel::updateChatList()` [1]

- one in `ChatListController::viewDidLoad()` -> `ChatListViewModel::refreshData()` -> `ChatListViewModel::updateChatList()` [1]

as this seems to significantly slow down startup, we treat that as a bug.

nb: these two calls are blocking the main thread, subsequent updates are scheduled to background using  
`ChatListController::refreshInBg()` -> `ChatListViewModel::refreshData()` ->   `ChatListViewModel::updateChatList()` [1] -
maybe we can also do the initialisation non-blocking. but at least one call should go away.

---

[1] `ChatListViewModel::updateChatlist()` then calls    `DcContext::getChatlist()` -> `dc_get_chatlist()` in a blocking way

cyBerta:
closes #1499 

testing on an emulator, the speed gain is marginal thogh.

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -73,7 +73,7 @@ class ChatListController: UITableViewController {
                     self.searchController.searchResultsUpdater = self.viewModel
                     self.searchController.searchBar.delegate = self
                 }
-                self.tableView.reloadData()
+                self.handleChatListUpdate()
             }
         }
     }
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -428,7 +428,7 @@ class ChatListController: UITableViewController {
     private func updateTitle() {
         if RelayHelper.sharedInstance.isForwarding() {
             titleView.text = String.localized("forward_to")
-            if isArchive {
+            if !isArchive {
                 navigationItem.setLeftBarButton(cancelButton, animated: true)
             }
         } else if isArchive {
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -89,8 +89,7 @@ class ChatListController: UITableViewController {
         let msg = dcContext.newMessage(viewType: DC_MSG_TEXT)
         msg.text = String.localized("update_1_28_android") + "\n\n" + String.localized("update_1_28_ios_extra_line")
         dcContext.addDeviceMessage(label: "update_1_28a_ios", msg: msg)
-
-        viewModel.refreshData()
+        handleEmptyStateLabel()
     }
 
     override func willMove(toParent parent: UIViewController?) {
@@ -427,7 +426,10 @@ class ChatListController: UITableViewController {
 
     func handleChatListUpdate() {
         tableView.reloadData()
+        handleEmptyStateLabel()
+    }
 
+    private func handleEmptyStateLabel() {
         if let emptySearchText = viewModel.emptySearchText {
             let text = String.localizedStringWithFormat(
                 String.localized("search_no_result_for_x"),
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -54,7 +54,7 @@ class ChatListController: UITableViewController {
 
     private lazy var emptyStateLabel: EmptyStateLabel = {
         let label = EmptyStateLabel()
-        label.isHidden = false
+        label.isHidden = true
         return label
     }()
 
@@ -96,7 +96,6 @@ class ChatListController: UITableViewController {
         let msg = dcContext.newMessage(viewType: DC_MSG_TEXT)
         msg.text = String.localized("update_1_28_android") + "\n\n" + String.localized("update_1_28_ios_extra_line")
         dcContext.addDeviceMessage(label: "update_1_28a_ios", msg: msg)
-        handleEmptyStateLabel()
     }
 
     override func willMove(toParent parent: UIViewController?) {
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -2,8 +2,9 @@ import UIKit
 import DcCore
 
 class ChatListController: UITableViewController {
-    let viewModel: ChatListViewModel
+    var viewModel: ChatListViewModel?
     let dcContext: DcContext
+    var isArchive: Bool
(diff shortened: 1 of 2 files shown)


r10s:
> testing on an emulator, the speed gain is marginal thogh.

do not underestimate that on a real database where a getChatlist() easily takes 500ms or longer :)

r10s:
hm, there are also still both calls mentioned at https://github.com/deltachat/deltachat-ios/issues/1499, cmp. https://gist.github.com/r10s/e4d7045eb4679d259017e4e61312843c (from flightmode, therefore the notification error), so, it is not clear to me what exactly this pr is about.

the second call, however, is in background, so, indeed, that would not add much "visible" improvement - but it is still a call that will slow down other things.

i would imagine, that we can do this "handleEmptyStateLabel" on init, instead of the first call to getChatlist(), the viewDidLoad could maybe stay as is. that way, the ui would pop up much faster and the chatlist would get visible as it is there, a moment later.

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -89,8 +89,7 @@ class ChatListController: UITableViewController {
         let msg = dcContext.newMessage(viewType: DC_MSG_TEXT)
         msg.text = String.localized("update_1_28_android") + "\n\n" + String.localized("update_1_28_ios_extra_line")
         dcContext.addDeviceMessage(label: "update_1_28a_ios", msg: msg)
-
-        viewModel.refreshData()
```
cyBerta, line 93: this call has been removed. It calls dcContext.getChatlist() internally and is executed on the main thread.

cyBerta:
the second call to refreshInBg() comes from applicationDidBecomeActive(_ notification: NSNotification). It is called whenever the app gets to foreground. I would like to avoid tweaking that method for the initialization case, also because it is not blocking the UI.

cyBerta:
So to answer the question what this PR is about: it reduces the amount of inital calls to dcContext.getChatlist() from 2 to 1, while a 3. call (now the 2.) is still executed, but on a background thread.

r10s:
thanks for more detailed explanations! ❤️ 

i am still wondering if we can get that down to one call.

maybe remove that one in `init()`, showing the emptyStateLabel there so that we do not block the ui. and do the background call always in viewWillAppear(). the 3rd call could be optimized away if we add debouncing as we are doing already for the search (debouncing makes sense anyway there, tbh, i thought we already did that - or have i overseen sth?)

but we can maybe better sync about that in an av.

cyBerta:
> the 3rd call could be optimized away if we add debouncing as we are doing already for the search (debouncing makes sense anyway there, tbh, i thought we already did that - or have i overseen sth?)

We implemented debouncing for `refreshInBg() `. It reduces multiple calls to the core to max. 2 within 500ms (the initial one is always immediately executed, all follwing ones are reduced to one call after 500 ms). 

viewWillAppear is not always executed, it is only exectuted if the view controller initially appears and if it reappears, after stacked viewcontrollers have been popped.  It is not called if the app goes to the background for some seconds. This case is covered by `applicationDidBecomeActive`. The mechanism was introduced to fix missing updates in that case. I wouldn't just remove it, because it would cause other regression.

cyBerta:
I'm not so happy with https://github.com/deltachat/deltachat-ios/pull/1514/commits/58a7b9d74a7ea581e446feb53ed92592101bf51b and would rather revert it. This commit reduces the number of initial calls to getChatList to one and it is always done in background. However the viewModel becomes optional, which increases the code complexity to some extend. It also breaks the viewModel approach, because I have to keep `isArchive` outside of the ChatlistViewModel so that the title bar and buttons can be immediately initialized correctly. To me the speed gain is not worth the decrease of code quality. If you have other good idea how to refactor this class, I would be happy to hear them.

r10s:
complexity is increased, yes. but the win is huge if you have some filled database, i did two runs with and without this pr, here are the logs:
without this pr: https://gist.github.com/r10s/3556534a02d74526a4bf65df04daf8e8
with this pr: https://gist.github.com/r10s/c06858fa621515b8cd650540849957ce

_processing_ time, background or not, is reduced from 12.6 seconds (partly in background, but that still slows down things) to 3.4 seconds - this is on quarter.

_startup_ time is also much faster, but that needs to be measured, i need to have a closer look at that.

so, i think, it is worth following this approach, but yes, there are some thing left  to do,  eg. the search functionaility is gone :)

cyBerta:
> so, i think, it is worth following this approach, but yes, there are some thing left to do, eg. the search functionaility is gone :)

I've fixed that and the initialization of the create button as well as an introduced bug wrt. forwarding (cancel button venished).

r10s:
> startup time is also much faster, but that needs to be measured.

i did some tests with a quite filled database on an iphone7: 

- total loading time improves **by 3 seconds** (from 7 seconds to 4 seconds)
- time, until the app is responsive, **improves by 6 seconds** (from 7 seconds to 1 seconds)

to have comparable results, i rebooted the iphone each time before testing. the latest core improvements are not yet included.

here is a video:

https://user-images.githubusercontent.com/9800740/161298837-e180aa76-b672-4119-a1df-73dd9e746c76.mp4

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -241,11 +248,14 @@ class ChatListController: UITableViewController {
...
-    
+    private var isInitial = true
     @objc func applicationDidBecomeActive(_ notification: NSNotification) {
         if navigationController?.visibleViewController == self {
-            startTimer()
-            refreshInBg()
+            if !isInitial {
+                startTimer()
+                refreshInBg()
+                isInitial = false
```
r10s, line 257: as a general precaution, it probably makes sense to set `isInitial = false` two lines further up:

```suggestion
                isInitial = false
                startTimer()
                refreshInBg()
```

r10s (approved):
lgtm now.

there is an issue left with `handleEmptyStateLabel()` - the emptyStateLabel does not disappear when the archive is opened:



i think, the empty state label should only appear when the list is finished loaded - otherwise, it would be just wrong (saying "there is no archive" for a moment and then display the archive)

but we can also fix that in a subsequent pr.

r10s (approved):
i added a commit that fixes the flashing of the "empty state" label, now everything is fine for me :)

PR #1524, opened 2022-04-05
add account-id to webxdc-url

### LINKED ISSUE #1507: add account id to webxdc-fake-url, avoid "app" wording
should be simple to do and avoids collisions, see https://github.com/deltachat/deltachat-android/issues/2229

when on that, we should also change the domain to [`.localhost`](https://en.wikipedia.org/wiki/.localhost) instead of `.local.app` unless there is a good reason for the latter.

so, full domain would be `acc123-msg456.localhost` - again, unless there are good reasons for sth. else.

r10s:
see commit message for details.

closes #1507


cyBerta (approved):
lgtm

PR #1525, opened 2022-04-05
adapt to new webxdc-api

### LINKED ISSUE #1495: adapt to new webxdc api
see https://github.com/deltachat/deltachat-core-rust/pull/3081 and https://github.com/deltachat/deltachat-android/pull/2223 for details.

r10s:
closes #1495 

see commit messages for some details.

### DIFF
--- a/deltachat-ios/Controller/WebxdcViewController.swift
+++ b/deltachat-ios/Controller/WebxdcViewController.swift
@@ -212,16 +212,18 @@ class WebxdcViewController: WebViewViewController {
         }
     }
 
-    var lastSerial: Int = 0
+    var lastSerial: Int?
     private func updateWebxdc() {
-        let statusUpdates = dcContext.getWebxdcStatusUpdates(msgId: messageId, lastKnownSerial: lastSerial)
-        if let data: Data = statusUpdates.data(using: .utf8),
-           let array = (try? JSONSerialization.jsonObject(with: data, options: [])) as? [Any],
-           let first = array.first as? [String: Any],
-           let maxSerial = first["max_serial"] as? Int {
-            lastSerial = maxSerial
+        if let lastSerial = lastSerial {
+            let statusUpdates = dcContext.getWebxdcStatusUpdates(msgId: messageId, lastKnownSerial: lastSerial)
+            if let data: Data = statusUpdates.data(using: .utf8),
+               let array = (try? JSONSerialization.jsonObject(with: data, options: [])) as? [Any],
+               let first = array.first as? [String: Any],
+               let maxSerial = first["max_serial"] as? Int {
+                self.lastSerial = maxSerial
+            }
+            webView.evaluateJavaScript("window.__webxdcUpdate(atob(\"\(statusUpdates.toBase64())\"))", completionHandler: nil)
         }
-        webView.evaluateJavaScript("window.__webxdcUpdate(atob(\"\(statusUpdates.toBase64())\"))", completionHandler: nil)
     }
 }
 
--- a/deltachat-ios/Controller/WebxdcViewController.swift
+++ b/deltachat-ios/Controller/WebxdcViewController.swift
@@ -152,12 +152,13 @@ class WebxdcViewController: WebViewViewController {
         ) { [weak self] notification in
             guard let self = self else { return }
             guard let ui = notification.userInfo,
-                  let messageId = ui["message_id"] as? Int,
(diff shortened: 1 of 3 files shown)


deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -227,10 +211,16 @@ class WebxdcViewController: WebViewViewController {
...
-        logger.debug("status updates: \(statusUpdates)")
+
+    var lastSerial: Int = 0
+    private func updateWebxdc() {
+        let statusUpdates = dcContext.getWebxdcStatusUpdates(msgId: messageId, lastKnownSerial: lastSerial)
+        if let data: Data = statusUpdates.data(using: .utf8),
+           let array = (try? JSONSerialization.jsonObject(with: data, options: [])) as? [Any],
+           let first = array.first as? [String: Any],
+           let maxSerial = first["max_serial"] as? Int {
+            lastSerial = maxSerial
```
r10s, line 222: @cyBerta here i also stumbled upon the `if let` thingie :) maybe we should take care that the `let` are really horizontally aligned, then it is at least a bit better readable - and swift does not complain and we still follow swift style.

deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -227,10 +211,16 @@ class WebxdcViewController: WebViewViewController {
             }
         }
     }
-    
-    private func updateWebxdc(statusId: Int) {
-        let statusUpdates = self.dcContext.getWebxdcStatusUpdates(msgId: messageId, statusUpdateId: statusId)
-        logger.debug("status updates: \(statusUpdates)")
+
+    var lastSerial: Int = 0
+    private func updateWebxdc() {
```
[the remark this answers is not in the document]
r10s: > why not do this in js like on the other platforms?

see commit message, however, also in more detail here:

handling `lastSerial` inside js would require to keep the bit complicated `async_calls` array; moreover,  also  `getStatusUpdates` would become more complex as it would require a second parameter (`invocationId` and `lastSerial` then). doable, but all that sounds a bit too complicated without much benefits.

handling `lastSerial` inside swift removes the  `async_calls`, it removes `invocationId` array and also simplifies the call to swift (getStatusUpdates is no longer needed)

so, i think, the swift approach makes the javascript part much smaller, and removes several lines of code and complexity in general. 

> also is the swift lastSerial cleared correctly?

lastSerial is set when `setUpdateListener` is set - after that, it is set to max_serial whenever updates come in. when should it be cleared?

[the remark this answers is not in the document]
Simon-Laux: when switching apps, I don't know how long those swift vars live or whether they are reused... so I thought it would be better to ask not that we miss sth.

deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -227,26 +212,33 @@ class WebxdcViewController: WebViewViewController {
...
+    var lastSerial: Int?
+    private func updateWebxdc() {
+        if let lastSerial = lastSerial {
+            let statusUpdates = dcContext.getWebxdcStatusUpdates(msgId: messageId, lastKnownSerial: lastSerial)
+            if let data: Data = statusUpdates.data(using: .utf8),
+               let array = (try? JSONSerialization.jsonObject(with: data, options: [])) as? [Any],
+               let first = array.first as? [String: Any],
+               let maxSerial = first["max_serial"] as? Int {
+                self.lastSerial = maxSerial
+            }
```
Simon-Laux, line 225: I don't really get this strange syntax:
what curly braces terminate which block?
what's with the indentation?

Simon-Laux (approved):
code looks good to me, though I don't understand the strange swift syntax.
have not tried it.

deltachat-ios/Controller/WebxdcViewController.swift
r10s: the block is only the single line `self.lastSerial = maxSerial` - but that is the default swift formatting, @cyBerta is already very used to that, i was also stumbling upon that, thinking the block also includes the `let`-lines. 

would be better if the `{` is in a single line and on the same indetation as `}` - but swift then complains (and it was not important enough to check how to tweak these linter things ;)

r10s:
i played around a bit with the adapted https://github.com/r10s/webxdc-poll and some other Webxdc - things work as expected so far :)

PR #1526, opened 2022-04-05
cleanup temporary Images and keep background image

### LINKED ISSUE #1498: custom wallpaper forgotten after some time
since the last version, one can define a custom wallpaper.

however, after some time (or after some special action i am not yet aware of), the wallpaper is just forgotten and the background is _plain white_ (_not:_ the default background).

i encountered that first time with "ipcloud photos" are enabled,however, that might be totally unrelated, @cyBerta say, we are doing a copy. so, then, it might be a caching issue. 

unless we see sth. obvious, i suggest to add some more debugging, eg. add the path of the image, including size, to the "debug info", so next time that happens we will know if the image is deleted or sth. else has happened.

cyBerta:
closes #1498

### DIFF
--- a/deltachat-ios/Helper/ImageFormat.swift
+++ b/deltachat-ios/Helper/ImageFormat.swift
@@ -157,4 +157,27 @@ extension ImageFormat {
         }
         return nil
     }
+
+    public static func deleteImage(atPath: String) {
+        if Thread.isMainThread {
+            DispatchQueue.global(qos: .background).async {
+                deleteFile(path: atPath)
+            }
+        } else {
+            deleteFile(path: atPath)
+        }
+    }
+
+    private static func deleteFile(path: String) {
+        let fileManager = FileManager.default
+        if !fileManager.fileExists(atPath: path) {
+            return
+        }
+
+        do {
+            try fileManager.removeItem(atPath: path)
+        } catch {
+            print("err: \(error.localizedDescription)")
+        }
+    }
 }
--- a/deltachat-ios/Helper/ImageFormat.swift
+++ b/deltachat-ios/Helper/ImageFormat.swift
@@ -100,7 +100,9 @@ extension ImageFormat {
         }
         var subdirectoryURL = directoryURL.appendingPathComponent(identifier)
         do {
-            try fileManager.createDirectory(at: subdirectoryURL, withIntermediateDirectories: true, attributes: nil)
+            if !fileManager.fileExists(atPath: subdirectoryURL.path) {
+                try fileManager.createDirectory(at: subdirectoryURL, withIntermediateDirectories: true, attributes: nil)
+            }
         } catch {
             print("err: \(error.localizedDescription)")
             return nil
--- a/deltachat-ios/Helper/ImageFormat.swift
+++ b/deltachat-ios/Helper/ImageFormat.swift
@@ -84,6 +84,8 @@ extension ImageFormat {
         return saveImage(data: data, name: name, suffix: suffix)
     }
 
+    // implementation is following Apple's recommendations
+    // https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/AccessingFilesandDirectories/AccessingFilesandDirectories.html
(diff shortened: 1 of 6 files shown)


r10s:
@cyBerta there are some conflicts now, this needs a rebase.

r10s:
is it known or on purpose, that the background is visible in the "Files" app deep inside the DBDebugToolkit folder?



location is DBDebugToolkit / Network / Response / var / mobile / Containers/  Data / Application / <hex-number> / Library / Application%20Support / chat.delta / BACKGROUBND_IMAGE.jpg_<number>

maybe this belongs to some setting i made and forgot about - otherwise, i find that weird and unexpected.

deltachat-ios/Helper/ImageFormat.swift
```diff
@@ -84,25 +84,55 @@ extension ImageFormat {
...
-        else { return nil }
+
+        // ensure directory exists (application support dir doesn't exist per default)
+        let fileManager = FileManager.default
+        let urls = fileManager.urls(for: directory, in: .userDomainMask) as [URL]
+        guard let identifier = Bundle.main.bundleIdentifier else {
+            print("err: Could not find bundle identifier")
+            return nil
+        }
+        guard let directoryURL = urls.last else {
```
r10s, line 97: why the last here? https://programmingwithswift.com/save-file-to-documents-directory-with-swift/ uses the first ... just asking, i am not sure which one is better :)

deltachat-ios/Helper/ImageFormat.swift
```diff
@@ -84,25 +84,55 @@ extension ImageFormat {
...
+            print("err: Could not find directory url for \(String(describing: directory)) in .userDomainMask")
+            return nil
+        }
+        var subdirectoryURL = directoryURL.appendingPathComponent(identifier)
```
r10s, line 101: why are we creating an extra subdirectory here? isn't applicationSupportDirectory already unique to our app?

in general, maybe add  a some link to some documentation or used source for this whole block.

deltachat-ios/Helper/ImageFormat.swift
```diff
@@ -129,4 +159,27 @@ extension ImageFormat {
...
+        if Thread.isMainThread {
+            DispatchQueue.global(qos: .background).async {
+                deleteFile(path: atPath)
+            }
+        } else {
+            deleteFile(path: atPath)
+        }
+    }
+
+    private static func deleteFile(path: String) {
```
r10s, line 173: why not also `atPath` here?

deltachat-ios/Helper/ImageFormat.swift
```diff
@@ -84,25 +84,55 @@ extension ImageFormat {
...
+        do {
+            if !fileManager.fileExists(atPath: subdirectoryURL.path) {
+                try fileManager.createDirectory(at: subdirectoryURL, withIntermediateDirectories: true, attributes: nil)
+            }
+        } catch {
+            print("err: \(error.localizedDescription)")
+            return nil
+        }
+
+        // Opt out from iCloud backup
```
r10s, line 111: is this really needed for `.cachesDirectory`?

r10s (approved):

deltachat-ios/Helper/ImageFormat.swift
cyBerta: I'll use the first one, it is also mentioned in the apple example https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/AccessingFilesandDirectories/AccessingFilesandDirectories.html

deltachat-ios/Helper/ImageFormat.swift
cyBerta: I sould stick to the recommendations in this case: https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/AccessingFilesandDirectories/AccessingFilesandDirectories.html
the identifier is generally used to avoid file collisions. I'll add a comment

deltachat-ios/Helper/ImageFormat.swift
cyBerta: no not necessarily, but for the .applicationSupportDirectory, which is the default argument in the method signature.

PR #1527, opened 2022-04-06
Send correct DC_EVENT_MSGS_CHANGED when returning from share extension

### LINKED ISSUE #1513: Stuff sent via share extention is not visible in the chat until I leave and re-enter it
- Device: iphone 6s
- Delta Chat version: 1.28.1
- Expected behavior: Shared Message is visible in the chat
- Actual behavior:  Shared Message is not visible in the chat until I go back to chatlist and then re-enter it.
1. open `saved messages` chat
2. switch to another app and share something to the `saved messages` chat
3. go back to delta-chat - message is not visible in the chat

r10s:
this fixes the underlying issue of #1513 - that an incomplete event is sent - and fixes the issue also for other views than the ChatViewController (profile etc.)

issue was that the id was left out instead of set to `0` as it happens when the event comes from the core.

fixes #1513

the refactoring of #1518 probably still makes sense, however #1518 fixes the #1513 only for the ChatViewController.

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -197,7 +197,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
                     NotificationCenter.default.post(
                         name: dcNotificationChanged,
                         object: nil,
-                        userInfo: [:]
+                        userInfo: [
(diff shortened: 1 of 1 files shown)


cyBerta (approved):
nice catch!

Issue #1467, opened 2022-02-08
sendUpdate() without summary pops up in chat


r10s:
updates without summaries should not be visible in the chat, however, they're shown on the sending device until the chat is re-entered.

cyBerta:
@r10s We receive both DC_EVENT_MSGS_CHANGED and DC_EVENT_MSG_DELIVERED. I'm not sure if that is really what we want, if we don't track the sent state for webxdc updates in the UI. 
I can filter those updates out, but I think  it should considered to adapt the core actually.

cyBerta:
> updates without summaries should not be visible in the chat

@r10s do we want any outgoing webxdc info messages?

r10s:
> @r10s do we want any outgoing webxdc info messages?
 
yes. the chat one sender and receiver side should have the same information. 
the core adds the info messages as needed - however, not every sendUpdate() results in an info-message.

PR #1529, opened 2022-04-11
fix missing and double webxdc info messages

### LINKED ISSUE #1519: webxdc status messages not shown immediately in chat
the chat is not updated when a webxdc status message is generated on the same device.

once the chat is left and re-entered, the status message is shown.

cyBerta:
* always update chat messages from core if webxdc info message was sent, avoids missing or double info message entries

* fixes #1519 

some context: for outgoing messages, we attempt to optimize things a little bit, which bites us here.  We append new messages instead of loading the whole list again from the core. I've removed that optimization for webxdc info messages. some more context: 
https://github.com/deltachat/deltachat-ios/issues/1467
https://github.com/deltachat/deltachat-ios/pull/1482

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -506,7 +506,14 @@ class ChatViewController: UITableViewController {
             if let ui = notification.userInfo {
                 let dcChat = self.dcContext.getChat(chatId: self.chatId)
                 if dcChat.canSend, let id = ui["message_id"] as? Int, id > 0 {
-                    self.updateMessage(id)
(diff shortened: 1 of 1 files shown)


r10s (approved):
i find the code really hard to read and to understand, but after some thinking, it is probably okay.

i think, at the very end of `updateMessage()` an `else refreshMessages()` is missing - otherwise, we may again forget to update sth. that does not met the other conditions. but this is no new bug.

maybe in another pr, we should get rid of updateMessage at all and move the code one level up where we can combine the conditionns from inside and outside updateMessage().

however, as it actually fixes a bug, i am happy to approve this anyway :)

PR #1530, opened 2022-04-11
Retry notification service

### LINKED ISSUE #1515: notification service does no retry connection
if on startup the notification service cannot be established eg. as there is no network, it stays as "not connected" all the time.

we should try to re-connect as we do with io.

to reproduce:
- enable flight mode and start Delta Chat
- disable flight mode and check "Connectivity View" -> Notifications are still shown as "Not connected"

this is a minor issue, however, as the apps are short-living anyway, so this is probably not related to the "big" notification issues we have.

cyBerta:
fixes #1515


r10s (approved):
lgtm, also tested once and it works as expected :)

PR #1534, opened 2022-04-13
Voice msg accessibility

### LINKED ISSUE #1497: voice-over output halted after sending a voice-message
with voice-over enabled, the system explains button etc. on tapping (via audio-outpout on the system speaker) (a double tap does the action then).

during recording of a voice-message, the audio-output is halted, which is correct (otherwise the explanations would be recorded as well)

the issue:

- newer versions do not restart audio-output after sending a voice-message (voice-over seems to be enabled as there is still a vibrate on tapping controls)

- on older versions, this has worked (so voice-over was restarted as expected)

the issue may be cause by a change in Delta Chat _or_ by iOS 15, this is not yet figured out; issue reported by @missytake's  interviewee.

cyBerta:
fixes #1497


r10s (approved):
lgtm, as far as i could understand it :) did not test, however.

PR #1536, opened 2022-04-13
suspend media playback instead of pausing it WebxdcViewController gets dismissed

### LINKED ISSUE #1488: audio playback of closed webxdc is still "visible"
while with https://github.com/deltachat/deltachat-ios/pull/1477 audio playback is paused - it is only that: paused. you can resume it eg, from the lockscreen, even when delta chat is left :)

resuming does work for a while, but at some point it does no longer work at all, but the controls are still there. may also be a bug in ios, however, for now i suspect  Delta Chat more :)

cyBerta:
fixes #1488


r10s (approved):
great - i also tested, works like a charm :)

PR #1542, opened 2022-04-16
[do not merge] try to debug background fetch better

r10s:
this pr is not for merging, it only adds some additional debugging statements to get a grip on where the time wasting figured out at https://github.com/deltachat/deltachat-ios/pull/1461 comes from.

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -473,10 +473,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
             self.pushToDebugArray("s4")
             if !self.appIsInForeground() {
-                self.pushToDebugArray("s5:"+String(format: "%.3fs", Double(Date().timeIntervalSince1970)-nowTimestamp))
+                self.pushToDebugArray("s5|"+String(format: "%.3fs", Double(Date().timeIntervalSince1970)-nowTimestamp))
                 self.dcAccounts.stopIo()
             }
-            self.pushToDebugArray("s6:"+String(format: "%.3fs", Double(Date().timeIntervalSince1970)-nowTimestamp))
+            self.pushToDebugArray("s6|"+String(format: "%.3fs", Double(Date().timeIntervalSince1970)-nowTimestamp))
 
             // to avoid 0xdead10cc exceptions, scheduled jobs need to be done before we get suspended;
             // we increase the probabilty that this happens by waiting a moment before calling completionHandler()
@@ -489,7 +489,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
                     return
                 }
 
-                self.pushToDebugArray("d2:"+String(format: "%.3fs", Double(Date().timeIntervalSince1970)-nowTimestamp))
+                self.pushToDebugArray("d2|"+String(format: "%.3fs", Double(Date().timeIntervalSince1970)-nowTimestamp))
                 completionHandler(.newData)
                 self.pushToDebugArray("d3")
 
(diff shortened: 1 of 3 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -428,6 +433,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         // if at some point we do per-message-push-notifications, we need to tweak this gate.
         let nowTimestamp = Double(Date().timeIntervalSince1970)
         if nowTimestamp < bgIoTimestamp + 60 {
+            self.pushToDebugArray("e2")
             logger.info("➡️ fetch was just executed, skipping")
             completionHandler(.newData)
```
r10s, line 438: we are skipping calls far too often, the condition seems to be wrong:



does not happen always, screenshot is from iphone7, the other testphones do not have that issue yet.  once we are in this state, however, we're kind of stuck.

deltachat-ios/AppDelegate.swift
```diff
@@ -458,23 +468,30 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
                 return
             }
 
+            self.pushToDebugArray("s4")
             if !self.appIsInForeground() {
+                self.pushToDebugArray("s5:"+String(format: "%.3fs", Double(Date().timeIntervalSince1970)-nowTimestamp))
                 self.dcAccounts.stopIo()
```
r10s, line 474: by some measurements, the time is not wasted here (as i suspected) but somewhere before:

deltachat-ios/AppDelegate.swift
```diff
@@ -49,7 +49,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         // setupCrashReporting() may create an additional handler, but we do not want to rely on that
         signal(SIGPIPE, SIG_IGN)
 
-        bgIoTimestamp = Double(Date().timeIntervalSince1970)
+        self.pushToDebugArray("🏁")
```
r10s, line 52: setting `bgIoTimestamp` here seems to be wrong in case the app was not freezed by actually killed (what happens more often on older phones than on newer)

deltachat-ios/AppDelegate.swift
r10s: `bgIoTimstamp` must probably not be set on init, see comment above.

r10s: not sure at the end as also `didFinishLaunchingWithOptions` starts IO, however, maybe does not run long enough. we need more information ...

deltachat-ios/AppDelegate.swift
```diff
@@ -141,8 +142,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         }
         
         if let notificationOption = launchOptions?[.remoteNotification] {
+            self.pushToDebugArray("📡'")
             logger.info("Notifications: remoteNotification: \(String(describing: notificationOption))")
             increaseDebugCounter("notify-remote-launch")
+            performFetch(completionHandler: dummyCompletionHandler)
```
r10s, line 148: not sure if that is working as expected ... we need to observe that for another day ...

r10s: `didFinishLaunchingWithOptions:remoteNotification` may or may not be called before `didReceiveRemoteNotification`, might be that behavior has changed in iOS 14, cmp. https://stackoverflow.com/questions/59990130/push-notifications-sometimes-call-didfinishlaunchingwithoptions-instead-of-didre

what is still weird - i do not manage to start the background thread in performFetch() in all cases, seems esp. the call from `didFinishLaunchingWithOptions:remoteNotification` often fails.

r10s: i will try over with DispatchQueue.global(qos: .default).async https://stackoverflow.com/questions/42452437/swift-3-0-async-dispatch-global-does-not-run

r10s:
i'll create a  new pr with only the relevant changes and without all the testing steps, this pr stays for reference.

deltachat-ios/AppDelegate.swift
```diff
@@ -447,38 +462,46 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
                 return
             }
 
+            // we're in background, run IO for a little time
+            self.dcAccounts.startIo()
+            self.pushToDebugArray("s2")
+            self.dcAccounts.maybeNetwork()
+            self.pushToDebugArray("s3")
+
+            _ = fetchSemaphore.wait(timeout: .now() + 10)
```
r10s, line 480: this is still where the app sometimes hangs and the semaphore never returns.

might be ios takes the waking as a chance to freeze the app again. _much_ too late, the app returns then, without getting messages in time.

this does not happen often, but it happens. maybe w can try if a look with some sleep() works better.

PR #1546, opened 2022-04-20
Lint fixes

### LINKED ISSUE #1540: empty enum argument violation
i have no idea what these warnings wants to say to me:

WelcomeViewController.swift:

AppDelegate.swift:

these are just two examples, there are more corresponding warnings.

imu, the functions have just no arguments. is this a false positive of the linter?

cyBerta:
closes #1540 and some other lint (and real) issues.

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -882,7 +882,7 @@ public class DcChat {
         return DcUtils.copyAndFreeArray(inputArray: dc_get_chat_contacts(dcContext.contextPointer, UInt32(id)))
     }
 
-    public lazy var profileImage: UIImage? = { [weak self] in
+    public lazy var profileImage: UIImage? = {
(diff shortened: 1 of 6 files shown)


r10s (commented):
adding 31mb binary data to the repo looks like an accident to me ...?

cyBerta:
> adding 31mb binary data to the repo looks like an accident to me ...?
> 
> not sure, however, already now, there are 40mb in the Pods directory, maybe this is just the way ...

This extra MBs is only fetched locally,  but not pushed to the repository, or?! I think that's the way it is.

r10s:
> This extra MBs is only fetched locally, but not pushed to the repository, or?! 

it is in this pr:

Screen Shot 2022-04-22 at 12 52 50

however, yes, if that is the way to go, then that's not great but okay.

cyBerta:
> however, yes, if that is the way to go, then that's not great but okay.

that's the way to go, unforunately.

r10s (approved):

Issue #1516, opened 2022-03-27
too much scrolling in message view


r10s:
the message view still does much too much scrolling.

eg. when sending a message, it happens regulary, that the view does not just scroll _up_ but scrolls down to the new message.

there are many more similar situations, on sending etc. i still think, we are doing too much :)

cyBerta:
One thing that really annoys me is if I read a message and the message jumps after a minute, because the time label was updated and the list is already near the bottom. I will take care of that issue in a separate PR now. 

Another improvement would be that the list isn't scrolled down automatically as long as the user drags the list.

r10s:
tracked in https://github.com/deltachat/deltachat-ios/issues/1737 now

PR #1549, opened 2022-04-21
less scrolling to the bottom in chat view

cyBerta:
related to #1516

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -422,7 +422,7 @@ class ChatViewController: UITableViewController {
                     guard let self = self else { return }
                     self.highlightedMsg = nil
                     self.isInitial = false
-                    self.messageInputBar.scrollDownButton.isHidden = self.isLastRowVisible()
+                    self.updateScrollDownButtonVisibility()
                 }
             })
         } else {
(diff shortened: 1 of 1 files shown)


r10s:
the issues described in https://github.com/deltachat/deltachat-ios/issues/1516https://github.com/deltachat/deltachat-ios/issues/1516 are unchanged for me - here is a video to clarify what i mean:

1. the keyboard flickers (any maybe bumps the whole screen up and down which results in 2.)
2. the messages appear as being  scrolled  in from the top
3. the "down" button is flickering (it should not be there at all)

r10s:
the issues described in https://github.com/deltachat/deltachat-ios/issues/1516https://github.com/deltachat/deltachat-ios/issues/1516 are unchanged for me - here is a video to clarify what i mean:

1. the keyboard flickers (any maybe bumps the whole screen up and down which results in 2.)
2. the messages appear as being  scrolled  in from the top
3. the "down" button is flickering (it should not be there at all)

cyBerta:
It takes too long to upload videos here. That's why I'm just describing what exactly this PR is fixing. It relates to: https://github.com/deltachat/deltachat-ios/issues/1516#issuecomment-1103974251

With this PR, scrolling to the bottom when new messages are received only happens if the bottom of last message is currently visible. Otherwise the scroll-down message appears. 
Without the PR always the tableview is always scrolled down if new messages are received for a thread. This bothers the user (/me) when reading messages, especially longer ones.

It mentioned fix also applies to other situations than inserting new message, like updating cells. Without the fix, sometimes the messages scroll down to the bottom after about a minute. That is when the receive date of all messages are updated in the UI and the cells are reloaded.

Also this PR fixes the scroll-down button showing up in case the keyboard is shown and the last message is not visible. The behavior is now the same if the keyboard is hidden or shown. 
Without this PR, the scroll-down button only shows up when the last message has been scrolled out of the screen, not out of the visible area (below the message input bar).

r10s (approved):

PR #1550, opened 2022-04-25
Apply bg fetch test outcomes

r10s:
this is the extract of https://github.com/deltachat/deltachat-ios/pull/1542 and also picks up the semaphore idea of https://github.com/deltachat/deltachat-ios/pull/1461

the pr improves getting notifications a lot on my test devices, there were some real bugs in, one issue is open, see code-comment `TOCHECK` - but that can be done in a subsequent pr and needs some more testing.

see commit messages for details, targets https://github.com/deltachat/deltachat-ios/issues/1157

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -144,6 +144,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         if let notificationOption = launchOptions?[.remoteNotification] {
             logger.info("Notifications: remoteNotification: \(String(describing: notificationOption))")
             increaseDebugCounter("notify-remote-launch")
+            performFetch(completionHandler: { (_) -> Void in })
         }
 
         if dcAccounts.getSelected().isConfigured() && !UserDefaults.standard.bool(forKey: "notifications_disabled") {
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -435,6 +435,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             return
         }
         bgIoTimestamp = nowTimestamp
+        increaseDebugDoingFetch()
 
         // make sure to balance each call to `beginBackgroundTask` with `endBackgroundTask`
         var backgroundTask: UIBackgroundTaskIdentifier = .invalid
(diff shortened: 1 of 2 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -143,6 +144,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         if let notificationOption = launchOptions?[.remoteNotification] {
             logger.info("Notifications: remoteNotification: \(String(describing: notificationOption))")
             increaseDebugCounter("notify-remote-launch")
+            performFetch(completionHandler: { (_) -> Void in })
```
cyBerta, line 147: can't you just pass nil here?

cyBerta: ok, I see now currently you can't but still normally they _should_ be optional. I think it makes sense to fix it, it's our code as far as i can see.

r10s: hm, we are using the same syntax as for the completion handler we are forwarding from `performFetchWithCompletionHandler`/`didReceiveRemoteNotification` - changing that to `@escaping ((UIBackgroundFetchResult) -> Void)?` results in an error, maybe because of the @escaping (whatever that is for) - so, maybe things are just fine and the empty completion handler is the minor issue :)

r10s: @escaping seems to be needed https://www.donnywals.com/what-is-escaping-in-swift/

(sorry for pinging you, @escaping :)

deltachat-ios/AppDelegate.swift
```diff
@@ -568,19 +570,20 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
-            var slidingTimeframe: [Double]
-            if timestamps != nil, let timestamps = timestamps as? [Double] {
-                slidingTimeframe = timestamps.filter({ nowTimestamp < $0 + 60 * 60 * 24 })
-            } else {
-                slidingTimeframe = [Double]()
-            }
-            slidingTimeframe.append(nowTimestamp)
-            UserDefaults.standard.set(slidingTimeframe, forKey: Constants.Keys.notificationTimestamps)
+    // Values calculated for connectivity view
+    private func increaseDebugDoingFetch() {
```
cyBerta, line 581: The method name confuses me a little bit, I thought it is actually fetching, but it adding timestamps.

r10s: happy to rename it - do you have a suggestion?

cyBerta: addDebugTimestamp ?

cyBerta: addDebugFetchTimestamp

PR #1551, opened 2022-04-25
hotfix: appIsInForeground() must not be called from non-main thread

r10s:
this sneaked in by last-minute changes of #1550 :)

(if whitespace is disabled by review, only few lines have changed)


r10s:
@cyBerta i just merged that in as otherwise bg-fetch crashes always, if you have code comments, i do that in another pr :)

PR #1552, opened 2022-04-25
signal when connectivity is "connected" again

r10s:
this finishes the background-fetch much faster,
instead of always 11 seconds, things finish in less than 2 seconds usually
and take longer only when here is really something to download.

EDIT: after testing the pr for a day, there are some open points:
- [x] i agree with @cyBerta - we should check all accounts for reaching connected state - otherwise, if your active account is "fast", slower background account won't even get connected
- [x] we should not only abort on DC_CONNECTIVITY_CONNECTED but also for DC_CONNECTIVITY_NOT_CONNECTED - we may have no network or the server may be down. i had bad network at some place, and run always for 20 seconds therfore

fetchSemaphore is signalled when we receive the DC_CONNECTIVITY_CHANGED event
with connectivity set to DC_CONNECTIVITY_CONNECTED - we have called maybeNetwork()
just before which sets connectivity to CONNECTING,
we know pretty sure that one round is done.

we check that on the selected context only,
that is maybe not perfect, but probably good enough
(other accounts are checked as well, but may be stopped too soon)

this implements ideas already drafted at https://github.com/deltachat/deltachat-ios/pull/1461 and finally closes https://github.com/deltachat/deltachat-ios/issues/1157

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -9,6 +9,7 @@ public class DcAccounts {
     let applicationGroupIdentifier = "group.chat.delta.ios"
     var accountsPointer: OpaquePointer?
     public var logger: Logger?
+    public var fetchSemaphore: DispatchSemaphore?
 
     public init() {
     }
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -56,6 +56,10 @@ public class DcAccounts {
         dc_accounts_maybe_network_lost(accountsPointer)
     }
 
+    public func isAllWorkDone() -> Bool {
+        return dc_accounts_all_work_done(accountsPointer) != 0
+    }
+
     public func startIo() {
         dc_accounts_start_io(accountsPointer)
     }
--- a/DcCore/DcCore/DC/events.swift
+++ b/DcCore/DcCore/DC/events.swift
@@ -214,6 +214,9 @@ public class DcEventHandler {
                 return
             }
             dcContext.logger?.info("network: DC_EVENT_CONNECTIVITY_CHANGED: \(dcContext.getConnectivity())")
+            if let sem = dcAccounts.fetchSemaphore, dcContext.getConnectivity() == DC_CONNECTIVITY_CONNECTED {
+                sem.signal()
+            }
             DispatchQueue.main.async {
                 let nc = NotificationCenter.default
                 nc.post(
--- a/DcCore/DcCore/DC/events.swift
+++ b/DcCore/DcCore/DC/events.swift
@@ -210,13 +210,13 @@ public class DcEventHandler {
             }
 
         case DC_EVENT_CONNECTIVITY_CHANGED:
+            if let sem = dcAccounts.fetchSemaphore, dcAccounts.isAllWorkDone() {
+                sem.signal()
+            }
             if dcContext.id != dcAccounts.getSelected().id {
                 return
             }
             dcContext.logger?.info("network: DC_EVENT_CONNECTIVITY_CHANGED: \(dcContext.getConnectivity())")
-            if let sem = dcAccounts.fetchSemaphore, dcContext.getConnectivity() == DC_CONNECTIVITY_CONNECTED {
-                sem.signal()
-            }
             DispatchQueue.main.async {
                 let nc = NotificationCenter.default
                 nc.post(
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -450,8 +450,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             }
         }
 
-        let fetchSemaphore = DispatchSemaphore(value: 0)
-
         // move work to non-main thread to not block UI (otherwise, in case we get suspended, the app is blocked totally)
         // (we are using `qos: default` as `qos: .background` or `main.asyncAfter` may be delayed by tens of minutes)
         DispatchQueue.global().async { [weak self] in
@@ -461,7 +459,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             self.dcAccounts.startIo()
             self.dcAccounts.maybeNetwork()
 
-            _ = fetchSemaphore.wait(timeout: .now() + 10)
+            // create a new semaphore to make sure the received DC_CONNECTIVITY_CONNECTED really belongs to maybeNetwork() from above
+            // (maybeNetwork() sets connectivity to DC_CONNECTIVITY_CONNECTING, when fetch is done, we're back at DC_CONNECTIVITY_CONNECTED)
+            self.dcAccounts.fetchSemaphore = DispatchSemaphore(value: 0)
+            _ = self.dcAccounts.fetchSemaphore?.wait(timeout: .now() + 10)
+            self.dcAccounts.fetchSemaphore = nil
 
             // TOCHECK: it seems, we are not always reaching this point in code,
             // the semaphore.wait does not exit after 10 seconds and the app gets suspended -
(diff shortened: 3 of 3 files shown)


DcCore/DcCore/DC/events.swift
```diff
@@ -214,6 +214,9 @@ public class DcEventHandler {
                 return
             }
             dcContext.logger?.info("network: DC_EVENT_CONNECTIVITY_CHANGED: \(dcContext.getConnectivity())")
+            if let sem = dcAccounts.fetchSemaphore, dcContext.getConnectivity() == DC_CONNECTIVITY_CONNECTED {
+                sem.signal()
+            }
```
r10s, line 219: this seems to be easier and faster than creating an event handler etc. that may have issues on its own wrt lifetime etc.

deltachat-ios/AppDelegate.swift
```diff
@@ -461,7 +459,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             self.dcAccounts.startIo()
             self.dcAccounts.maybeNetwork()
 
-            _ = fetchSemaphore.wait(timeout: .now() + 10)
+            // create a new semaphore to make sure the received DC_CONNECTIVITY_CONNECTED really belongs to maybeNetwork() from above
+            // (maybeNetwork() sets connectivity to DC_CONNECTIVITY_CONNECTING, when fetch is done, we're back at DC_CONNECTIVITY_CONNECTED)
+            self.dcAccounts.fetchSemaphore = DispatchSemaphore(value: 0)
+            _ = self.dcAccounts.fetchSemaphore?.wait(timeout: .now() + 20)
```
cyBerta, line 465: why 20 s?

cyBerta: ok, I read it in the commit msg.

deltachat-ios/AppDelegate.swift
```diff
@@ -461,7 +459,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             self.dcAccounts.startIo()
             self.dcAccounts.maybeNetwork()
 
-            _ = fetchSemaphore.wait(timeout: .now() + 10)
+            // create a new semaphore to make sure the received DC_CONNECTIVITY_CONNECTED really belongs to maybeNetwork() from above
+            // (maybeNetwork() sets connectivity to DC_CONNECTIVITY_CONNECTING, when fetch is done, we're back at DC_CONNECTIVITY_CONNECTED)
+            self.dcAccounts.fetchSemaphore = DispatchSemaphore(value: 0)
```
cyBerta, line 464: from what I understood, there are cases where perform fetch can be called twice right after each other. it would lead to one semaphore is thrown away immediately. I'm not sure about the implications in depth. It might be ok. However we could also add a null check here.

r10s: i think, this is just okay, would result in one thread waiting for the max. time and not getting a signal (the situations where there are really two concurrent performFetch running are very rare)

i'd like to avoid an `fetchSemaphore != nil` check to avoid weird other happenings as never getting performFetch running - same if we would reuse semaphores and there are too many/few singal() raised.

cyBerta:
What I don't understand is why we're using DC_EVENT_CONNECTIVITY_CHANGED for the check if the account has finished fetching. 
Another question i have regards to the multi-account setup. From my understanding this PR potentially worsens the situation in muti-account situations. while a fixed time of 10 seconds might be enough time for all accounts to fetch all data completely, we now finish after the first account finished connecting?!

cyBerta:
We can test it on the devices, but I guess that's will cause some other issues with multi-account. As a fix we would have to wait for all accounts to finish, the semaphore would be a member of appDelegate in that case instead of a single account.

r10s:
as far as i understand, stopIo() does not really interrupt ongoing downloads, so, a message being downloaded will not fail as being half downloaded (cc @link2xt - or is it different?)

so, yes, non-active-accounts do not get the time they may need, however, there is no notification for them anyway, so, at least for now, things are maybe "good enough". mid-term, it would be better to stopIo if we know of all accounts that they are done (we could probably achieve that by just counting the number of DC_CONNECTIVITY_CONNECTED we got) - but i also like to have things simple for now and see how things behave at all.

r10s:
btw since https://github.com/deltachat/deltachat-ios/pull/1551 the notify-fetch-durations on one testing device get far too big again - so, it seems pretty reprodocible that `DispatchQueue.main.async` (which is needed for the foreground check) causes this problem. it does not affect fetching as such, however, as that is already done.

cyBerta:
> btw since #1551 the notify-fetch-durations on one testing device get far too big again - so, it seems pretty reprodocible that `DispatchQueue.main.async` (which is needed for the foreground check) causes this problem. it does not affect fetching as such, however, as that is already done.


Just an idea: shall we chek if we can track the foregroundstate by ourselves, so that we can revert #1551 but also avoid the crash?

deltachat-ios/AppDelegate.swift
```diff
@@ -481,12 +484,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
                 // to avoid 0xdead10cc exceptions, scheduled jobs need to be done before we get suspended;
                 // we increase the probabilty that this happens by waiting a moment before calling completionHandler()
-                DispatchQueue.global().async { [weak self] in
-                    guard let self = self else { completionHandler(.failed); return }
-
-                    _ = fetchSemaphore.wait(timeout: .now() + 1)
+                DispatchQueue.global().async {
+                    usleep(1_000_000)
```
cyBerta, line 488: have you read that this is better than asyncAfter or it more trial and error?

r10s: i figured that out by testing: `DispatchQueue.main.asyncAfter` is very often executed only when the app back in foreground, so after tens of minutes. same for `DispatchQueue.global(qos: .background).async`.

best guarantee i found that things are really executed in time from a backround thread is a non-mainthread with higher qos, i tested `.default`, therefore `DispatchQueue.global().async`.

note: that all is true if the app is in background only, in foreground all variants are just fine.

cyBerta: > best guarantee i found that things are really executed in time from a backround thread is a non-mainthread with higher qos, i tested `.default`, therefore `DispatchQueue.global().async`.

oh I ment `DispatchQueue.global().asyncAfter` not `DispatchQueue.main.asyncAfter`

cyBerta:
> > btw since #1551 the notify-fetch-durations on one testing device get far too big again - so, it seems pretty reprodocible that `DispatchQueue.main.async` (which is needed for the foreground check) causes this problem. it does not affect fetching as such, however, as that is already done.
> 
> Just an idea: shall we check if we can track the foregroundstate by ourselves, so that we can revert #1551 but also avoid the crash?

to illustrate my idea:
```

func applicationWillEnterForeground(_: UIApplication) {
   logger.info("➡️ applicationWillEnterForeground")
   applicationInForeground = true
   ...
```
 ```
func applicationDidEnterBackground(_: UIApplication) {
      logger.info("⬅️ applicationDidEnterBackground")
      applicationInForeground = false
  }
```

just to make sure in most of the cases we use the already established mechnisms we could additionally do sth. like:
 ```
func appIsInForeground() -> Bool {
        if Thread.isMainThread {
            switch UIApplication.shared.applicationState {
            case .background, .inactive:
                applicationInForeground = false
            case .active:
                applicationInForeground = true
            }
        }
        
        return applicationInForeground
    }


```

deltachat-ios/AppDelegate.swift
r10s: by the removal of two thread dispatching (by allowing to call appIsInForeground() from background), this question is no longer relevant :)

deltachat-ios/AppDelegate.swift
```diff
@@ -215,13 +218,20 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         }
     }
 
+    // applicationDidBecomeActive() is called on initial app start _and_ after applicationWillEnterForeground()
```
cyBerta, line 221: 👍

cyBerta (approved):
let's try it :)

DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -55,6 +56,10 @@ public class DcAccounts {
         dc_accounts_maybe_network_lost(accountsPointer)
     }
 
+    public func isAllWorkDone() -> Bool {
+        return dc_accounts_all_work_done(accountsPointer) != 0
```
cyBerta, line 60: btw. I wasn't aware we have this API !

r10s: i was involved in the design that time, but i also forgot about that :)

PR #1554, opened 2022-04-26
always refresh chatlist in background if app became active again

### LINKED ISSUE #1553: chatlist timstamps not updated sometimes
to reproduce:
- ~~send~~ receive? a message
- wait some minutes
- -> after some more minutes, the timestamp still shows "Now" where it should show "10 min" or so

EDIT: i just tried sending, this works as expected. but it happens to me just this morning that a received messages shows "Now" hours later. the message was received in background - maybe there is some condition when adding the timer?

cyBerta:
fixes #1553

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -250,11 +250,11 @@ class ChatListController: UITableViewController {
     private var isInitial = true
     @objc func applicationDidBecomeActive(_ notification: NSNotification) {
         if navigationController?.visibleViewController == self {
-            if !isInitial {
+            if isInitial {
                 isInitial = false
                 startTimer()
-                refreshInBg()
             }
+            refreshInBg()
         }
     }
 
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -252,8 +252,8 @@ class ChatListController: UITableViewController {
         if navigationController?.visibleViewController == self {
             if !isInitial {
                 startTimer()
+                handleChatListUpdate()
             }
-            handleChatListUpdate()
             isInitial = false
         }
     }
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -250,11 +250,11 @@ class ChatListController: UITableViewController {
     private var isInitial = true
     @objc func applicationDidBecomeActive(_ notification: NSNotification) {
         if navigationController?.visibleViewController == self {
-            if isInitial {
-                isInitial = false
+            if !isInitial {
                 startTimer()
             }
-            refreshInBg()
+            handleChatListUpdate()
+            isInitial = false
         }
     }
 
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/ChatListController.swift
```diff
@@ -250,11 +250,11 @@ class ChatListController: UITableViewController {
     private var isInitial = true
     @objc func applicationDidBecomeActive(_ notification: NSNotification) {
         if navigationController?.visibleViewController == self {
-            if !isInitial {
+            if isInitial {
                 isInitial = false
                 startTimer()
-                refreshInBg()
             }
+            refreshInBg()
```
r10s, line 257: i think `refreshInBg` is too much here, no need to reload the chatlist, this is driven by the events already. it should be sufficient to redraw the cells.

(the bug kind of proves that - applicationDidBecomeActive has never done any update - and the chatlist was always okay in my experience - apart of the timestamps :)

cyBerta: It's not really affecting the perfomance since the whole thing runs in the background and is even debounced. This is also the same what we currently do in the timer that is started by `startTimer()`, so it's also more consistent right now than calling directly handleChatListUpdate().

r10s: sure, it does not block ui, but doing an expensive request that costs maybe 2 seconds processor time every 60 seconds will still affect overall performance and resources. and, if the call is not needed, it seems to be worthwhile to avoid it.

if `handleChatListUpdate()` just forces the redraw - what's wrong with calling it? 

@cyBerta  we can also discuss this in an av, i am available today :)

cyBerta: After reviewing it, the timer registration logic is just wrong

cyBerta: > After reviewing it,I realized the timer registration logic is just wrong

fixed with last commit

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -251,10 +251,10 @@ class ChatListController: UITableViewController {
     @objc func applicationDidBecomeActive(_ notification: NSNotification) {
         if navigationController?.visibleViewController == self {
             if !isInitial {
-                isInitial = false
                 startTimer()
-                refreshInBg()
             }
+            handleChatListUpdate()
```
r10s, line 256: i think, the `handleChatListUpdate()` can go back to the `!isInitial` condition - if it is the initial load, things are created correctly anyway, calling `handleChatListUpdate()` explicitly in this case would result in one unneeded update.

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -442,8 +442,16 @@ class ChatListController: UITableViewController {
...
+        if Thread.isMainThread {
+            tableView.reloadData()
+            handleEmptyStateLabel()
+        } else {
+            DispatchQueue.main.async { [weak self] in
+                guard let self = self else { return }
+                self.tableView.reloadData()
+                self.handleEmptyStateLabel()
+            }
+        }
```
r10s, line 454: in a subsequent pr, we can remove the thread-checks for the other calls of `handleChatListUpdate()`.

r10s (approved):
some minors, also testing the pr, works like expected :)

deltachat-ios/Controller/ChatListController.swift
cyBerta: no, better not, or we would have to add them in the ChatListController used in the Share extension as well.

deltachat-ios/Controller/ChatListController.swift
cyBerta: 🤦 right

deltachat-ios/Controller/ChatListController.swift
r10s: k, then better leave things as is.

PR #1555, opened 2022-04-27
fix multi-select message layout

### LINKED ISSUE #1053: date/delivery display sometimes hidden in new multi-select mode
when starting multi-select mode, the bubbles get a bit less width to make place for the checkmarks.

as the height of the rows stays as is (which is correct!), depending on the bubble-content, text may be wrapped differently, resulting in the last line becoming invisible.

as we cannot modify the height of the rows (that would make things shifting around and result in really bad ux), we could:

1. leave also the width of the bubbles but reduce the margin, that should make enough place for the checkmarks (the margin is used to distinguish better between incoming and outgoing bubbles, for a nicer view)

2. move bubbles a bit right outside (that could be needed in addition to 1. as well)

3. ignore the issue

not sure how hard it is, 3. is only an option if that is really hard :) otherwise i would be in favor to 1.

EDIT: here are two screenshots (i was disturbed while filing the issue . if i would have added them right away, @Hocuri would be less confused probably :)

as you can see below, the first and the last date/delivery display in selection mode is missing.

what i would also consider: unless there are reasons to do otherwise, i think, both modes should have the same input-bar-height. this would lead is smoother animations when hitting eg. cancel as nothing needs to be shifted  vertically (currently, there are some pixel difference)

cyBerta:
fixes #1053

### DIFF
--- a/deltachat-ios/Chat/Views/Cells/BaseMessageCell.swift
+++ b/deltachat-ios/Chat/Views/Cells/BaseMessageCell.swift
@@ -3,11 +3,16 @@ import DcCore
 
 public class BaseMessageCell: UITableViewCell {
 
+    // horizontal message constraints for received messages
     private var leadingConstraint: NSLayoutConstraint?
     private var trailingConstraint: NSLayoutConstraint?
-    private var leadingConstraintCurrentSender: NSLayoutConstraint?
+    private var trailingConstraintEditingMode: NSLayoutConstraint?
     private var leadingConstraintGroup: NSLayoutConstraint?
+    // horizontal message constraints for sent messages
+    private var leadingConstraintCurrentSender: NSLayoutConstraint?
+    private var leadingConstraintCurrentSenderEditingMode: NSLayoutConstraint?
     private var trailingConstraintCurrentSender: NSLayoutConstraint?
+
     private var mainContentBelowTopLabelConstraint: NSLayoutConstraint?
     private var mainContentUnderTopLabelConstraint: NSLayoutConstraint?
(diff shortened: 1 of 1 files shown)


r10s (commented):
while this seems to be better as before, the width of the messages bubbles still alter when entering edit mode, resulting in changing height and overall relayout.

i assume, this is because the "no edit mode left/right" 50+50 and "edit mode left/right" are only estimations?

whatsapp/signal manage to do this by keeping things vertically unchanged, which is pretty smart. maybe we should have a look if signal uses the same controller base and have a look what they are doing there?

cyBerta:
Signal implements their own view controllers: https://github.com/signalapp/Signal-iOS/blob/master/Signal/src/ViewControllers/ConversationView/ConversationViewController.swift, https://github.com/signalapp/Signal-iOS/blob/c77939a907430d8f00d05c0b3025c3c5ec22e76e/SignalUI/ViewControllers/OWSViewController.m
I cannot find where they actually set the edit multiselect mode, but is all build from scretch it seems, without relying too much on iOS' default ViewControllers

cyBerta:
> i assume, this is because the "no edit mode left/right" 50+50 and "edit mode left/right" are only estimations?

True, there's no documentation abount the width of the edit mode. However by testing I came to the conclusion that it's roughly 50 dip. minor shifts might still occour, it seems to be rounding issues. However I tested with a wide range of messages, and all messages are shown completely now. So that's already a huge improvement.

r10s (approved):

PR #1559, opened 2022-04-27
use new dc_provider_new_from_email_with_dns() api

### LINKED ISSUE #1557: use dc_provider_new_from_email_with_dns()
use use dc_provider_new_from_email_with_dns() instead of dc_provider_new_from_email(), see https://github.com/deltachat/deltachat-core-rust/pull/3256 for details and reasoning, requires recent master.

r10s:
see https://github.com/deltachat/deltachat-core-rust/pull/3256 for details, this pr only changes the api call so that everything is as before. 

closes #1557

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -542,8 +542,8 @@ public class DcContext {
         return Int(dc_add_device_msg(contextPointer, label, msg.cptr))
     }
 
-    public func getProviderFromEmail(addr: String) -> DcProvider? {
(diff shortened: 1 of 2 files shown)


cyBerta (approved):

cyBerta:
Why do we want DNS queries on iOS again? Wasn't this an oauth related thing?

r10s:
> Why do we want DNS queries on iOS again? 

it is about showing "@gmail.com information" also for custom domains. as i do not know of much issues with the old provider-info-check, i think, we should just keep it (the pr does not change anything but the function name)

PR #1560, opened 2022-04-27
improve bg logging

r10s:
this is an excerpt of the things tried out at https://github.com/deltachat/deltachat-ios/pull/1542

### DIFF
--- a/deltachat-ios/Controller/SettingsController.swift
+++ b/deltachat-ios/Controller/SettingsController.swift
@@ -698,6 +698,14 @@ internal final class SettingsViewController: UITableViewController, ProgressAler
             info += "\(name)=\(cnt)x\(startStr)\(timestampStr)\n"
         }
 
+        info += "notify-timestamps="
+        if let timestamps = UserDefaults.standard.array(forKey: Constants.Keys.notificationTimestamps) as? [Double] {
+            for currTimestamp in timestamps {
+                info += DateUtils.getExtendedAbsTimeSpanString(timeStamp: currTimestamp) + " "
+            }
+        }
+        info += "\n"
+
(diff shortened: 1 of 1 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -598,13 +608,14 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
         let values = UserDefaults.standard.array(forKey: name)
-        var slidingValues = [Double]()
-        if values != nil, let values = values as? [Double] {
-            slidingValues = values.suffix(16)
+        var slidingValues = [String]()
+        if values != nil, let values = values as? [String] {
+            slidingValues = values.suffix(512)
         }
-        slidingValues.append(value)
+        slidingValues.append(value+"@"+DateUtils.getExtendedAbsTimeSpanString(timeStamp: Double(Date().timeIntervalSince1970)))
```
cyBerta, line 618: minor thing "\(variable)@\(functionCall)" is more swiftish

r10s: happy to use any other format that makes the time distinguishable when skimming over :) what do you suggest?

r10s: i switched to `/` which was alrady in use at printing `3`

cyBerta: Markdown formatting ate a backslash. What I ment is the formatting above (now correct)

cyBerta (approved):

r10s:
thanks for approving! i think, we get that to master now, when weird things happen to the notifications in the future, we will have some data, even if in weird format :)

PR #1561, opened 2022-04-29
[do not merge] no semaphores in background and other experiments

r10s:
while `performFetchWithCompletionHandler` and `didReceiveRemoteNotification` is meanwhile quite okay, the third entry point, `launchOptions?[.remoteNotification]` is not.

some background:

- `launchOptions?[.remoteNotification]` is emitted on some devices sometimes instead of `didReceiveRemoteNotification`
- often, but not always, `launchOptions?[.remoteNotification]` is followed by `performFetchWithCompletionHandler` or `didReceiveRemoteNotification`
- we have some detailed logging meanwhile, eg., we [print `2/time` before](https://github.com/deltachat/deltachat-ios/blob/master/deltachat-ios/AppDelegate.swift#L478) we start waiting for io to finish and [`3/time` afterwards](https://github.com/deltachat/deltachat-ios/blob/master/deltachat-ios/AppDelegate.swift#L494)

what's weird:

- we _sometimes_ on _some devices_ do not get that `3/time` at all
- one assumption was, that the waiting for the semaphore may put the app again to suspend mode, i replaced the semaphor by a loop in this pr, in fact, things seems to be little better, but still not good: https://gist.github.com/r10s/901ab95e44ed376ebad817cb31aa97f9 - also, this is a weird explanation

i am wondering, what exactly `launchOptions?[.remoteNotification]` is meant to be used for. it seems to be emitted in case the app was killed before, as a replacement for `didReceiveRemoteNotification` - but the issues here and also the missing option to call a completion handler make me doubt that. maybe we should not try a `performFetch()` on `launchOptions?[.remoteNotification]` - but then we would not wake up for hours, see the log.

maybe there are just internal crashes?
EDIT: in fact i found one, https://github.com/deltachat/deltachat-core-rust/issues/3264 , however, this does not explain all of failing wakeups, at least i cannot see a crash for every wakeup.

EDIT: some more log from iphone7 _running this pr_ for several hours: https://gist.github.com/r10s/b5b9ccc5a103d65af47dd7a5c37c0191
EDIT: and some log from the same phone _not running_ this pr: https://gist.github.com/r10s/9539b5e3fdb08f9c255b3e73ddd34a8f (not totally sure if they overlap, but the last 12h should be definetely without the pr)

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -473,6 +473,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             guard let self = self else { completionHandler(.failed); return }
 
             // we're in background, run IO for a little time
+            self.dcAccounts.doExitPerformFetch = false
             self.dcAccounts.startIo()
             self.dcAccounts.maybeNetwork()
             self.pushToDebugArray("2")
@@ -481,9 +482,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
             // create a new semaphore to make sure the received DC_CONNECTIVITY_CONNECTED really belongs to maybeNetwork() from above
             // (maybeNetwork() sets connectivity to DC_CONNECTIVITY_CONNECTING, when fetch is done, we're back at DC_CONNECTIVITY_CONNECTED)
-            self.dcAccounts.fetchSemaphore = DispatchSemaphore(value: 0)
-            _ = self.dcAccounts.fetchSemaphore?.wait(timeout: .now() + 20)
-            self.dcAccounts.fetchSemaphore = nil
+            for _ in 0..<20 {
+                usleep(500_000)
+                if self.dcAccounts.doExitPerformFetch {
+                    break
+                }
+            }
 
             // TOCHECK: it seems, we are not always reaching this point in code,
             // semaphore?.wait() does not always exit after the given timeout and the app gets suspended -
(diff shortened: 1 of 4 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -48,7 +50,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     // `didFinishLaunchingWithOptions` is _not_ called
     // when the app wakes up from "suspended" state
     // (app is in memory in the background but no code is executed, IO stopped)
-    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+    func application(_ application: UIApplication, willFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
```
cyBerta, line 53: moving  `let notificationOption = launchOptions?[.remoteNotification]` to `willFinishLaunchingWithOptions` shouldn't work according to the documentation: https://developer.apple.com/documentation/uikit/uiapplication/launchoptionskey/1622967-remotenotification (see the Discussion part)

cyBerta, line 53: the link was wrong of the post that I deleted, but I still think from reading docs the launch options should be read from `didFinishLaunchingWithOptions`

r10s: yip, i just want to do some experiments.

but i am still wondering if it is correct that we start the ui unconditionally, so, also on background fetch. or i misunderstood the code :)

PR #1565, opened 2022-05-04
Revamp welcome screen

### LINKED ISSUE #1563: revamp welcome screen
- [x] revamp welcome screen similar to https://github.com/deltachat/deltachat-android/pull/2284 , however, keep ios button styles as is (so blue background for primary, blue text without border for secondary)

- [ ] if easily doable, move "restore from backup" to the welcome-screen (it is currently hidden inside "log in") (if that is too complicated, we should do that after 1.30)

cyBerta:
closes #1563 

as discussed in #1563 this PR doesn't contain the move of the backup import functionality to the welcome screen.


deltachat-ios/Controller/WelcomeViewController.swift
```diff
@@ -94,8 +94,11 @@ class WelcomeViewController: UIViewController, ProgressAlertHandler {
     override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
         super.viewWillTransition(to: size, with: coordinator)
         welcomeView.minContainerHeight = size.height - view.safeAreaInsets.top
-        scrollView.setContentOffset(CGPoint(x: 0, y: 0), animated: true)
-    }
+        scrollView.setContentOffset(CGPoint(x: 0, y: 0), animated: false)
+        coordinator.animate(alongsideTransition: { [weak self] _ in
+            self?.scrollView.scrollToBottom(animated: true)
+        })
+     }
```
r10s, line 101: what is this change doing?

r10s: ah, i see scrolling the view to bottom when switching from portrait to landscape. 

as this would be most important when the view is _started_ in landscape, but it does not happen then, i think, we can remove this extra complexity. by the  smaller image, at least one button is even visible on iphoneSE.

r10s (approved):

PR #1570, opened 2022-05-05
Fix wrong text top of chat list

OzancanKaratas:
I noticed it while testing the app today. I hope I fixed the right thing.

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -435,7 +435,7 @@ class ChatListController: UITableViewController {
             navigationItem.setLeftBarButton(nil, animated: true)
            
         } else {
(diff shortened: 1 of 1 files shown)


r10s (changes requested):
thank you for bringing up that issue. i assume this is about accessibility.

however, the title "Chats" is correct. the title above a page shows what the shown page is about, in this case, the chat list is shown. showing "Connectivity" as the title would be wrong,

that the title can be tapped is just a side-effect, a shortcut for "settings / connectivity" also without voice over, you can be surprised at the first moment.

but i assume that can be solved by leaving the title as is and adding "Tap to show connectivity" or so as the accessibility hint.

cyBerta:
> but i assume that can be solved by leaving the title as is and adding "Tap to show connectivity" or so as the accessibility hint.

I think that is a good idea

r10s:
superseded by https://github.com/deltachat/deltachat-ios/pull/1571

PR #1577, opened 2022-05-11
Move import backup btn to welcome screen

### LINKED ISSUE #1568: move 'restore from backup' to welcome screen
this is the open point of https://github.com/deltachat/deltachat-ios/issues/1563 after merging https://github.com/deltachat/deltachat-ios/pull/1565.

maybe we can also think over removing the scroll-down on orientation-change, https://github.com/deltachat/deltachat-ios/pull/1565#discussion_r865860477 - it seems not to be needed, and experience show that one easily gets in trouble when trying to be too smart on ios ;)

cyBerta:
closes #1568


r10s (approved):
lgtm, also tested a bit. hopefully this does not add additional issues wrt backups ;)

PR #1587, opened 2022-05-21
show webxdc document name where appropriate

### LINKED ISSUE #1582: support webxdc document names
UI part of https://github.com/deltachat/deltachat-core-rust/pull/3317:

- if `document` is set, in the bubble view or gallery, it should appear in front of the WebxdcName . if things are truncated here, this does not matter, both strings are under the control of the Webxdc anyway.
- due to limited space, in WebxdcViewController, i suggest to show the `document` _instead_ of  `name` - otherwise the group name will not be visible on most phones.
- ~~in the `WebxdcViewController`, it probably makes sense to truncate `document` resp.  `name` so that the group name is most times visible and target [#3316](https://github.com/deltachat/deltachat-core-rust/issues/3316) that way~~ let's first see how it goes in practise, see comment https://github.com/deltachat/deltachat-core-rust/issues/3316#issuecomment-1128627826

things need to be updated when DC_EVENT_MSGS_CHANGED arrive - for the bubble-/gallery-view, this is probably already working as  summary-changes emit the same event.

r10s:
- show document name _instead_ of webxdc name in titles,
  (there is also the chat name shown,
  and three names are definetely too much for typical phones)

- how document name _before_ webxdc name nearby icons

closes #1582
counterpart of https://github.com/deltachat/deltachat-android/pull/2300

### DIFF
--- a/deltachat-ios/Controller/WebxdcViewController.swift
+++ b/deltachat-ios/Controller/WebxdcViewController.swift
@@ -138,9 +138,13 @@ class WebxdcViewController: WebViewViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         let msg = dcContext.getMessage(id: messageId)
+        let dict = msg.getWebxdcInfoDict()
+
+        let document = dict["document"] as? String ?? ""
+        let webxdcName = dict["name"] as? String ?? "ErrName" // name should not be empty
         let chatName = dcContext.getChat(chatId: msg.chatId).name
-        let webxdcName = msg.getWebxdcInfoDict()["name"] as? String ?? ""
-        self.title = webxdcName + " – " + chatName
+
+        self.title = document.isEmpty ? "\(webxdcName) – \(chatName)" : "\(document) – \(chatName)"
     }
     
     override func willMove(toParent parent: UIViewController?) {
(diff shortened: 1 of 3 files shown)


deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -138,9 +138,13 @@ class WebxdcViewController: WebViewViewController {
...
         let msg = dcContext.getMessage(id: messageId)
+        let dict = msg.getWebxdcInfoDict()
+
+        let document = dict["document"] as? String ?? ""
+        let webxdcName = dict["name"] as? String ?? "ErrName" // name should not be empty
         let chatName = dcContext.getChat(chatId: msg.chatId).name
-        let webxdcName = msg.getWebxdcInfoDict()["name"] as? String ?? ""
-        self.title = webxdcName + " – " + chatName
+
+        self.title = document.isEmpty ? "\(webxdcName) – \(chatName)" : "\(document) – \(chatName)"
```
cyBerta, line 147: Isn't that already too long? 
We could use a 2 line custom title view, I think we have that already somewhere and reuse it, I could look it up.

cyBerta (approved):
lgtm

PR #1588, opened 2022-05-23
support large text in document gallery view

### LINKED ISSUE #1585: Large Text support in document gallery view
- similar to #1584 large text support is still missing in the document gallery view

cyBerta:
closes #1585

the cell hight is the same as for the contact and chat cells (in the chat list). Since it has a very similar layout (image, title, subtitle), I think that makes sense.

### DIFF
--- a/deltachat-ios/View/Cell/DocumentGalleryFileCell.swift
+++ b/deltachat-ios/View/Cell/DocumentGalleryFileCell.swift
@@ -7,10 +7,10 @@ class DocumentGalleryFileCell: UITableViewCell {
 
     static var cellHeight: CGFloat {
         let textHeight = UIFont.preferredFont(forTextStyle: .headline).pointSize + UIFont.preferredFont(forTextStyle: .subheadline).pointSize + 24
-        if textHeight > 74.5 {
+        if textHeight > 60 {
             return textHeight
         }
-        return 74.5
+        return 60
     }
 
     private let fileImageView: UIImageView = {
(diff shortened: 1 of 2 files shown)


deltachat-ios/View/Cell/DocumentGalleryFileCell.swift
```diff
@@ -5,6 +5,14 @@ class DocumentGalleryFileCell: UITableViewCell {
 
     static let reuseIdentifier = "document_gallery_file_cell"
 
+    static var cellHeight: CGFloat {
+        let textHeight = UIFont.preferredFont(forTextStyle: .headline).pointSize + UIFont.preferredFont(forTextStyle: .subheadline).pointSize + 24
+        if textHeight > 74.5 {
+            return textHeight
+        }
+        return 74.5
```
r10s, line 13: why raise the minimum from 60 to 74.5 here?

for normal and small font sizes, this wastes much screen space so that the user does not really benefit from smaller sizes.

so, in normal font size quite fewer content is displayed: before (11.5 rows) vs after (9 rows) :
 

not that i am not suggesting to change the size of the icon, that seems fine in all cases.

r10s: > It's just the same cell height we're having in contact cells and chat list as well, we can reduce the height though.

i mainly came over that, as we are doing some calculations on the row height, but in practice, it is just always the same :)

i think, we should use roughly the old number of row as a reference, and go to slightly smaller/larger heights when font sizes change - i assume, this was also the gist, but at least for iphone13 mini, i end up always at the same row height.

cyBerta: using a simulator you can easily switch between font sizes with command option + or -

The last changes:

Simulator Screen Recording - iPhone 13 - 2022-06-13 at 13 29 26

r10s (approved):

Issue #1322, opened 2021-07-30
implement search in WebView


cyBerta:
Especially when reading long HTML emails a search functionality as implemented in Android is very useful.
My use case: I've subscribed a mailing list and get a summary email containing all messages sent over the mailing list each day. The email has a first section showing the topics and below each message added. Having a search allows to easily jump to the message I'm interested.

Also for the help pages / FAQ, a search feature would be a nice improvement.

r10s:
what is the idea ui-wise? maybe we can also target that together with an in-chat search, things as the up/down arrows, maybe a hit-counter are similar; on android, we also targeted [these two things together](https://github.com/deltachat/deltachat-android/issues/1217).

i filed an additional issue for the in-chat-search: https://github.com/deltachat/deltachat-ios/issues/1323 (@cyBerta if you agree on targeting these issues together, feel free to close #1323, we can track both things in one issue then)

cyBerta:
I think it's fine to keep both issues, they target different areas in our code even though the UX ideas shall be consistent. 
That said, I think for the WebView a scroll-over gesture should be sufficient to open the search. Contrary to the chat, in the WebViews the scrollable content starts from the top, so that scroll-over gesture is easily accessible.

cyBerta:
there's an api for iOS 14+ to implement search in WKWebview without any external libraries or extension like https://github.com/sstahurski/SearchWKWebView.

I would prefer to implement that feature for iOS14+ only.

r10s:
yip, having that for  iOS 14+ only is totally fine.

r10s:
moving issue to [resurrection](https://github.com/orgs/deltachat/projects/19) due to [march2022 cleanup](https://github.com/orgs/deltachat/projects/48).

cyBerta:
Reopening, I would like to keep that on my radar, the lack of search in the webview annoys me too much :)

PR #1589, opened 2022-05-24
search in WebViews

cyBerta:
* [x] search in Help
* [x] search in FullMessageView
* [x] don't allow search in WebxdcViewController
* [x] don't allow search in ConnectivityViewController

The search is only available for iOS 14+, because we can allow preshipped js while rejecting any external js from that version on. We use js to highlight and move to selected search results. 

closes https://github.com/deltachat/deltachat-ios/issues/1322


r10s (approved):
lgtm, great you figured out all this stuff! that looks quite complicated and quite some work!

PR #1592, opened 2022-05-31
easy fix missing chat list updates after sharing

### LINKED ISSUE #1591: chatlist not always updated after sharing
while the chatlist is updated when eg. sharing from gallery -> delta chat, the update is missing when sharing from delta chat -> delta chat.

example:

- go to profile / files of a chat with some webxdc
- long tap a webxdc, select "share to" and share to some other chat

-> the chatlist is not updated.

this may also be buggy for sharing files or other things this way, i tested only for webxdc.

EDIT: also, the message-list is not updated when sharing to the same chat.

cyBerta:
reintroduce refreshInBg to keep Chat list updated after sharing from DC to DC

fixes #1591

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -126,6 +126,7 @@ class ChatListController: UITableViewController {
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         startTimer()
+        refreshInBg() // at least required for sharing from DC -> DC
     }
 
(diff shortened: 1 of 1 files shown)


r10s (changes requested):
this would result in some performance regression and also back ux and flickering, see comment at  https://github.com/deltachat/deltachat-ios/issues/1591

r10s:
i created an alternative at https://github.com/deltachat/deltachat-ios/pull/1596 that does not have the drawbacks.

PR #1593, opened 2022-05-31
implement source code menu entry in webxdc views

### LINKED ISSUE #1581: webxdc: support `source_code_url`
see https://github.com/deltachat/deltachat-core-rust/pull/3314

simple implementation:

- if `source_code_url` is set, in the `WebxdcViewVontroller` in the upper right corner the icon "three horizontal dots in a circle" (as used in the welcome screen) should appear. 
- when tapping, `UIAlertController` with `.safeActionSheet` and the entries "Source Code" and "Cancel" should appear.

in the future, there may be more in this menu, eg. "Full screen" is an obvious idea.

cyBerta:
closes #1581


r10s:
> which one supports it currently?

eg. "Hextris" and some others from https://webxdc.github.io/website/

r10s (approved):
nice!

i added two commits with some nitpicking stuff.

PR #1595, opened 2022-05-31
implement bot command support in chat messages

### LINKED ISSUE #1594: support bot commands
* low hanging fruit and nice improvement for better bot UX

cyBerta:
closes #1594


r10s (approved):
great!

i added a minor commit that adds a space after the command and allows easier adding of command parameters. this is how it is done on android.

nb: android does use the command also when the input field is not empty, however, not sure is intended and if there are really bots that would need that. so, i find this precaution we have here reasonable.

PR #1596, opened 2022-05-31
emit DC_MSGS_CHANGED event after sharing from Delta to Delta

r10s:
alternative to #1592, fixes https://github.com/deltachat/deltachat-ios/issues/1591 for both, message-list and chat-list and without performance and ux regressions.

we are doing this by checking the flag `UserDefaults.hasExtensionAttemptedToSend` as soon as possible after using the share-function.


cyBerta (approved):
nice

PR #1597, opened 2022-06-01
Mailto in webxdc

### LINKED ISSUE #1579: support `mailto:`-linkls in webxdc
tapping a `mailto:`-link in an webxdc should open/create a chat, subject/body preset, similar to as if the link is opened from external or from inside delta chat.

see https://github.com/deltachat/deltachat-core-rust/issues/3312 for some more background.

cyBerta:
closes #1579


r10s (approved):
works as expected.

but i find the code hard to follow meanwhile.

maybe we can cleanup that in a subsequent pr, eg. we could ask at `AppDelegate.application(open url)`, get rid of `askToChatWithMailto` and have only two states left at handleMailto()

also, the "shortcut" `handleMailto(contact)` is used only once and, as a one-liner, not helpful wrt readabilty.

grown code :)

PR #1598, opened 2022-06-01
fix scroll-down btn visibility when opening a chat with many unread messages

cyBerta:
* fix initial calculation of messageInputBars top y-coordinate
* fixes visibility of the scroll-down button when opening a chat from the global search or when opening chats with many unread messages

I haven't created a bug ticket for it, but I consider this a bug fix :)

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -1042,17 +1042,17 @@ class ChatViewController: UITableViewController {
 
         let rectOfCellInTableView = tableView.rectForRow(at: lastIndexPath)
         // convert points to same coordination system
-        let inputBarTopInWindow = messageInputBar.convert(CGPoint(x: 0, y: messageInputBar.bounds.minY), to: window)
+        let inputBarTopInWindow = window.bounds.maxY - (messageInputBar.intrinsicContentSize.height + messageInputBar.keyboardHeight)
         var cellTopInWindow = tableView.convert(CGPoint(x: 0, y: rectOfCellInTableView.minY), to: window)
         cellTopInWindow.y = floor(cellTopInWindow.y)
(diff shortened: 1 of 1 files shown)


r10s:
currently trying that out ...

one note already now:

the scroll-down-button is always visible also when the chat is empty, which is a bit annoying (this bug is not introduced by this pr, but maybe it can be fixed here, at least we should see where the issue comes from :)

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1042,17 +1042,17 @@ class ChatViewController: UITableViewController {
 
         let rectOfCellInTableView = tableView.rectForRow(at: lastIndexPath)
         // convert points to same coordination system
-        let inputBarTopInWindow = messageInputBar.convert(CGPoint(x: 0, y: messageInputBar.bounds.minY), to: window)
+        let inputBarTopInWindow = window.bounds.maxY - (messageInputBar.intrinsicContentSize.height + messageInputBar.keyboardHeight)
```
r10s, line 1045: is it really correct to remove the convert() function here?

further down, we are comparing a convert()'ed value with an unconverted.

just wondering, i am not saying it is wrong :)

cyBerta: it should be fine, all views are converted to the coordination system of the window. the new change also calculates the coordinates based on the window coordination system, however in a different way.

cyBerta:
> the scroll-down-button is always visible also when the chat is empty, which is a bit annoying (this bug is not introduced by this pr, but maybe it can be fixed here, at least we should see where the issue comes from :)

I will add a fix for that tomorrow in this PR.

deltachat-ios/Chat/ChatViewController.swift
r10s: k, then i assume, it is correct :)

thanks for explanations!

cyBerta:
@r10s fixed

r10s (approved):

PR #1599, opened 2022-06-03
don't allow swipe gesture to navigate back in WebxdcViewController

cyBerta:
don't allow swipe to navigate back in WebxdcViewController, because the gesture interferes with interactions in webxdcs 

fixes my highscores in several games


r10s (approved):
nice! 

this also destroyed some highscores for me (excuses ... :)

PR #1602, opened 2022-06-06
update changelog

r10s:
counterpart of https://github.com/deltachat/deltachat-android/pull/2315


r10s:
we probably go directly for a 1.32

PR #1604, opened 2022-06-14
Chatlist multiselect

### LINKED ISSUE #1603: Multi-Select in chat list
bring the same functionality to iOS that Android already has:
* long tap to enable multi-select
* show a panel at the bottom of the screen, similar to the multi-select UI in chats
* offer options to pin, delete, archive and select all chats

cyBerta:
closes #1603


r10s:
this is a huge improvement, great!

i did not look much at the code yet, however, i noticed a not-so-great visual/haptic feedback on long-press:

- when long-tapping a row, (1) the row turns highlighted, which is good, and then (2) back to normal state - while still tapping the row.

- it does not matter where you move your finger now - on releasing, after (2) you always end in multiselect mode but you need to release your finger

it would be nicer if one would go directly to multiselect mode when tapping "long enough", so at (2), at best combined with some vibration/haptic feedback.

compare entering multi-select in the message view, this is much clearer there. same for other messengers (whatsapp, imessage do not have multiselect (really?) but long-tapping is possible)

r10s:
oh, and another thing: the "new chat" icon is gone forever once multiselect is entered a single time :)

cyBerta:
Thanks for the feedback! I will adapt the long-tap behavior tomorrow.

PR #1612, opened 2022-06-21
implement reply privately

### LINKED ISSUE #1609: implement reply privately
cmp. Android

- not in Contact Requests
- only in Groups
- only for received messages

- creates a new 1-1 chat if there's none existing with message sender
- part of the long-press menu of a single message

cyBerta:
closes #1609 

tested on ios12 and ios15


r10s:
function test, did not look at the code much:

- works as expected, nice!

some ui nitpicks:

- ui-wise, i think the menu item should be below "reply" - and if "reply" is not available, it should be the first one - so always above forward.

- the used icon is not clear i think, usually "reply" is a "swiped-left-arrow" - as this one is already used, maybe (1) use no icon, (2) a variant of the icon, maybe outlined or (3) do as in thunderbird and use the typical "double-swiped-arrow-arrow" for "reply" which makes the "swiped-left-arrow" free for "reply privately"

@cyBerta i'd like to browse the asset icons a bit, maybe there is something fitting - but i forgot where i can browse them - can you hive me a hint :)

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -154,11 +154,25 @@ class ChatViewController: UITableViewController {
...
-        let copyItem = ContextMenuProvider.ContextMenuItem(
-        title: String.localized("global_menu_edit_copy_desktop"),
-        imageName: "doc.on.doc",
-        action: #selector(BaseMessageCell.messageCopy),
-        onPerform: { [weak self] indexPath in
+        let dcChat = dcContext.getChat(chatId: chatId)
+        let config = ContextMenuProvider()
+        if #available(iOS 13.0, *), dcChat.canSend {
+            let mainMenu = ContextMenuProvider.ContextMenuItem(submenuitems: [replyItem, forwardItem, infoItem, copyItem, replyPrivatelyItem, deleteItem])
+            config.setMenu([mainMenu, selectMoreItem])
```
r10s, line 161: maybe take the chance to streamline this code - or is there a reason to use `mainMenu` and then anyway add another item on `setMenu`? the other branches are more straight forward,

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -154,11 +154,25 @@ class ChatViewController: UITableViewController {
...
+        } else if dcChat.canSend {
+            config.setMenu([forwardItem, infoItem, copyItem, replyPrivatelyItem, deleteItem, selectMoreItem])
```
r10s, line 163: seems ios<13 does not have the "reply" function - what was the reason for that? and is it okay to allow replyPrivately? you tested on ios12, so probably fine - just asking :)

cyBerta: I think we decided against it back in the days, because the horizontal space for items are limited. you need to press an arrow icn to see additional options, and repply was easily accessible with the swipe gesture

deltachat-ios/Chat/ChatViewController.swift
cyBerta: yes, we do this on purpose to get the space between the upper menu items and selectMore

cyBerta:
> is suggest to use `arrowshape.turn.up.left.fill` for "reply"

what shall we do with the icon for swipe-to-reply? That feature is available on pre-iOS13 and SF Symbols are not supported there. I we can keep the old reply icon for pre iOS13 and use `arrowshape.turn.up.left.fill` for swipe to reply, to have some visual consistency.

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1673,14 +1709,28 @@ class ChatViewController: UITableViewController {
...
+        if #available(iOS 13.0, *) {
+            return
+        }
+
+        if isHidden {
+            UIMenuController.shared.menuItems = nil
+        } else if isGroupChat && !message.isFromCurrentSender {
+            UIMenuController.shared.menuItems = contextMenu.menuItems
+        } else {
+            // filter out replyPrivatelyItem
```
r10s, line 1722: this comment seems superfluous as it just says repeats what is visible in the code

r10s (approved):

PR #1614, opened 2022-06-21
allow selection of info messages

### LINKED ISSUE #1610: never select info messages
Currently it's possible to select info messages in a chat, if you long tap on a normal message sent by a user to enter the multi-select-mode. While the selection is not shown in the UI, an info message can internally still get selected by tapping on it in multi-select mode. In that case the forwarding and deletion options aren't disabled anymore, if even if apparantly no message is selected.

cyBerta:
fixes #1610 by allowing selection of info-messages

* you still cannot initially long-tap on a info-message to open the menu to select info / copy / delete / select more like on android. In order to implement this https://github.com/deltachat/deltachat-ios/pull/1612 needs to be merged in.

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -1819,9 +1819,17 @@ class ChatViewController: UITableViewController {
     }
 
     func handleEditingBar() {
-        if let rows = tableView.indexPathsForSelectedRows,
-           !rows.isEmpty {
+        if let indexPaths = tableView.indexPathsForSelectedRows,
+           !indexPaths.isEmpty {
+            var hasInfoMessageSelected = false
+            for indexPath in indexPaths {
+                if dcContext.getMessage(id: messageIds[indexPath.row]).isInfo {
+                    hasInfoMessageSelected = true
+                    break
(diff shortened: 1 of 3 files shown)


r10s (changes requested):
i think, info-messages should be multi-selectable as other message so that a user can delete them together with other messages 

this is how it works on android, desktop does not have multi-select but allows deletion of a single info-message.

for forwarding: android hides the option if one of the message is an info-message, desktop allows forwarding. as info-messags are not forwarded as such (only text is forwarded), i think, the latter is also fine - and probably easier to do.

(of course daymakers and n-new-message markers cannot be deleted)

cyBerta:
ok, I'll allow mutli-selecting info messages then, if they don't have the message  ID of  a daymarker or new message marker.

r10s:
as core allows forwarding info-messages, i added a test to make sure, they're not forwarded as info-messages but only as text: https://github.com/deltachat/deltachat-core-rust/pull/3446 there is also some reasoning why forwarding info-messages may be not-so-good, however :) let's discuss there and leave this open until it is resolved there :)

r10s (approved):

PR #1620, opened 2022-06-22
[WIP] keep search bar accessory view in webviews

### LINKED ISSUE #1617: keep bottom search panel in WebView
If the keyboard was dismissed during an active search in a full-message view, the bottom search panel (which is attached to the keyboard), disappears, too. It would be nicer if the bottom panel stays visible until the search was cancelled, even if the keyboard was dismissed by swiping it down.

cyBerta:
closes #1617


cyBerta:
closing for now, since I didn't follow up with this open PR yet. Unfortunately it's harder to archieve with the given architecture and requires bigger refactorings. (It seems we cannot rely on the viewsController accessoryView to avoid the bottom bar disappears.)

PR #1623, opened 2022-06-24
show number of selected chats in title

r10s:
this bit is missing somehow, having the wording "selected" in the title also underlines that one is in "select" mode.

i was also thinking about saying "N Chats Selected" in the title bar, however, this will get easily to wide in translations, so, i decided for the short variation. also, that way the string can be reused in similar views ;)

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -389,6 +389,8 @@ class ChatListController: UITableViewController {
             editingBar.showUnpinning = viewModel.hasOnlyPinnedChatsSelected(in: tableView.indexPathsForSelectedRows)
             if tableView.indexPathsForSelectedRows == nil {
                 setLongTapEditing(false)
+            } else {
+                updateMultiSelectSelectionCount()
             }
         }
     }
@@ -400,6 +402,7 @@ class ChatListController: UITableViewController {
         }
         if tableView.isEditing {
             editingBar.showUnpinning = viewModel.hasOnlyPinnedChatsSelected(in: tableView.indexPathsForSelectedRows)
+            updateMultiSelectSelectionCount()
             return
         }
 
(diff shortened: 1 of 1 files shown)


cyBerta:
nice idea

cyBerta:
in order to make sure, we handle the title correctly in all possible states (forwarding, archive, multi-select, normal -> connection states)  I would prefer if the handling would happen within updateTitle(). It is a little bit easier to follow, and a little bit less likely to introduce regressions later on.

cyBerta (changes requested):
in order to ensure we handle the title correctly in all possible states (forwarding, archive, multi-select, normal -> connection states) I would prefer if the handling would happen within updateTitle(). It is a little bit easier to follow, and a little bit less likely to introduce regressions later on.

r10s:
> in order to ensure we handle the title correctly in all possible states (forwarding, archive, multi-select, normal -> connection states) I would prefer if the handling would happen within updateTitle(). It is a little bit easier to follow, and a little bit less likely to introduce regressions later on.

i would also prefer that, however, i hestiated as it was unclear to me if `setLeftBarButton()` etc. are idempotent when applying the same object over and over (which would happen if we leave the existing logic as is).

having a closer look at https://developer.apple.com/documentation/uikit/uinavigationitem/1624954-setleftbarbutton this discusses only applying the same object to different views, however, one can probably assume that applying the same object to the same view is okay as well. 

@cyBerta do we have experiences with that in otherview? or what is your stance on this?

cyBerta:
it should do nothing, e.g. regarding animations, if the same object is applied to the the same UIBarButton.

r10s:
@cyBerta i added a commit, now all title changes are done in updateTitle(). the function is still a bit confusing wrt state changes etc, however, if we want to clean up further there, that should go to another pr (i tried, but that would make review of this pr much harder :)

cyBerta:
great, thanks for this nice improvement

PR #1629, opened 2022-07-04
immediately update chatlist on swipe actions

### LINKED ISSUE #1621: speed regression in pin/archive/delete chat
on swipe-left-chat in chatlist to pin/archive/delete, these action were done nearly instantly in the past, eg. at 82a8b2ca9ce01ad1709967df47532a5475a14a01 .

current master with multi-select (see #1604, #1613, #1618, #1619) is way slower, there is a delay of nearly a second, maybe because some animation is played, not sure.

for testing, compare eg. `git checkout 82a8b2ca9ce01ad1709967df47532a5475a14a01` with `git checkout master`.

cyBerta:
fixes #1621 

the update is immediately happening now, so no swipe-back animation is shown if you tap on pin or archive. Delete is left as it was.

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -445,13 +445,15 @@ class ChatListController: UITableViewController {
         let archiveActionTitle: String = String.localized(archived ? "unarchive" : "archive")
 
         let archiveAction = UITableViewRowAction(style: .destructive, title: archiveActionTitle) { [weak self] _, _ in
-            self?.viewModel?.archiveChatToggle(chatId: chatId)
+            self?.viewModel?.archiveChatToggle(chatId: chatId, notifyListener: true)
+            self?.setEditing(false, animated: true)
         }
         archiveAction.backgroundColor = UIColor.lightGray
 
         let pinned = chat.visibility==DC_CHAT_VISIBILITY_PINNED
         let pinAction = UITableViewRowAction(style: .destructive, title: String.localized(pinned ? "unpin" : "pin")) { [weak self] _, _ in
-            self?.viewModel?.pinChatToggle(chatId: chat.id)
+            self?.viewModel?.pinChatToggle(chatId: chat.id, notifyListerner: true)
+            self?.setEditing(false, animated: true)
         }
         pinAction.backgroundColor = UIColor.systemGreen
 
@@ -465,6 +467,7 @@ class ChatListController: UITableViewController {
 
     override func setEditing(_ editing: Bool, animated: Bool) {
         super.setEditing(editing, animated: animated)
+        tableView.setEditing(editing, animated: animated)
         viewModel?.setEditing(editing)
     }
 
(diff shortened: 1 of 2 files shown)


Simon-Laux (approved):
don't quite understand the swift code completly yet, but works for me

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -445,13 +445,15 @@ class ChatListController: UITableViewController {
         let archiveActionTitle: String = String.localized(archived ? "unarchive" : "archive")
 
         let archiveAction = UITableViewRowAction(style: .destructive, title: archiveActionTitle) { [weak self] _, _ in
-            self?.viewModel?.archiveChatToggle(chatId: chatId)
+            self?.viewModel?.archiveChatToggle(chatId: chatId, notifyListener: true)
+            self?.setEditing(false, animated: true)
```
r10s, line 449: i do not really get why `setEditing(false)` needs to be called here - this "edit" mode refers to the new multi-select seems unrelated to the "old swipe" action.

is it for a quick redraw?

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -465,6 +467,7 @@ class ChatListController: UITableViewController {
 
     override func setEditing(_ editing: Bool, animated: Bool) {
         super.setEditing(editing, animated: animated)
+        tableView.setEditing(editing, animated: animated)
         viewModel?.setEditing(editing)
     }
```
r10s, line 472: we set the edit-mode for _three_ objects here.

why do we set the edit-mode only for _two_ objects a few lines below in `setLongTapEditing` - but cancel again _three_ objects.

this looks like some imbalance.

sorry for nitpicking, i just try to understand :)

deltachat-ios/Controller/ChatListController.swift
cyBerta: exactly, it is for a quick redraw. otherwise setEditing is set to false by the viewcontroller itself (internally) once the swipe back animation has finished.

r10s: so, the "edit mode" is not only for multi-select but also for the swipe actions?

i am still wondering why before multi-select the swipe-to-pin and swipe-to-archive was faster even with the "closing animation".

deltachat-ios/ViewModel/ChatListViewModel.swift
```diff
@@ -228,25 +228,25 @@ class ChatListViewModel: NSObject {
         INInteraction.delete(with: ["\(dcContext.id).\(chatId)"])
     }
 
-    func archiveChatToggle(chatId: Int) {
+    func archiveChatToggle(chatId: Int, notifyListener: Bool = false) {
```
r10s, line 231: unless it is on purpose, the `notifyListener` parameter can be dropped, it is always true. same for `pinChatToggle`.

deltachat-ios/Controller/ChatListController.swift
r10s: thanks! i was just seeing 2 there and 3 there without understanding details :)

deltachat-ios/Controller/ChatListController.swift
cyBerta: because always all list updates have been done immediately. With multi-select we postpone list-updates until editing is done. This way we avoid that the selected items change (we use the tableViewController own mecahnism to track the selected cells - not the selected messages).

Issue #1504, opened 2022-03-20
imported backups / account may be forgotten


r10s:
a once imported backup, that works for some days or hours may be forgotten, resulting in the "welcome" screen shown again.

not sure what the reason is here, initially reported via dm, however, ~~i could meanwhile also reproduce that~~ it also happens to me once.

i've seen that only for encrypted databases yet, however, not sure if that happens also to normal ones.

as similar reports are not present for android/desktop, maybe that is related to ios, maybe getting the secret from the system does not always work or so.

@cyBerta i could debug a bit: somehow the logic seems to think, there is no configured account. the account presented to the "welcome view controller" was actually empty, and there was no other account (accountIds == 1). to be honest, i played around with adding/removing accounts first, adding the same account one time encrypted and another time unencrypted. maybe that mixed up some things - however, that was a bit before the "welcome view controller" was displayed suddenly.

plaintextdesign:
chiming in here to confirm this issue. the original backup was *not* encrypted but it was imported via "create encrypted account" and then "import backup". since then it has been (normally) imported via "log in to your server" and then "import backup". but that did not stop this behavior.

side note: encrypted accounts should be created/accessed differently!

r10s:
> but it was imported via "create encrypted account" and then "import backup"

this is what i meant, it is currently the only way to create an encrypted account

> side note: encrypted accounts should be created/accessed differently!

yip. the current, experimental, implementation is only a step. it is intentionally a bit hidden and flagged as "experimental" to get aware of bugs as this one and to fix them before using things on a wider user base.

might be that encrypted accounts just become the default.

cyBerta:
thanks for debugging @r10s 

> i could debug a bit: somehow the logic seems to think, there is no configured account. the account presented to the "welcome view controller" was actually empty, and there was no other account (accountIds == 1).

Could you confirm that account with the id 1 is the empty one?

cyBerta:
I think I managed to reproduce the issue. For some reason an empty context is selected and shown. In my case the cancel button on the upper left of the AccountSetupController was visible and clickable. After clicking on it I went back to the configured account. 

I will try to figure out how to fix that.

r10s:
> I think I managed to reproduce the issue

can you remember how you reproduced that? i am totally failing in getting the issue again :/

r10s:
maybe related: https://github.com/deltachat/deltachat-ios/issues/1541

cyBerta:
[...]

PR #1636, opened 2022-07-05
implement an error screen for closed account

Simon-Laux:
This pr includes:
- display closed accounts as `(closed account)` instead of falsely displaying them as `(unconfigured)`
- guard against deleting the account if it is closed in the welcome screen
- an error screen instead of the welcome screen when we encounter the problem
    - there is also an experimental "try fix without restart" button in the error screen, that worked in my testing

solves the "account may be forgotten" part of #1504, but the bug of a closed encrypted account is still there

### DIFF
--- a/deltachat-ios/Controller/ClosedAccountErrorViewController.swift
+++ b/deltachat-ios/Controller/ClosedAccountErrorViewController.swift
@@ -33,10 +33,11 @@ class ClosedAccountErrorViewController: UIViewController {
     private lazy var titleLabel: UILabel = {
         let label = UILabel()
         label.text = "Error: Account is closed"
-        label.textColor = DcColors.grayTextColor
+        label.textColor = .red
         label.textAlignment = .center
         label.numberOfLines = 1
         label.font = UIFont.systemFont(ofSize: 24, weight: .bold)
+        label.translatesAutoresizingMaskIntoConstraints = false
         return label
     }()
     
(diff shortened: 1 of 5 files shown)


deltachat-ios/Controller/ClosedAccountErrorViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+import DcCore
+
+class ClosedAccountErrorView: UIView {
+    private lazy var titleLabel: UILabel = {
+        let label = UILabel()
+        label.text = "Error: Account is closed"
+        label.textColor = DcColors.grayTextColor
+        label.textAlignment = .center
+        label.numberOfLines = 1
+        label.font = UIFont.systemFont(ofSize: 24, weight: .bold)
```
cyBerta, line 20: we are setting the constraints manually in the code, therefore all UIElements need to have the flag
`translatesAutoresizingMaskIntoConstraints` set to false

deltachat-ios/Controller/ClosedAccountErrorViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+    }()
+    
+    
+    public init () {
+        super.init(frame: .infinite)
+        
+        addSubview(titleLabel)
+        addSubview(errorDescription1)
+        addSubview(errorDescription2)
+
```
cyBerta, line 51: here you need to add your constraints - how do you want to arrange your UIElements?

deltachat-ios/Controller/ClosedAccountErrorViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+        label.textAlignment = .center
+        label.numberOfLines = 1
+        label.font = UIFont.systemFont(ofSize: 24, weight: .bold)
+        return label
+    }()
+    
+    private func setupSubviews() {
+        view.addSubview(closedAccountErrorView)
+        view.addSubview(titleLabel)
+    }
```
cyBerta, line 98: here as well

cyBerta:
I think this PR can be closed, the check https://github.com/deltachat/deltachat-ios/pull/1636/files#diff-f06dfba6a1bb192e6de55c8b3205d07ad69e82b47610fedfca051bf2acead007R239 has been implemented in the meanwhile, and the root cause (too early key chain access) was fixed, too.

PR #1638, opened 2022-07-06
ensure encrypted accounts are always opened when resuming from background

cyBerta:
another attempt to fix https://github.com/deltachat/deltachat-ios/issues/1504

* We change the access policy here, the app is allowed to access the keychain after the device was unlocked for the first time. 
* This PR also ensures no accounts gets deleted if they are in a closed state and the user presses cancel in the Welcome screen. 
* it also improves error logging for better debugging in the future

### DIFF
--- a/DcCore/DcCore/Helper/KeychainManager.swift
+++ b/DcCore/DcCore/Helper/KeychainManager.swift
@@ -46,6 +46,7 @@ public class KeychainManager {
           kSecAttrAccount as String: "\(id)",
           kSecClass: kSecClassGenericPassword,
           kSecAttrAccessGroup as String: KcM.sharedKeychainGroup as AnyObject,
+          kSecAttrAccessible: kSecAttrAccessibleAfterFirstUnlock,
         ] as CFDictionary
 
         let status = SecItemAdd(keychainItemQuery, nil)
--- a/DcCore/DcCore/Helper/KeychainManager.swift
+++ b/DcCore/DcCore/Helper/KeychainManager.swift
@@ -15,6 +15,11 @@ public class KeychainManager {
     private static let sharedKeychainGroup = "\(KcM.teamId).group.chat.delta.ios"
 
     public static func getAccountSecret(accountID: Int) throws -> String {
+        if let buildVersion = Bundle.main.infoDictionary?["CFBundleVersion"] as? Int,
+           buildVersion <= 74 {
+            upgradeAccountSecret(id: accountID)
+        }
+
         do {
             return try queryAccountSecret(id: accountID)
         } catch KeychainError.noPassword {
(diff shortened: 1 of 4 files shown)


r10s:
still WIP, is see :) anyway, can you write a few more details how that may fix the issue? just a precaution? (which would be fine!)

in general, imu, `didFinishLaunchingWithOptions` was always called before `didReceiveRemoteNotification`/`performFetchWithCompletionHandler` - if the account is not open at that point that'd be good to figure out why.

cyBerta:
the root cause _might_ be related to a new ios 15 feature called prewarming [https://developer.apple.com/documentation/uikit/app_and_environment/responding_to_the_launch_of_your_app/about_the_app_launch_sequence#3894431](https://github.com/deltachat/deltachat-ios/pull/1638#). 

It seems we cannot reliably call the keychain in didFinishLaunchingWithOptions, because the device might still be locked. Prewarming can happen minutes before the app is actually started by the user. 

Another interesting read on that issue: [https://sourcediving.com/solving-mysterious-logout-issues-on-ios-15-8b818c089466](https://github.com/deltachat/deltachat-ios/pull/1638#)

cyBerta:
also the default access policy seems to be:  `kSecAttrAccessibleWhenUnlocked`

currently we don't explicitely define the access control. According to https://developer.apple.com/documentation/security/keychain_services/keychain_items/restricting_keychain_item_accessibility we probably want to use [ksecattraccessibleafterfirstunlock](https://developer.apple.com/documentation/security/ksecattraccessibleafterfirstunlock) instead.

deltachat-ios/AppDelegate.swift
```diff
@@ -207,6 +207,14 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         }
     }
 
+    func applicationProtectedDataDidBecomeAvailable(_ application: UIApplication) {
+        logger.info("➡️ applicationProtectedDataDidBecomeAvailable")
+    }
+
+    func applicationProtectedDataWillBecomeUnavailable(_ application: UIApplication) {
+        logger.info("➡️ applicationProtectedDataWillBecomeUnavailable")
```
r10s, line 215:
```suggestion
        logger.info("⬅️ applicationProtectedDataWillBecomeUnavailable")
```

DcCore/DcCore/Helper/KeychainManager.swift
```diff
@@ -15,6 +15,11 @@ public class KeychainManager {
     private static let sharedKeychainGroup = "\(KcM.teamId).group.chat.delta.ios"
 
     public static func getAccountSecret(accountID: Int) throws -> String {
+        if let sharedUserDefaults = UserDefaults.shared,
+            !sharedUserDefaults.bool(forKey: "\(UserDefaults.upgradedKeychainEntry)\(accountID)") {
+            upgradeAccountSecret(id: accountID)
```
r10s, line 20: i would skip this upgrade.

- it adds complexity, i like to have less keychain code :)
- most accounts that are theoretically upgradeable are probably broken anyway meanwhile (experimental :)
- the code should be able to work with `kSecAttrAccessibleWhenUnlocked` as well as with `kSecAttrAccessibleAfterFirstUnlocked` anyway (difference is only background fetch from lock-screen when the app is not frozen but terminated)

cyBerta: I removed the migration as we discussed before

Issue #1576, opened 2022-05-11
allow scanning an `DCACCOUNT:` qr-code from the system camera and by the internal scanner


r10s:
- currently, if you  scan a `DCACCOUNT:` qr-code using the system camera, nothing happens.
it would be nice if instead Delta Chat is opened and the user is asked "Create new e-mail address on 'provider.org' and log in there? bla bla ..." (we can use `qraccount_ask_create_and_login_another` as it is used on android already).

- similar when scanning a `DCACCOUNT:` qr-code from within Delta Chat's qr-code-button in the lower left corner of the chatlist - there it says "The scanned QR code is for setting up a new account [...]" - it could also show the same question and forward to the welcome screen.  

(nb: scanning a qr code from the camera app and opening delta chat already works for secure-join and join-group qr-codes(`OPENPGP4FRP:`), so we probably can copy that pattern :)

PR #1640, opened 2022-07-06
register for `dcaccount:` scheme

Simon-Laux:
tested with system camera, tap on link from browser (webxdc.org qr code), new welcome view qr code reader and global QR-View qr code reader

closes https://github.com/deltachat/deltachat-ios/issues/1576

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * fix accessibility hint in multi-select chat list title
 * update translations, revise english source
 * update to core90
+* register on system for `dcaccount:` scheme
 
 
 ## v1.31.0 Testflight
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,6 @@
 * fix accessibility hint in multi-select chat list title
 * update translations, revise english source
 * update to core90
-* register on system for `dcaccount:` scheme
-
 
 ## v1.31.0 Testflight
 2022-07
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -197,7 +197,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
                            .replacingOccurrences(of: "openpgp4fpr", with: "OPENPGP4FPR", options: .literal, range: nil)
                            .replacingOccurrences(of: "%23", with: "#", options: .literal, range: nil)
 
-            self.appCoordinator.handleQRCode(urlString)
+            self.appCoordinator.handleQRCodeForCurrentAccount(urlString)
             return true
         case "mailto":
             return self.appCoordinator.handleMailtoURL(url)
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -174,7 +174,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         appFullyInitialized = true
     }
 
-    public var accountCreationQrCode: String = ""
     // `open` is called when an url should be opened by Delta Chat.
     // we currently use that for handling oauth2 and for handing openpgp4fpr.
     //
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -188,7 +188,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
         switch url.scheme?.lowercased() {
         case "dcaccount":
-            self.reloadDcContext(accountCode: url.absoluteString)
+            appCoordinator.handleQRCode(url.absoluteString)
             return true
         case "openpgp4fpr":
             // Hack to format url properly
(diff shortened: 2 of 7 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -173,6 +174,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         // }
 
         switch url.scheme?.lowercased() {
+        case "dcaccount":
+            _ = self.dcAccounts.add()
```
cyBerta, line 178: Why don't you implement it similar to openpgp4fpr or handleMailtoURL, delegating it to appCoordinator?

deltachat-ios/Controller/WelcomeViewController.swift
```diff
@@ -90,6 +90,16 @@ class WelcomeViewController: UIViewController, ProgressAlertHandler {
         }
         navigationItem.rightBarButtonItem = moreButton
     }
+    
+    override func viewDidAppear(_ animated: Bool) {
+        if let appDelegate = UIApplication.shared.delegate as? AppDelegate {
+            if !appDelegate.accountCreationQrCode.isEmpty {
```
cyBerta, line 96: it would be nice if we could avoid keeping the state (`accountCreationQrCode`) directly in appDelegate.

cyBerta:
thanks for pushing this forward! I will have a more in depth look into it tomorrow.

deltachat-ios/Controller/WelcomeViewController.swift
Simon-Laux: tried that first with passing it though the functions but it looked more annoying/complicated/too-verbose.
Is there another alternative that I don't know of yet?

deltachat-ios/AppDelegate.swift
Simon-Laux: just tried to make it work at all for now :) , if I were to make it nice I would show the confirmation dialog first regardless of screen and don't even create the account if the user aborts/cancels. ;)

cyBerta:
I've discussed how to proceed with the dcaccount url scheme handling with @r10s and we thought it might be a good idea to fix some more issues when we're already at it:
- currently there are 2 different methods handling qr codes, in the WelcomeView for dcaccounts and in the AppCoordinator for all other qr codes. Instead  we would like to have one central method reponsible for QR-code handling in the AppCoordinator
- as a result dcaccount qr codes can be read currently onlyfrom the WelcomeViewController but not from DC's general QR code reader. that should be improved, too.

Instead of having an extra handling here: https://github.com/deltachat/deltachat-ios/pull/1640/files#diff-fc4377fbb4c60bfc9a11f53fd448de91bbf34787141509ddb85242515fea9e23R178 we would just pass the url to `self.appCoordinator.handleQRCode(urlString)` just as it is done for `openpgp4fpr`.

That requires a little bit more of refactoring but it's worth the effort.

r10s:
@cyBerta you go for it, right?

Simon-Laux:
I agree with the plan, don't have time to do it myself soon so I would also be glad if @cyBerta takes it from here.

deltachat-ios/Controller/WelcomeViewController.swift
```diff
@@ -316,7 +322,9 @@ extension WelcomeViewController: QrCodeReaderDelegate {
     }
 
     private func confirmAccountCreationAlert(accountDomain domain: String, qrCode: String) {
-        let title = String.localizedStringWithFormat(String.localized("qraccount_ask_create_and_login"), domain)
+        let title = String.localizedStringWithFormat(
+            String.localized(dcAccounts.getAll().count > 1 ? "qraccount_ask_create_and_login_another" : "qraccount_ask_create_and_login"),
```
cyBerta, line 326: @r10s I've found in the resources that there's an extra string for the case there's already an account configured.We didn't use that string before.

deltachat-ios/Controller/QrPageController.swift
```diff
@@ -184,11 +186,11 @@ extension QrPageController: QrCodeReaderDelegate {
             let nameAndAddress = dcContext.getContact(id: qrParsed.id).nameNAddr
             let msg = String.localizedStringWithFormat(String.localized(state==DC_QR_ADDR ? "ask_start_chat_with" : "qrshow_x_verified"), nameAndAddress)
             let alert = UIAlertController(title: msg, message: nil, preferredStyle: .alert)
+            alert.addAction(UIAlertAction(title: String.localized("cancel"), style: .default, handler: nil))
             alert.addAction(UIAlertAction(title: String.localized("start_chat"), style: .default, handler: { _ in
                 let chatId = self.dcContext.createChatByContactId(contactId: qrParsed.id)
                 self.showChat(chatId: chatId)
             }))
-            alert.addAction(UIAlertAction(title: String.localized("cancel"), style: .default, handler: nil))
```
cyBerta, line 191: I unified the buttons of qr code related alerts: left cancel, right action

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * fix accessibility hint in multi-select chat list title
 * update translations, revise english source
 * update to core90
+* register on system for `dcaccount:` scheme
```
r10s, line 16: this is wrong here, maybe just drop the entry, we write the changelog on releases currently.

r10s (approved):
code looks good as far as i can tell, did not test much yet.

the CHANGLOG entry should be removed.

PR #1643, opened 2022-07-07
[WIP] feat: add new backup transfer mechanism

dignifiedquire:
Based on https://github.com/deltachat/deltachat-core-rust/pull/3489, see that for details

This barely works, it fails on first start, only on switching back and forth to the qr-code will it work, no progress indication.


r10s:
superseeded by https://github.com/deltachat/deltachat-ios/pull/1649 , which aim to work a bit better and has some ui :)

PR #1650, opened 2022-07-15
delete account encryption keys from keychain if account was deleted

### LINKED ISSUE #1645: Delete encryption keys on account deletion
Currently account encryption keys are not deleted in the keychain if the account was deleted. That should be fixed.

cyBerta:
closes #1645

### DIFF
--- a/DcCore/DcCore/Helper/KeychainManager.swift
+++ b/DcCore/DcCore/Helper/KeychainManager.swift
@@ -35,6 +35,20 @@ public class KeychainManager {
         return status == errSecSuccess || status == errSecItemNotFound
     }
 
+    /**
+     * Tries to deletes the secret for a given account id, errors will be ignored
+     */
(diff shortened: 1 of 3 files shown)


DcCore/DcCore/Helper/KeychainManager.swift
```diff
@@ -35,6 +35,20 @@ public class KeychainManager {
         return status == errSecSuccess || status == errSecItemNotFound
     }
 
+    /**
+     * Tries to deletes the secret for a given account id, errors will be ignored
+     */
+    public static func deleteAccountSecret(id: Int) {
+        let query = [
+          kSecValueData: createRandomPassword().data(using: .utf8)!,
```
cyBerta, line 43: proably that should be removed. I'll double check that

r10s (approved):
i think, the password needs to be removed from query, otherwise lgtm.

r10s (approved):

PR #1655, opened 2022-07-26
finish app initialization as soon as the access to keychain is allowed

### LINKED ISSUE #1648: Refactoring AppDelegate
- as a follow-up of #1638 we want make sure that access to the keychain and thus opening DcContext is handled correctly despite new features like [prewarming](https://developer.apple.com/documentation/uikit/app_and_environment/responding_to_the_launch_of_your_app/about_the_app_launch_sequence#3894431). 
- Appearantly it is not guaranteed that accessing the keychain in `didFinishLaunchingWithOptions` succeeeds. Therefore we want to encapsulate much of the functionality in `didFinishLaunchingWithOptions` in a separate method that can be called idempotentaly to finish the initial app state setup either in `didFinishLaunchingWithOptions` or in [applicationProtectedDataDidBecomeAvailable](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623044-applicationprotecteddatadidbecom?language=objc).

cyBerta:
closes #1648

* it's not calling the new method `finishAppInitialization` idempotently, but checks in `applicationProtectedDataDidBecomeAvailable` if calling the method is really necessary

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -23,6 +23,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     var window: UIWindow?
     var notifyToken: String?
     var applicationInForeground: Bool = false
+    private var launchOptions: [UIApplication.LaunchOptionsKey: Any]?
 
     // purpose of `bgIoTimestamp` is to block rapidly subsequent calls to remote- or local-wakeups:
     //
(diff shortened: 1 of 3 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -49,13 +51,20 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     // when the app wakes up from "suspended" state
     // (app is in memory in the background but no code is executed, IO stopped)
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+        pendingAppInitialization = true;
```
r10s, line 54:
```suggestion
        pendingAppInitialization = true
```

superfluous semicolon, also, there seems to be no need to set it here.

deltachat-ios/AppDelegate.swift
```diff
@@ -49,13 +51,20 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
+        logger.info("➡️ didFinishLaunchingWithOptions")
```
r10s, line 55: that needs to go after `logger.addDestination(console)` - otherwise, logging has not much effect. 

i'd also move the `SDImage` initialisation part a bit down again so that issues there will be logged.

deltachat-ios/AppDelegate.swift
```diff
@@ -23,6 +23,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     var window: UIWindow?
     var notifyToken: String?
     var applicationInForeground: Bool = false
+    private var launchOptions: [UIApplication.LaunchOptionsKey: Any]?
+    private var pendingAppInitialization = true;
```
r10s, line 27:
```suggestion
    private var pendingAppInitialization = true
```

semicolon not needed.

i would also prefer a more concrete naming for this boolean, sth. as `appFullyInitialized` that is set to `false` here and set to `true` at the end of `finishAppInitialization`

deltachat-ios/AppDelegate.swift
```diff
@@ -64,6 +73,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         dcAccounts.openDatabase()
         migrateToDcAccounts()
 
+        self.launchOptions = launchOptions
+        finishAppInitialization()
+        return true
+    }
+
+    // finishes the app initialization which depends on the successful access to the keychain
+    func finishAppInitialization() {
```
r10s, line 82: naming suggestion: `continueDidFinishLaunchingWithOptions`, that's what it is - and this is also nicely grep'able.

r10s (approved):
some minor suggestions, in general, that should already do the job.

one question: is it guaranteed that `didFinishLaunchingWithOptions` and `applicationProtectedDataDidBecomeAvailable` is called from the main thread so that races are impossible? i am pretty sure, this is the case, but can you please double check? thanks!

deltachat-ios/AppDelegate.swift
cyBerta: java routine....

cyBerta:
From what I read the delegate methods are called from the main thread, yes.

PR #1656, opened 2022-07-26
Encrypted Account improvements

### LINKED ISSUE #1644: Encrypted Account improvements
currently we immediately switch to the login view after tapping on create encrypted account.

Instead we should remain on the welcome screen and indicate the encrypted account in the title. 

This way we improve the UX for users who want to import backups into encrypted accounts or create an encrypted account by using a qr code.

cyBerta:
don't open the AccountSetupController after switching to an encrypted account

@r10s shall we localize the strings? the whole feature is not yet localized, but since we're improving this feature step by step, and since we don't know yet if and when this feature will become the default, we could also start translating the required strings.

closes #1644

### DIFF
--- a/deltachat-ios/Controller/WelcomeViewController.swift
+++ b/deltachat-ios/Controller/WelcomeViewController.swift
@@ -213,7 +213,7 @@ class WelcomeViewController: UIViewController, ProgressAlertHandler {
                 logger.error("Failed to open account database for account \(dcContext.id)")
                 return
             }
-            showAccountSetupController()
+            self.navigationItem.title = "Add encrypted account"
         } catch KeychainError.unhandledError(let message, let status) {
             logger.error("Keychain error. Failed to create encrypted account. \(message). Error status: \(status)")
         } catch {


r10s:
>  shall we localize the strings? the whole feature is not yet localized, but since we're improving this feature step by step, and since we don't know yet if and when this feature will become the default, we could also start translating the required strings.

i'd say this is not needed currently, it is all experimental, so it seems okay to have this prototyped in english only. during experimental, strings are changing more often as usual - we can spare the translators and us the additional effort.

deltachat-ios/Controller/WelcomeViewController.swift
```diff
@@ -213,7 +213,7 @@ class WelcomeViewController: UIViewController, ProgressAlertHandler {
                 logger.error("Failed to open account database for account \(dcContext.id)")
                 return
             }
-            showAccountSetupController()
+            self.navigationItem.title = "Add encrypted account"
```
r10s, line 216: that looks easy :)

r10s (approved):

PR #1659, opened 2022-07-28
chat view multi-select improvements

### LINKED ISSUE #1632: add copy to multi-select in chat
* the option should be added to the bottom edit panel
* on tap on the copy icon the text of all selected messages should be copied to the clipboard
* the cancel button should be moved to the upper left corner

cyBerta:
add copy to multi-select editing bar and move cancel button to the ViewController's upper left corner

closes #1632

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -153,6 +153,13 @@ class ChatViewController: UITableViewController {
         return UIBarButtonItem(customView: initialsBadge)
     }()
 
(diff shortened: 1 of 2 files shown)


Simon-Laux (approved):

r10s (approved):
nice - i did not look at the code but just testing it :)

one thing: we should show in the title "N Selected" as for the chat list (chat title and subtitle would be hidden then, the icon can stay). this would make the selection mode much more outstanding and viewing the number of selected items is also a good thing. however, that can also go to a subsequent pr/issue as this one is already reviewed :)

PR #1663, opened 2022-08-02
Chat multi-select title

cyBerta:
follow-up of https://github.com/deltachat/deltachat-ios/pull/1659

* also reduces the calls to core to request dcChat
* dcChat will be initlialized lazily and updated on `DC_EVENT_CHAT_MODIFIED`
* chat title will be initialized lazily too and reused.


r10s (approved):
nice! also the speed improvements!

PR #1664, opened 2022-08-02
implement resend message feature

### LINKED ISSUE #1607: implement message resent
cmp. Android https://github.com/deltachat/deltachat-android/pull/2278

only for sent messages for now

- cancel button to the top left, EDIT: done
- title: x selected, EDIT: done
- bottom resend

cyBerta:
closes #1607 

ftr: a fix for the copy icon enabled state has been implemented in #1663


r10s (approved):
works great!

ux-wise, i think the disabled "..." button is not so nice. in general disabled things raise the questions _why_ this is disabled. we could either hide the button - or move also "Copy" to "..." so that the button is just always there. but both should not block merging this pr :)

PR #1665, opened 2022-08-03
Webxdc shortcuts

### LINKED ISSUE #1660: create webxdc shortcuts to home screen
We want to figure out how we can archieve webxdc shortcuts, research is required in first place. Useful findings will be reported here.

Some pointers:
* compare https://github.com/deltachat/deltachat-android/pull/2353
* https://swiftrocks.com/adding-deeplink-shortcuts-to-the-ios-home-screen.html
* https://medium.com/wolox/ios-deep-linking-url-scheme-vs-universal-links-50abd3802f97
* https://medium.com/wolox/ios-deep-linking-url-scheme-vs-universal-links-50abd3802f97

So the obvious first step is to implement another urlscheme for deep linking. We could also use Universal Links but it would require us to give an URL endpoint at delta.chat that returns a json with information about the ownership of the app. It requires us to run an additional central service, so let's strike out that option. 

https://swiftrocks.com/adding-deeplink-shortcuts-to-the-ios-home-screen.html suggests to use safari's add to home-screen feature, which creates shortcuts for url links on the home screen. 
We would have to embed a small http server in delta chat that serves a shortcut landing page on localhost with instructions how to add the the shortcut in safari. that landing page contains a little bit of JS which checks if the webpage was openend in Safari or as a Shortcut (PWA) in which case it redirects to Delta chat. 

A longer explanation can be found here as well: https://stackoverflow.com/questions/28042152/link-to-safari-add-to-home-screen-from-inside-app

A demo how this can look like is

https://user-images.githubusercontent.com/2614900/182612420-e26e0c14-86b1-498e-842c-017b8f0952c6.mp4

cyBerta:
closes #1660

there's still one known issue: I haven't figured out yet how the upper left back button responsible to switching back to the former app can be tweaked.

EDIT: It seems like the untitled back button is an  long standing iOS bug: https://stackoverflow.com/questions/72053298/standalone-pwa-shows-untitled-instead-of-actual-pwa-name

**EDIT2:**
After spending some time trying to resolve the bread crumb loop (aka. the upper left back button "< Untitled" reopening the PWA which in turn directly forwards to the corresponding deltachat webxdc), I realized that whenever the user taps on the bread crumb back button, the PWA is reloaded. This is appearantly considered a feature, even though a [controversional one](https://news.ycombinator.com/item?id=16850614). This fact makes it hard to distinguish between the PWA being started from the home screen or the bread crumb menu. Consequently, implementing different behavior - suppressing a forwarding to the delta chat in case of returning to the PWA from the bread crumb menu - is difficult.

What we can consider to do is to store timestamps in a local cache every time the PWA is loaded. If the user elapsed time is too short, the PWA shows a hint how to kill the PWA and button to manually trigger the fowarding to Delta Chat's webxdc. 
If the timeout was exceeded, the user will be automatically forwarded instead. 
@r10s what do you think about that solution?

I was also considering other more complex solutions, similar to[ this one](https://jakub-kozak.medium.com/how-to-share-state-data-between-a-pwa-in-ios-safari-and-standalone-mode-64174a48b043),  making use of service workers on WKWebview in order to be able share data over the [cache](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage) between Delta Chat and the PWA in order to distinguish the return-from-bread-crumb and the start-from-home-screen case. In the meanwhile I doubt that this is doable at all.


Simon-Laux:
it would make sense to document the new uri scheme in https://github.com/deltachat/interface/blob/master/uri-schemes.md

r10s:
> I haven't figured out yet how the upper left back button responsible to switching back to the former app can be tweaked.

yip, that is annoying. also the "dummy app" that exists in the app overview. maybe one can terminate it after Delta Chat has launched? maybe a stupid idea, i did not look to code and concepts deeply yet :)

also, the ”start via shortcut" is a bit flickering, but that is a minor issue and can maybe be mitigated by adapting the startup screen (eg. no delta logo that is visible only for a fraction of a second)

cyBerta:
> "dummy app" that exists in the app overview

What's the dummy app?

Simon-Laux:
removing shortcuts might not be possible on certain manged (school or company) devices, atleast that was the case for me once with iOS > 12 on a such a managed ipad, I could create the link, but not delete it.
maybe this isn't relevant anymore, just wanted to tell the tale, and even if its still the case it's not like we could do anything about it..

deltachat-ios/en.lproj/Localizable.strings
```diff
@@ -894,3 +894,6 @@
 
 "webxdcs" = "Apps";
 "webxdc_empty_hint" = "Webxdc apps you have received or sent will appear here. You can tap on files to select downloaded apps.";
+"shortcut_share_btn" = "Click the share button.";
+"shortcut_add_to_home_description" = "Select - To home screen - to add the shortcut to your home screen!";
```
r10s, line 898:
```suggestion
"shortcut_share_btn" = "Click the share button";
"shortcut_add_to_home_description" = "Select \"Add to Home Screen\" to add the app to your home screen";
```

this should also better go to `untranslated.xml`, or?

cyBerta: right

cyBerta: Escaped quotes screw the leayout and I couldn't yet figure out how to properly escape them. `&quot;` is just printed literally, native JS encoding `\x22` didn't work either out of the box. I appreciate any hints how to fix that. For now I reverted the escaped quotes and use again dashes (that was also the reason I used them in first place)

cyBerta:
Btw. with [203a22a](https://github.com/deltachat/deltachat-ios/pull/1665/commits/203a22aafd662d1b8d10a41e5a65f189577a5eb8) I've fixed the bread crumb loop dismissing the current state of an already opened webxdc.

r10s (approved):
great stuff, thanks for going through all these issues and pitfalls.

the html-part is semantically a git weird, however, it is tested and works, so, i'd leave that as is for now.

PR #1668, opened 2022-08-11
Webxdc selector

### LINKED ISSUE #1661: webxdc app selector
Similar to a sticker selector, we want to have a simple UI to select and paste webxdc's into a chat. 

UI wise we will use a grid layout similarly to the current image selector on iOS. Each grid tile should show the app icon and the application name, the entries in the grid should be sorted by timestamp. 

Changes in core are needed so that 0 for the chat ID can be passed to [dc_get_chat_media](https://c.delta.chat/classdc__context__t.html#a344a82b9f288b5eb5d39c7cf475cddb7). The returned array should contain all messages of a given account containing the requested message type. The corresponding core ticket: https://github.com/deltachat/deltachat-core-rust/issues/3527

Tapping on a grid tile should add the selected webxdc to the draft area, so that it can be edited before it is sent.

cyBerta:
requires https://github.com/deltachat/deltachat-core-rust/pull/3528

closes #1661

### DIFF
--- a/deltachat-ios/en.lproj/Localizable.strings
+++ b/deltachat-ios/en.lproj/Localizable.strings
@@ -892,3 +892,4 @@
 // device messages for updates
 "update_1_30" = "Faster. More stable.\n\nFor 1.30 releases, we focused on speed and reliability, fixing dozens of bugs. Check our changelogs if your favorite one is fixed: https://get.delta.chat/#changelogs 🚀";
 
+"webxdcs" = "Apps";
--- a/deltachat-ios/en.lproj/Localizable.strings
+++ b/deltachat-ios/en.lproj/Localizable.strings
@@ -893,3 +893,4 @@
 "update_1_30" = "Faster. More stable.\n\nFor 1.30 releases, we focused on speed and reliability, fixing dozens of bugs. Check our changelogs if your favorite one is fixed: https://get.delta.chat/#changelogs 🚀";
 
 "webxdcs" = "Apps";
+"webxdc_empty_hint" = "Webxdc apps you have received or sent will appear here. You can tap on files to select downloaded apps.";
--- a/scripts/untranslated.xml
+++ b/scripts/untranslated.xml
@@ -3,5 +3,5 @@
     <!-- iOS specific untranslated strings -->
     <string name ="a11y_connectivity_hint">Double tap to view connectivity details.</string>
     <string name ="webxdcs">Apps</string>
-    <string name ="webxdc_empty_hint">Webxdc apps you have received or sent will appear here. You can tap on files to select downloaded apps.</string>
+    <string name ="webxdc_empty_hint">Received or sent apps will appear here. Tap \"Files\" to select downloaded apps.</string>
 </resources>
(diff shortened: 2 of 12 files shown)


r10s (commented):
some quick functionality check: when there are no apps:
- an assert is logged `deltachat-ios[71734:885763] [Assert] negative or zero item sizes are not supported in the flow layout`
- a wrong placeholder message is shown ("Images and videos shared in this chat will appear here.")

r10s (approved):
no blockers, however, maybe target the easy things already in this pr, then we can merge that to test it ourselves together with the webxdc-icon-to-homescreen :)

cyBerta:
> * an assert is logged `deltachat-ios[71734:885763] [Assert] negative or zero item sizes are not supported in the flow layout`

I still need to fix that. It seems that there's a bug with prefetching items.

r10s (changes requested):
the recent changes are great!

however, a bug sneaked in: when you try to attach a webxdc using "Apps / Files" or just the old "Files" entry, things do not work, the app is shown as an error then:

deltachat-ios/en.lproj/Localizable.strings
```diff
@@ -892,3 +892,5 @@
 // device messages for updates
 "update_1_30" = "Faster. More stable.\n\nFor 1.30 releases, we focused on speed and reliability, fixing dozens of bugs. Check our changelogs if your favorite one is fixed: https://get.delta.chat/#changelogs 🚀";
 
+"webxdcs" = "Apps";
+"webxdc_empty_hint" = "Webxdc apps you have received or sent will appear here. You can tap on files to select downloaded apps.";
```
r10s, line 896:
```suggestion
"webxdc_empty_hint" = "Received or sent apps will appear here. Tap \"Files\" to select downloaded apps.";
```

r10s: ah, that one is generated, should go to untranslated.xml

scripts/untranslated.xml
```diff
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <!-- iOS specific untranslated strings -->
-    "a11y_connectivity_hint" = "Double tap to view connectivity details.";
+    <string name ="a11y_connectivity_hint">Double tap to view connectivity details.</string>
+    <string name ="webxdcs">Apps</string>
+    <string name ="webxdc_empty_hint">Webxdc apps you have received or sent will appear here. You can tap on files to select downloaded apps.</string>
```
r10s, line 6:
```suggestion
    <string name ="webxdc_empty_hint">Received or sent apps will appear here. Tap &quot;Files&quot; to select downloaded apps.</string>
```

cyBerta:
> the recent changes are great!
> 
> however, a bug sneaked in: when you try to attach a webxdc using "Apps / Files" or just the old "Files" entry, things do not work, the app is shown as an error then:
> 
> 
> 
> worse: even the preview of eg. PDFs are missing now. there seems to be sth. broken with attaching files in general :)

Does the issue persist for you with the latest version? It doesn't happen to me. I suspect you were trying with ios 15.X right?

r10s:
for the title: great, this is much better. 

two minors:

- can we have a bigger font? ideally the same size and weight as used when selecting "Files" in icon mode? (list mode is maybe a bit too big)
- i'd add a bit more padding between the cells, maybe the font height. that way, it is also visually clearer where the title belongs to (as the padding between icon and title is unchanged)

PR #1681, opened 2022-09-06
show "Low Data Mode" and "Low Power Mode" as possible reasons for disabled notifications

### LINKED ISSUE #1677: detect "low data mode" and "low power mode" and show it in connectivity
enabled "low data mode" may be a reason for massively delayed notifications, however, this is not always clear to the user (i myself was wondering in holidays :)

same for "low power mode".

we should try to detect these modes and show that to the user in the "connectivity", similar to `disabled_in_system_settings` and `bg_app_refresh_disabled`.

cyBerta:
closes #1677 

The strings may need some adaption. Suggestions welcome.

### DIFF
--- a/deltachat-ios/Controller/ConnectivityViewController.swift
+++ b/deltachat-ios/Controller/ConnectivityViewController.swift
@@ -123,7 +123,7 @@ class ConnectivityViewController: WebViewViewController {
         if ProcessInfo.processInfo.isLowPowerModeEnabled {
             return "<span class=\"disabled dot\"></span>"
                 .appending(title)
-                .appending(String.localized("connectivity_low_battery_mode"))
+                .appending(String.localized("connectivity_low_power_mode"))
         }
 
         let timestamps = UserDefaults.standard.array(forKey: Constants.Keys.notificationTimestamps) as? [Double]
--- a/deltachat-ios/Controller/ConnectivityViewController.swift
+++ b/deltachat-ios/Controller/ConnectivityViewController.swift
@@ -115,7 +115,7 @@ class ConnectivityViewController: WebViewViewController {
         }
 
         if isLowDataMode {
-            return "<span class=\"red dot\"></span>"
+            return "<span class=\"disabled dot\"></span>"
                 .appending(title)
                 .appending(String.localized("connectivity_low_data_mode"))
         }
--- a/deltachat-ios/Controller/ConnectivityViewController.swift
+++ b/deltachat-ios/Controller/ConnectivityViewController.swift
@@ -121,7 +121,7 @@ class ConnectivityViewController: WebViewViewController {
         }
 
         if ProcessInfo.processInfo.isLowPowerModeEnabled {
-            return "<span class=\"red dot\"></span>"
+            return "<span class=\"disabled dot\"></span>"
                 .appending(title)
                 .appending(String.localized("connectivity_low_battery_mode"))
         }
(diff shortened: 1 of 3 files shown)


deltachat-ios/en.lproj/Localizable.strings
```diff
@@ -896,3 +896,5 @@
 "webxdc_empty_hint" = "Received or sent apps will appear here. Tap \"Files\" to select downloaded apps.";
 "shortcut_share_btn" = "Click the share button";
 "shortcut_add_to_home_description" = "Select - Add to Home Screen - to add the app to your home screen.";
+"connectivity_low_data_mode" = "Low Data Mode: background message fetching is disabled.";
```
r10s, line 899: maybe 

```suggestion
"connectivity_low_data_mode" = "Disabled by system's \"Low Data Mode\"";
```

this fits better as it basically reads "Notifications: Disabled ..." - and also the other strings are similar ("Disabled in Delta Chat" or "Disabled om system settings").

and yes, according to https://support.apple.com/en-us/HT210596 it is actually disabled.

scripts/untranslated.xml
```diff
@@ -6,4 +6,6 @@
     <string name ="webxdc_empty_hint">Received or sent apps will appear here. Tap \"Files\" to select downloaded apps.</string>
     <string name ="shortcut_share_btn">Click the share button</string>
     <string name ="shortcut_add_to_home_description">Select - Add to Home Screen - to add the app to your home screen.</string>
+    <string name="connectivity_low_data_mode">Low Data Mode: background message fetching is disabled.</string>
+    <string name="connectivity_low_battery_mode">Low Power Mode: background message fetching is disabled.</string>
```
r10s, line 10:
```suggestion
    <string name="connectivity_low_power_mode">Disabled by system's \"Low Power Mode\"".</string>
```

i'd name the string correctly here, and adapt translation similar to the other strings.

according to https://support.apple.com/en-us/HT205234 - things are "reduced or affected" - so, maybe we can also say "Delayed by system's \"Low Power Mode\"", however, i have the feeling it is always close to "Disabled", so maybe better saying that, even if it is too pessimistic in some cases.

r10s (commented):
review of the strings only ...

deltachat-ios/Controller/ConnectivityViewController.swift
```diff
@@ -85,6 +114,18 @@ class ConnectivityViewController: WebViewViewController {
                 .appending(String.localized("connectivity_not_connected"))
         }
 
+        if isLowDataMode {
+            return "<span class=\"red dot\"></span>"
```
r10s, line 118: i'd say, this is no error, it is disabled as for the other reasons.

```suggestion
            return "<span class=\"disabled dot\"></span>"
```

deltachat-ios/Controller/ConnectivityViewController.swift
```diff
@@ -85,6 +114,18 @@ class ConnectivityViewController: WebViewViewController {
...
+                .appending(title)
+                .appending(String.localized("connectivity_low_data_mode"))
+        }
+
+        if ProcessInfo.processInfo.isLowPowerModeEnabled {
+            return "<span class=\"red dot\"></span>"
```
r10s, line 124: same here:

```suggestion
            return "<span class=\"disabled dot\"></span>"
```

r10s (approved):
some minor suggestions wrt wording and colors - otherwise, this lgtm.

PR #1682, opened 2022-09-07
drag-and-drop into chats

### LINKED ISSUE #1671: (Drag and) Drop support
Mainly drag files into the composer, a bit like what desktop has.

For iPhone this helps when I make a screenshot it hovers in the side and I expect to be able to drop it into the composer to send it directly (or better put it into the draft of course). (also you can switch apps which dragging an image/file, which I just found out)

For iPad this is useful when multitasking like files or gallery app one one side and deltachat on the other side.

For running the iOS app on macOS this makes the app more complete and intuitive to use.

https://developer.apple.com/design/human-interface-guidelines/patterns/drag-and-drop
https://developer.apple.com/documentation/uikit/drag_and_drop

cyBerta:
closes #1671


Simon-Laux (approved):
works nicely!
(tested on iPhone and on M1 desktop)

Issue #1188, opened 2021-04-30
make links in error messages clickable


r10s:
error messages sent from the server may contain links as "click here to do this-and-that", therefore, if we show the error messages after configure fails, we should make the links inside the message clickable, if possible. 

on android we did that at https://github.com/deltachat/deltachat-android/pull/1282

this was the outcome of https://github.com/deltachat/deltachat-ios/issues/1034

cyBerta:
This is not possible on iOS without using private APIs in a UIAlertController. 
Our choices are
- using private APIs, unlikely (doesn't affect's the users privacy for example) but not impossible that we get rejected when publishing a release
- using a replacement for UIAlertController, like https://github.com/SwiftKickMobile/SwiftMessages (or an own implementation)

cyBerta:
using a framework like https://github.com/SwiftKickMobile/SwiftMessages would also help us here:  #878

cyBerta:
> using a framework like https://github.com/SwiftKickMobile/SwiftMessages would also help us here: #878

also here: #1039

r10s:
if we do not want to add more frameworks for minor things, assuming this is mainly and issue in the configure-dialog and assuming most times there is only one url, we could also just get the first url from the message and add a "Open Link" button.

r10s:
moving stale feature request to resurrection

Issue #1039, opened 2021-01-25
improve forward workflow


r10s:
- when forwarding is cancelled already in the "select-forward-destination chatlist", one should get back to the chat where forwarding was started.  
currently, one just stays in the "main chatlist".  
(cancelling forwarding later in a selected chat is fine - that brings one back to the "select-forward-destination chatlist")

- when "saved messages" is selected in the "select-forward-destination chatlist", the message is saved without further questions,  which is great.  
however, it would be even better when, after saving, the user comes back to the original chat from where forwarding was started. this allows "saving" a message quickly by two taps without leaving the context.  

- there should be some feedback on the screen after "saving" - [on android, we're using a "toast"](https://github.com/deltachat/deltachat-android/pull/1651), that does not exist on ios, however, there are probably alternatives (sth. easy without extra additional libs, if possible)

esp. the last two points will give usability a boost :)

cyBerta:
> there should be some feedback on the screen after "saving" - on android, we're using a "toast", that does not exist on ios, however, there are probably alternatives (sth. easy without extra additional libs, if possible)

that's another usecase for https://github.com/SwiftKickMobile/SwiftMessages

having in mind https://github.com/deltachat/deltachat-ios/issues/1188 and https://github.com/deltachat/deltachat-ios/issues/878 I think it is worth it to use the library for our needs.

r10s:
moving stale issue to [resurrection](https://github.com/orgs/deltachat/projects/19) due to [march2022 cleanup](https://github.com/orgs/deltachat/projects/48).

Issue #878, opened 2020-07-22
improve autocrypt-setup-message dialog


cyBerta:
Currently there's an alert dialog containing a long input field for entering the fingerprint / validation code. 
It's hard to keep track of the long code while entering since there's no visual help.

We should implement an autocrypt setup dialog as it was done on Android. The latter has multiple input field blocks, separated by some space.

r10s:
moving stale issue to [resurrection](https://github.com/orgs/deltachat/projects/19) due to [march2022 cleanup](https://github.com/orgs/deltachat/projects/48).

PR #1687, opened 2022-09-18
add quick switch account avatar button to chatlist header

Simon-Laux:
Highly experimental, layout still needs to be done to make it look good, also there is no custom menu on tapping it 
yet, it's just a modified version of the alert menu to only allow account switching(removed the possible destructive 
options for adding or 
removing account for now) and added unread count for each account.

What is this and why did you do it?
~~I'm planning to draw some sketches and do a real feature proposal for this over at out forum soon. (I also want it on 
android)~~
https://support.delta.chat/t/account-selector-on-avatar-on-chat-list/2242

### Progress

- [X] Logic
- [ ] make the Avatar Button with unread counter look nice
- [ ] custom nice looking context menu that also shows avatars, the android design would already be good, but I hope 
for something like what telegram is doing.

#### Bugs
- [x] custom icon replaces back button for archived chats, it should not do that, it should just be on the root/home view of the chatlist
- [x] reuse old function -> also show add and delete for now until we have a custom menu

## some sneak peaks

Bildschirmfoto 2022-09-18 um 20 19 34

Bildschirmfoto 2022-09-18 um 20 07 55

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -536,7 +536,7 @@ class ChatListController: UITableViewController {
     }
     
     lazy var accountButtonAvatar: InitialsBadge = {
-        let badge = InitialsBadge(size: 42)
+        let badge = InitialsBadge(size: 40)
         badge.setColor(UIColor.lightGray)
         badge.isAccessibilityElement = false
         return badge
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -549,7 +549,6 @@ class ChatListController: UITableViewController, AccountSwitcherHandler {
         return view
     }()
 
-
     private lazy var accountButton: UIBarButtonItem = {
         let containerView = UIView(frame: CGRect(x: 0, y: 0, width: 37, height: 37))
         containerView.addSubview(accountButtonAvatar)
@@ -563,7 +562,6 @@ class ChatListController: UITableViewController, AccountSwitcherHandler {
     
     private func updateAccountButton() {
         let unreadCount = getUnreadCounterOfOtherAccounts()
-        // accountButton.title = "Account" + (unreadCount == 0 ? "" : " [" + String(unreadCount) + "]")
         accountButtonUnreadMessageCounter.setCount(unreadCount)
         accountButtonUnreadMessageCounter.isHidden = unreadCount == 0
         
@@ -573,8 +571,6 @@ class ChatListController: UITableViewController, AccountSwitcherHandler {
         if let image = contact.profileImage {
             accountButtonAvatar.setImage(image)
         }
-        
-        
     }
     
     private func getUnreadCounterOfOtherAccounts() -> Int {
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -654,12 +654,11 @@ class ChatListController: UITableViewController {
                     titleView.accessibilityHint = "\(String.localized("connectivity_connected")): \(String.localized("a11y_connectivity_hint"))"
                 }
             }
+            navigationItem.setLeftBarButton(accountButton, animated: false)
+            updateAccountButton()
         }
         titleView.isUserInteractionEnabled = !tableView.isEditing
         titleView.sizeToFit()
-        
-        navigationItem.setLeftBarButton(accountButton, animated: false)
-        updateAccountButton()
     }
 
     func handleMultiSelectionTitle() -> Bool {
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -1,10 +1,10 @@
 import UIKit
 import DcCore
 
-class ChatListController: UITableViewController {
+class ChatListController: UITableViewController, AccountSwitcherHandler {
     var viewModel: ChatListViewModel?
     let dcContext: DcContext
-    private let dcAccounts: DcAccounts
+    internal let dcAccounts: DcAccounts
     var isArchive: Bool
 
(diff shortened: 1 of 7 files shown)


Simon-Laux:
https://github.com/KaoruMuta/IGStoryButtonKit could be used as a reference on how to make custom buttons in uikit

cyBerta:
https://github.com/SwiftKickMobile/SwiftMessages is on my wish-list for longer, e.g. to solve https://github.com/deltachat/deltachat-ios/issues/1188, https://github.com/deltachat/deltachat-ios/issues/878 and https://github.com/deltachat/deltachat-ios/issues/1039

Would you be up to try that lib for the account switch alert?

r10s:
better finish this pr without adding dependencies and do discussions about that afterwards in a subsequent pr, with a broader look on different features. maybe there are also system popup menus etc. that could do the job.

i discussed this with @Simon-Laux already, so, just use the same, boring switch dialog, adding "[123]" unread numbers and focus on functionality and switch avatar here (the latter already looks good to me). the dialog from the second screenshot can be removed as well then (this will be replaced by sth. nicer).

r10s:
there is a memory leak in account switching - it was there before, however, gets much more visible by making account switching more accessible, see https://github.com/deltachat/deltachat-ios/issues/1695

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -522,6 +534,75 @@ class ChatListController: UITableViewController {
...
+        view.isHidden = true
+        return view
+    }()
+
+
+    private lazy var accountButton: UIBarButtonItem = {
+        // todo make pretty
+        
+        let rect = CGRect(x: 0, y: 0, width: 70, height: 50)
+        let myView = UIView(frame: rect)
```
cyBerta, line 557: nitpicking: could you rename it to sth. like containerView? myView is a little bit unusual in this repo.

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -522,6 +534,75 @@ class ChatListController: UITableViewController {
...
+        view.backgroundColor = DcColors.unreadBadge
+        view.isHidden = true
+        return view
+    }()
+
+
+    private lazy var accountButton: UIBarButtonItem = {
+        // todo make pretty
+        
+        let rect = CGRect(x: 0, y: 0, width: 70, height: 50)
```
cyBerta, line 556: I would prefer if we set manually constraints when composing this view, just as we do elsewhere. This line would become obsolete in that case.

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -522,6 +534,75 @@ class ChatListController: UITableViewController {
...
+        let myView = UIView(frame: rect)
+        
+        myView.addSubview(accountButtonAvatar)
+        myView.addSubview(accountButtonUnreadMessageCounter)
+        
+        // this line is only to visualize the rect bounds for visual debugging
```
cyBerta, line 562: that could be cleaned up

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -69,8 +71,9 @@ class ChatListController: UITableViewController {
 
     private var editingConstraints: NSLayoutConstraintSet?
 
-    init(dcContext: DcContext, isArchive: Bool) {
+    init(dcContext: DcContext, dcAccounts: DcAccounts, isArchive: Bool) {
```
cyBerta, line 74: we don't need to pass both dcAccounts and DcContext there, dcAccounts would be sufficient

deltachat-ios/Handler/AccountSwitcherHandler.swift
```diff
@@ -0,0 +1,86 @@
+//
+//  AccountSwitcherHandler.swift
+//  deltachat-ios
+//
+//  Created by bb on 30.09.22.
+//  Copyright © 2022 Jonas Reinsch. All rights reserved.
```
cyBerta, line 6: that should be removed, too :)

deltachat-ios/Handler/AccountSwitcherHandler.swift
```diff
@@ -0,0 +1,86 @@
...
+//
+
+import Foundation
+import UIKit
+import DcCore
+import DBDebugToolkit
+import Intents
+
+protocol AccountSwitcherHandler: UIViewController {
```
cyBerta, line 15: Nice!

cyBerta (approved):
I have only minor proposals.

deltachat-ios/Handler/AccountSwitcherHandler.swift
r10s: really wondering where this comes from, is there some old template deep in xcode?

deltachat-ios/Controller/ChatListController.swift
r10s: +1, i did the same in the commit i just added.

deltachat-ios/Controller/ChatListController.swift
r10s: i tweaked size in the latest commit, but it is probably still a bit handwavy @cyBerta feel free to add a commit to change to "manually constraints"

deltachat-ios/Controller/ChatListController.swift
r10s: i removed that in recent commit.

deltachat-ios/Handler/AccountSwitcherHandler.swift
r10s: removed with the latest commit.

r10s (approved):

r10s:
k, let's merge that in to move forward. i am unsure about the position of the unread-counter (ios us upper-right in similar situations), but it is good enough and definitely an improvement. thanks a lot!

PR #1689, opened 2022-09-19
bigger chat avatar in chatview

Simon-Laux:
feature proposal: https://support.delta.chat/t/larger-profile-picture-in-the-action-bar/2226

tried to make it a similar size to telegram & whatsapp.

image_2022-09-19_12-57-19

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -143,7 +143,7 @@ class ChatViewController: UITableViewController {
 
     private lazy var initialsBadge: InitialsBadge = {
         let badge: InitialsBadge
-        badge = InitialsBadge(size: 38, accessibilityLabel: String.localized("menu_view_profile"))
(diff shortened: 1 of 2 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -143,7 +143,7 @@ class ChatViewController: UITableViewController {
 
     private lazy var initialsBadge: InitialsBadge = {
         let badge: InitialsBadge
-        badge = InitialsBadge(size: 28, accessibilityLabel: String.localized("menu_view_profile"))
+        badge = InitialsBadge(size: 38, accessibilityLabel: String.localized("menu_view_profile"))
```
r10s, line 146: the avatar appeared a bit too large to be pleasing, a tiny bit too few whitespace. i compared pixel-wise with telegram, and `37` seems what they're using, so we should go for that as well.

```suggestion
        badge = InitialsBadge(size: 37, accessibilityLabel: String.localized("menu_view_profile"))
```

r10s (approved):
nice!

PR #1692, opened 2022-09-21
Import backup improvements

### LINKED ISSUE #872: Improve import backup onboarding experience
When importing a backap after a fresh DC installation, a an alert message appears if no backup has been found. The alert message tells the path to the directory DC will look for an backup file. At that point the directory is not created yet. Therefore the user will need to remember the path in the alert message and create manually the directory, move a backupfile there and retry the import process. 

My proposal is that  
a) the directory will be automatically generated as soon as the user tries to import an backup and if that directory is not yet existing. This directory will be used later on for other DC config files anyways (wal sqlite file for example). This way, the user will find the target folder where to move the backup to (using the Files app) more easily.

b) the user can choose the file from a file picker: the file will be copied to the apps documents folder (which will be created if it's not yet existing) and the backup starts immediately afterwards. This implies additional error handling, if the app has not the permission to access the selected file. The documents picker logic could be reused from the attach documents context.

@r10s @nayooti I would like to know what you think about it

cyBerta:
closes #872 

* automatically restores the latest backup in the DC backup folder
* if no backup is available there, a document picker shows up, the user can select a backup from arbitrary locations, also the cloud
* if manual backup selection was cancelled, the alert appears with a hint where to put the backup as before
* it is ensured the DC folder exists and can be found in the finder app
* also: cancelling an ongoing backup attempt replaced the half-configured account with an empty new one, cancelling was half-baked before
* cancelling an ongoing backup attempt into an encrypted database falls back to an unencrypted empty account

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -555,15 +555,6 @@ public class DcContext {
         dc_imex(contextPointer, what, directory, passphrase)
     }
 
-    public func imexHasBackup(filePath: String) -> String? {
-        var file: String?
-        if let cString = dc_imex_has_backup(contextPointer, filePath) {
-            file = String(cString: cString)
-            dc_str_unref(cString)
-        }
-        return file
-    }
-
     public func isSendingLocationsToChat(chatId: Int) -> Bool {
         return dc_is_sending_locations_to_chat(contextPointer, UInt32(chatId)) == 1
     }
--- a/deltachat-ios/Controller/WelcomeViewController.swift
+++ b/deltachat-ios/Controller/WelcomeViewController.swift
@@ -313,13 +313,21 @@ class WelcomeViewController: UIViewController, ProgressAlertHandler {
             forName: dcNotificationImexProgress,
             object: nil,
             queue: nil
-        ) { notification in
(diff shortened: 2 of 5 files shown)


r10s (commented):
a great improvement!

there may be the situation that _two_ backups exist - in this case, imexHasBackup() will return the newest and the user does not has the option to chose another.

if we think, that this is an issue, we could (a) maybe just show _always_ show selector. this would also make the bit weird alert dialog on "cancel" superfluous. or (b) show the selector in case of multiple backups only.

(b) sounds a bit too less straight-forward to me. sometimes the selector will appear, sometimes not.

cc @Simon-Laux - what do you think on that?

if we go for (a) - maybe we can somehow preselect that folder? for older os: does the fileselector work on all ios versions we target?

Simon-Laux:
I'd vote for a) always show the selector.
As for when older OS'ses should not support it, then show an option list like the ones we use for account switching, or something simple like that?

cyBerta:
the document picker should be available on all supported iOS versions. It's using the same code as we do for attaching files to chats.

r10s (approved):
lgtm; maybe also remove the now unused `imexHasBackup()` completely, but this is no blocker, one can also argue to keep it for now :)

PR #1699, opened 2022-09-27
[WIP] Fixing memory leak in ChatListController

### LINKED ISSUE #1695: memory leak when switching accounts
when switching accounts, either with the old switcher or with the new from https://github.com/deltachat/deltachat-ios/pull/1687, the old chatlist-viewcontrollers seem not to be removed, but only new ones are added atop. we had sth. like that in the past, it leads to slowness the longer you use the app.

for testing, i added a `self?.dcContext.logger?.info("🎃🎃🎃🎃🎃🎃 HI!")` to the `msgChangedObserver` of `ChatListController` - it should be printed only once eg. when deleting a chat - but if you switch ten times or so accounts, it gets fired multiple times, same for the chatlist-loading (which leads to the slowness)

larger log: https://gist.github.com/r10s/383f748186d4db75bc802aa0bdd57936

cyBerta:
Ensure that tabbar root controllers always call willMove before they get replaced during account switch. 

closes #1695

### DIFF
--- a/deltachat-ios/Coordinator/AppCoordinator.swift
+++ b/deltachat-ios/Coordinator/AppCoordinator.swift
@@ -235,12 +235,17 @@ class AppCoordinator {
     }
 
     func resetTabBarRootViewControllers() {
+        // call `willMove()` for the root view controllers of each tab, after popping to root.
+        // this is not always done by `setViewControllers()` the documentation is vague on this point:
+        // <https://developer.apple.com/documentation/uikit/uitabbarcontroller/1621177-setviewcontrollers>
+        // (calling `willMove()` is needed eg. to remove observers - otherwise we have a memory leak)
         self.tabBarController.viewControllers?.forEach { controller in
             if let navController = controller as? UINavigationController {
                 navController.popToRootViewController(animated: false)
                 navController.viewControllers[0].willMove(toParent: nil)
             }
         }
+
         self.tabBarController.setViewControllers([createQrNavigationController(),
                                                   createChatsNavigationController(),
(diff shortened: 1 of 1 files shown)


r10s (approved):
works and and fixes the issue, great!

however, why is `willMove` not called by ios in this case? is that documented somewhere? or an ios bug? maybe we are doing things wrong at another place and hacking around that here?  
at least i would have expected `willMove` being called for `popToRootViewController` ...

the potential damage if `willMove(null)` is called twice, however, is currently low, subsequent calls would just do nothing - so, all these questions are no blockers.

@cyBerta but i would like to get your opinion on the questions, maybe double check, and maybe we should add some comment.

cyBerta:
> at least i would have expected `willMove` being called for `popToRootViewController` ...
Indeed that should be the case. However the rootViewControllers are the ChatListController, the SettingsController and the QRViewController. So `willMove` is not called in these controllers. 

Appearantly `self.tabBarController.setViewControllers` replaces the existing  view controllers without `willMove` is called in these cases. I can resarch a little bit on behavior.

r10s:
thanks for clarification. in any case (`setViewControllers()` is buggy or it is the defined behavior) i'd add a comment that this is needed as `setViewControllers()` does not call  `willMove()` - and if that changes in a future ios version, `willMove` may be called twice.

cyBerta:
I'm still trying to find a better solution. back to WIP

r10s:
as discussed in chat, we merge this fix. if research leads to a better solution, we do a follow up.

PR #1700, opened 2022-09-27
show 'back button' in preview again

### LINKED ISSUE #1696: ios16: full-screen-image-view cannot be closed
the "Done" button of the full-screen-image-view is gone on ios16 (tested on iphone13mini)

maybe there is a gesture or sth. like that, but i could not manage to close the view at all.

r10s:
`QLPreviewController` seems not to show a 'back' button if `present()`ed inside a `UINavigationController` - in the past, there was a hack that adds the button manually in this case.

this hack stopped working with ios16 -
resulting having no way to leave preview.

this pr removes the hack, stops using `present()`
and switches to `pushViewController()` -
this seems to be the natural way here and shows the proper back buttons.

moreover, the pr removes the sometimes used wrapper ("betterPreviewController"), not sure what was the reasinging for that, probably needed for some fullscreen-mode however, pushing a `UINavigationController` to a `UINavigationController` is not allowed, so it just cannot be used in this form anymore -
and we're "fullscreen" as usual anyway.

closes #1696

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -1546,10 +1546,8 @@ class ChatViewController: UITableViewController {
     }
 
     private func showMediaGallery(currentIndex: Int, msgIds: [Int]) {
-        let betterPreviewController = PreviewController(dcContext: dcContext, type: .multi(msgIds, currentIndex))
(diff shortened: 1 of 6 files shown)


cyBerta (approved):
lgtm. Thanks for figuring this out.

PR #1702, opened 2022-09-27
add support for dclogin also fixes #1701

### LINKED ISSUE #1701: `qrErrorAlert()` does not show account creation error from core
https://github.com/deltachat/deltachat-ios/blob/00614004a88e590d8eeeb86e06a0f80c48f78fed/deltachat-ios/Controller/WelcomeViewController.swift#L372-L374

It should show the error too, like android does (android phishes it out of the last events, sure not ideal but I see no other quick way, it's not like we would migrate the whole iOS to jsonrpc over this that would be insane. maybe a new cffi call that writes the error in an error variable or returns a result struct, but I guess that is also too much work for now.).

Especially for `DCLOGIN:` it is needed because it tells the user to update if the version is not supported.

Simon-Laux:
adds support for dclogin scheme.
also fixes #1701

needs core with https://github.com/deltachat/deltachat-core-rust/pull/3541 merged.

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -187,7 +187,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         // }
 
         switch url.scheme?.lowercased() {
-        case "dcaccount":
+        case "dcaccount", "dclogin":
             appCoordinator.handleQRCode(url.absoluteString)
             return true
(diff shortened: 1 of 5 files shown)


r10s (approved):
lgtm, beside core, this pr requires https://github.com/deltachat/deltachat-android/pull/2366 to be merged first. https://github.com/deltachat/deltachat-android/pull/2366 has some open tiny things (funcrtion name, wording) to be checked before merging.

PR #1705, opened 2022-10-01
enlarge profile action buttons

r10s:
this makes the 'search' and 'mute' buttons a bit larger, so that they're easier to tap
and also easier to recognize as a button at all
(not only a state icon)

before / after:


cyBerta (approved):
you're so fast ;)

PR #1706, opened 2022-10-02
show webxdc-selector and -profile-tabs only if there are webxdc

r10s:
see dev group for reasoning and discussion.


cyBerta (approved):
lgtm

PR #1710, opened 2022-10-04
fix disappearing message input bar on ios16

### LINKED ISSUE #1709: input bar sometimes missing when switching reply source
when changing the parent a message shall become a reply to, the input field may disappear completely. this happens only when the keyboard is opened _and_ when using "reply" from the long-tap-menu, not for swipe-to-reply.

to reproduce:

- long-tap a message and select "Reply"
- make sure the keyboard is open
- long-tap _another_ message and select "Reply"

-> the input bar is gone.

cyBerta:
closes #1709


cyBerta:
I only have an simulator right now to test this. Please verify with a real device.

r10s (approved):
fixed for me :)

(tested on iphone13 with ios16)

PR #1711, opened 2022-10-04
move "recently seen dot" lower right

r10s:
this moves the "recently seen dot" to the corner where it is shown also for many other apps: to the lower right corner.

as the lower right corner was already occupied by the "verified checkmark", that one is moved to the upper right corner, with a little bit of padding to the top. i played around with different positions, that one looked best.

also the combination of both indicators look fine now.

it is anyway a bit unclear what happens to the "verified" indicator for contacts, it may go away completely and being replaced by a line as "Verified by NAME", so it is good enough for now :)

@cyBerta i am not so much in these constraints - please push to this pr if i did sth wrong :)

### DIFF
--- a/DcCore/DcCore/Views/InitialsBadge.swift
+++ b/DcCore/DcCore/Views/InitialsBadge.swift
@@ -2,7 +2,6 @@ import UIKit
 
 public class InitialsBadge: UIView {
 
-    private let verificationViewPadding: CGFloat = 2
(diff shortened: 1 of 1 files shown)


adbenitez (approved):
not sure about the padding at the top for the verified icon, but all in all better now

Simon-Laux (approved):

Issue #1703, opened 2022-09-29
sharing to a one-to-one siri-shortcut faulty


paulaluap:
to reproduce: 
- share a webpage from safari
- select a one-to-one siri-shortcut
- subsequent dialog shows "saved messages" instead of the selected one-to-one chat

same with sharing an image from the gallery and probably other apps.

sharing to a group works.

- Device: iphone x
- iOS version: 16
- Delta Chat version: 1.32

cyBerta:
I can only reproduce that bug in a multi-account scenario if the selected account is not the one the one-on-one chat belongs to. Is that the case for you as well @paulaluap ?

paulaluap:
> I can only reproduce that bug in a multi-account scenario if the selected account is not the one the one-on-one chat belongs to. Is that the case for you as well @paulaluap ?

thanks for taking care. no, i am not using a multi-account.

cyBerta:
stepping back from this ticket for a bit, to focus on #1709, which I can reproduce

r10s:
i found the failing line:

https://github.com/deltachat/deltachat-ios/blob/bf38b19e329c59cbadff66ad2e2cfc5cb97ddd4a/DcShare/Controller/ShareViewController.swift#L103

here, `if let intent = self.extensionContext?.intent as? INSendMessageIntent` fails with `self.extensionContext?.intent` being `nil` - maybe the intent is of another class?

r10s:
i am also failing to reproduce "new chats" with the issue. also, when donating a previously failing contact again, things are fine. maybe this is some hickup from the ios15 -> ios16 upgrade?

r10s:
there is no actionable item left, issues seems to come from previous versions. please reopen, if you manage to reproduce issue on a new installation.

PR #1714, opened 2022-10-05
switch accounts if nececssary for direct share shortcuts

cyBerta:
part of #1703 

not yet implemented: UI changes. 
We could add the account avatar to the navigation bar, e.g. left of the cancel button or left of the send button (maybe that get's more attention, before the user hits the send button). 
We can also add another row to the bottom showing the account name there (unfortunately but without an icon using the basic iOS share framework) and could give an option to switch accounts there. That might become especially useful in case the user didn't select a direct share shortcut, but the general delta chat share icon. 
Opinions welcome :)

### DIFF
--- a/DcShare/Controller/ShareViewController.swift
+++ b/DcShare/Controller/ShareViewController.swift
@@ -84,47 +84,63 @@ class ShareViewController: SLComposeServiceViewController {
     override func presentationAnimationDidFinish() {
         dcAccounts.logger = logger
         dcAccounts.openDatabase()
-        if !dcContext.isOpen() {
-            do {
-                let secret = try KeychainManager.getAccountSecret(accountID: dcContext.id)
-                if !dcContext.open(passphrase: secret) {
-                    logger.error("Failed to open database.")
-                }
-            } catch KeychainError.unhandledError(let message, let status), KeychainError.accessError(let message, let status) {
-                logger.error("KeychainError. \(message). Error status: \(status)")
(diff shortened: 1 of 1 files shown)


cyBerta:
added icon: 
Simulator Screen Shot - iPhone SE (1st generation) - 2022-10-07 at 17 12 54
Simulator Screen Shot - iPhone SE (1st generation) - 2022-10-07 at 17 14 12

Raiden1982:
I really like the idea of being able to switch accounts when sharing. Maybe it would look better if the avatar is placed in the middle between cancel and send.

Raiden1982:
I really like the idea of being able to switch accounts when sharing. Maybe it would look better if the avatar is placed in the middle between "_Cancel_" and "_Send_"?
However, if it should be to the left, _iMessage_ could perhaps be an example. The top left corner is blank in the _iMessage_ share screen. The avatar could be displayed here. This design is probably more complex to implement.
I created a mockup screenshot for both options.

Avatar in the middle:
Bildschirmfoto 2022-10-07 um 20 42

Share screen derived from iMessage:
Bildschirmfoto 2022-10-08 um 11 50

cyBerta:
@r10s Let's get that in. It fixes a bug in context of multi-account and gives additionally a hint from which account the message is going to be shared. Further improvements wrt. to UI can be done in a subsequent PR.

cyBerta:
We can proceed with another PR

PR #1715, opened 2022-10-06
fix emojis in webxdc

Simon-Laux:
- fix emojis in display name
- get webxdc update json over web request this is a solution fto circumvent or possible xss attacks (because executing the data concatenated into a command is possibly dangerous - the base64 encoding woraround broke the emojis)
- remove base64 string extention that is now unused

### DIFF
--- a/deltachat-ios/Controller/WebxdcViewController.swift
+++ b/deltachat-ios/Controller/WebxdcViewController.swift
@@ -55,6 +55,11 @@ class WebxdcViewController: WebViewViewController {
     """
     
     lazy var webxdcbridge: String = {
+        let addr = dcContext.addr?
+            .addingPercentEncoding(withAllowedCharacters: CharacterSet.urlQueryAllowed)
+        let displayname = (dcContext.displayname ?? dcContext.addr)?
+            .addingPercentEncoding(withAllowedCharacters: CharacterSet.urlQueryAllowed)
+        
         let script = """
         window.webxdc = (() => {
           let setUpdateListenerPromise = null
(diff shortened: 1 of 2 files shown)


deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -371,6 +371,18 @@ extension WebxdcViewController: WKScriptMessageHandler {
 extension WebxdcViewController: WKURLSchemeHandler {
     func webView(_ webView: WKWebView, start urlSchemeTask: WKURLSchemeTask) {
         if let url = urlSchemeTask.request.url, let scheme = url.scheme, scheme == INTERNALSCHEMA {
+            if url.path == "/webxdc-update.json" || url.path == "webxdc-update.json" {
```
cyBerta, line 374: are both conditions needed or was this just for testing?

cyBerta (approved):

deltachat-ios/Controller/WebxdcViewController.swift
Simon-Laux: the first is needed, the second might be (I don't know maybe other web view versions), I kept it for redundancy. we could remove it but I think it won't hurt

deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -271,7 +276,7 @@ class WebxdcViewController: WebViewViewController {
                let maxSerial = first["max_serial"] as? Int {
                 self.lastSerial = maxSerial
             }
-            webView.evaluateJavaScript("window.__webxdcUpdate(atob(\"\(statusUpdates.toBase64())\"))", completionHandler: nil)
+            webView.evaluateJavaScript("window.__webxdcUpdate(\(String(lastSerial)))", completionHandler: nil)
```
r10s, line 279: at a first glance, i fear that here may be a race condition leading to double updates:

if after this line, core receives a new update, this will be returned by `webxdc-update.json?SERIAL` _without_ updating `lastSerial`. when then `updateWebxdc()` is called again, the same update is returned again.

but please @Simon-Laux , @cyBerta  double-check, might be that the race is impossible by some reasons i do not have in mind.

Simon-Laux: ok let's move the whole last serial shenanigans to the js side to not have this issue?

Simon-Laux: moved the stuf to js https://github.com/deltachat/deltachat-ios/pull/1718

PR #1716, opened 2022-10-06
missing inserted messages

cyBerta:
this PR reduces the code complexitiy. We just always call refreshMessagees()/dc…Context.getChatMsgs() on DC_EVENT_MSGS_CHANGED, DC_EVENT_MSG_READ, DC_EVENT_MSG_DELIVERED and DC_EVENT_MSG_FAILED now.

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -563,20 +563,19 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
                 logger.debug(">>> msgChangedObserver: \(String(describing: ui["message_id"]))")
                 if self.dcChat.canSend, let id = ui["message_id"] as? Int, id > 0 {
                     let msg = self.dcContext.getMessage(id: id)
-                    if msg.isInfo,
-                       let parent = msg.parent,
-                       parent.type == DC_MSG_WEBXDC {
-                        self.refreshMessages()
-                    } else {
-                        self.updateMessage(msg)
+                    if msg.type == DC_MSG_WEBXDC,
+                       msg.chatId == self.chatId {
+                        // webxdc draft got updated
+                        self.draft.draftMsg = msg
+                        self.configureDraftArea(draft: self.draft, animated: false)
+                        return
                     }
-                } else {
                     self.refreshMessages()
                     DispatchQueue.main.async {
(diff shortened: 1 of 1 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -563,20 +563,19 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
                 logger.debug(">>> msgChangedObserver: \(String(describing: ui["message_id"]))")
                 if self.dcChat.canSend, let id = ui["message_id"] as? Int, id > 0 {
```
r10s, line 564: we still need an update for `id == 0` which is [emitted for batch changes](https://c.delta.chat/group__DC__EVENT.html#ga0f52cdaad70dd24f7540abda6193cc2d).

also the `canSend` looks wrong to me here, so, in general, it looks to me as if an `else` is missing.

otoh, i would move the `msg.chatId == self.chatId` check up, so that the whole thing is not exectuted for foreign changes - that way, we can also avoid loading the message from the database in most cases.

i can do a commit with these changes, but one question before: why is the DRAFT handling needed here at all? this is a bit unclear to me.

cyBerta: IIRC, it's related to webxdc drafts that we initilize with values before sending

cyBerta: > we still need an update for `id == 0` which is [emitted for batch changes](https://c.delta.chat/group__DC__EVENT.html#ga0f52cdaad70dd24f7540abda6193cc2d).

that might be the root cause for the bug, we never adapted to that change in core. (I think that behavior didn't exist all the time)

r10s: the behavior exists forever, but maybe also the bug is there since then :)

for the draft: still a bit unclear to me, the draft is not updated from the core "just so" - otoh, it's fine, we do not have issues with the drafts, so maybe leave that part alone for now :)

cyBerta: If you find time for it, please feel free to add any commits here!

r10s: yip, i'll do it!

cyBerta:
closing in favor of #1767

PR #1717, opened 2022-10-06
fix phaser sprite loading bug

Simon-Laux:
fix phaser sprite loading bug in webxdc the reason was the missing http status code

### DIFF
--- a/deltachat-ios/Controller/WebxdcViewController.swift
+++ b/deltachat-ios/Controller/WebxdcViewController.swift
@@ -392,28 +392,19 @@ extension WebxdcViewController: WKURLSchemeHandler {
                 data = dcMsg.getWebxdcBlob(filename: file)
             }
             let mimeType = DcUtils.getMimeTypeForPath(path: file)
-            if !data.isEmpty {
-                urlSchemeTask.didReceive(HTTPURLResponse(
-                    url: url,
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -392,9 +392,28 @@ extension WebxdcViewController: WKURLSchemeHandler {
...
+            if !data.isEmpty {
+                urlSchemeTask.didReceive(HTTPURLResponse(
+                    url: url,
+                    statusCode: 200,
+                    httpVersion: "HTTP/1.1",
+                    headerFields: [
+                        "Content-Type": mimeType,
+                        "Content-Length": "\(data.count)"
+                    ]
+                )!)
```
cyBerta, line 404: we normally try to program a little bit more defensively, avoiding force unwrapping. Is it assured this assertion is always true? 
I would propably move the HTTPURLResponse into a `if let` statement

r10s (approved):
seems okay, in theory we could differ between file-not-found and empty-file, however, as this is only about webxdc in the .xdc archive, empty files do not really make sense;  also, at least android has the same behavior, so we make discouraging empty files in the archive is fine.

great you figured out this bug, @Simon-Laux!

PR #1718, opened 2022-10-06
improve webxdc js to prevent races

Simon-Laux:
that were possibly introduced by https://github.com/deltachat/deltachat-ios/pull/1715#discussion_r989488545

### DIFF
--- a/deltachat-ios/Controller/WebxdcViewController.swift
+++ b/deltachat-ios/Controller/WebxdcViewController.swift
@@ -62,24 +62,37 @@ class WebxdcViewController: WebViewViewController {
         
         let script = """
         window.webxdc = (() => {
-          let setUpdateListenerPromise = null
           var log = (s)=>webkit.messageHandlers.log.postMessage(s);
         
           var update_listener = () => {};
 
-          window.__webxdcUpdate = async (lastSerial) => {
+          let should_run_again = false;
+          let running = false;
+          let lastSerial = 0;
+          window.__webxdcUpdate = async () => {
+            if (running) {
+                should_run_again = true
+                return
+            }
+            if (should_run_again) {
+                should_run_again = false
+            }
+            running = true;
             try {
-                const updates = await fetch("webxdc-update.json?"+lastSerial).then((response) => response.json())
+                var updates = await fetch("webxdc-update.json?"+lastSerial).then((response) => response.json())
                 updates.forEach((update) => {
                   update_listener(update);
+                  if (lastSerial < update["max_serial"]){
+                    lastSerial = update["max_serial"]
+                  }
                 });
             } catch (e) {
                 log("json error: "+ e.message)
             } finally {
-              if (setUpdateListenerPromise) {
-                 setUpdateListenerPromise()
-                 setUpdateListenerPromise = null
-              }
+                running = false;
+                if (should_run_again) {
+                    window.__webxdcUpdate()
+                }
             }
           }
 
@@ -90,11 +103,7 @@ class WebxdcViewController: WebViewViewController {
         
             setUpdateListener: (cb, serial) => {
                 update_listener = cb
-                const promise = new Promise((res, _rej) => {
-                   setUpdateListenerPromise = res
-                })
-                webkit.messageHandlers.setUpdateListener.postMessage(typeof serial === "undefined" ? 0 : parseInt(serial));
-                return promise
+                return window.__webxdcUpdate()
             },
 
             getAllUpdates: () => {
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -62,24 +62,37 @@ class WebxdcViewController: WebViewViewController {
...
+          let running = false;
+          let lastSerial = 0;
+          window.__webxdcUpdate = async () => {
+            if (running) {
+                should_run_again = true
+                return
+            }
+            if (should_run_again) {
+                should_run_again = false
+            }
```
r10s, line 79: that could be just `should_run_again = false` unconditionally, or?

r10s, line 79:
```suggestion
            should_run_again = false
```

i think, the condition is not needed.

deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -62,24 +62,37 @@ class WebxdcViewController: WebViewViewController {
...
+            running = true;
             try {
-                const updates = await fetch("webxdc-update.json?"+lastSerial).then((response) => response.json())
+                var updates = await fetch("webxdc-update.json?"+lastSerial).then((response) => response.json())
```
r10s, line 82: why is `const` no longer correct?

r10s (commented):
i think, the race is gone by this pr; alone that there is only one call to `getWebxdcStatusUpdates` left lgtm. also, `__webxdcUpdate()` has the same signature as on android.

this is definitely an improvement, +1

however, it is a pity that now two new states (`running` and `should_run_again`) are needed.

i am also not sure, that they are correct in this form: the doc of [`setUpdateListener`](https://docs.webxdc.org/spec.html#setupdatelistener) says _"The returned promise resolves when the listener has processed all the update messages known at the time when setUpdateListener was called."_ - now, there could be more updates processed if during processing new updates come in.  
maybe this is okay, it is not so easy to think about all this stuff :) also not sure what to do on that without adding more complexity. 

what exactly was the initial issue with base64? :D

deltachat-ios/Controller/WebxdcViewController.swift
Simon-Laux: const is correct, but apple gave me an unhelpful message about the await keyword (I forgot the async keyword above and it did not tell me that instead it complained about the fetch keyword, so I was confused and tried all kinds of things out, switching to var was one of those things

deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -62,24 +62,37 @@ class WebxdcViewController: WebViewViewController {
...
             } catch (e) {
                 log("json error: "+ e.message)
             } finally {
-              if (setUpdateListenerPromise) {
-                 setUpdateListenerPromise()
-                 setUpdateListenerPromise = null
-              }
+                running = false;
+                if (should_run_again) {
+                    window.__webxdcUpdate()
```
Simon-Laux, line 94: > i am also not sure, that they are correct in this form: the doc of [setUpdateListener](https://docs.webxdc.org/spec.html#setupdatelistener) says "The returned promise resolves when the listener has processed all the update messages known at the time when setUpdateListener was called." - now, there could be more updates processed if during processing new updates come in.

should be fixed by adding a second await here?
```suggestion
                    await window.__webxdcUpdate()
```

deltachat-ios/Controller/WebxdcViewController.swift
Simon-Laux:
```suggestion
                const updates = await fetch("webxdc-update.json?"+lastSerial).then((response) => response.json())
```

deltachat-ios/Controller/WebxdcViewController.swift
Simon-Laux: yes, you are correct here, thanks

Simon-Laux:
> however, it is a pity that now two new states (running and should_run_again) are needed.

it is how it is, how else could you insure the function only has one running instance at a time?

cyBerta:
> > however, it is a pity that now two new states (running and should_run_again) are needed.
> 
> it is how it is, how else could you insure the function only has one running instance at a time?

one could use an update counter, but I think it's fine to have two booleans instead

deltachat-ios/Controller/WebxdcViewController.swift
r10s: can you elaborate why the `await` solves the issue?

r10s:
> it is how it is, how else could you insure the function only has one running instance at a time?

yip, if it is needed at the end, then this is it, yes. however, it is still unclear to me how base64 can fail so badly, i mean, this seems the main reason for the whole effort.

deltachat-ios/Controller/WebxdcViewController.swift
Simon-Laux: I'm sorry, but I don't really see any problem here.

r10s: i am not saying that there are problems.

i just want to understand - imu, this await will wait until the promise is resolved and then return the promise from the outer `__webxdcUpdate()` call - this seems wronger than without the `await` as there could be more updates processed if during processing new updates come in.

what happens if the `await` was left out as before? would the function be scheduled at some later point? that would probably what we want to have here.

we can also do an av on that.

r10s (approved):
probably okay :)

PR #1721, opened 2022-10-10
fix switch account layout

### LINKED ISSUE #1719: Account switch icon character is wrong
The new switch avatar icon shows an "M", probably for 'me' instead of the first character of the display name like in the settings. 

correct:
bug-01a

wrong:
bug-01b

### LINKED ISSUE #1720: Account Switch Layout Bug on iOS 12
Simulator Screen Shot - iPhone SE (1st generation) - 2022-10-07 at 16 32 39

that doesn't look right :)
We need to check the backwards compatibility for other iOS versions and could consider to remove it completely for iOS 12 and earlier, if adding constraints doesn't help as [proposed](https://github.com/deltachat/deltachat-ios/pull/1687#discussion_r985087398)

cyBerta:
* moves the unread message counter to the upper right of the avatar, adding some padding so that more of the avatar is still visible when the unread message counter is shown
* fixes layout issues on iOS 12 (fixes #1720)
* fixes wrong initals in avatar (fixes #1719)

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -551,8 +551,9 @@ class ChatListController: UITableViewController, AccountSwitcherHandler {
         accountButtonAvatar.setUnreadMessageCount(getUnreadCounterOfOtherAccounts())
         
         let contact = dcContext.getContact(id: Int(DC_CONTACT_ID_SELF))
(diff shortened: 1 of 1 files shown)


Simon-Laux (approved):

r10s:
great!

PR #1723, opened 2022-10-17
Switch account layout

### LINKED ISSUE #1712: Account Switch View Controller
With https://github.com/deltachat/deltachat-ios/pull/1687 we introduced a way to quickly switch accounts from the chat list ui. This ticket is a follow-up to replace the currently shown alert controller with a custom UIViewController with a similar UI to: 
[Safari's Tab selector](https://support.delta.chat/uploads/default/original/2X/9/986bbba3e9f0e2ad7fc9298610ec5cf440acd4a8.png).

When we're at it, we can also change the location of the unread message counter on the avatar to the upper right, which seems a little bit more common on iOS.

cyBerta:
closes #1712


Simon-Laux (approved):

r10s:
looks good already! great you push that forward.

one thing that came immediately to my mind is that it is not always obvious to see that there are more accounts, so that you actually can scroll - depending on the screen size, eg. four accounts fit perfectly, no hint for the fifth - until you scroll.

apple solved this issue in the "tab switcher" by having a border around the items to select, this has rounded corners, so it is clear, that if one does not see the bottom rounded corners, there is sth. to scroll:



the example has two groups, we would have only one, maybe we can put the "add account" to an extra group, but this is cosmetic only, not important.

i think, we should do sth similar. also i agree with @Simon-Laux and would do the account switch lines a bit narrower. maybe take the upper left account avatar as a pattern and reuse that size - this would make it also more clear that these are no chats but accounts and there would be a clearer distinction to the partly visible chatlist.

Simon-Laux:
There also was a crash when deleting an account.
image

infos to reproduce it:
while deleting an account that still fetches messages, I encountered this crash. the deleted account was the one that was currently selected. pressing cancel did not work, and sometimes I could still see the deleted account and delete it again, so that is not the only bug it seems.

r10s:
looks much nicer already, much better with the rounded corner and the smaller heights.

wondering, however, if we missed some system support for these kind of controls; eg. usually the colors are swapped, so outside is gray and controls are white (see example above) and also when tapping the hilighting is not nice (it is larger than the control with the rounded corners).  
so, maybe double check if there is some more system-like-way to do this.

cyBerta:
> wondering, however, if we missed some system support for these kind of controls; eg. usually the colors are swapped, so outside is gray and controls are white (see example above)

Even though these elements look pretty standard I don't think they are available out of the box for app developers. Of course we can adapt the colors. 

We didn't have yet the trash icon. In other places we use a "Delete" as text as an alternative for pre iOS13. We could do that here as well, however that wouldn't fix the checkmark, having "checked" as a textual representation doesn't look very pretty.

cyBerta:
I will try to improve the selected state for the cells

r10s:
> There's a default system layout beginning from iOS13 that allows pretty [exactly what we want](https://developer.apple.com/documentation/uikit/uitableview/style/insetgrouped). What do we do with pre iOS13 though? no background at all?

cool you figured that out, i'd go for that. and i would not think too much about ios13 if it is not perfect there.

cyBerta:
@r10s I've adapted the PR accordingly

r10s (approved):
looks great!

PR #1725, opened 2022-10-21
rename settings sourcecode and group them

Simon-Laux:
- move settings to own group/folder
- rename files and sort them also document the naming rules that I applied


deltachat-ios/Controller/Settings/SettingsViewController.swift
```diff
@@ -572,7 +572,7 @@ internal final class SettingsViewController: UITableViewController, ProgressAler
     }
 
     private func showVideoChatInstance() {
-        let videoInstanceController = SettingsVideoChatViewController(dcContext: dcContext)
+        let videoInstanceController = VideoChatViewController(dcContext: dcContext)
```
r10s, line 575: better call this `VideoChatInstanceViewController` or `VideoChatSettingsViewController` or so, otherwise, at a first glance one would think that this shows the video etc.

r10s (commented):
great improvement!

- as mentioned, an `Off` setting would be welcome, esp. this is a nicer default when the option was never used before.

- the `ProfileSettingsViewController` should show the selected name and not the URL.

finally, maybe better merge this to `master`, so this gets independent from the renaming discussion and does not make the other pr too large.

deltachat-ios/Controller/Settings/Pages/README.md
```diff
@@ -0,0 +1,10 @@
+#  Settings Pages
+
+This group is for settings pages that combine multiple settings on one page.
```
r10s, line 3: by this definition, `AutodeletionOverviewViewController` is a page and not a single setting, or?

tbh, i am not convinced by trying to sort all setting view controllers into "single" and "pages" as artificial categories, it adds bureaucracy and discussions, for a handful of files. 

so, +1 for streamlining the names, but a flat structure seems "good enough" here.

Simon-Laux: I think it's faster to find, if you search for a page, but yeah `AutodeletionOverviewViewController` should be moved.

r10s: it is only faster to find if you know beforehand, if sth. is categorized as page or a single option. and you have to understand these categories. also, things may move between them ... eg. the VideoChatViewController you just updated may easily become a "page". we can think over it if we have several dozens of files and controllers, let us just keep things simple here :)

I understand the goal you have here, but i don't think i agree that it hits the correct set of tradeoffs.

r10s:
replaced by https://github.com/deltachat/deltachat-ios/pull/1739 which pulls in much of the things discussed here.

PR #1726, opened 2022-10-21
add default providers we have in desktop

Simon-Laux:
they were asked for permission as far as I know.

Edit: checkmark works now as expected, we have now an "Off" option and also note that custom value get persisted in `ui.custom_webrtc_instance` so users don't loose what they typed in there when trying default instances out (we should add that behaviour on desktop too).

### DIFF
--- a/deltachat-ios/Controller/SettingsVideoChatViewController.swift
+++ b/deltachat-ios/Controller/SettingsVideoChatViewController.swift
@@ -2,18 +2,17 @@ import UIKit
 import DcCore
 
 class PredefinedVideoChatOptionCell: UITableViewCell {
-    
+
     public var url: String
-    
+
     init(label: String, url: String) {
         self.url = url
         super.init(style: UITableViewCell.CellStyle.subtitle, reuseIdentifier: label)
         self.textLabel?.text = label
         self.detailTextLabel?.text = url
         self.detailTextLabel?.textColor = .lightGray
-        
     }
-    
+
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
@@ -29,13 +28,13 @@ class SettingsVideoChatViewController: UITableViewController {
         PredefinedVideoChatOptionCell(label: "Systemli", url: "https://meet.systemli.org/$ROOM"),
         PredefinedVideoChatOptionCell(label: "Autistici", url: "https://vc.autistici.org/$ROOM"),
     ]
-    
+
     private lazy var offCell: UITableViewCell = {
         let cell = UITableViewCell(style: UITableViewCell.CellStyle.default, reuseIdentifier: "off")
         cell.textLabel?.text = String.localized("off")
         return cell
     }()
-    
+
     private lazy var customInstanceCell: TextFieldCell = {
         let cell = TextFieldCell.makeConfigCell(labelID: String.localized("custom"),
                                                 placeholderID: String.localized("videochat_instance_placeholder"))
@@ -58,7 +57,7 @@ class SettingsVideoChatViewController: UITableViewController {
     required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
-    
+
     // MARK: - Table view data source
 
     override func numberOfSections(in tableView: UITableView) -> Int {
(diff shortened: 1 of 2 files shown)


deltachat-ios/Controller/Settings/SingleSettings/VideoChatViewController.swift
```diff
@@ -1,21 +1,50 @@
...
 class VideoChatViewController: UITableViewController {
 
     private var dcContext: DcContext
 
-    private lazy var videoInstanceCell: TextFieldCell = {
-        let cell = TextFieldCell.makeConfigCell(labelID: String.localized("videochat_instance"),
+    private let defaultOptions = [
+        PredefinedVideoChatOptionCell(label: "Jitsi", url: "https://meet.jit.si/$ROOM"),
+        PredefinedVideoChatOptionCell(label: "Systemli", url: "https://meet.systemli.org/$ROOM"),
+        PredefinedVideoChatOptionCell(label: "autistici", url: "https://vc.autistici.org/$ROOM"),
```
r10s, line 30:
```suggestion
        PredefinedVideoChatOptionCell(label: "Autistici", url: "https://vc.autistici.org/$ROOM"),
```

deltachat-ios/Controller/Settings/SingleSettings/VideoChatViewController.swift
```diff
@@ -31,23 +60,74 @@ class VideoChatViewController: UITableViewController {
...
+            customInstanceCell.accessoryType = .checkmark
+            customInstanceCell.textField.text = currentUrl
+        } else {
+            customInstanceCell.accessoryType = .none
+            customInstanceCell.textField.text = dcContext.getConfig("ui.custom_webrtc_instance")
+        }
     }
 
     override func tableView(_ tableView: UITableView, titleForFooterInSection section: Int) -> String? {
         return String.localized("videochat_instance_explain_2") + "\n\n" + String.localized("videochat_instance_example")
```
r10s, line 124: wondering if the example string is still needed - `basicwebrtc`, as mentioned in the example string is not really a thing currently - and the jitsi example is shown already above.

r10s (commented):
this is a great improvement!

- [ ] as said the first option should be "Off", as on desktop - and also to have a nicer default when opening the dialog for the very first time

- [ ] better rebase that on `master` so we can merge this indepentently of renaming discussions - and without makeing the renaming pr too large

- [ ] the overview page should show the Name instead of the URL if one of the predefined option is selected (iterate over the predefined values, on match, show name - otherwise, for the custom setting, show the URL)

deltachat-ios/Controller/Settings/SingleSettings/VideoChatViewController.swift
Simon-Laux: probably not, but I'm too lazy to remove it. maybe we should explain what `$ROOM` does instead?

r10s: yip, let's keep the example string and update the string for all platforms at some point.

cyBerta:
Btw. the tech collective's name is actually Autistici/Inventati or A/I, see also https://www.inventati.org or https://www.autistici.org/

cyBerta:
> they were asked for permission as far as I know.

👍

r10s (approved):
lgtm, i rebased the pr to master and will probably also adapt the preview issue mentioned above.

deltachat-ios/Controller/SettingsVideoChatViewController.swift
```diff
@@ -1,53 +1,156 @@
...
+        return cell
+    }()
+
+    private lazy var customInstanceCell: TextFieldCell = {
+        let cell = TextFieldCell.makeConfigCell(labelID: String.localized("custom"),
                                                 placeholderID: String.localized("videochat_instance_placeholder"))
         cell.textField.autocapitalizationType = .none
         cell.textField.autocorrectionType = .no
         cell.textField.textContentType = .URL
+        cell.textField.addTarget(self, action: #selector(setCustom), for: .editingDidBegin)
```
cyBerta, line 44: why this one? It will save the existing settings again, before it was changed. I doubt this is necessary

r10s: it makes sure that the checkmark is moved to "custom" when the text field is tapped. otherwise, you have to tap a second time on the label or so which is not so intuitive.

cyBerta (approved):

PR #1734, opened 2022-10-31
handle mailto addresses in WebViewControllers

### LINKED ISSUE #1724: support `mailto:` links inside html messages
this is needed for bots, however, also useful in general.

cyBerta:
closes #1724

### DIFF
--- a/deltachat-ios/Controller/WebViewViewController.swift
+++ b/deltachat-ios/Controller/WebViewViewController.swift
@@ -1,5 +1,6 @@
 import UIKit
 import WebKit
+import DcCore
 
 class WebViewViewController: UIViewController, WKNavigationDelegate {
 
@@ -46,6 +47,7 @@ class WebViewViewController: UIViewController, WKNavigationDelegate {
     private var debounceTimer: Timer?
     private var initializedSearch = false
     open var allowSearch = false
+     var dcContext: DcContext
 
     open var configuration: WKWebViewConfiguration {
         let preferences = WKPreferences()
@@ -59,7 +61,8 @@ class WebViewViewController: UIViewController, WKNavigationDelegate {
         return config
     }
 
-    init() {
+    init(dcContext: DcContext) {
+        self.dcContext = dcContext
         super.init(nibName: nil, bundle: nil)
         hidesBottomBarWhenPushed = true
     }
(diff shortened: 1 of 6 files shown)


deltachat-ios/Controller/WebViewViewController.swift
```diff
@@ -192,6 +202,47 @@ class WebViewViewController: UIViewController, WKNavigationDelegate {
...
         updateAccessoryBar()
     }
+
+    func openChatFor(url: URL) {
+        guard let appDelegate = UIApplication.shared.delegate as? AppDelegate,
+              let emailAddress = parseEmailAddress(from: url) else {
+            return
+        }
+
+        // FIXME: lookupContactIdByAddress is still broken
```
r10s, line 212: what precisely is broken?

deltachat-ios/Controller/WebViewViewController.swift
```diff
@@ -192,6 +202,47 @@ class WebViewViewController: UIViewController, WKNavigationDelegate {
...
+        // let contactId = dcContext.lookupContactIdByAddress(emailAddress)
+
+        // workaround:
+        let contacts: [Int] = dcContext.getContacts(flags: DC_GCL_ADD_SELF, queryString: emailAddress)
+        let index = contacts.firstIndex(where: { dcContext.getContact(id: $0).email == emailAddress }) ?? -1
+        var contactId = 0
+        if index >= 0 {
+            contactId = contacts[index]
+        }
```
r10s, line 221:
```suggestion
        let contactId = index >= 0 ? contacts[index] : 0
```

unmutable is always better :)

r10s (approved):
also tested a bit, works as expected :)

deltachat-ios/Controller/WebViewViewController.swift
r10s: /me suggesting to add these information to the code comment :)

cyBerta: from what I tested, ` dcContext.lookupContactIdByAddr(addr)` just always returns 1.  Exact the same work-around has been used on iOS in a different spot already, that's why 'still'.

r10s: hm, we should file an issue for that, from core-side, that should work.

android uses `lookupContactIdByAddr` as well and checks if a chat with that contact exists (which is more accurate than checking if the contact exist in the addressbook):

https://github.com/deltachat/deltachat-android/blob/master/src/org/thoughtcrime/securesms/NewConversationActivity.java#L156

PR #1739, opened 2022-11-03
sort and rename some ViewControllers

r10s:
- sort ViewControllers to "Settings" dir (new) and to "AccountSetup" (already existant)

- review names of the settings ViewControllers; to keep the scope narrow, the other names stay as is

this pr replaces https://github.com/deltachat/deltachat-ios/pull/1725


cyBerta (approved):
lgtm

PR #1743, opened 2022-11-10
[postponed] better background picker, that uses the phone screen size as aspect ratio for croping

### LINKED ISSUE #1502: Chat-background-selection is not intuitive
The selection crops out a square while the chat background area is a high rectangle.
This makes it impossible for users to select what exact section of an image they want to select.

image_2022-03-19_12-30-42

### LINKED ISSUE #1688: setting chat background crops to square
probably because it reuses cropping for avatar or sth?

it should crop to screen/device aspect ratio in portrait mode instead, because that would make the preview more accurate to the actual result.

Simon-Laux:
fixes #1502
fixes #1688


r10s:
i see the gist of this pr, and maybe that makes sense, see discussions in the corresponding issues. nice that selection in landscape is possible a well (ipad) so that this mode is not deteriorated. so, for the changed functionality i am +/-0, even slightly positive :)

what is more questionable is that we pull in a [bigger dependency](https://github.com/TimOliver/TOCropViewController) - even though it is used by bigtech and also looks maintained in general, there are quite obvious bugs (rotate two times, pixels are lost), and probably some more hidden ones, this can cause quite some work wrt support, bugs etc.

apart from that, we would have two crop thingies, which is not nice wrt UX and maintainability.

i am not totally sure where the existing crop mechanism one comes from and why this crops always squares - but we should at least consider to just tweak that part. @cyBerta do you know more about that?

@Simon-Laux i did not dived through the code in detail - but the rendering inside the viewController is not changed, right?

Simon-Laux:
> there are quite obvious bugs (rotate two times, pixels are lost)

probably because it zooms in to keep the fixed aspect ratio?


> apart from that, we would have two crop thingies, which is not nice wrt UX and maintainability.

I plan to replace the other one as well eventually, this lib has even a round one that would fit well with avatar picture selection.

> i am not totally sure where the existing crop mechanism comes from and why this crops always squares

it comes from the systems image picker and it is not possible to change the aspect ratio, at least my search resulted in nothing, just people telling it's impossible on stackoverflow.

>  but the rendering inside the viewController is not changed, right?

it's a modal that is presented on top of the background screen if that is what you mean? though we could eventually do a custom one like telegram that represents a chat screen?

r10s:
nb: what whatsapp & co are doing here, is to skip the cropper view completely - and say in the selection view that you can  pinch-to-zoom.

r10s:
> I plan to replace the other one as well eventually, this lib has even a round one that would fit well with avatar picture selection.

there are but reports with that as well, https://github.com/TimOliver/TOCropViewController/issues/530

also quite some other bugs sound serious, freeze on ios, m1 building issues - and also the bug of rotating i mentioned above is filed but not commented on since Aug22 (maybe new with some ios update).

so, i do not think, adding this dependency hits the correct set of tradeoffs, as discussed above. using the system library seem to be better here: less bugs, no dependency, more secure, future proof. also, it looks more integrated. in general, we try to get rid of dependencies and add new ones only if it makes things _much_ better - and not better at one point and worse at others.

but it was still good to check the library out, definitely.

PR #1746, opened 2022-11-15
webxdc internet_access flag

### LINKED ISSUE #1744: take webxdc `internet_access` flag into account
UIs should check the field `internet_access` from [`dc_msg_get_webxdc_info()`](https://c.delta.chat/classdc__msg__t.html#aee526949983962ed25adf33a2f5e6878) - if that is true, internet access should be allowed for that webxdc, see https://github.com/deltachat/deltachat-core-rust/pull/3516 for discussions and reasoning.

respecting the flag is concretely needed to drive some experiments forward.

cmp. https://github.com/deltachat/deltachat-android/pull/2378, https://github.com/deltachat/deltachat-desktop/issues/3005

cyBerta:
load content without any restrictions if webxdc has set internet_access flag and is sent to 'Saved Messages'
closes #1744

### DIFF
--- a/deltachat-ios/Controller/WebxdcViewController.swift
+++ b/deltachat-ios/Controller/WebxdcViewController.swift
@@ -15,6 +15,7 @@ class WebxdcViewController: WebViewViewController {
     var webxdcUpdateObserver: NSObjectProtocol?
     var webxdcName: String = ""
     var sourceCodeUrl: String?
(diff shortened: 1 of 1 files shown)


r10s (approved):
great, straight forward, nice!

i tested that with https://github.com/r10s/maps - works like a charm from "Saved MessagesQ"

PR #1749, opened 2022-11-17
precheck with DC_INFO_WEBXDC_INFO_MESSAGE

r10s:
this saves one database call if the info is not a webxdc info message, only a minor improvement, but still :)

in theory, the additional checks against `isInfo` and `type`, not totally sure in practice, so i left them for resilience :)


cyBerta (approved):
You could remove parent.type == DC_MSG_WEBXDC. That should work.

PR #1755, opened 2022-11-24
replace DBDebugKit with simple ViewController

cyBerta:
* all we keep is a simple view contoller showing our custom variables which we use for debugging
* this PR removes a dependency that doesn't work 100% (crash log functionality doesn't work) in our case and might cause crashes if the app was started from the background and is about to be shut down after fetching messages in the background.


cyBerta:
removing **24,689** lines of code :)

r10s (approved):
nice, less dependencies are always good!

did not test due to my macbook being in repair, however, the code looks good. so please merge in as needed - iirc, next step may be to get rid of swifty beaver, which is also not really used by us.

PR #1757, opened 2022-11-25
Fix crash on copy message with iOS 14.8

rosano:
It seems that setting the pasteboard content to `nil` crashes consistently on iOS 14.8. Setting it to an empty string should not conflict with other things?


cyBerta (approved):
nice catch! <3

r10s:
thanks for quick fixing and merging!

PR #1758, opened 2022-11-26
show audio recorder on half screen

### LINKED ISSUE #1747: half screen voice message recording controller
With https://github.com/deltachat/deltachat-ios/issues/1712 https://github.com/deltachat/deltachat-ios/issues/1661 and https://github.com/deltachat/deltachat-ios/issues/1662 the voice message recording UI could also appear only on half a screen, especially since the audio recording screen itself has only a few UI components. 

In contrast to that little change, that makes the UI a little bit more streamlined, reworking the whole audio recording UI is a much bigger task. So it's a low hanging fruit until that can be easily archieved.

cyBerta:
closes #1747


r10s (approved):
lgtm

Issue #1691, opened 2022-09-20
just sent message is not always added to chat view


r10s:
sometimes, a just sent message is not shown the chat view (it is not only not scrolled down).

leaving and re-entering the chat shows the message and the message is also sent as expected.

i do not know how to reproduce that, but that happens to me every day, so maybe for 1 of 50 messages.

also, i can say that the issue is quite new, i noticed that the last weeks.

cyBerta:
I wonder if that could be related to a missing event in core?

Simon-Laux:
just had this, could also be a scrolling bug, because it was there once I scrolled down.

Simon-Laux:
in #1969 there is a state where events are broken, there it happens all the time, so I also think this is about a missing event.

Issue #1737, opened 2022-11-01
Bottom of message gets cut off


snan:
If I have an external keyboard plugged in over USB, I can't see the lowest part of the most recent message. I can pull it up temporarily to read it, but as soon as I let go, it snaps back. Most importantly, I can not follow links in there.

![IMG_1697](https://user-images.githubusercontent.com/309417/199307338-596bf56d-e418-45df-bd6d-72f04baf4e70.png)

cyBerta:
it looks like this issue also exists on some iphones, which have the long black bottom swipe UI element (I don't recall the name right now) which acts as an replacement of the home btn.  And it seems to be related to ios 16

r10s:
yip, it is not only ipad os meanwhile ... needs to be fixed soon and should go to the next update :)

MrPetovan:
I have witnessed some improvement with regards to this issue: The default view still cuts off the bottom of the conversation, but scrolling it into view now keeps it into view, there's no pull back behind the input bar.

Simon-Laux:
![image](https://github.com/deltachat/deltachat-ios/assets/18725968/a4f8bb3d-dd99-4a18-818b-758d2f12cdf2)
another example of this bug

PR #1767, opened 2022-12-08
Draft: improve scrolling behavior

cyBerta:
Trying to figure out the root cause for the weird scroll behavior. This PR is only to discuss possible changes. 
The reason I started investigating this is #1691. The fix for that issue is fairly simple, see [af5cefe](https://github.com/deltachat/deltachat-ios/pull/1767/commits/af5cefee713af559e90af2b18d27df3bb214ed3c), if we take the DC Android as an template. However the weird scrolling down and back up when sending a message always happens now. I will explain a little bit more in the comments.

Generally I've noticed that the scrolling-down and up happens, because the messageInputBar looses it's focus when the user taps on the send button. As a result normally the keyboard would disappear. We mitigate that by [re-requesting the focus for the messageInputBar](https://github.com/deltachat/deltachat-ios/pull/1767/files#diff-bf4dd3d619f49250044e05ac5777f82c9b4029e49d8d175f6e2bbfefc1fe4685R2461) when the messageInputBar's UITextViewDelegate calls back textViewShouldEndEditing(). Normally that is sufficient to keep the keyboard in place, on some devices and OS versions you may see a very small flickering. 
However the frame size of the UITableView immediately changes and all cells at the bottom get moved for a short moment to the bottom of the screen, before the size frame size regains it's previous size - above the keyboard.

While experimenting and reading on [stack overflow for possible solutions](https://stackoverflow.com/questions/50459798/increase-the-height-of-the-tableview-when-the-keyboard-disappears) I learned that we can use `contentInset` to avoid the content is dragged down to the bottom of the screen. I vaguely remember we already worked with `contentInset`s in the past in this controller, but scratched that approach because of other issues we had with it like wrong initialization.

My current idea is - contrary to what we tried in the past regarding `contentInset`s - to set the content inset _only temporarily_ at the bottom before the keyboard events are thrown / the tableViewFrame changes and reset it as soon as the reload() was done and the new message bubble appeared (but is still hidden below the keyboard). 

What's not yet clear to me is if any call to scrollDown would be still necessary after resetting the contentInset, or maybe after animating the contentInset value.

What I can confirm so far is that setting the contentInset in `textViewShouldEndEditing` (= before the keyboard events are thrown / before the tableViewFrame size changes) indeed prevents the messages being moved down if I [comment out scrolling completely.](https://github.com/deltachat/deltachat-ios/pull/1767/files#diff-bf4dd3d619f49250044e05ac5777f82c9b4029e49d8d175f6e2bbfefc1fe4685L1149-L1160)

### DIFF
--- a/deltachat-ios.xcodeproj/project.pbxproj
+++ b/deltachat-ios.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		21D6C941260623F500D0755A /* NotificationManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21D6C9392606190600D0755A /* NotificationManager.swift */; };
+		3001B560294265C200816B0C /* AtomicBoolean.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3001B55F294265C200816B0C /* AtomicBoolean.swift */; };
 		3008CB7224F93EB900E6A617 /* AudioMessageCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3008CB7124F93EB900E6A617 /* AudioMessageCell.swift */; };
 		3008CB7424F9436C00E6A617 /* AudioPlayerView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3008CB7324F9436C00E6A617 /* AudioPlayerView.swift */; };
 		3008CB7624F95B6D00E6A617 /* AudioController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3008CB7524F95B6D00E6A617 /* AudioController.swift */; };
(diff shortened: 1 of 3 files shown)


cyBerta:
closes also #1691

r10s:
ftr: we decided to tackle this together with the other https://github.com/deltachat/deltachat-ios/labels/message-list issues beginning of february.

r10s:
after playing around a little, this pr does not really fix the message-list issued, it maybe mitigates them, but there is still:

- jumping up/down
- input bar laying over bubbles sometimes (this is easily reproducible by entering a chat, swiping a little up)
- keyboard does not always appear

we will probably go for a bit larger rewrite in #1810 therefore.

r10s:
closing stale draft, superseded by #1810

r10s:
cmp https://github.com/deltachat/deltachat-ios/issues/1737

PR #1770, opened 2022-12-13
Expand setup instructions

rosano:
Putting all the necessary commands in a visible order so newcomers can more easily follow them, and highlighting the workspace file more.


cyBerta (approved):
nice improvements

PR #1773, opened 2022-12-15
use regular heartbeat server also in DEBUG mode

### LINKED ISSUE #1772: drop heartbeat sandbox usage for DEBUG
the sandbox is used only if DEBUG is set, this is true at most for self-compiled apps created by some devs (testflight uses the same binaries as regular store).

however, if devs do use another heartbeat server than users, we will not get aware if things are down (or up again :)

r10s:
using the sandbox on a few developer devices does not make sense as that way devs do not get aware of issues
if the regular heatbeat server is down - or up again.

also, getting rid of the sandbox saves us some maintainance effort.

closes #1772 

@missytake feel free to shut down the sandbox server :)


cyBerta (approved):
bye bye sandbox

PR #1781, opened 2023-01-03
Move rust-toolchain file from the core repository

link2xt:
This way updating the core does not automatically update toolchain version and any toolchain can be used as long as it can build the core.

Similar change for Android: https://github.com/deltachat/deltachat-android/pull/2440


r10s (approved):
thanks for taking care!

i adapted the path, force-pushed, and tested if it builds (it does :)

Issue #1780, opened 2022-12-30
No success feedback on backup export


narFnarF:
- Device: iPhone 6s
- iOS version:15.7.2
- Delta Chat version: 1.34.1
- Steps to reproduce the problem:
    1. In Delta Chat, go to Settings, Ecport Backup
    2. Tap Start backup
    3. Wait for the backup to be done
- Actual behavior: Delta Chat goes back to the list of messages. Nothing confirms if the backup was successful or failed. No instructions are provided on how/where to get or export this backup.
- Expected behavior: 
    1. A message confirm that the backup was successful
    2. The app provides instruction on how to access the backup or save it or transfer it
- Screenshots:
- Logs:

narFnarF:
I just found out that the backups are actually saved on the iphone. Go to the Files app, go in the Browse tab and tap the Back button until you're at the beginning and you see "On my iPhone". Tap on that, open the Delta Chat folder. Your backups are in there. You can use the Files app to share them, for example email them to yourself.

r10s:
@cyBerta i think, that fits somehow into the other backup issues you're currently targeting, can you please have a look at this one as well?

Simon-Laux:
> You can use the Files app to share them, for example email them to yourself.

only that they are unencrypted and contain your private key (and are large normally like 100mb to a few gb's depending on how much files you received/sent), so should be ideally transferred directly via usb.

Basically the reverse of https://support.delta.chat/t/import-backup-to-ios/1628.

We already have logic for encrypting backups, we just need to implement the field/dialog to ask the user for the password.

Also to solve this issue would a small text explanation in a dialog box suffice?

cyBerta:
@Simon-Laux I think it is a good idea to show the path of the exported account in a dialog box.

cyBerta:
The UI of for- and decrypting of backups need to be coordinated with all client platforms. That's beyond the scope here.

narFnarF:
When will I be able to try this improvement? In the next iOS release?

PR #1787, opened 2023-01-09
Backup Success dialog

cyBerta:
closes https://github.com/deltachat/deltachat-ios/issues/1780


r10s:
i think, the message dialog should get a title (just `export_backup_desktop`) so that the message goes to the message field and the dialog looks nicer and more normal.

instead of passing the title around with the message - i am wondering why we're not just showing the dialog in the already existing  `onSuccess` handler - that seems to be much less code changes we can more easily tweak the success dialog.

cyBerta:
You're right. I've adopted the MR

r10s (approved):

PR #1795, opened 2023-01-25
show verified marker behind chat title

### LINKED ISSUE #1759: show verification marker after contact/group name
from nov22 consistency discussions:

- Android always puts the v-mark after contact or group name,  and also in profile
- iOS always puts the v-mark on the avatar (group or contact), also in profile
- Desktop puts v-mark on Chat-avatar, and after contact names,  and v-mark in group-avatar in the group profile but not in the contact profile

Solution: make all platforms do the same thing, namely show v-mark after contact and group names, also in profiles and add a add a byline "verified" ([later "verified by ..."](https://github.com/deltachat/deltachat-core-rust/issues/3786)) to the profiles.

This also solves the visual conflict between "online" and v-marks.

cmp. https://github.com/deltachat/deltachat-desktop/issues/3035, https://github.com/deltachat/deltachat-core-rust/issues/3786

cyBerta:
closes #1759

### DIFF
--- a/deltachat-ios/View/ChatTitleView.swift
+++ b/deltachat-ios/View/ChatTitleView.swift
@@ -37,6 +37,7 @@ class ChatTitleView: UIView {
         subtitleLabel.translatesAutoresizingMaskIntoConstraints = false
         subtitleLabel.font = UIFont.systemFont(ofSize: 12)
         subtitleLabel.textAlignment = .center
+        subtitleLabel.setContentCompressionResistancePriority(.required, for: .horizontal)
         return subtitleLabel
     }()
 
--- a/deltachat-ios/View/ChatTitleView.swift
+++ b/deltachat-ios/View/ChatTitleView.swift
@@ -15,7 +15,7 @@ class ChatTitleView: UIView {
 
     private lazy var verifiedView: UIImageView = {
         let imgView = UIImageView()
-        let img = UIImage(named: "verified")?.scaleDownImage(toMax: 18)
+        let img = UIImage(named: "verified")?.scaleDownImage(toMax: 14.4)
         imgView.isHidden = true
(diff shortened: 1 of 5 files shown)


deltachat-ios/View/ChatTitleView.swift
```diff
@@ -3,20 +3,41 @@ import DcCore
...
         titleLabel.font = UIFont.systemFont(ofSize: 16, weight: .semibold)
         titleLabel.textAlignment = .center
+        titleLabel.setContentHuggingPriority(.defaultHigh, for: .horizontal)
         return titleLabel
     }()
 
-    private var subtitleLabel: UILabel = {
+    private lazy var verifiedView: UIImageView = {
+        let imgView = UIImageView()
+        let img = UIImage(named: "verified")?.scaleDownImage(toMax: 18)
```
r10s, line 18: this seems a bit too large, maybe we  can also use sth. as `titleLabel.font.pointSize * 0.9` here - that way it would also scale together with font size.

r10s (approved):
looks great!

visually, the only thing i've noticed is that the "verified marker" behind the chat  title in the chat is a little bit too large:

Screenshot 2023-02-06 at 16 15 26

_left: chat list, right: chat aka message list_

the left  one is fine, the right one should be smaller.

deltachat-ios/View/ChatTitleView.swift
cyBerta: Font sizes for the title bar are hard coded since UIController titles usually change their font size only minimally (mostly capable to reduce the seize, but not increase). I reduce the hard-coded value to 0.9 x 16 points (font size of title text size).

r10s (approved):

PR #1804, opened 2023-02-07
use menu- instead of reload-icon

### LINKED ISSUE #1803: "load images" icon in help view confusing
this is the "reload" icon, which maybe makes some sense, sth. like "my images are not shown, reload" -- however, most persons i speak to are just wondering :)

we should replace that, if in doubt, with the "three dot menu" icon used also elsewhere.

r10s:
the idea of the "reload" icon was
"oh, my images are missing, i hit reload, maybe they appear then", however, this idea did not really work out,
we we figured out by seveial user discussions :)

(the old icon is also read as "sync", which is even more confusing)

just use the menu icon, i think, this will lead to fewer questions (also, we may add more options that way :)

closes #1803

### DIFF
--- a/deltachat-ios/Controller/FullMessageViewController.swift
+++ b/deltachat-ios/Controller/FullMessageViewController.swift
@@ -5,9 +5,14 @@ import DcCore
 class FullMessageViewController: WebViewViewController {
 
     var loadButton: UIBarButtonItem {
-        let button = UIBarButtonItem(barButtonSystemItem: UIBarButtonItem.SystemItem.refresh, target: self, action: #selector(showLoadOptions))
+        let image: UIImage?
+        if #available(iOS 13.0, *) {
+            image = UIImage(systemName: "ellipsis.circle")
+        } else {
+            image = UIImage(named: "ic_more")
+        }
+        let button = UIBarButtonItem(image: image, style: .plain, target: self, action: #selector(showLoadOptions))
         button.accessibilityLabel = String.localized("load_remote_content")
-        button.tintColor = DcColors.primary
         return button
     }
 


deltachat-ios/Controller/FullMessageViewController.swift
```diff
@@ -76,14 +83,15 @@ class FullMessageViewController: WebViewViewController {
         let alert = UIAlertController(title: String.localized("load_remote_content"),
                                       message: String.localized("load_remote_content_ask"),
                                       preferredStyle: .safeActionSheet)
-        let alwaysAction = UIAlertAction(title: "\(alwaysCheckmark)\(String.localized("always"))", style: .default, handler: alwaysActionPressed(_:))
-        let neverAction = UIAlertAction(title: "\(neverCheckmark)\(String.localized("never"))", style: .default, handler: neverActionPressed(_:))
-        let onceAction = UIAlertAction(title: "\(onceCheckmark)\(String.localized("once"))", style: .default, handler: onceActionPressed(_:))
-
+        if isContactRequest {
+            alert.addAction(UIAlertAction(title: "\(neverCheckmark)\(String.localized("no"))", style: .default, handler: neverActionPressed(_:)))
```
cyBerta, line 87: If I read this block correctly, it can be deduplicated a bit further while still keeping readable code. 
e.g.
```
let reject = "\(neverCheckmark)\(String.localized(isContactRequest ? "no" : "never"))"
... adding defaults
if isContactRequest {
... adding last action
}

r10s: i avoided that initially as i do not wanted to have two `isContactRequest` conditions, however, i agree, this is a bit smarter overall, i added a commit. thanks!

Issue #1653, opened 2022-07-22
add an option to copy qr code data


r10s:
copying/pasting qr codes around is a simple and useful fallback if qr code scanning is not possible. adding this to ios is just feature parity with android/desktop in this area.

to copy a qr code to clipboard:

- in the upper right corner of the global "QR Invite Code" and also of the per-group-invite codes, add a "(...)" icon (same icon as eg. in webxdc or welcome screen)
- tapping the icon, open an action sheet with the options "Copy To Clipboard" and "Cancel"
- content to copy comes from `dc_get_securejoin_qr()`

~~to paste a qr code from clipboard:~~

- ~~in the upper right corner of the global "Scan QR Code", add a "(...)" icon~~
- ~~tapping the icon, open an action sheet with the options "Paste From Clipboard" and "Cancel"~~
- ~~the pasted code just goes through the existing check/process pipeline as if a qr code was just scanned.~~

EDIT: paste becomes less important, see discussion below

r10s:
i think, even by the new http://i.delta.chat , it still makes sense to target at least "copy to clipboard", to have the same experience on all platforms, cmp. https://github.com/deltachat/invite/issues/10 - switching later to a direct "copy invite-link to clipboard" seems to be easy then.

"paste from clipboard", oth seems not to be super-important with the new invite links; still makes sense for some corner-cases, sure, however, nothing to keep as an open issue (it was not really missed up to now).

PR #1805, opened 2023-02-08
copy qr code data to clipboard

r10s:
closes https://github.com/deltachat/deltachat-ios/issues/1653

needed for https://github.com/deltachat/invite/pull/12

### DIFF
--- a/deltachat-ios/Controller/QrViewController.swift
+++ b/deltachat-ios/Controller/QrViewController.swift
@@ -17,6 +17,16 @@ class QrViewController: UIViewController {
         return view
     }()
 
+    private lazy var moreButton: UIBarButtonItem = {
+        let image: UIImage?
+        if #available(iOS 13.0, *) {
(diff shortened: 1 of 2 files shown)


deltachat-ios/Controller/QrViewController.swift
```diff
@@ -43,6 +43,14 @@ class QrViewController: UIViewController {
         title = String.localized("qrshow_title")
         setupSubviews()
         view.backgroundColor = DcColors.defaultBackgroundColor
+
```
cyBerta, line 46: because of style reasons and consistency over the project I would initialize these UI elements lazily.

r10s: hm, yes, maybe, i also thought about that. but - we are not doing that everywhere and, tbh, i do not really get the gist of tearing code artificially that way :)

but regarding overall style and consistency, it is probably better.

cyBerta (approved):

PR #1808, opened 2023-02-08
try-to-add-longpress-handler-and-failed

Simon-Laux:


cyBerta:
alternative approach could be to always ask on a simple tap on a url link to either copy or open the link. That might make sense anyways, so that tapping on a link not automatically leads to opening a link in a browser.

r10s:
> alternative approach could be to always ask on a simple tap on a url link to either copy or open the link

hm, while understanding the idea, this is nothing what ppl usually expect when tapping a link; esp. as the link is already shown and there is no surprise in what will be opened.

we could maybe add a "Copy Link" item to the existing context menu when the text contains a link (any maybe more items if there are more links) ... but i think, that is also an ugly workaround.

i am positive that we figure out how to deal with long-taps at some point, however, this is noting urgent in any way just now.

r10s:
> this is also what telegram does for example (asking to open a link). So UX wise users might also know this behavior from other apps.

weird, have never seen that. maybe telegram want ppl to stay in their cage :)

cyBerta:
hmmm... the pop-up shows up if I long-tap in TG iOS, my fault.

Simon-Laux:
I would really appreciate having this feature in my day to day use, @cyBerta do you have an idea on where to listen for the long tap event on links?

r10s:
moving stale draft to resurrection

PR #1810, opened 2023-02-15
Draft: add basic ChatView UIViewController 

cyBerta:
* ... containing a tableView and a dummy TextInputField. 
* Animate TextInputField position on keyboard appearing/disappearing

TODO:

* [ ] fix message input bar max textview height on orientation change
* [ ] ...

### DIFF
--- a/DcCore/DcCore/Extensions/UIView+Extensions.swift
+++ b/DcCore/DcCore/Extensions/UIView+Extensions.swift
@@ -281,6 +281,14 @@ public extension UIView {
         return constraint
     }
 
+    func constraintMinHeightTo(_ height: CGFloat, priority: UILayoutPriority? = .none) -> NSLayoutConstraint {
+        let constraint = heightAnchor.constraint(greaterThanOrEqualToConstant: height)
+        if let priority = priority {
+            constraint.priority = priority
+        }
+        return constraint
+    }
+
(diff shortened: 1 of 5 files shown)


r10s:
tested only in the emulator, showing the new chat controller, there is already the lag at the bottom of the screen, so, things do not get displayed instantly (as true eg. in the settings or in other apps) - i think, we should get a grip on that first:

have a look to the bottom part of the screen: although, also for the old controller, the bubbles are already there, the overlay changes half a second after the screen is there:

https://user-images.githubusercontent.com/9800740/219119599-5f9fad51-98f2-4d20-9515-8567e0e42adc.mov

cyBerta:
I haven't seen this issue before, because I tested with emulators and devices that don't have this touchbar outside of the safe layout area of the screen. The reason why this bug appears seems to be related to the fact that these outside-areas are not updated during the transition. Please have a look at the current state again.  I kept the commits small, so that the intention for the changes become more obvious.

r10s:
in the emulator, things look good - where they were buggy before as well. so, with high chance, the last commits changed the correct things :)

i will try my real device tomorrow or monday, currently checking there if i got crashed wrt https://github.com/deltachat/deltachat-ios/issues/1202#issuecomment-1435329483

r10s:
closing stale pr, we'll try to fix https://github.com/deltachat/deltachat-ios/issues/1737 more on point

PR #1817, opened 2023-02-25
Rephrase to imperative, not past tense

snan:


r10s:
thanks a lot, @snan! i corrected the string on transifex and pulled the strings from there, see https://github.com/deltachat/deltachat-ios/commit/0af034c5fdde28f545cf35d799567955f53e2082 (otherwise, your change to https://github.com/deltachat/deltachat-ios/pull/1817#diff-d69dc2bb9500f86057f63d7dd1c06102a50ee1655c52da641c132003537e72ee would have been temporary only as this is a generated file)

snan:
I always forget about transifex 💁🏻‍♀️

r10s:
np, i am just happy that things get fixed :) thanks again!

PR #1819, opened 2023-02-28
prefer pasteboard.hasImages over pasteboard.image != nil

### LINKED ISSUE #1528: Delta Chat pasted from …
If I have something in my clipboard from any other app, even if it's just text from Safari or Notes, but also any other kind of data, I get these messages constantly as soon as I enter Delta Chat. Several in a row until the originating app swaps out.

Screenshot 2022-04-10 at 08 17 47

It comes across as if Delta Chat were snooping on the clip board + it's annoying + it's wasteful (in this case it's a 3d scene with tons of polygons and rigs and angles).

r10s:
this has performance advantages
and also the "has"-properties should not trigger a "User pasted from clipboard" warning
(if i understood https://developer.apple.com/documentation/uikit/uipasteboard correctly)

moreover, this pr does some cleanup.

closes #1528

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -2026,10 +2026,7 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
     }
 
     private func copyToClipboard(ids: [Int]) {
-        let pasteboard = UIPasteboard.general
-        pasteboard.string = ""
         var stringsToCopy = ""
-
         if ids.count > 1 {
             let sortedIds = ids.sorted()
             var lastSenderId: Int = -1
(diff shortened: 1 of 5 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -2026,10 +2026,7 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
     }
 
     private func copyToClipboard(ids: [Int]) {
-        let pasteboard = UIPasteboard.general
-        pasteboard.string = ""
```
rosano, line 2030: I remember setting string to blank to avoid some crash in https://github.com/deltachat/deltachat-ios/commit/f2d987ae29ad84add427f46cdd9a4253853b9f2e, but maybe it's not necessary when the code is organized this way.

rosano (approved):
Yay for less `let` 👍🏽 I'll approve but without running the code (if I should run before approving in the future let me know)

deltachat-ios/Chat/ChatViewController.swift
r10s: good pointer. however, i think that time the bug could also have been fixed by removing the line.

we're now building the string and after that touching the clipboard - this is what we're doing at other places as well, so i think, that should be okay.

PR #1820, opened 2023-03-04
update mute icon on CHAT_MODIFIED

r10s:
check if an update is needed,
if so, call updateTitle()

nb: updateTitle() is done on viewWillAppear(),
however, self.dcChat is not always up to date at that point, this is not changed by this pr:
in general, we should aim to avoid updates on subsequent shows anyway and let all updates be driven by events
as this makes navigation between controller much faster and also allows the state to be changed by other reasons.

i assume, #1810 will pull in the existing event handlers at some point,
so this pr should not make #1810 harder and can be merged independently

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -617,6 +617,12 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
                         self.messageInputBar.isHidden = true
                     }
                 }
+
+                let oldMuted = self.navigationItem.rightBarButtonItems?.contains(self.muteItem)
+                let newMuted = self.dcChat.isMuted
+                if oldMuted != newMuted {
+                    self.updateTitle()
+                }
             }
         }
 


rosano:
Looks fine to me but I would prefer to try it on a device and don't have the option right now.

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -617,6 +617,12 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
                         self.messageInputBar.isHidden = true
                     }
                 }
+
+                let oldMuted = self.navigationItem.rightBarButtonItems?.contains(self.muteItem)
+                let newMuted = self.dcChat.isMuted
+                if oldMuted != newMuted {
+                    self.updateTitle()
```
cyBerta, line 624: I would just call `updateTitle()` here. On a first glance the extra code around it seems to optimize at a place where there's not so much to gain.

r10s: i was hesitant as things are a bit of a mess for other things, eg. the member list counter or name changes are updated somehow somewhere else ... and i wanted to avoid double updates.

but maybe the correct way is to remove these updates (so, probably mostly the call to updateTitle() on viewWillAppear()

r10s: i added a commit

cyBerta (approved):

PR #1833, opened 2023-04-05
fix out of bounds error in search

Simon-Laux:
I came across this error while testing the search field on mac.
could have something to do with pressing backspace when the search field is empty


r10s (approved):
thanks for fixing! and yes, might error out on mac-only, but looks as if this is a useful precaution in any case.

PR #1837, opened 2023-04-06
make it build again on new xcode 14.3

Simon-Laux:
I'm not sure if we should merge it as is, the diff seems like it changed something else too?


r10s:
yeah, let's better redo this as needed without side changes.

ftr, i am also using xcode14.3, so the real issue is probably not even the target versions of some frameworks but sth else

PR #1843, opened 2023-04-26
add "All Media"

r10s:
this pr adds an "All Media" view, similar to https://github.com/deltachat/deltachat-android/pull/2489, was less effort than intitally expected :)

- [x] add icon to tabbar
- [x] reuse existing `GalleryViewController` and `DocumentGalleryController`
- [x] set title of "Files" tab to "Files and Apps" if there are actually apps. or: add a separate "Apps" tab
- [x] set fallback-icon for older ios
- [x] the title is currently transparent, check if opaque  isn't better - EDIT: added a commit to make the title translucent
- [x] check, that the huge views are actually only instantiated when used
- [x] use correct empty-placeholder, `tab_all_media_empty_hint`
- [x] gallery and document views should hide tabbar - as already true for apps and most other views (it is hidden coincidentally when opened from profiles)

there are also some issues with the gallery, #1844, #1845, #1846 however, that can be targeted in subsequent prs as well as that affects also the profile-galleries

Screenshot 2023-04-27 at 20 44 25

<details>
<summary>tap for old icon view</summary>
this was the first iteration, without adapting the existing icons:<br>

</details>

while this is not the most important thing to do, it is a good practise to get a bit more used again to the source details :)

for, now, the view has the same tabs as the profiles, that can be changed in a subsequent pr, as needed. but i also have the feeling it is still good to tune down "Apps", by saying they're less important than files.


Simon-Laux:
I'm not sure whether the icon in the tab bar is a good place...

Simon-Laux (approved):
code looks fine to me

r10s:
> I'm not sure whether the icon in the tab bar is a good place...

what would you suggest?

adbenitez (approved):

PR #1849, opened 2023-04-28
use translucent qr-code title bar

r10s:
the old, totally transparent title bar led to more or less unreadable tabs, depending on what is shown to the camera.

after figuring out the translucent thingie at #1843, this is a nice improvement for the other view using a tabbar in the title.

moreover, this pr adds an animation for both tabs, things really seem to be fast enough,
also, the system seem to do UIViewController creation in parallel anyways.

before/after - it is not always that bad, but you get the idea:

 &nbsp; 

finally, i have to clean my screen

### DIFF
--- a/deltachat-ios/Controller/QrPageController.swift
+++ b/deltachat-ios/Controller/QrPageController.swift
@@ -74,6 +74,10 @@ class QrPageController: UIPageViewController {
             animated: true,
             completion: nil
         )
+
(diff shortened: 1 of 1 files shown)


adbenitez (approved):
nice! I request changes in your screen once you do a clean-up commit then it is ok to merge

r10s:
cleaner screen:

IMG_2949

Issue #1845, opened 2023-04-27
consider loading gallery async to make things snappier


r10s:
`GalleryViewController` is now also used for the global gallery in #1843, which works surprisingly well without much adaptions.

however, already the in the past, the startup of `GalleryViewController` takes a few ms as it is not done async, maybe #1843 is another reason for improvements here.

of course, this is noticable only on galleries with some 1000 images.

r10s:
while it would be an improvement, it is also no so badly slow currently.

closing for now, as this is no longer an actionable item to be targeted just now.

PR #1860, opened 2023-05-03
snappier qr-code and all-media tabs

r10s:
this pr starts the tab-change-animation already
when the user starts dragging the page.

this feels much snappier,
however, requires to revert the selection if the user aborts the dragging.

looking at the api, i think, this is also the gist.

targets https://github.com/deltachat/deltachat-ios/issues/1845


r10s:
a general note: if i merge things in to move forward, this does not mean you can comment on a pr, we can always change things :)

PR #1877, opened 2023-05-08
open poi in organic maps app

Simon-Laux:
image

works, but organic maps has some bugs that degrade the experience, read more at https://support.delta.chat/t/allow-to-open-map-points-in-organic-maps/2561

https://github.com/organicmaps/organicmaps/issues/5106
https://github.com/organicmaps/organicmaps/issues/5107

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -600,6 +600,22 @@ public class DcContext {
     public func setLocation(latitude: Double, longitude: Double, accuracy: Double) {
         dc_set_location(contextPointer, latitude, longitude, accuracy)
     }
+    
+    public func getLocationStringForMessage(chatId: Int, msgId: Int) -> String? {
+        let locationsArray = dc_get_locations(contextPointer, UInt32(chatId), 0, 0, 0)
+        let len = dc_array_get_cnt(locationsArray)
+        var result: String?
+        for i in 0 ..< len {
+            if dc_array_get_msg_id(locationsArray, i) == msgId {
+                let lat = Double(dc_array_get_latitude(locationsArray, i))
+                let lng = Double(dc_array_get_longitude(locationsArray, i))
+                result = "\(lat),\(lng)"
+                break
+            }
+        }
+        dc_array_unref(locationsArray)
+        return result
+    }
 
     public func searchMessages(chatId: Int = 0, searchText: String) -> [Int] {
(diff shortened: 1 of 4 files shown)


dignifiedquire:
Why not open in the default map application that the user has installed? I personally would like to open these in `Apple Maps` and not some other map application I don't use.

Simon-Laux:
Is there a concept of default applications in iOS for something else than email clients and Browsers?
Or would we need to handle this default selection?

r10s:
this pr is more an experiment, in reality, it is not about opening an external app (which, of course may be better than nothing)

we want some embedded functionality at the end, adding pois, showing tracks, live updates and all that.

i fear, we currently do not have resources to dive into the system map api on all different platforms, we are already struggling with mapbox a lot :) also, map data from osm are often better.

however, there is an already working approach for that in https://github.com/deltachat/deltachat-ios/pull/1912 ,  idea is to get that on android and maybe desktop as well at some point.

so, i do not think, it makes much sense to invest in this pr

Simon-Laux:
don't forget that this button then should open the internal leaflet map, we still need it for a smooth user experience.
Also an option to open a point in your preferred mapping app is still very useful, as those map apps have routing and can show you the way to a point. Also often they are also integrated with your car's dashboard.

r10s:
sure, additional buttons ui etc. can be added at any time, however, first step is to get a shared thing across platforms at all

dignifiedquire:
using this would avoid having to embed additional data/maps  https://developer.apple.com/documentation/mapkit/

r10s:
> using this would avoid having to embed additional data/maps https://developer.apple.com/documentation/mapkit/

data/maps are not embedded in #1912

and i am pretty sure, mapkit pod/lib does not add much less overhead than the 50k needed for the solution at #1912. not to speak about additional coding effort on three platforms, less flexibility and potentially worse maps ... also note, we currently do not have a single ios fulltime dev (not even halftime :), and there are much more important things to get right on ios :)

Simon-Laux:
Telegram seems to offer opening with other installed apps: https://github.com/TelegramMessenger/Telegram-iOS/blob/ae998eb91ef05dfec5c5fa2dfa3c78371992a482/submodules/OpenInExternalAppUI/Sources/OpenInOptions.swift#L89

apple allows for checking schemes for up to 50 different schemes/apps that were defined in the plist under `LSApplicationQueriesSchemes`. I think this would be a nice addition to the map & context menu when we add the map

PR #1883, opened 2023-05-28
show existing saved-messages hint in profile

r10s:
a little, cheap help. inspired by https://github.com/deltachat/deltachat-desktop/pull/3234

Screenshot 2023-05-29 at 00 43 57

moreover, this pr hides "last seen" for device messages and hides the label "Signature Text" for both, device messages and saved messages (the label does not make much sense for the explanation)


Simon-Laux (approved):
Looks a bit weird on first glance without a label for the area, but maybe it does not matter so much... otherwise great

PR #1893, opened 2023-06-05
disable "mark all read" button when it does nothing

Simon-Laux:
This also helps with understanding what it does,
because you can now only click it if there are messages that can be marked as read.

Basically it disables the button if there are no unread archived chats.

Bildschirmfoto 2023-06-05 um 21 12 10
Bildschirmfoto 2023-06-05 um 21 11 52

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -661,6 +661,7 @@ class ChatListController: UITableViewController {
                 navigationItem.setLeftBarButton(nil, animated: true)
                 navigationItem.setRightBarButton(markArchivedReadButton, animated: true)
             }
+            updateMarkArchivedReadButton()
         } else {
             titleView.text = DcUtils.getConnectivityString(dcContext: dcContext, connectedString: String.localized("pref_chats"))
             if !handleMultiSelectionTitle() {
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/ChatListController.swift
```diff
@@ -674,6 +674,7 @@ class ChatListController: UITableViewController {
         }
         titleView.isUserInteractionEnabled = !tableView.isEditing
         titleView.sizeToFit()
+        updateMarkArchivedReadButton()
```
r10s, line 677: that should go to the `if isArchive` branch above - otherwise, the button is created and (expensively) updated without reasons in the normal chatlist.

deltachat-ios/Controller/ChatListController.swift
```diff
@@ -703,6 +704,11 @@ class ChatListController: UITableViewController {
                 self.handleEmptyStateLabel()
             }
         }
+        updateMarkArchivedReadButton()
```
r10s, line 707: not sure if that is needed here, and a good place, did not test. if, it should be in an `if isArchive` branch

r10s (approved):
a great improvement!

deltachat-ios/Controller/ChatListController.swift
Simon-Laux: it needs to be updated if new messages come in. I also just tested it it is triggered.

r10s (approved):

PR #1895, opened 2023-06-05
early prototype of split ui view

Simon-Laux:
we probably need to conditionally,
open the view in splitview or in normal navigation,
becaus eon newer iphones the secondary split view column is shown as a weird overlay

### DIFF
--- a/deltachat-ios/Controller/ChatListController.swift
+++ b/deltachat-ios/Controller/ChatListController.swift
@@ -862,7 +862,8 @@ class ChatListController: UITableViewController {
             searchController.searchBar.resignFirstResponder()
         }
         let chatVC = ChatViewController(dcContext: dcContext, chatId: chatId, highlightedMsg: highlightedMsg)
-        navigationController?.pushViewController(chatVC, animated: animated)
+        // navigationController?.pushViewController(chatVC, animated: animated)
+        splitViewController?.showDetailViewController(chatVC, sender: self)
     }
 
     public func showArchive(animated: Bool) {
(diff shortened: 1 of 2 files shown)


deltachat-ios/Controller/ChatListController.swift
```diff
@@ -862,7 +862,8 @@ class ChatListController: UITableViewController {
             searchController.searchBar.resignFirstResponder()
         }
         let chatVC = ChatViewController(dcContext: dcContext, chatId: chatId, highlightedMsg: highlightedMsg)
-        navigationController?.pushViewController(chatVC, animated: animated)
+        // navigationController?.pushViewController(chatVC, animated: animated)
+        splitViewController?.showDetailViewController(chatVC, sender: self)
```
Simon-Laux, line 866: like have a condition here where to open it, or a function on the parent that wants a view and decides where to add it.

deltachat-ios/Coordinator/AppCoordinator.swift
```diff
@@ -126,6 +133,9 @@ class AppCoordinator {
 
     func showChat(chatId: Int, msgId: Int? = nil, openHighlightedMsg: Bool = false, animated: Bool = true, clearViewControllerStack: Bool = false) {
         showTab(index: chatsTab)
+        let dcContext = dcAccounts.getSelected()
+        let chatVC1 = ChatViewController(dcContext: dcContext, chatId: chatId, highlightedMsg: msgId)
+        splitViewController.showDetailViewController(chatVC1, sender: self)
```
Simon-Laux, line 138: I have not tried this here, but it is a reminder that there are multiple places where we need to consider how to open views.

r10s:
thanks for the draft, for a real implementation, however, there is a long way to go and this is currently not planned. we can recondider when other, more important issues are fixed

closing stale pr

PR #1898, opened 2023-06-08
fix webxdc.js that broken after adding importFiles()

r10s:
there was just a comma missing, that broke _some_ apps (for whatever reason, not _all_ apps)

ftr: in theory, adding a trailing comma is always a good idea, make adding things easier.
however, JavaScript and JSON is a mess here:
the former allows adding a trailing comma, the latter does not allow that - resulting in devs using both often to not add the comma.


adbenitez (approved):
"haha, classic" :smoking:

PR #1900, opened 2023-06-09
speedup opening of attach menu

r10s:
the check, if there were any webxdc apps ever is comparable expensive on larger databases and if there are webxdc apps.
in this case, check only once (otherwise the checke is not that expensive).

came over that which trying to get a grip on a random crash when opening the menu.


adbenitez (approved):

r10s:
upps, sorry, there was a commit fixing #1900 to this pr accidentally, wanted to add that to another pr, but well, then it's that :)

PR #1904, opened 2023-06-15
do not exist apps unexpectedly

r10s:
the current "forward" implementation does not allow to "cancel" the action and return to the app (as on android).
this results in apps being terminated quite unexpectedly and undoable, which is esp. surprising if the user taps around to check functinalities or hits a "Share" button accidentally.

this pr adds a little dialog as the next better approach; (i also thought about explaining more in the message, however, this is also not what we're doing on android, so this is out of scope)

for desktop, we need sth. like that or as on android as well


adbenitez (approved):

r10s:
i will file a dedicated issue for desktop

PR #1908, opened 2023-06-22
prepare 1.38.2

r10s:


adbenitez (approved):

r10s:
on its way to the app store now (first to testflight, directly afterwards to the app store)

PR #1912, opened 2023-08-02
add experimental map

r10s:
this PR adds a map to show locations to iOS.

the PR makes use of the new "Webxdc Integration" feature of https://github.com/deltachat/deltachat-core-rust/pull/5461 (which need to be merged first).

therefore, review should concentrate on the buttons etc. added to iOS - the map itself is totally independent and can then be improved independently of iOS.

# still, some map impressions

image_2023-06-03_13-18-23 image_2023-06-03_13-19-06
image_2023-06-03_13-18-35 image_2023-06-03_13-19-19 

(the third image is a result from location-streaming-api some years ago, however, how we have nice test-data by that :)

the current mapx.xdc uses leaflet, and the API defined at `webxdc/maps_integration.rs` in core. some random thoughts:

- mapbox was a a source of all kind of issues in the last (telemetry hard to disable, being flagged as "bad" on fdroid and things like that)
- mapx.xdc is only 50k in side (android mapbox is about 4mb)
- leaflet is much faster when it comes to many poi
- at least our usage of mapbox is slow in general, startup time on desktop is about 10 seconds, https://github.com/deltachat/deltachat-desktop/issues/3768
- idea is that the webxdc can be replaced by the end-user later (eg. by sending a own variation to saved-messages)

~~main open point is that the xdc is currently visible in the saved-messages chat,  we should consider if it is better to hide it somehow. EDIT: this will be done subsequently, for now, @nicodh and me decided that a message describing what is going on will do~~ EDIT: default maps.xdc is no longer visible due to help of core


r10s:
back to draft as the core part is close to being ready and we can remove all the update-hacks here soon

r10s:
although maps for iOS will stay experimental for 1.46, we aim to get this in to move forward, to get bug reports etc. the UX improvement is huge - there is just no map before (we aimed to get that in even with the much more hacky swift-but-rust implementation :)

deltachat-ios/Controller/AllMediaViewController.swift
```diff
@@ -78,6 +86,11 @@ class AllMediaViewController: UIPageViewController {
         }
     }
 
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        navigationItem.rightBarButtonItem = chatId == 0 && UserDefaults.standard.bool(forKey: "location_streaming") ? mapButton : nil
```
zeitschlag, line 91: What's the reason that this is a button in the `NavigationBar` and not another option in the segmented control?

deltachat-ios/Controller/MapViewController.swift
```diff
@@ -0,0 +1,33 @@
+import UIKit
+import WebKit
+import DcCore
+
+class MapViewController: WebxdcViewController {
+    private let isGlobalMap: Bool
+
+    init(dcContext: DcContext, chatId: Int) {
+        isGlobalMap = chatId == 0
```
zeitschlag, line 9: I think I asked this before, but what does `chatId == 0` mean again? Would it make sense to put this somewhere, i.e. a computed property/variable/... with a fitting name?

deltachat-ios/Controller/ContactDetailViewController.swift
```diff
@@ -95,6 +95,20 @@ class ContactDetailViewController: UITableViewController {
         return cell
     }()
 
+    private lazy var locationsCell: UITableViewCell = {
+        let cell = UITableViewCell(style: .value1, reuseIdentifier: nil)
```
zeitschlag, line 99: I'm so used to using `UIListContentConfiguration` these days :D

deltachat-ios/Controller/MapViewController.swift
```diff
@@ -0,0 +1,33 @@
...
+        super.init(dcContext: dcContext, messageId: webxdcInstanceId)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        navigationItem.rightBarButtonItem = nil
```
zeitschlag, line 26: I wondered why this is necessary, but the `WebxdcViewController` is setting a `More`-button and this is to override it, right?

zeitschlag (approved):
LGTM and does what it says! I added some questions here and there, but nothing blocking. Thanks again!

deltachat-ios/Controller/MapViewController.swift
r10s: yes. the webxdc menu with "source code" and esp. "share" does not fit and is not needed

deltachat-ios/Controller/AllMediaViewController.swift
r10s: pragmatism :) the reason is that we want to reuse the WebxdcViewController, at least for now, and not deal with fragments or whatnot. but i also find it not too bad meanwhile

deltachat-ios/Controller/MapViewController.swift
r10s: it usually means "no filter, show everything". this pattern us used at many places, eg. also for global gallery etc. but maybe changing the parameter to `chatIdFilter: Int? = nil` makes it more obvious

deltachat-ios/Controller/ContactDetailViewController.swift
r10s: `UIListContentConfiguration` is what renders the settings eg. in the iOS system settings? anyways, requires iOS14, so not today :)

PR #1920, opened 2023-09-04
use core email validation function

### LINKED ISSUE #1918: duplicated email validation code (swift and core)
We have email validation in swift:
https://github.com/deltachat/deltachat-ios/blob/bad499e7bd21ce424f19425abcdfe394b4db8fa5/deltachat-ios/Helper/Utils.swift#L7-L18

And in core: `dc_may_be_valid_addr`

which is backed by this rust function: https://github.com/deltachat/deltachat-core-rust/blob/f94d34c94b5f00aaf3c3571065d3bcf5a4c2021a/src/tools.rs#L517

I propose that we should use the rust function on all platforms and improve it to do better / more correct parsing if necessary.

Simon-Laux:
closes #1918

The core email validation function should be improved to be atleast as good as the regex that is replaced by this pr, before merging 
this pr.

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -1546,3 +1546,7 @@ func strToBool(_ value: String?) -> Bool {
 
     return false
 }
+
+public func isValidEmail(email: String) -> Bool{
+    return dc_may_be_valid_addr(email) != 0
+}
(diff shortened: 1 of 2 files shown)


r10s:
thanks for the hint!

i created a successor at #1929 that:
- uses a better name - we generally name things the same as in core unless there are really good reasons to diverge
- avoids additional wrapping

> the core email validation function should be improved to be atleast as good as the regex that is replaced by this pr, before merging this pr.

well, the core function is in use on android for a long time, i think, it is good enough. 

in any case it is good to have the same behaviour among systems

PR #1921, opened 2023-09-08
fix: blurry icons in messagelist

### LINKED ISSUE #1646: "Lock" and "√" icons appear blurry in message list
- Device:
- iOS version:
- Delta Chat version:
- Expected behavior:
- Actual behavior:
- Screenshots:
- Logs:

Simon-Laux:
fixes #1646

### DIFF
--- a/deltachat-ios/Assets.xcassets/padlock.imageset/Contents.json
+++ b/deltachat-ios/Assets.xcassets/padlock.imageset/Contents.json
@@ -1,0 +1,23 @@
+{
+  "images" : [
+    {
+      "filename" : "padlock.svg",
+      "idiom" : "universal",
+      "scale" : "1x"
+    },
+    {
+      "filename" : "padlock.svg",
+      "idiom" : "universal",
+      "scale" : "2x"
+    },
+    {
+      "filename" : "padlock.svg",
+      "idiom" : "universal",
+      "scale" : "3x"
+    }
+  ],
+  "info" : {
+    "author" : "xcode",
+    "version" : 1
+  }
+}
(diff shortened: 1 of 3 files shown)


deltachat-ios/Assets.xcassets/padlock.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
...
+      "filename" : "padlock.svg",
+      "idiom" : "universal",
+      "scale" : "3x"
+    }
+  ],
+  "info" : {
+    "author" : "xcode",
+    "version" : 1
+  }
+}
```
r10s, line 23: imagesets do not support svg on older system we're still supporting. so we need to use bitmaps with single, double and triple size (as we're doing for all the other images)

in fact, the old `ic_lock` already is in this format, however, probably not in the optimal resolution

r10s (changes requested):

deltachat-ios/Assets.xcassets/padlock.imageset/Contents.json
Simon-Laux: >  the old ic_lock already is in this format

yes, I noticed but it was always selecting the worst resolution when displaying. Do you know what is the minimum target for svg images icons?

r10s: ios13 afaik

r10s:
closing as it does not fix the issue, in general, svg is not the solution, even stock images appear blurry.

the issue probably is the NSTextAttachment, cmp https://stackoverflow.com/questions/27281945/blurry-image-nstextattachment-on-nsattributedstring

Simon-Laux:
> closing as it does not fix the issue, in general, svg is not the solution, even stock images appear blurry.

I did not change the other icons, those worked without changing to svg.
Maybe we can just pick those changes of the pr and deal with the padlock later?

PR #1922, opened 2023-09-18
update SDWebImage, libwebp and other Pods

r10s:
ran `pod update` (after installing CocoaPods, previously, this was done by @cyBerta ) according to https://guides.cocoapods.org/using/using-cocoapods.html and as done in the past eg. at #1211 (see also discussion wrt "pods in git" there)


r10s:
hm, getting the error `No such module 'SDWebImage'` now, maybe somehow related to the extension, idk. @cyBerta did you do sth more special than `pod update` on the last updates of Pods? maybe there are even some notes somewhere?

Screenshot 2023-09-19 at 00 16 32

Simon-Laux:
Maybe we can migrate the usage of some packages to the built in swift package manager in the future and get rid of the pods stuff?

r10s:
whatever sucks less and is simpler is welcome ...

> built in swift package manager 

... do you have a link or so for that? or some click paths? just searching brings up lots of probably unrelated stuff.

Simon-Laux:
https://www.swift.org/package-manager/
https://wwdcbysundell.com/2019/xcode-swiftpm-first-look/
https://developer.apple.com/documentation/xcode/swift-packages

afaik as I understand this the library needs to support this installation method, but many do. I used it in https://github.com/deltachat/apple_webxdc_quick_look_thumbnail_provider for installing the zip library right inside of Xcode.

r10s:
thanks - so it is up to the library to support "Swift Packet Manager" it seems.

i had a look at the used libraries, seem 4 of 5 would support this:

https://github.com/ashleymills/Reachability.swift - ok
https://github.com/stefanceriu/SCSiriWaveformView - no Packet Manager
https://github.com/SDWebImage/SDWebImage - ok
https://github.com/SDWebImage/SDWebImageSVGKitPlugin - ok, but somehow the repo is archived
https://github.com/SVGKit/SVGKit - ok

however, the issue is also related to the "share extension", so it is unclear even, if a complete switch would solve that. also, we have even less knowledge about that as for pods  - where the updates work in the past without issues.

i am still trying out pod stuff, i am still pretty sure, i just missed sth.

r10s:
superseded by #1926

PR #1931, opened 2023-10-07
raise deployment target from ios11 to ios12

r10s:
this is needed as otherwise there are weird crashes, eg. when opening connectivity view
(there are sth. simple as `let monitor = NWPathMonitor()`)

as reported eg. at
https://github.com/fluttercommunity/plus_plugins/pull/2169 it seems xcode15 does not support ios<12 fully

fortunately,
this does not change our minimum supported devices as all of them get get ios12.


r10s:
merging that in quickly to allow to do a 1.40.3 testflight soon

Issue #1930, opened 2023-10-07
ios 16.6.1 crashes when opening the connectivity view


r10s:
pretty sure, this did not happen before the last ios update, will have a look ...

r10s:
similar issue: https://github.com/fluttercommunity/plus_plugins/issues/1955#issuecomment-1649361001 , seems to be like that since 16.6.1

r10s:
the fix suggested at https://github.com/fluttercommunity/plus_plugins/pull/2169 also works for us: raise the deployment target to ios 12 - fortunately, this does not change the number of supported devices, so this is an easy decision. will do a pr

r10s:
closed by https://github.com/deltachat/deltachat-ios/issues/1930

PR #1932, opened 2023-10-07
prepare 1.40.3

r10s:
this fixes a crash probably introduced by one of the last ios updates, see https://github.com/deltachat/deltachat-ios/issues/1930


adbenitez (approved):
ah so that was why it was 93!

r10s:
:)

PR #1938, opened 2023-10-24
fix swift and build warnings

r10s:
see commit messages for details and reasoning; this pr also removes unused and never maintained test files

together with https://github.com/deltachat/deltachat-ios/pull/1937 we're down to ZERO warnings,  resulting in new warnings being much more visible


adbenitez (approved):
remove the leaves to expose the bugs :ok_hand:

Issue #1880, opened 2023-05-18
adding webxdc to home screen stopped working


r10s:
as there are no recent changes in the code, i suspect the iOS 16.4 update i did recently on my iphone14.

EDIT: yip, it seems to be a change in iOS, the same Delta Chat version can add shortcuts on iphoneSE1 with iOS 15.7.3

when using "Add to homescreen" from the webxdc's menu, the following error is shown in the browser (the browser is needed as we are creating a "mini pwa" as only they are allowed to exist on the homescreen - this needs an http-server during adding, see https://github.com/deltachat/deltachat-ios/pull/1665 - and that seems not to work)

<img width=320  src=https://github.com/deltachat/deltachat-ios/assets/9800740/917fd607-96f8-49f2-b3d4-b8d67c963b25>

r10s:
the issue in upstream: https://github.com/httpswift/swifter/issues/537

it is still a pity, that there is [no way](https://stackoverflow.com/questions/21672873/opening-dataxxx-urls-in-safari) to pass a `data:` url directly; in theory, the whole server part should not be needed at all :/

r10s:
from ios-dev chat:

> idea is that we can tell users in a short tutorial how to use the shortcut app for this.
like this app does it: https://apps.apple.com/app/apple-store/id1532875441
Basically tell user to copy a code (contains account & messageid) and then create a shortcut, open safari url and paste the link there, then add the shortcut to the homescreen. The coding (logic) work would be very minimal because we might be able to just use safari / standard open url action. The harder part is the design of the tutorial, since it only effects newer versions we could use swiftui or just a website in a webview if we think that would be easier.

Simon-Laux:
> privacy wise, that may not be worse than using apple shortcuts

if you don't share shortcuts I don't know what's bad for privacy in using them. But for convenience I agree, the Webserver could be a solution, but it means
- you need to be online to create the link
- we need to maintain a new server and mention it in our privacy policy
- links may only work when user is online, because offline PWA's sometimes were sometimes suddenly not cached anymore in my experience with them.

So in essence the webserver approach would serve a custom PWA for the user that has a custom icon, so it is probably not possible or very hard to do this in a static website, I bet we need a dynamic site for that.

Benefits and downsides:

|Webserver | Shortcut tutorial |
|-----------|------------------|
|+ easy for users | - more clicks for users |
|- need to be online to create the link | + works offline |
[...]

PR #1939, opened 2023-10-25
remove nonworking 'Add To Home Screen' hack

r10s:
roughly reverts https://github.com/deltachat/deltachat-ios/pull/1665 and closes https://github.com/deltachat/deltachat-ios/issues/1880 ,   which also shows some alternatives to removal, however it is unclear what is working and/or requires lots of work; this time is spend better in other areas acurrently


r10s:
a little smarter and easier to review approach will be to disable shortcuts for ios16 only, that way, we won't take away things ... i'll do a pr for that

Issue #1909, opened 2023-06-24
no error message for unsupported video formats


JoeMcNuggets:
Attachment with webm videos are not playable: white screen shown instead of player

- Device: iPhone 12
- iOS version: 16.5
- Delta Chat version: 1.32.8
- Expected behavior: received video could be played
- Actual behavior: white screen is shown
- Steps to reproduce the problem: receive video/webm attachment

r10s:
seems as if webm is not yet supported on ios (only macos), see https://discussions.apple.com/thread/252503189 , maybe there are some libs or so we could use for playback, however, i do not think we have resources to dive into that the next months.

there are more things that do not work on ios, also eg. some ogg formats, word documents or telegram stickers - in general, we cannot guarantee that all external formats work in delta chat - we try to make sure, however, formates generated by delta chat work on all systems (eg, videos recorded on android are playable on ios)

however, we can probably at least improve the error messages - and maybe also improve passing the file over to an app that supports the corresponding file - eg. vlc in this case probably has support for webm

PR #1941, opened 2023-10-25
give instructions how to play unsupported video formats

r10s:
this first commit reduces unneeded code complexity and removes one full message-db load.

the second commit shows an alert with instructions in case a video-preview could not be generated (with high chance, also Preview will fail to play the file)

closes https://github.com/deltachat/deltachat-ios/issues/1909

of course, this can be improved further, however, regarding that .webm video is not an officially supported format, this seems good enough and already a good improvement, giving the ppl an idea how to play that at all (it was not even clear to me :)


adbenitez (approved):
:ok_hand:

PR #1947, opened 2023-10-28
add option to deactivate codes

r10s:
this better guides users to option; scanning for disabling and re-enabling is still possible

counterpart of https://github.com/deltachat/deltachat-android/issues/2763


adbenitez (approved):
great!

PR #1966, opened 2023-11-11
more in-app help

r10s:
this pr prepares for opening verified-chats-help in-app, see https://github.com/deltachat/deltachat-android/issues/2829 

- allow opening in-app help at given fragment
- use multi-device-troubleshouting in-app-help


adbenitez (approved):
wow, always amazed by how simple the changes look in the ios code base :o

PR #1969, opened 2023-11-14
try to fix notifications

### LINKED ISSUE #1202: background handler terminated by 0xdead10cc sometimes

### LINKED ISSUE #1830: stop event loop in a thread safe way
See the comment here: https://github.com/deltachat/deltachat-core-rust/issues/2280#issuecomment-1154843563

Once https://github.com/deltachat/deltachat-core-rust/pull/4233 is merged, the code in this repository should be checked to use FFI in a thread safe way.

Similar issue for Android: https://github.com/deltachat/deltachat-android/issues/2515

### LINKED ISSUE #1979: Notifications: Crash in background activity -> File Locks
As already mentioned in https://github.com/deltachat/deltachat-core-rust/issues/4420#issuecomment-1809675907, many of our background crashes are because of the file locks we hold, so we need to shutdown the database and account manager additionally to stopIo to remove the locks and fully prepare the app for suspense without being killed by the system for still holding locks.

> Specifically accounts/accounts.lock and accounts/23cc2440fa7e45599169db4cd0a2a121/dc.db so iOS app needs to close both the database and the whole account manager to release accounts.lock.

I think this affects both encrypted accounts (sqlcipher) and the new lock file created but the account manager.
Why do I think it's only for encrypted accounts? because I only have one encrypted account on the test device, while having around 4 unencrypted accounts, I assume there must be more locks otherwise (unless this is about account selection, but @link2xt told me that sqlcipher is known for creating locks).

What points to this issue:
The log that I pulled from the iPhone and viewed on the Console Mac app:

```
standard	08:29:32.448158+0100	runningboardd	[application<chat.delta>:6570] Set darwin role to: None
standard	08:29:32.448230+0100	runningboardd	[application<chat.delta>:6570] set Memory Limits to Hard Inactive (2098)
standard	08:29:32.448348+0100	runningboardd	[application<chat.delta>:6570] check if suspended process is holding locks
standard	08:29:32.460998+0100	runningboardd	[application<chat.delta>:6570] Found locked file lock: /var/mobile/Containers/Shared/AppGroup/BEA5D4F2-D7ED-422A-88CF-7D471FAB8CAE/accounts/accounts.lock
standard	08:29:32.465152+0100	runningboardd	[application<chat.delta>:6570] Found locked file lock: /var/mobile/Containers/Shared/AppGroup/BEA5D4F2-D7ED-422A-88CF-7D471FAB8CAE/accounts/23cc2440fa7e45599169db4cd0a2a121/dc.db
fehler	08:29:32.465698+0100	runningboardd	[application<chat.delta>:6570] suspended with locked system files:
/var/mobile/Containers/Shared/AppGroup/BEA5D4F2-D7ED-422A-88CF-7D471FAB8CAE/accounts/accounts.lock
/var/mobile/Containers/Shared/AppGroup/BEA5D4F2-D7ED-422A-88CF-7D471FAB8CAE/accounts/23cc2440fa7e45599169db4cd0a2a121/dc.db
fehler	08:29:32.465761+0100	runningboardd	[application<chat.delta>:6570] locked files not in allowed directories: /var/mobile/Containers/Data/Application/0B8944B3-B7FF-417C-AAE2-0F63C7DD42E7
/var/mobile/Containers/Data/Application/0B8944B3-B7FF-417C-AAE2-0F63C7DD42E7/tmp
standard	08:29:32.465934+0100	runningboardd	[application<chat.delta>:6570] Terminating with context: <RBSTerminateContext| domain:15 code:0xDEAD10CC explanation:[application<chat.delta>:6570] was suspended with locked system files:
/var/mobile/Containers/Shared/AppGroup/BEA5D4F2-D7ED-422A-88CF-7D471FAB8CAE/accounts/accounts.lock
/var/mobile/Containers/Shared/AppGroup/BEA5D4F2-D7ED-422A-88CF-7D471FAB8CAE/accounts/23cc2440fa7e45599169db4cd0a2a121/dc.db
not in allowed directories:
/var/mobile/Containers/Data/Application/0B8944B3-B7FF-417C-AAE2-0F63C7DD42E7
/var/mobile/Containers/Data/Application/0B8944B3-B7FF-417C-AAE2-0F63C7DD42E7/tmp reportType:CrashLog maxTerminationResistance:Absolute>
standard	08:29:32.465976+0100	runningboardd	[application<chat.delta>:6570] terminate_with_reason() success
standard	08:29:32.466099+0100	kernel	deltachat-ios[6570] Corpse allowed 1 of 5
```

BTW if we should use the unified logging system (`os_log`, in #1973) our log messages would also be readable with the console Mac app.

The other indicator for this is the crash statistics we got from apple:
Bildschirmfoto 2023-11-15 um 03 11 09
as you can see it is about file locks.

related to: #1202

Simon-Laux:
- don't use logger from context outside of context -> r10s did that in #1994
- context wrapper objects that are created for each event can not be used to save state so I commented out those non working statements -> now also in https://github.com/deltachat/deltachat-ios/pull/1976
- try to add real shutdown for event handler and closing the db (#1979)

Does not crash as often anymore, but:
- [ ] delay after waking up / unsuspeding, link2xt said probably because of missing cache, also might just be sqlcipher taking long to open the db
- [X] events are not received by the UI anymore, althrough I sucessfully restarted the handler that the events are printed to the console.
- [ ] there is a performance warning because I wait in the main thread for the event thread to finish: "Thread Performance Checker: Thread running at QOS_CLASS_USER_INTERACTIVE waiting on a lower QoS thread running at QOS_CLASS_BACKGROUND. Investigate ways to avoid priority inversions"
- [X] TODO: eventHandlerActive boolean should proabably be an atomic? -> seems to be not necessary
- [ ] it wastes time when going in the background, sometimes just waits while doing nothing, https://github.com/deltachat/deltachat-core-rust/issues/4420 could be useful here. - though keeping open a bit for the case that user comes back to the app is a feature, not a bug

## Also See

- https://github.com/deltachat/deltachat-core-rust/issues/4420

closes #1979
closes #1830
closes #1202 (basically the same as #1979)

if crashes still continue, we should open new, on-point issues

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -566,7 +566,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             self.pushToDebugArray(String(format: "3/%.3fs", Double(Date().timeIntervalSince1970)-nowTimestamp))
 
             if !self.appIsInForeground() {
-                self.dcAccounts.stopIo()
                 // to avoid 0xdead10cc exceptions, we need to make sure there are no locks on files.
                 self.closeDB()
             }
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -353,11 +353,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         if !dcAccounts.isOpen() {
             logger.debug("⏫ openDatabase again, because it is closed")
             dcAccounts.openDatabase(writeable: true)
-            self.openAccountsWithKeychain()
-            logger.debug("⏫ openDatabase done")
-            if !eventHandlerActive {
-                installEventHandler()
-            }
+            openAccountsWithKeychain()
+            logger.debug("➡️ openDatabase done")
+            installEventHandler()
             reloadDcContext()
         }
     }
@@ -638,14 +636,17 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         }
     }
 
-    var eventHandlerActive = false
-    func installEventHandler() {
+    private var eventHandlerActive = false
+    private func installEventHandler() {
+        if eventHandlerActive {
+            return
+        }
+        eventHandlerActive = true
         DispatchQueue.global(qos: .background).async { [weak self] in
             guard let self = self else { return }
             let eventHandler = DcEventHandler(dcAccounts: self.dcAccounts)
             let eventEmitter = self.dcAccounts.getEventEmitter()
-            eventHandlerActive = true
-            logger.info("⏫ event emitter started")
+            logger.info("➡️ event emitter started")
             while true {
                 if shouldShutdownEventLoop {
                     break
@@ -653,7 +654,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
                 guard let event = eventEmitter.getNextEvent() else { break }
                 eventHandler.handleEvent(event: event)
             }
-            logger.info("⏬ event emitter finished")
+            logger.info("⬅️ event emitter finished")
             eventShutdownSemaphore.signal()
             eventHandlerActive = false
         }
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -570,6 +570,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
                 self.closeDB()
             }
 
+            // to avoid 0xdead10cc exceptions, scheduled jobs need to be done before we get suspended;
+            // we increase the probabilty that this happens by waiting a moment before calling completionHandler()
+            usleep(1_000_000)
             logger.info("⬅️ fetch done")
             completionHandler(.newData)
             if backgroundTask != .invalid {
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -67,6 +67,33 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         continueDidFinishLaunchingWithOptions()
         return true
     }
+    
+    func openAccountsWithKeychain() {
+        logger.info("🔐 openAccountsWithKeychain")
+        let accountIds = dcAccounts.getAll()
+        for accountId in accountIds {
+            let dcContext = dcAccounts.get(id: accountId)
+            if !dcContext.isOpen() {
+                do {
+                    logger.info("🔐 start")
+                    let secret = try KeychainManager.getAccountSecret(accountID: accountId)
+                    logger.info("🔐 got secret")
+                    if !dcContext.open(passphrase: secret) {
+                        logger.error("Failed to open database for account \(accountId)")
+                    }
+                    logger.info("🔐 done")
+                } catch KeychainError.accessError(let message, let status) {
+                    logger.error("Keychain error. \(message). Error status: \(status)")
+                    return
+                } catch KeychainError.unhandledError(let message, let status) {
+                    logger.error("Keychain error. \(message). Error status: \(status)")
+                } catch {
+                    logger.error("\(error)")
+                }
+            }
+        }
+        logger.info("🔐 openAccountsWithKeychain done")
+    }
 
     // finishes the app initialization which depends on the successful access to the keychain
     func continueDidFinishLaunchingWithOptions() {
@@ -83,27 +110,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         } catch {
             // TODO: Handle
         }
-
-        let accountIds = dcAccounts.getAll()
-        for accountId in accountIds {
-            let dcContext = dcAccounts.get(id: accountId)
-            if !dcContext.isOpen() {
-                do {
-                    let secret = try KeychainManager.getAccountSecret(accountID: accountId)
-                    if !dcContext.open(passphrase: secret) {
-                        logger.error("Failed to open database for account \(accountId)")
-                    }
-                } catch KeychainError.accessError(let message, let status) {
-                    logger.error("Keychain error. \(message). Error status: \(status)")
-                    return
-                } catch KeychainError.unhandledError(let message, let status) {
-                    logger.error("Keychain error. \(message). Error status: \(status)")
-                } catch {
-                    logger.error("\(error)")
-                }
-            }
-        }
-
+        
+        openAccountsWithKeychain()
+        
         if dcAccounts.getAll().isEmpty, dcAccounts.add() == 0 {
            fatalError("Could not initialize a new account.")
         }
@@ -211,9 +220,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     // applicationWillEnterForeground() is _not_ called on initial app start
     func applicationWillEnterForeground(_: UIApplication) {
         logger.info("➡️ applicationWillEnterForeground")
+        maybeStart()
         applicationInForeground = true
         dcAccounts.startIo()
-
+        
         DispatchQueue.global(qos: .background).async { [weak self] in
             guard let self = self else { return }
             if let reachability = self.reachability {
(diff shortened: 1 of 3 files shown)


r10s:
great to push this forward!

i know, this is still a draft, but still:

for easier review, merging and more on point discussions, please do not put unreleated things into one pull requests or at least do on-point commits.

also, please do not shift blocks around, this also makes it very hard to find the gist of the change. this whole area is very sensible and complicated already

Simon-Laux:
yeah would have made more sense to split the changes over multiple commits.
creating a dedicated pr for moving the logger away from context is a good idea, by the way it it still needed that it is pluggable/changable? or could we just import the simple logger everywhere?

Simon-Laux:
> [seems like this fixed the resume issue](https://github.com/deltachat/deltachat-ios/pull/1969/commits/bf344695d56b5e89221ea9b8d77b6167535ad644)

it did not, just got it again (now event handler is false though)

Simon-Laux:
since last commit resume works reliable for me.

r10s:
i simplified the logging part at #1994 - once #1994 is merged, this pr should be rebased and become much simpler

deltachat-ios/AppDelegate.swift
```diff
@@ -323,6 +333,34 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
+        // somehow wait for the eventhandler to be closed
+        eventShutdownSemaphore.wait()
+        // to avoid 0xdead10cc exceptions, we need to make sure there are no locks on files.
+        self.dcAccounts.closeDatabase()
+        shouldShutdownEventLoop = false
+    }
+    
+    func maybeStart() {
+        logger.debug("⏫ maybe openDatabase again 🤔")
+        if !dcAccounts.isOpen() {
```
r10s, line 353: how can this ever happen? imu ...

- `isOpen()` checks if `accountsPointer != nil`

- on startup, `accountsPointer` is already set in `didFinishLaunchingWithOptions` (which should be always called before or instead any performFetch())

- on shutdown, `accountsPointer` is set to `nil` only when `AppDelegate`, where `maybeStart() `is part of, gets destroyed

have you ever seen this code being executed? otherwise, i suggest to revert the change as it add unneeded complexity

Simon-Laux: > on shutdown, accountsPointer is set to nil only when AppDelegate, where maybeStart() is part of, gets destroyed

closeDB (which internally calls `self.dcAccounts.closeDatabase()`)  sets it to nil to shut it down and remove the locks.

Simon-Laux: > have you ever seen this code being executed?

If you mean if I have seen its log messages? yes, quite often actually

r10s: k, now i get the things slowly :) might have be easier to understand, with less code moved around, btw. so leaving continueDidFinishLaunchingWithOptions at its place

deltachat-ios/AppDelegate.swift
```diff
@@ -527,11 +567,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
             if !self.appIsInForeground() {
                 self.dcAccounts.stopIo()
```
r10s, line 569: is this stopIo() on purpose here? graceful shutdown incl. stopIo() is done one line below

deltachat-ios/AppDelegate.swift
```diff
@@ -323,6 +329,28 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             }
         }
     }
+    
+    private var shouldShutdownEventLoop: Bool = false
+    private var eventShutdownSemaphore = DispatchSemaphore(value: 1)
```
r10s, line 334: are you sure, the initial value is correct here?

the [condition for `wait()`](https://developer.apple.com/documentation/dispatch/1453087-dispatch_semaphore_wait) seems to be always always fulfilled, whether [`signal()`](https://developer.apple.com/documentation/dispatch/1452919-dispatch_semaphore_signal) is called or not

(i did not test or investigate deeper, it is just the rule-of-thumb that a semaphore with intial=1 is used for critical-sections and not for inter-thread-communicatio)

Simon-Laux: I just copied this from somewhere and it did what I wanted, I did not dive far into the documentation.
https://developer.apple.com/documentation/dispatch/dispatchsemaphore

r10s: >  it did what I wanted, I did not dive far into the documentation

well, did it really ever wait as you wanted? without the change, it may be that accounts are closed while still being in the event loop - maybe it just did never happen (can it?) with the change to value=0, you'll also get a warning that a user-qos-threads should not wait for a background-qos-thread (what, however, is okay in case of shutdown), indicating that there is really some wait taking effect

Simon-Laux: > with the change to value=0, you'll also get a warning that a user-qos-threads should not wait for a background-qos-thread (what, however, is okay in case of shutdown),

I always get this warning even with value 1.
Isn't this just the start value? but maybe 0 is better?

Simon-Laux: 0 seems to be more correct from what I read online, maybe just change it to 0 then.

Simon-Laux: yeah let's change it to 0

r10s:
unfortunately, this PR in its current form leads to issues when the app is restarted after being suspended to memory before (so if `applicationWillTerminate()` is _not_ called and we land in `maybeStart()` on restart):

in this case, the whole app stays in memory - while all context objects get invalidated by `closeDB()`. later in `maybeStart()`, the accounts object is recreated, but the context objects stay invalid. this leads to crashes or all kind of weird behaviours.

the account suspend/resume idea from before would have avoided that situation.

otoh, suspend/resume would introduce another state in core, which will have issues on its own, so i tend to follow the approach starting in this PR of calling `dc_accounts_unref()` on suspend.

**so, this PR is necessary, but not sufficient:** additionally, we need to invalidate all context objects returned by `dc_accounts_get*()` - maybe that is not that-hard as the swift wrapper objects can stay as is, we could eg. use a function to get the account object that fulfils the conditions

r10s:
after some chatting with core devs deep into things, this PR does basically two things:

1. graceful shutdown event handler, fix crashes with previous shutdown of event handlers

2. avoid 0xdead10cc crashes that happens as the account's lock file is kept open to avoid only one account instance to add accounts. these  0xdead10cc crashes are avoided by calling `dc_accounts_unref()` - which has side effects and other issues as described above

other than i have thought initially, this PR does not change anything on `dc_context_t` and possible 0xdead10cc crashes related to them: accounts that you get with `dc_accounts_get_*()` should stay valid even after `dc_accounts_unref()` they are not closed or so; you can still use them and have to dc_context_unref() them in the end

todo:

- `dc_accounts_t` 0xdead10cc crashes should be fixed by #2009 

- `dc_context_t` 0xdead10cc crashes are at least partly related to encrypted accounts; this will be fixed by #2010 

- this PR should be closed (kept for reference!) and being replaced by a PR that cherry picks "1. only graceful shutdown event "handler

PR #1972, opened 2023-11-14
ge2ee: switch from staging-help to local-help

r10s:


adbenitez (approved):

r10s:
added a commit to use more iOS looking images ...

PR #1981, opened 2023-11-15
non-actionable download states hide action button

### LINKED ISSUE #1980: Check for DC_DOWNLOAD_UNDECIPHERABLE
This constant is documented in the core in https://github.com/deltachat/deltachat-core-rust/pull/5018

Corresponding Android PR: https://github.com/deltachat/deltachat-android/pull/2864

Current code does not check for it, it should hide the button as if in the DC_DOWNLOAD_DONE state:
https://github.com/deltachat/deltachat-ios/blob/0cbb359ab4400f73ce4d34127d890ad220ec15ac/deltachat-ios/Chat/Views/Cells/BaseMessageCell.swift#L393-L412

r10s:
this is achieved by showing the button on known actions instead of duplicating some logic as before
(which led to DC_DOWNLOAD_UNDECIPHERABLE being regarded as an action)

closes #1980

### DIFF
--- a/deltachat-ios/Chat/Views/Cells/BaseMessageCell.swift
+++ b/deltachat-ios/Chat/Views/Cells/BaseMessageCell.swift
@@ -393,21 +393,27 @@ public class BaseMessageCell: UITableViewCell {
         let downloadState = msg.downloadState
         let hasHtml = msg.hasHtml
         let hasWebxdc =  msg.type == DC_MSG_WEBXDC
-        isActionButtonHidden = !hasWebxdc && !hasHtml && downloadState == DC_DOWNLOAD_DONE
         
         switch downloadState {
         case DC_DOWNLOAD_AVAILABLE:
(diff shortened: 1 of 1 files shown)


deltachat-ios/Chat/Views/Cells/BaseMessageCell.swift
```diff
@@ -393,21 +393,27 @@ public class BaseMessageCell: UITableViewCell {
         let downloadState = msg.downloadState
         let hasHtml = msg.hasHtml
         let hasWebxdc =  msg.type == DC_MSG_WEBXDC
-        isActionButtonHidden = !hasWebxdc && !hasHtml && downloadState == DC_DOWNLOAD_DONE
```
r10s, line 396: don't make me think - the logic below is much easier to read and understand :)

adbenitez (approved):
don't make me think :monkey:

Issue #1973, opened 2023-11-14
Better Logging


Simon-Laux:
Our current logging system needlessly passes around a logger instance and thus combines logging with dc core wrappers which makes code more complex without providing any benefit.

## A. Unified System Loging
https://developer.apple.com/documentation/os/logging/generating_log_messages_from_your_code - iOS system logging
Pro:
+ used by system and is what apple wants you to use
+ ability to redact private information
+ nice tutorial on https://developer.apple.com/videos/play/wwdc2020/10168/

Contra:
- different syntax for using it
- no pretty emojis I guess.
- the new friendly api is only available since iOS 14, while we still target iOS 11, but os_log is already available (but not so nice syntax and we should not wrap it according to this site: https://steipete.com/posts/logging-in-swift/)

## B. Customisable Logging with swift-log
https://github.com/apple/swift-log - logging in swift like the rust log crate (an interface with many implementations)
one implementation is https://github.com/sushichop/Puppy, which has many options (saving to apples unified logging system, file, stderr and so on) that can also be combined.
Pro:
+ many options
+ we could probably keep our pretty logging output

Contra:
- different syntax for using it
- extra packages to install

## C. Keep out logging system, but turn it into a singleton

our current logging system is passed around through dc contexts, this is not needed as the login system has no dependencies and just is a wrapper function for `print()`.

It would simplify code not to pass it through context, but instead use it globally.

Simon-Laux:
since A is only really nice if we update to min target iOS 14 and looks like this if we don't
```
// Log a message to the default log and default log level.
   os_log("This is a default message.")
            
   // Log a message to the default log and debug log level
   os_log("This is a debug message.", log: OSLog.default, type: .debug)
            
   // Log an error to a custom log object.
   let customLog = OSLog(subsystem: "com.your_company.your_subsystem", 
            category: "your_category_name")
   os_log("An error occurred!", log: customLog, type: .error)
``` from https://developer.apple.com/documentation/os/logging/generating_log_messages_from_your_code#3665951

I would prefer B or C. But I would also be fine with using os_log directly.

Simon-Laux:
![image](https://github.com/deltachat/deltachat-ios/assets/18725968/f053387a-8263-46e2-b6ad-6bde2a165267)
So emojis are possible also in `os_log` so I would opt for using os_log directly.

r10s:
[...]

Issue #1977, opened 2023-11-15
Events do not log the account id they belong to


Simon-Laux:
This makes debugging with multiple accounts harder.

Simon-Laux:
If we use a logging library (#1973) we could use a scope while logging (like `chat.delta.account.${accountId}`) but maybe it could already be enough to just log it similar to how desktop does it:

As number:
<img width="768" alt="Bildschirmfoto 2023-11-15 um 02 47 16" src="https://github.com/deltachat/deltachat-ios/assets/18725968/dbce6a2d-2d1e-411f-98e1-c8b2c5753fa9">
Or number with emoji (emoji depending on wether the account is the selected one, though desktop also uses css to style it):
<img width="768" alt="Bildschirmfoto 2023-11-15 um 02 49 48" src="https://github.com/deltachat/deltachat-ios/assets/18725968/fbcba270-dc85-4759-a2a2-654be0c52680">

r10s:
we decided in one-to-one chat that this is no bug and we may decide to add logging information as needed and if there is really a benefit

closing for now, if required, we would directly open an issue

PR #1988, opened 2023-11-17
try out migrating to os_log for logging

Simon-Laux:
with this old syntax I'm not so sure about it, if you mess up the formatting string you get a runtime crash.
maybe we should rather use a library until we can upgrade the target to iOS 14. (see #1973)

### DIFF
--- a/DcCore/DcCore/DC/events.swift
+++ b/DcCore/DcCore/DC/events.swift
@@ -1,5 +1,6 @@
 import UIKit
 import UserNotifications
+import OSLog
 
 public let dcNotificationChanged = Notification.Name(rawValue: "MrEventMsgsChanged")
 public let dcNotificationIncoming = Notification.Name(rawValue: "MrEventIncomingMsg")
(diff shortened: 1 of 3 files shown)


r10s:
this seems to be too noisy and not really any improvement.

can't we just store the logger thingie globally and do

```
logger.info("network: \(event.data2String)")
```

instead of

```
dcContext.logger?.info("network: \(event.data2String)")
```

we won't log the account Id by that, but that's good enough for now. to get a grip on the current issues we won't use multiple accounts at the first place, and we can always iterate.

Simon-Laux:
> we won't log the account Id by that

I was just also trying out emojis and fixing #1977 already too. We can also log account it with old logger

r10s:
> We can also log account it with old logger

but that would again make logging much harder as you have to pass the accountID to the function (and we do not even have that at hand often). sure we _can_ do anything ... i am afk the weekend, let's postpone this discussion, or go for the agreed parts

Simon-Laux:
what about a simple `logger.info("\(accountId) network: \(event.data2String)")`?
or it could have a function to get an instance with a specified prefix.

r10s:
if we have the accountId at hand (and only then!) we can surely decide to log it und given circumstances (we're maybe even doing it here and there)

but i do not want a discussion about changing all existing log lines currently

initially, the idea was to use only a single logging object and maybe to use os_log() instead of print() - which both make sense - but are also not critical or urgent or seem to fix anything currently

but let's postpone that, and see next week

Simon-Laux:
> if we have the accountId at hand (and only then!) we can surely decide to log it und given circumstances (we're maybe even doing it here and there)

this is only for event logging for now and there we always have it on hand as it is a property of the event ;)

PR #2001, opened 2023-11-26
align options ephemeral options to android/desktop for consistency

### LINKED ISSUE #2000: use "5 weeks" as maximum for "disappearing messages"
and also make sure, other values >1 seconds are not shown as "off"

also minimum and maybe some other things needs to be changed, this is what the wanted state is:

Screenshot 2023-11-25 at 16 47 42

r10s:
- add "5 minutes", "30 minutes", remove "30 seconds", see https://github.com/deltachat/deltachat-android/pull/1780 for reasoning
- change "4 weeks" to "5 weeks", see https://github.com/deltachat/deltachat-android/pull/1786 for reasoning
- for timespans, that are not available as options, chose the smaller one just before (before, we've always shown "off" which was much worse)

closes #2000

### DIFF
--- a/deltachat-ios/Controller/EphemeralMessagesViewController.swift
+++ b/deltachat-ios/Controller/EphemeralMessagesViewController.swift
@@ -32,7 +32,20 @@ class EphemeralMessagesViewController: UITableViewController {
         self.dcContext = dcContext
         self.chatId = chatId
         super.init(style: .grouped)
-        self.currentIndex = self.options.index(of: dcContext.getChatEphemeralTimer(chatId: chatId)) ?? 0
+
+        // select option close to the timespan (that may no be available as an option eg. in case option have changed)
+        self.currentIndex = 0
+        let timespan = dcContext.getChatEphemeralTimer(chatId: chatId)
+        if timespan > 0 {
+            self.currentIndex = options.count - 1
+            for i in 2...options.count - 1 {
+                if timespan < options[i] {
+                    self.currentIndex = i - 1
+                    break
+                }
+            }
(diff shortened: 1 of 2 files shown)


deltachat-ios/Helper/Constants.swift
```diff
@@ -20,15 +20,13 @@ struct Constants {
 }
 
 struct Time {
-    static let thirtySeconds = 30
     static let oneMinute = 60
-    static let twoMinutes = 2 * 60
     static let fiveMinutes = 5 * 60
-    static let thirtyMinutes = 30 * 6
+    static let thirtyMinutes = 30 * 60
```
r10s, line 25: thirtyMinutes was wrongly defined before, however also not used (neither by ephemeral nor by mute)

deltachat-ios/Controller/EphemeralMessagesViewController.swift
```diff
@@ -32,7 +32,20 @@ class EphemeralMessagesViewController: UITableViewController {
         self.dcContext = dcContext
         self.chatId = chatId
         super.init(style: .grouped)
-        self.currentIndex = self.options.index(of: dcContext.getChatEphemeralTimer(chatId: chatId)) ?? 0
+
+        // select option close to the timespan (that may no be available as an option eg. in case option have changed)
```
r10s, line 36: this is the same what we're doing on android, and it seems to be good enough, esp. as this is only for cornercases as removed options - and most chat members do not tweak this option anyways

Simon-Laux (approved):

PR #2002, opened 2023-11-27
immedite feedback when tapping chat title

r10s:
iOS usually shows immediate feedback when tapping a button, eg. disable the button for a moment (you can see that on many "back" or "ok" buttons)

for the "title bar taps", this was just missing (still good to speed up loading, however, even then it is good to give immeduate feedback before animation starts)


Simon-Laux (approved):
code looks hacky to me, but if it works it's ok I guess

r10s:
in contrasr to css, these kind of things usually do not happen automatically on native os development (windows, mac, android - and ios)

PR #2003, opened 2023-11-27
move 'Password and Account' one level up

r10s:
(an idea that came out of recent desktop discussions)

this aligns to desktop, to help
and also to to many other messengers
that have 'Account' settings somewhere in the settings - and not in the 'Profile'.

still, it is tuned down by using a position further down and by using an "standard, unobtrusive" icon

EDIT: moved account related items (add second device, password and account) together

EDIT: grouped "info" things together

### DIFF
--- a/deltachat-ios/Controller/Settings/SelfProfileViewController.swift
+++ b/deltachat-ios/Controller/Settings/SelfProfileViewController.swift
@@ -11,13 +11,7 @@ class SelfProfileViewController: UITableViewController, MediaPickerDelegate {
     private let section1Status = 2
     private let section1RowCount = 3
 
-    private let section2 = 1
-    private let section2AccountSettings = 0
-    private let section2RowCount = 1
-
-    private let sectionCount = 2
-
-    private let tagAccountSettingsCell = 1
+    private let sectionCount = 1
 
     private lazy var mediaPicker: MediaPicker? = {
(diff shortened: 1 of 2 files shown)


r10s:
maybe it is more logical to move it one position up, so directly after "Add Second Device", which is strongly account-related as well

r10s:
> Just an idea.
> Maybe it would make more sense to move “Password and Account” to the top?

it is like that on desktop, however, this very top "in your face" position does not align with its importance - if you logged in, you can forget that option for  months, maybe forever. it is not uncommon that you _never_ go there again, esp. for automatically generated accounts

on android/ios it was hidden in the profile settings therefore, and no one really complained - so, moving it one level up makes sense, however, then, to a "tuned down" position

anyways, i think, one can argue for many ordering -- the idea of this pr is not to revamp and to reorder things completely (which is also an ux issue wrt habits), but more to do small improvements and align things among the systems

r10s:
let's move this is when 1.42 is settled (at least one other update with core 1.131.8 will be needed

r10s:
i created a more top-level discussion for that at https://github.com/deltachat/interface/issues/57 , let's figure things out there first and then just apply that on ios

r10s:
currently all systems are in-sync; let's park this PR in resurrection until https://github.com/deltachat/interface/issues/57 is resolved;

r10s:
moved to "advanced" on all platforms; due to "chatmail" changes this became less important

PR #2012, opened 2023-12-07
thread-safe event-shutdown

r10s:
this PR cherry-picks event-shutdown handling from https://github.com/deltachat/deltachat-ios/pull/1969/files authored by @Simon-Laux 

shutdown is done only on termination (so just before the app quits comparable to `exit()`), there are no hints iOS cares about open connections on suspend (in contrast to file locks leading to 0xdead10cc)

@link2xt stopIO() may be called also before halting the event handler, is that okay?

@link2xt i do not really understand point 5. from https://github.com/deltachat/deltachat-core-rust/blob/main/deltachat-ffi/deltachat.h#L198 ("Keep calling dc_get_next_event() in a loop until it returns NULL, indicating that the contexts are deallocated." - is this needed also when using `dc_accounts_t`? i also do not see that in the "reference" implementation from https://github.com/deltachat/deltachat-core-rust/pull/3289/files

do we need to add a `while eventEmitter.getNextEvent() != nil { /*do nothing*/}` at the end of `applicationWillTerminate()` just before the app terminates?

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -271,6 +271,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
     func applicationWillTerminate(_: UIApplication) {
         logger.info("⬅️ applicationWillTerminate")
+        uninstallEventHandler()
         dcAccounts.closeDatabase()
         if let reachability = reachability {
             reachability.stopNotifier()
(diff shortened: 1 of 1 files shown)


Simon-Laux (approved):

r10s:
let's merge that in and do a testflight, even if there are some unanswered questions left. it is an improvement anyways, and is already quite tested

PR #2013, opened 2023-12-07
prepare 1.42.7

r10s:


r10s:
let's merge this, testflight is on its way

PR #2016, opened 2023-12-11
document push certificate renewal

r10s:
i had a similar documentation locally already, however, it makes sense to share that "knowledge"

also i find that local documentation always hard to find (it  is used too rarely), i think, here in the `docs` folder, it is more visible.

@missytake if there is sth similar in some server-docs, maybe you can point to this document - and we can also do the other way round in step 4.

### DIFF
--- a/docs/renew-push-certificate.md
+++ b/docs/renew-push-certificate.md
@@ -1,0 +1,45 @@
+# how to renew "Apple Push Services Certificate"
+
+About two times a year, we get an email from Apple
+reminding to renew the "Apple Push Services Certificate"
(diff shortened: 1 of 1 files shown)


Simon-Laux (approved):
maybe also link to https://github.com/deltachat/notifiers (server software source code, you might want to find it and it's not always obvious where things are)

adbenitez (approved):

missytake:
I linked to it in https://github.com/deltachat/sysadmin/tree/master/page#2023-12-12---notificationsdeltachat-update-apple-certificate (where I documented the cert renewal)

PR #2017, opened 2023-12-12
use core background fetch

Simon-Laux:
use https://github.com/deltachat/deltachat-core-rust/pull/5104 instead of the semaphore that waited for the event.
I was conservative with removing code, maybe we can remove even more code now, because we now only have one blocking core call that has 
timeout built-in.

### DIFF
--- a/DcCore/DcCore/DC/Wrapper.swift
+++ b/DcCore/DcCore/DC/Wrapper.swift
@@ -10,7 +10,6 @@ public class DcAccounts {
     /// The ID is created in the apple developer portal and can be changed there.
     let applicationGroupIdentifier = "group.chat.delta.ios"
     var accountsPointer: OpaquePointer?
-    public var fetchSemaphore: DispatchSemaphore?
 
     public init() {
     }
@@ -65,6 +64,10 @@ public class DcAccounts {
 
     public func stopIo() {
         dc_accounts_stop_io(accountsPointer)
+    }
+    
+    public func backgroundFetch(timeout: UInt64) -> Bool {
+        return dc_accounts_background_fetch(accountsPointer, timeout) == 1
     }
 
     public func select(id: Int) -> Bool {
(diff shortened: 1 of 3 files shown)


r10s:
this is much cleaner code, indeed! nice!

i have one question wrt timing: when does `dc_accounts_background_fetch_with_timeout()` return? is that comparable to when `dc_all_work_done()` returns `true`?

asking as processing is done by events which handlers may still run if `dc_accounts_background_fetch_with_timeout()` returns much sooner that `dc_all_work_done()`  is fired (not saying that for for `dc_all_work_done()` things are perfect wrt timing, just want to bring up that aspect)

Simon-Laux:
> is that comparable to when dc_all_work_done() (which can be removed btw) returns true?

yes, maybe even slightly faster because it does less. It always takes as long as the slowest account, so 1-8 seconds in my test. (the riseup provider was sometimes slow, the average/normal time of the other providers was sth. like 2 seconds)


> asking as processing is done by events which handlers may still run if dc_accounts_background_fetch_with_timeout() returns much sooner that dc_all_work_done() is fired (not saying that for for dc_all_work_done() things are perfect wrt timing, just want to bring up that aspect)

There is still the sleep second after, that should give enough time to process all events (empty the queue), but we could also do a semaphore to empty the event queue to be absolutely sure. Good catch, though also note that relying on the connectivity event which changed behaviour brought us the bug in the first place, so I would be careful about relying on events like that (that we could break by accident again).

We could also make a new `BackgroundFetchCompletedForAllAccounts` event that the core function could add to the event queue at the end / or when it fails, then we could have a semaphore for reaching that exact event and don't need any additional core check "if work is already fully done".

Simon-Laux:
https://github.com/deltachat/deltachat-core-rust/pull/5104 now has an `BackgroundFetchCompletedForAllAccounts` event, this pr still needs to be updated, when we use this event, we might be able to shorten or even remove the second of sleep.

r10s:
this PR is superseded by https://github.com/deltachat/deltachat-ios/pull/2017, that incorporates suggestions from this PR - thanks for your initial work!

PR #2019, opened 2023-12-14
fix some images not shown in gallery

### LINKED ISSUE #1911: Some images are not displayed in "All Media" and the local media library
Some images are not displayed in "_All Media_" and the local media library. 
Instead, there is just an empty field. If you tap on it, the picture will open normally. 
All images are displayed in the list view and in the chat history.
I didn't notice this bug before, so I'm not sure if it could have something to do with yesterday's iOS 16.6 update.

Device: iPhone 14
iOS version: 16.6
Delta Chat version: 1.38.2

Addendum:
The bug also occurs in DC-iOS 1.40 (iOS 17.01).
When I create a backup and load it into DC-Desktop 1.40.4, all images are displayed normally in the media library there.

r10s:
with this PR, images can be loaded in gallery that failed before.

not sure about the _why_, however, SDAnimagedImage seems to fail sometimes where UIImage succeeded - maybe also the other way round, images are complicated in hard.

the failing images where always rendered in the bubbles btw - but there we did not use SDAnimagedImage directly - but SDAnimagedImageView - might be that that wrapper also calls UIImage if in doubt.

anyways, the nice bears (test images that were failing before) are rendered correctly now :)

Screenshot 2023-12-14 at 23 55 34

closes #1911

### DIFF
--- a/deltachat-ios/Helper/ImageFormat.swift
+++ b/deltachat-ios/Helper/ImageFormat.swift
@@ -48,17 +48,12 @@ extension ImageFormat {
         return .unknown
     }
 
-
-    // Theoretically, SDAnimatedImage should be able to read data of all kinds of images.
-    // In practive, JPG files haven't been read correctly, for now we're using SDAnimatedImage
-    // only for file formats that support animated content.
     static func loadImageFrom(data: Data) -> UIImage? {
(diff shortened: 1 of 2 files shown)


r10s:
merging to move forward with #1330 that partly uses code changes from this PR

PR #2025, opened 2023-12-20
delete unused resources

r10s:
this PR deletes more than 250 resource images.

more than half of the resource bytes that were left behind over time, cleaning up 1.7mb from 3.3mb.

as the resources are mostly references by strings on iOS, it may even be that the app shrinks by that amount. but if not, it is much nicer and easier maintainable without the legacy mess, partly even totally unused android icons ...

i commented some steps and reasonings in the commit messages.

the images were partly created by previous contributors (@cyBerta , @dignifiedquire, @nayooti, @jonasreinsch and others) and i checked carefully if they're not in use, hopeful not to have overseen sth :)


deltachat-ios/Extensions/UIImage+Extension.swift
```diff
@@ -65,9 +65,3 @@ extension UIImage {
         return finalImage
     }
 }
-
-public enum ImageType: String {
-    case play
-    case pause
-    case disclouser
-}
```
r10s, line 73: this is just unused

deltachat-ios/Helper/ShortcutManager.swift
```diff
@@ -80,7 +80,7 @@ public class ShortcutManager {
         }
 
         let infoDict = message.getWebxdcInfoDict()
-        let iconData = scaledDownLogo?.pngData() ?? UIImage(named: "appicon")?.pngData() ?? nil
+        let iconData = scaledDownLogo?.pngData() ?? nil
```
r10s, line 83: a fallback beside "nil" for "no webxdc icon" is not needed - core already provides a fallback (the webxdc icon)

Simon-Laux (approved):
we can always read them later if sth breaks

Issue #2009, opened 2023-12-06
use new core that avoids 0xdead10cc on dc_accounts_t


r10s:
see https://github.com/deltachat/deltachat-core-rust/pull/5081

this replaces the accounts deletion and creation of #1969

Issue #1882, opened 2023-05-23
no notifications or delayed notifications


mytickets617:
- Device: iphone, ipad various versions
- iOS version: 16.5
- Delta Chat version: 1.36.4
- Server: Dovecot, push enabled
- Expected behavior: receive a notification on new IM
- Actual behavior: no notifications
- Description:
There are multiple DC support and github threads, previously closed, reporting same issue. Not clear why they were closed though as same issue remains unresolved. This renders the app useless as a Chat tool. The suggested AppleMail workaround is not feasable. Here is the latest, which references several prior tickets:
https://support.delta.chat/t/notifications-missed-updates-on-ios-mailbox-org/2559

r10s:
notifications are tricky on ios when not using a central server or at least the help of the mail servers.

current expected behaviour is that we _really_ make use of the time slots the system provides to us - at best, this is every ~20 minutes, when the app is used less often, even less.

can you please provide more information about what is shown in "settings / connectivity view" and in  the log "settings / advanced / view log".

cmp. https://github.com/deltachat/deltachat-core-rust/issues/4420 for ongoing improvements (we have this issue on the radar all the time btw, but we usually track only really actionable items)

r10s:
> Notifications: Connected

that should show more information, sth. as:

<img width=320 src=https://github.com/deltachat/deltachat-ios/assets/9800740/84ee82d6-02f9-4c1a-930d-cf73537f2701>



> Which Log keys do you need to see?

to get a grip on your issue, all parts of the log may be of interest (in fact, it is not a log but mainly a list of variables), you can of course  remove personal information as name or email. esp. the part above arch=64 is interesting:

<img width=320 src=https://github.com/deltachat/deltachat-ios/assets/9800740/9f68cfbf-7b6c-451d-9f28-84e48e27c995>

mytickets617:
Debug value was reported above.
The iphone device was sitting without anyone touching for the past ~2hr. It received 1 notification just now surprisingly. Also, now "Notifications:" lists "Delayed, Checked at 06:17 PM, 05:52 PM, 05:26 PM, On average every 25 minutes.

r10s:
yip, the "sattelite" and other icon stuff is important as well as everything above (that shows times)

JoeMcNuggets:
I've lost a log file, but I'm sure I've seen the next case:
- all the checks successfully passed in the time message arrived
- no notification received

I have to idease:
[...]

PR #2029, opened 2023-12-23
shorten notification setup, fix potential crashes

r10s:
- improve error handling when notification registering fails
- force dispatching to main, this could be a source of crashes as `registerForNotifications()` was called at least partly on a background thread from the connectivity handler
- when looking on any example, the additional check to getNotificationSettings() is redundant

cmp #2009, #1882, #1202

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -324,7 +324,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
     // MARK: - background fetch and notifications
 
-    // asks the user if they want to get notifiations shown.
+    // asks the user if they want to get notifications shown.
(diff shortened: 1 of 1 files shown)


adbenitez (approved):

deltachat-ios/AppDelegate.swift
```diff
@@ -324,46 +324,26 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
     // MARK: - background fetch and notifications
 
-    // `registerForNotifications` asks the user if they want to get notifiations shown.
-    // if so, it registers for receiving remote notifications.
+    // asks the user if they want to get notifiations shown.
```
adbenitez, line 327:
```suggestion
    // asks the user if they want to get notifications shown.
```

Issue #1728, opened 2022-10-24
send reactions


r10s:
with recent [core master](https://github.com/deltachat/deltachat-core-rust/pull/3644), we have apis to send reactions, there is a separate issue about [showing reactions](https://github.com/deltachat/deltachat-ios/issues/1727), this on comes _after that_ and is about sending reactions.

most bigger messengers offer to send an reaction by **long-tapping** the message. the possible reactions are shown near or as a menu then:

<img width=300 src=https://user-images.githubusercontent.com/9800740/197546398-fb391f62-9383-489e-9800-68fd99d6e808.png> _whatsapp_

<img width=300 src=https://user-images.githubusercontent.com/9800740/197546433-412ac71f-1413-4786-8a21-d8858e5804c1.png> _signal_

<img width=300 src=https://user-images.githubusercontent.com/9800740/197546430-22434fef-5cb4-4198-866f-fd4b3e8a9d70.png> _imessage_

<img width=300 src=https://user-images.githubusercontent.com/9800740/197546428-04b1bf9d-e023-4fdc-880d-8e9e48161483.png> _element_

<img width=300 src=https://user-images.githubusercontent.com/9800740/197546421-bbfe05c2-c73c-44d7-87b9-1be2e0b5bc6e.png> _telegram_

our current context menu looks quite similar to whatsapp/signal/imessage/telegram, so that seems to be a good base.

if easily possible, we can go for sth. similar, so showing the reaction options **opposite to the rest of the menu**.

but i would also be fine in integrating them into the menu, so we would still have one place for the menu. integration, however, would still mean direct access and that the reaction options are show side-by-side, not below each other.

i'd say, whatever is easier.

then:
- the real set of reactions needs to be discussed, assume a handful of default reactions _plus_ the option to pick any emoji.
- if the user already reacted, their reaction should possibly be hilited (see the "heart" in the whatsapp image above)
- the context menu is already quite large in height, however, we can save one line by moving "Reply" and "Reply Privately" to a submenu - reply is probably done most time by swiping and "Reply Privately" is used not so often. the name of the submenu would be just "Reply". ideally, the submenu would look like the one in safari ios at "Tabs / Move to Tab Group": <img width=170 src=https://user-images.githubusercontent.com/9800740/197552567-0bb13664-e32b-4eb0-ba74-680677d10f69.PNG>

## timing

there us no urgency, there is even more time as for #1727

cmp. https://github.com/deltachat/deltachat-desktop/pull/2964, #1727

Issue #1727, opened 2022-10-24
show reactions


r10s:
with recent [core master](https://github.com/deltachat/deltachat-core-rust/pull/3644), we have apis to get reactions from core; we should display them near the bubbles.

i think, it makes sense to show then **below the text**, possibly a **bit outside** the bubble, on the bubble border. this makes clear the reaction is not part of what was initially sending and also is logic wrt "top-down".

most, but not all bigger messengers follow these guidelines, here are some examples how reactions are displayed by others:

<img width=300 src=https://user-images.githubusercontent.com/9800740/197538476-55189149-a4a9-4e06-b125-5c0b0c57f963.png> _whatsapp_

<img width=300 src=https://user-images.githubusercontent.com/9800740/197538472-df9ebe02-7ff2-4d7f-9cf8-3ce8d72e0c28.png> _signal_

<img width=300 src=https://user-images.githubusercontent.com/9800740/197538483-2318aaec-1f31-4b5a-a396-4b85fd2a594d.png>  _imessage_

<img width=300 src=https://user-images.githubusercontent.com/9800740/197538481-2642d2bc-1665-4dd6-b70a-3be4bbbf8793.png> _element_

<img width=300 src=https://user-images.githubusercontent.com/9800740/197538465-9e845a53-227f-4ef5-9412-8890fa7b352c.png> _telegram_

- the reaction line should display the "reaction emoji", followed by a counter (if >1), space, next "reaction emoji"
- a "reaction emoji" may be any kind of short text, although that is not encouraged to be sent, we have to deal with that
- there may be more reactions than fit into one line, this is probably rare, so we should do sth. easy (maybe just cut, details are available as, see next point)
- a tap in the reaction line should open reaction details (cmp. signal, whatsapp); some implementation also allow to send reactions from there, but i think, we should offer a unique way for that (long tap, see dedicated issue).

## timing

there are still some issues for the message list, leading to too much scrolling and display annoyances. targeting reaction layout is tempting, we should resist and target the bugs first - getting a grip on them won't be easier with complexity added to the message list.

also, there is no urgency, we want to ship a round of releases that have reactions in core but no option to send them.

cmp. https://github.com/deltachat/deltachat-desktop/pull/2964, https://github.com/deltachat/deltachat-ios/issues/1728

r10s:
[...]

PR #2039, opened 2024-01-09
add reactions api

r10s:
this PR adds two new apis, `sendReaction()` and `getReaction()`.

to make them work, jsonrpc is needed and added as well by this PR. while the json result of `getReaction()` is parsed gracefully to a structure, ~~the input json is just written as a string for now.~~ EDIT: both are created properly now

nb: for testing, i added the line `dcContext.sendReaction(messageId: message.id, reaction: "🤘")` before line 49 of MessageInfoViewController  - that way, you send a reaction whenever you long tap a message and select "Info" :)

targets #1727 and #1728


DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -456,6 +480,21 @@ public class DcContext {
...
 
+    public func getMessageReactions(messageId: Int) -> DcReactions? {
+        if let data = dcAccounts.blockingCall(method: "get_message_reactions", params: "\(id), \(messageId)") {
+            return try? JSONDecoder().decode(DcReactionResult.self, from: data).result
+        }
+        return nil
+    }
+
+    public func sendReaction(messageId: Int, reaction: String?) {
+        if let reaction = reaction {
```
zeitschlag, line 494:
```suggestion
        if let reaction {
```

The `= reaction` is not needed anymore since [Swift 5.7](https://www.swift.org/blog/swift-5.7-released/)

DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -1530,6 +1572,21 @@ public class DcProvider {
...
+}
+
+public struct DcReactions: Decodable {
+    let reactions: [DcReaction]
+    let reactionsByContact: [Int: [String]]
+}
+
+struct DcReactionResult: Decodable {
+    let result: DcReactions
+}
```
zeitschlag, line 1588: This is basically mapped from [`reactions.rs`](https://github.com/deltachat/deltachat-core-rust/blob/main/deltachat-jsonrpc/src/api/types/reactions.rs), right?

DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -95,12 +99,35 @@ public class DcAccounts {
...
+        }
+
+        if accountsPointer != nil {
+            dc_accounts_unref(accountsPointer)
+            accountsPointer = nil
+        }
+    }
+
+    @discardableResult
+    public func blockingCall(method: String, params: [AnyObject]) -> Data? {
```
zeitschlag, line 120: Is there a reason to use `JSONSerialization` and `AnyObject` instead of `Codable` and `JSONEncoder`?

DcCore/DcCore/DC/Wrapper.swift
```diff
@@ -2,23 +2,27 @@ import Foundation
 import UIKit
 import AVFoundation
 
+private let dcAccounts = DcAccounts()
 private var encryptedDatabases: [Int: Bool] = [:]
 
+public func getDcAccounts() -> DcAccounts {
+    return dcAccounts
+}
```
zeitschlag, line 10: What's the idea behind this?

zeitschlag (commented):
LGTM, I added a few questions and done remark regarding `if let` since Swift 5.7

DcCore/DcCore/DC/Wrapper.swift
r10s: yes.

DcCore/DcCore/DC/Wrapper.swift
r10s: is the one considered better as the other?

`JSONEncoder` did not work for me at the first place, maybe i did sth. wrong. however, as `JSONSerialization` works out with very few code as well, i also did not investigate further as i found a solution.

DcCore/DcCore/DC/Wrapper.swift
r10s: dcAccounts is a global handle; it is there exactly one time on the app. however, one cannot access such globals by variables across modules, therefore a simple getter

zeitschlag: What do you think about adding a `shared`-property to the `DcAccounts`-class to make it more object-oriented/iOS-y?

zeitschlag (approved):

DcCore/DcCore/DC/Wrapper.swift
r10s: would be fine for me as well. i am not that deep in these iOS-y things, so, best is if you push sth. directly to this pr

zeitschlag: Nah, this is your PR, if I contributed, reviewing my own code wouldn't make a lot of sense to me. I'd refactor this on the go somewhen.

DcCore/DcCore/DC/Wrapper.swift
r10s: i just double-checked: `JSONEncoder` cannot encode `[AnyObject]` as not derived from `Codeable`; it would require some additional recursive wrapper. as we're already using `JSONSerialization` in other parts, it is probably fine to stay with that for now; tweaking things here seems not worth the effort currently, we can iterate over that at any point if that becomes an issue

PR #2044, opened 2024-01-17
Split wrapper into seperate files

zeitschlag:


deltachat-ios/DC/DcAccount.swift
```diff
@@ -0,0 +1,134 @@
+//
+//  DcAccount.swift
+//  DcCore
+//
+//  Created by Nathan Mattes on 17.01.24.
+//  Copyright © 2024 merlinux GmbH. All rights reserved.
+//
+
```
r10s, line 8: also here: no such headers (same in other files)

r10s (approved):
if the copyright-headers are removed, this lgtm - i assume code is not really changed and bespoken :)

zeitschlag:
> i assume code is not really changed

@r10s I have to admit: I changed some code, like rewriting the `strToBool`-function to a method in a extension of [`String?`](https://github.com/deltachat/deltachat-ios/pull/2044/files#diff-46b0af42419def7a6fc92e2ea65c6483db316e2745a22ed91f6fdfd64748a472) and also `DcAccounts` and `DcContext` don't share `encryptedDatabase` anymore. Can you please give me 👍 or 👎 if the code looks fine for you? I'd wait with the merge for your feedback :)

DcCore/DcCore/Extensions/String+Extensions.swift
```diff
@@ -56,5 +56,13 @@ public extension String {
         }
         return attributedText
     }
+}
+
+extension String? {
+    /// returns `true` if `"1"`, in case of `nil` or any other string: return false
```
r10s, line 62: the following makes my mind comments easier to parse:

```suggestion
    /// returns `true` if `"1"`, returns `false` in case of `nil` or any other string
```

i was also thinking about a more descriptive function name (this is generally better than comments), but i did not came to sth better :)

imu, the "numeric" points out somehow that strings as `"true"` are _not_ parse and that it is a bit more special :)

r10s (approved):
thanks for the pointers to changed code, i've overseen that :) but still lgtm, the changes make sense and as divided by clean commits, this is a also a reasonable way to go (one big refactor commit, some on-point-changes)

PR #2046, opened 2024-01-18
update reaction view on DC_EVENT_REACTIONS_CHANGED

r10s:
handling of this event was just missing.
for simplicity, the event is handled that in the same way as the other more explicit events;
if really needed, we can optimize updates later.

this should also prepare for sending reactions,
as DC_EVENT_REACTIONS_CHANGED is also emitted for self-reactions (so, one can just call sendReaction(), no need to update anything)


zeitschlag (approved):
Ah, LGTM, thank you @r10s

PR #2048, opened 2024-01-19
Send Reactions

### LINKED ISSUE #1728: send reactions

zeitschlag:
Adds a view to send emoji-reactions. When you can't send a message in that very chat, you also won't be able to send an emoji-reaction. Please note, that this works only for iOS 13+.

For now, there are five default emojis: 👍, 👎, ❤️, ~😀~ 😂, 🙁. It is planned for the foreseable future to add an Emoji-selection. Your own reactions are visually marked. If you tap them again, your reaction disappears. And it looks like this on iPad:

## Screenshots

1728_ipad

On iPhone, it looks like this:

1728_iphone

The different blurry backgrounds come from the operating system (AFAIK). Another change is that only the message-bubble is highlighted, not the entire cell anymore. I also sneaked in some cleanups (see `UIView+Extensions`).

Closes #1728.

### DIFF
--- a/DcCore/DcCore/Extensions/UIView+Extensions.swift
+++ b/DcCore/DcCore/Extensions/UIView+Extensions.swift
@@ -7,19 +7,19 @@ public extension UIView {
         self.layer.borderWidth = 2
     }
 
-    func alignLeadingToAnchor(_ anchor: NSLayoutXAxisAnchor, paddingLeading: CGFloat = 0.0, priority: UILayoutPriority? = .none) {
+    func alignLeadingToAnchor(_ anchor: NSLayoutXAxisAnchor, paddingLeading: CGFloat = 0.0, priority: UILayoutPriority = .required) {
         _ = constraintAlignLeadingToAnchor(anchor, paddingLeading: paddingLeading, priority: priority)
     }
 
-    func alignTrailingToAnchor(_ anchor: NSLayoutXAxisAnchor, paddingTrailing: CGFloat = 0.0, priority: UILayoutPriority? = .none) {
+    func alignTrailingToAnchor(_ anchor: NSLayoutXAxisAnchor, paddingTrailing: CGFloat = 0.0, priority: UILayoutPriority = .required) {
         _ = constraintAlignTrailingToAnchor(anchor, paddingTrailing: paddingTrailing, priority: priority)
     }
 
-    func alignTopToAnchor(_ anchor: NSLayoutYAxisAnchor, paddingTop: CGFloat = 0.0, priority: UILayoutPriority? = .none) {
+    func alignTopToAnchor(_ anchor: NSLayoutYAxisAnchor, paddingTop: CGFloat = 0.0, priority: UILayoutPriority = .required) {
         _ = constraintAlignTopToAnchor(anchor, paddingTop: paddingTop, priority: priority)
     }
 
-    func alignBottomToAnchor(_ anchor: NSLayoutYAxisAnchor, paddingBottom: CGFloat = 0.0, priority: UILayoutPriority? = .none) {
+    func alignBottomToAnchor(_ anchor: NSLayoutYAxisAnchor, paddingBottom: CGFloat = 0.0, priority: UILayoutPriority = .required) {
         _ = constraintAlignBottomToAnchor(anchor, paddingBottom: paddingBottom, priority: priority)
     }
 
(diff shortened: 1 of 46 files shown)


DcCore/DcCore/Extensions/UIView+Extensions.swift
```diff
@@ -7,19 +7,19 @@ public extension UIView {
         self.layer.borderWidth = 2
     }
 
-    func alignLeadingToAnchor(_ anchor: NSLayoutXAxisAnchor, paddingLeading: CGFloat = 0.0, priority: UILayoutPriority? = .none) {
+    func alignLeadingToAnchor(_ anchor: NSLayoutXAxisAnchor, paddingLeading: CGFloat = 0.0, priority: UILayoutPriority = .required) {
```
r10s, line 10: i am sure, it is correct, just asking for understanding: what is the change from `.none` to `.required` about? or was the default just never used?

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1847,28 +1843,90 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
...
+    private func targetedPreview(for configuration: UIContextMenuConfiguration, reactionsHidden: Bool) -> UITargetedPreview? {
         guard let messageId = configuration.identifier as? NSString else { return nil }
         guard let index = messageIds.firstIndex(of: messageId.integerValue) else { return nil }
         let indexPath = IndexPath(row: index, section: 0)
+        let message = dcContext.getMessage(id: messageId.integerValue)
+
+        guard let cell = tableView.cellForRow(at: indexPath) as? BaseMessageCell,
+              let messageSnapshotView = cell.messageBackgroundContainer.snapshotView(afterScreenUpdates: false) else { return nil }
+
+        cell.messageBackgroundContainer.isHidden = (reactionsHidden == false)
```
r10s, line 1864: are you avoiding `!reactionsHidden` because the `!` if is easy to overlook?  

(i personally find the standard boolean negation more readable and have rarely overseen it :)

deltachat-ios/Chat/Send Reaction/SendReactionsView.swift
```diff
@@ -0,0 +1,134 @@
...
+    case heart
+    case haha
+    case sad
+
+    var emoji: String {
+        switch self {
+        case .thumbsUp: return "👍"
+        case .thumbsDown: return "👎"
+        case .heart: return "❤️"
+        case .haha: return "😀"
```
r10s, line 16: i've just noticed, we're using 😂 on android/ios; we should use the same here then

```suggestion
        case .haha: return "😂"
```

r10s (approved):
some minor comments and questions, the "haha" emoji should be changed all other comments can be ignored :)

great, accurate PR, also thanks a lot for testing on different devices etc.

deltachat-ios/Chat/ChatViewController.swift
zeitschlag: > because the ! if is easy to overlook?

Exactly!

DcCore/DcCore/Extensions/UIView+Extensions.swift
zeitschlag: It's about the Optional in combination with the later `if let priority`:

The [default priority of a constraint](https://developer.apple.com/documentation/uikit/nslayoutconstraint/1526946-priority) is `required`. If you don't set/change it, it'll be `required`. So why using an Optional here if there's a default?

r10s:
strike through emoji ~😀~ looks like "Salvator Dali" :)

PR #2049, opened 2024-01-19
move 'pinned' indicator right for cleaner UI

r10s:
'pinned' chats are always shown together below each other; moving the 'pinned' indicator to the same position in each chat makes the UI a little less cluttered and nicer
(all pinned indicators will form a column if you will)

this PR is an outcome of the feature-proposal at https://support.delta.chat/t/three-small-changes-to-make-delta-chat-better/2886


Simon-Laux (approved):
looks not 100% aligned but probably wasn't before either, so good to be merged

PR #2050, opened 2024-01-20
do not add "Changelog" for new accounts

r10s:
on new installations or accounts,
the "What's new?" messages only clutter the device messages with unimportant information
(either the user is new to the app, so everyrhing is new and nothing is changed, or an existing user has created a new account, they have seen the changelog anyways)

in other words,
the changelog will be added only for updates.

when reading https://c.delta.chat/classdc__context__t.html#a1a2aad98bd23c1d21ee42374e241f389 , this was always the gist, however, somehow forgotten or removed on the way :)

technically, it was easier to track the creation in an array, as the user flow from creation to finally addind the device message is a bit too complicated to pass flags around.

counterpart of https://github.com/deltachat/deltachat-android/pull/2935 and https://github.com/deltachat/deltachat-desktop/issues/3637

### DIFF
--- a/deltachat-ios/Controller/ChatListViewController.swift
+++ b/deltachat-ios/Controller/ChatListViewController.swift
@@ -134,9 +134,14 @@ class ChatListViewController: UITableViewController {
 
         // update messages - for new messages, do not reuse or modify strings but create new ones.
         // it is not needed to keep all past update messages, however, when deleted, also the strings should be deleted.
-        let msg = dcContext.newMessage(viewType: DC_MSG_TEXT)
-        msg.text = String.localized("update_1_42_common") + "\n\n" + String.localized("more_info_desktop") +  ": https://get.delta.chat/#changelogs"
-        dcContext.addDeviceMessage(label: "update_1_42b_ios", msg: msg)
+        let deviceMsgLabel = "update_1_42b_ios2"
+        if !dcAccounts.isJustAdded(id: dcContext.id) {
+            let msg = dcContext.newMessage(viewType: DC_MSG_TEXT)
+            msg.text = String.localized("update_1_42_common") + "\n\n" + String.localized("more_info_desktop") +  ": https://get.delta.chat/#changelogs"
+            dcContext.addDeviceMessage(label: deviceMsgLabel, msg: msg)
+        } else {
+            dcContext.addDeviceMessage(label: deviceMsgLabel, msg: nil)
+        }
 
         if dcContext.isAnyDatabaseEncrypted() {
             let msg = dcContext.newMessage(viewType: DC_MSG_TEXT)
--- a/deltachat-ios/Controller/ChatListViewController.swift
+++ b/deltachat-ios/Controller/ChatListViewController.swift
@@ -135,7 +135,7 @@ class ChatListViewController: UITableViewController {
         // update messages - for new messages, do not reuse or modify strings but create new ones.
         // it is not needed to keep all past update messages, however, when deleted, also the strings should be deleted.
         let deviceMsgLabel = "update_1_42b_ios2"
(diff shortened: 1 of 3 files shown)


deltachat-ios/DC/DcAccount.swift
```diff
@@ -12,6 +12,7 @@ public class DcAccounts {
     public var fetchSemaphore: DispatchSemaphore?
 
     private var encryptedDatabases: [Int: Bool] = [:]
+    private var justAdded: [Int: Bool] = [:]
```
zeitschlag, line 15: Instead of a dictionary, what do you think about using a simple list `freshlyAddedAccountIds: [Int] = []` as the `Bool` will never be `false`?

deltachat-ios/DC/DcAccount.swift
```diff
@@ -26,6 +27,7 @@ public class DcAccounts {
     public func add() -> Int {
         let accountId = Int(dc_accounts_add_account(accountsPointer))
         get(id: accountId).setConfig("verified_one_on_one_chats", "1")
+        justAdded[accountId] = true
```
zeitschlag, line 30: This would change to `freshlyAddedAccountIds.append(accountId)`

deltachat-ios/DC/DcAccount.swift
```diff
@@ -34,6 +36,10 @@ public class DcAccounts {
         return DcContext(contextPointer: contextPointer)
     }
 
+    public func isJustAdded(id: Int) -> Bool {
+        return justAdded[id] ?? false
```
zeitschlag, line 40: and this could be as simple as `freshlyAddedAccountIds.contains(accountId)`

deltachat-ios/Controller/ChatListViewController.swift
```diff
@@ -134,9 +134,14 @@ class ChatListViewController: UITableViewController {
...
-        let msg = dcContext.newMessage(viewType: DC_MSG_TEXT)
-        msg.text = String.localized("update_1_42_common") + "\n\n" + String.localized("more_info_desktop") +  ": https://get.delta.chat/#changelogs"
-        dcContext.addDeviceMessage(label: "update_1_42b_ios", msg: msg)
+        let deviceMsgLabel = "update_1_42b_ios2"
+        if !dcAccounts.isJustAdded(id: dcContext.id) {
+            let msg = dcContext.newMessage(viewType: DC_MSG_TEXT)
+            msg.text = String.localized("update_1_42_common") + "\n\n" + String.localized("more_info_desktop") +  ": https://get.delta.chat/#changelogs"
+            dcContext.addDeviceMessage(label: deviceMsgLabel, msg: msg)
+        } else {
+            dcContext.addDeviceMessage(label: deviceMsgLabel, msg: nil)
```
zeitschlag, line 143: Why is adding a `nil`-message necessary?

zeitschlag (approved):
LGTM 👍  I added some questions and a suggestion here and there. The PR does work, so please consider the suggestion a suggestion and feel free to reject/ignore it 🙂

deltachat-ios/DC/DcAccount.swift
r10s: totally makes sense and makes the gist clearer (i was just copying the pattern from `encryptedDatabases` databases without thinking much about it :)

i added a pr that picks up the idea

deltachat-ios/Controller/ChatListViewController.swift
r10s: a nil-message is added to mark the label as used; this avoids the message being added on the next start.

[`dc_add_device_msg()`](https://c.delta.chat/classdc__context__t.html#a1a2aad98bd23c1d21ee42374e241f389) tries to explain that maybe better :)

PR #2051, opened 2024-01-25
simplify message list updates

### LINKED ISSUE #1691: just sent message is not always added to chat view

r10s:
radically simplify message list updates, similar to https://github.com/deltachat/deltachat-android/blob/main/src/org/thoughtcrime/securesms/ConversationFragment.java#L977

this picks up ideas and comments up from #1716 and closes #1691 and fixes the accidentally added reaction-bubbles from #2048

it is needed to be tested a bit, and maybe some corner cases are missed and need to be tweaked, but at a first glance, things seem to work as expected

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -553,23 +553,23 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
             object: nil,
             queue: OperationQueue.main
         ) { [weak self] notification in
-            guard let self, let id = notification.userInfo?["message_id"] as? Int else { return }
-            if self.dcChat.canSend, id > 0 {
-                let msg = self.dcContext.getMessage(id: id)
-                if msg.isInfo,
-                   let parent = msg.parent,
-                   parent.type == DC_MSG_WEBXDC {
-                    self.refreshMessages()
-                } else {
-                    self.updateMessage(msg)
+            guard let self, let ui = notification.userInfo else { return }
+            if self.chatId == ui["chat_id"] as? Int {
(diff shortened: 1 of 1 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -553,39 +553,42 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
...
-            if self.dcChat.canSend, id > 0 {
-                let msg = self.dcContext.getMessage(id: id)
-                if msg.isInfo,
-                   let parent = msg.parent,
-                   parent.type == DC_MSG_WEBXDC {
-                    self.refreshMessages()
-                } else {
-                    self.updateMessage(msg)
+            guard let self, let ui = notification.userInfo else { return }
+            let chatId = ui["chat_id"] as? Int ?? 0
```
zeitschlag, line 557: What's so special about the `chatId 0`?

zeitschlag (approved):
LGTM, seems to work fine. I'm a bit confused about the chatId `0`

deltachat-ios/Chat/ChatViewController.swift
r10s: in theory, these are batch updates, affecting multiple chats. i assume, they're not really in use in practise, but i did not check and mimiced android-behavior

r10s:
replaced by https://github.com/deltachat/deltachat-ios/pull/2052

PR #2052, opened 2024-01-26
simplify message list updates

### LINKED ISSUE #1691: just sent message is not always added to chat view

r10s:
the existing code has bugs wrt not-appearing messages (closes #1691) and too many appearing messages (#1716 shows reactions temporarily as bubbles)

this PR is the same as https://github.com/deltachat/deltachat-ios/pull/2051 but can be merged to `main`; after that #1716 should be rebased on `main`


zeitschlag (approved):
I said everything in #2051 👍

PR #2058, opened 2024-01-29
share and copy invite-links

### LINKED ISSUE #2057: allow creating invite-links in-app
the QR code copy/paste options should deal with invite links

- for copying, invite-links can be created by a one-liner as in https://github.com/deltachat/deltachat-android/pull/2943/files#diff-3cd645a2c944b39351f145cd1d6fe006de4dde3b59438c945a19a581ffcfcd8aR73 (there is no core function for that currently)

- for pasting, probably not much to do as core currentl `main` accepts these links directly - but should be tested :)

r10s:
this PR let the **Copy to Clipboard** option of the 'QR Show' dialogs copy an invite-link instead of an openpgp4fpr-link (invite links are much more useful outside delta chat as they can be tapped and also show instructions about how to install delta chat)

moreover, this PR adds the option **Share** to the 'QR Show' dialogs, so that one can directly send the invite-links eg. via other messengers

 

on the receiving side, both, invite-links as well as openpgp4fpr-links can be pasted since https://github.com/deltachat/deltachat-core-rust/pull/5219 , as external protocol, for now openpgp4fpr is supported (which is shown on the invite page)

closes #2057

### DIFF
--- a/deltachat-ios/Controller/FilesViewController.swift
+++ b/deltachat-ios/Controller/FilesViewController.swift
@@ -307,9 +307,11 @@ extension FilesViewController {
         parentViewController.present(activityVC, animated: true, completion: nil)
     }
 
-    public static func share(text: String, parentViewController: UIViewController) {
-        let activityVC = UIActivityViewController(activityItems: [text], applicationActivities: nil)
-        parentViewController.present(activityVC, animated: true, completion: nil)
+    public static func share(url: String, parentViewController: UIViewController) {
+        if let url = URL(string: url) {
+            let activityVC = UIActivityViewController(activityItems: [url], applicationActivities: nil)
+            parentViewController.present(activityVC, animated: true, completion: nil)
+        }
     }
 }
 
--- a/deltachat-ios/Controller/FilesViewController.swift
+++ b/deltachat-ios/Controller/FilesViewController.swift
@@ -288,7 +288,6 @@ extension FilesViewController {
     }
 
     public static func share(message: DcMsg, parentViewController: UIViewController, sourceView: UIView) {
-        let activityVC: UIActivityViewController
         guard let fileURL = message.fileURL else { return }
         let objectsToShare: [Any]
         if message.type == DC_MSG_WEBXDC {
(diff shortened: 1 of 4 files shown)


deltachat-ios/Helper/Utils.swift
```diff
@@ -40,4 +41,16 @@ struct Utils {
         let window = UIApplication.shared.keyWindow
         return window?.safeAreaInsets.bottom ?? 0
     }
+
+    public static func getInviteLink(context: DcContext, chatId: Int) -> String? {
+        // convert `OPENPGP4FPR:FPR#a=ADDR&n=NAME&...` to `https://i.delta.chat/#FPR&a=ADDR&n=NAME&...`
+        if var data = context.getSecurejoinQr(chatId: chatId), let range = data.range(of: "#") {
```
zeitschlag, line 47: I was a bit confused by `data`, it's a `String`, right?

deltachat-ios/Controller/FilesViewController.swift
```diff
@@ -302,11 +301,18 @@ extension FilesViewController {
...
         activityVC.popoverPresentationController?.sourceView = sourceView
         parentViewController.present(activityVC, animated: true, completion: nil)
     }
+
+    public static func share(url: String, parentViewController: UIViewController) {
+        if let url = URL(string: url) {
+            let activityVC = UIActivityViewController(activityItems: [url], applicationActivities: nil)
+            parentViewController.present(activityVC, animated: true, completion: nil)
+        }
+    }
```
zeitschlag, line 315: What's the connection between `FilesViewController` and sharing QR-code? So does this sharing-code live in `FilesViewController`, which seems to be responsible to show files?

zeitschlag (approved):
As always: Looks like a pragmatic approach, I added some questions, but LGTM.

deltachat-ios/Helper/Utils.swift
r10s: it is a string containing "data" (and nothing human readable :)

i named it "qrCodeData" or so first, however, this is also not correct as it morphs to an URL, therefore i just used "data"

deltachat-ios/Controller/FilesViewController.swift
r10s: the only connection is that the other "share" function is there as well :)

keeping that together makes sense, but both could also go to Utils or so

PR #2060, opened 2024-01-30
Fix cell-hiding (#2056)

### LINKED ISSUE #2056: Messages disappear after short long-press
- Messages disappear after long-press them shortly
- Probable reason: `cell.messageBackgroundContainer.isHidden = (reactionsHidden == false)`
- But why isn't it triggered with `false` when canceling the long-press?

simulator_screenshot_CFE99236-CEE8-4EE4-B648-EB5D36C84BE0

zeitschlag:
Closes #2056

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -1945,9 +1945,6 @@ extension ChatViewController {
 
         return preview
     }
-
-
-
     // context menu for iOS 13+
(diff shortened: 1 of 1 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1927,8 +1947,8 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
     }
 
 
+
```
r10s, line 1950: with too many empty lines, we get a "whitespace" warning :)

```suggestion

```

r10s (approved):
lgtm, also tested, works for me(tm)

PR #2062, opened 2024-01-30
detailed error message on 'cannot encrypt'

### LINKED ISSUE #2054: improve status message for "Invalid Unencrypted"
- set correct string
- make the message tappable, similar to guaeranteed-e2ee-broken or cannot-add-not-green-checkmark
- show a dialog, again similar to guaeranteed-e2ee-broken or cannot-add-not-green-checkmark

cmp. <https://github.com/deltachat/deltachat-android/pull/2940>, <https://github.com/deltachat/deltachat-desktop/issues/3644>

r10s:
providers may not allow to send unencrypted messages. for this case, this PR adds a more detailed error message including away for the user to fix the situation.

 

closes #2054

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -735,6 +735,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         dcContext.setStockTranslation(id: DC_STR_CHAT_PROTECTION_ENABLED, localizationKey: "chat_protection_enabled_tap_to_learn_more")
         dcContext.setStockTranslation(id: DC_STR_CHAT_PROTECTION_DISABLED, localizationKey: "chat_protection_broken_tap_to_learn_more")
         dcContext.setStockTranslation(id: DC_STR_NEW_GROUP_SEND_FIRST_MESSAGE, localizationKey: "chat_new_group_hint")
+        dcContext.setStockTranslation(id: DC_STR_MESSAGE_ADD_MEMBER, localizationKey: "member_x_added")
+        dcContext.setStockTranslation(id: DC_STR_INVALID_UNENCRYPTED_MAIL, localizationKey: "invalid_unencrypted_tap_to_learn_more")
     }
 
     func appIsInForeground() -> Bool {
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -984,6 +984,8 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
                 showProtectionEnabledDialog()
             case DC_INFO_PROTECTION_DISABLED:
                 showProtectionBrokenDialog()
+            case DC_INFO_INVALID_UNENCRYPTED_MAIL:
+                showInvalidUnencryptedDialog()
             default:
                 break
             }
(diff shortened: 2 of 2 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -735,6 +735,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         dcContext.setStockTranslation(id: DC_STR_CHAT_PROTECTION_ENABLED, localizationKey: "chat_protection_enabled_tap_to_learn_more")
         dcContext.setStockTranslation(id: DC_STR_CHAT_PROTECTION_DISABLED, localizationKey: "chat_protection_broken_tap_to_learn_more")
         dcContext.setStockTranslation(id: DC_STR_NEW_GROUP_SEND_FIRST_MESSAGE, localizationKey: "chat_new_group_hint")
+        dcContext.setStockTranslation(id: DC_STR_MESSAGE_ADD_MEMBER, localizationKey: "member_x_added")
```
r10s, line 738: `DC_STR_MESSAGE_ADD_MEMBER` is unrelated to the error message, but was just missing

Simon-Laux (changes requested):
"scan qr code" opens own qr code tab instead of the scanner.
The other dialog action works fine (open help).

r10s:
already at other places (guarantee
-broken, contact-not-verified) we're using "Scan QR code" as a shortcut for "Join Contact using show and scan QR code or the other way round", so it is intended, you can do both.

technically, it is just a shortcut to the existing  ViewController that is nkt subjext to changes for 1.44

we're not going to "scan" by default as that tends to cause issues in the past in case camera has issues (we had that on android, but well)

also showing a QR code is more clear what this is about than just a camera

maybe the wording or other things can be improved, there are also ideas to move "scan" back to a button. but given the wordings shortness, it is also not bad as it is for now

Simon-Laux:
Your answer sounds like it is hard to implement opening the activity/screen at the "scan"-tab/mode.

I don't follow your arguments that it would be better to go to view the own qr code as:
- on iOS the QR code is a tab and very easy to find (might not be the case on android)
- the scan view has a subtitle telling the user to scan the QR code (so it's not completely unclear what to do)
   - if that's not enough we could add a qr code icon to the text so people get an additional hint.
- the wording invokes the user expectation to "scan" a QR code, not to view it.
- and you also told me to directly go to the scan view on desktop, if I remember correctly.

With a different design (scan button under own qr code) things might be different, but the wording issue still stands.

r10s:
i already wrote, that things can be improved. 

still, it is not a bigggie that should block anything

r10s:
thinking ot over, maybe going for the button "QR Invite Code" and force opening at "Show" is a little better and easy to do. this still fits the dialog explanation.

but we should do that in a subsrquent pr for all dialogs. and desktop/android should adapt (esp on desktop i think showing by default is better, even if i/we thought differently before)

Simon-Laux (approved):
Ok then I approve it, but I still think this unexpected UI behaviour should be fixed before the release.

r10s:
i filed an issue at https://github.com/deltachat/interface/issues/60

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1596,6 +1598,20 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
         navigationController?.present(alert, animated: true, completion: nil)
     }
 
+    private func showInvalidUnencryptedDialog() {
+        let alert = UIAlertController(title: String.localized("invalid_unencrypted_explanation"), message: nil, preferredStyle: .alert)
```
zeitschlag, line 1602: Just out of curiosity:

Is there a reason you put this (long) text into the title (and not the message)?

r10s: for this PR: just did as always :)

iirc, it otherwise rendered with a very small font (probably assuming you use title+message)

PR #2063, opened 2024-01-30
Add 'Scan QR Code' button to 'New Chat'

### LINKED ISSUE #2036: allow QR code scanning from "New Chat" dialog
the "New Chat" dialog is the main way we're offering to create new chats.

with chatmail becoming more into focus, scanning a QR code becomes more important and should be available from "New Chat" as well, cmp. https://github.com/deltachat/deltachat-desktop/issues/3623

r10s:
The "New Chat" dialog is the main way we're offering to create new chats. With chatmail getting more into focus,
scanning a QR code becomes more important
and is now available directly from "New Chat" as well

closes #2036

### DIFF
--- a/deltachat-ios/Controller/NewChatViewController.swift
+++ b/deltachat-ios/Controller/NewChatViewController.swift
@@ -6,10 +6,11 @@ class NewChatViewController: UITableViewController {
     private let dcContext: DcContext
 
     private let sectionNew = 0
-    private let sectionNewRowNewContact = 0
+    private let sectionNewRowScanQRCode = 0
     private let sectionNewRowNewGroup = 1
     private let sectionNewRowBroadcastList = 2
-    private var sectionNewRowCount: Int { return UserDefaults.standard.bool(forKey: "broadcast_lists") ? 3 : 2 }
+    private let sectionNewRowNewContact = 3
+    private var sectionNewRowCount: Int { return UserDefaults.standard.bool(forKey: "broadcast_lists") ? 4 : 3 }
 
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/NewChatViewController.swift
```diff
@@ -6,10 +6,11 @@ class NewChatViewController: UITableViewController {
     private let dcContext: DcContext
 
     private let sectionNew = 0
-    private let sectionNewRowNewContact = 0
+    private let sectionNewScanQRCode = 0
     private let sectionNewRowNewGroup = 1
     private let sectionNewRowBroadcastList = 2
-    private var sectionNewRowCount: Int { return UserDefaults.standard.bool(forKey: "broadcast_lists") ? 3 : 2 }
+    private let sectionNewRowNewContact = 3
+    private var sectionNewRowCount: Int { return UserDefaults.standard.bool(forKey: "broadcast_lists") ? 4 : 3 }
```
r10s, line 13: i've seen this nicer for other ViewControllers, but i do not want to do a refactoring for this small change :)
also, broadcast lists will  be non-experimental soon(tm) anyway, then it does not look _that_ hacky :)

Simon-Laux:
Is it intended that it goes to show own qr code? Should it not go to scan qr code instead?

r10s:
see https://github.com/deltachat/deltachat-ios/pull/2062

Simon-Laux (approved):
same as my review on #2062, besides the unexpected UX this is ok to be merged IMO

PR #2070, opened 2024-02-06
Use menu to send reactions

zeitschlag:
This reverts the UI of #2048 and replaces it with a menu-based approach for now. It looks and [behaves](https://github.com/deltachat/deltachat-ios/assets/2580019/869cc09e-d243-4bba-8d3b-b3b53c253aff) like this:

Simulator Screenshot - iPad Pro (11-inch) (4th generation) - 2024-02-06 at 17 53 44

simulator_screenshot_191F48E8-8DE3-4D92-94D6-000477662B25

But it comes at a price: The `Select more`-menu entry isn't a seperate menupoint (aka submenu) anymore. Please note that on iOS 12, you won't see a checkmark next to your selected reaction.

-------

At some point we should refactor `ContextMenuProvider` to deal with `// we only support 1 submenu layer for now`


r10s:
can we have the SF Symbol `face.smiling` as icon in the main menu?

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -158,22 +158,55 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
...
+                    dcContext.sendReaction(messageId: messageId, reaction: nil)
+                } else {
+                    dcContext.sendReaction(messageId: messageId, reaction: reaction.emoji)
+                }
+            }
+        }
+
+        let reactionsItem = ContextMenuProvider.ContextMenuItem(
+            title: String.localized("react"),
+            imageName: "heart.circle.fill",
```
r10s, line 189: this is not visible to me, i'd also prefer `face.smiling` as this better fits to the desktop UI using a similar icon for "React"

r10s (approved):
apart from the missing icon, code lgtm as well

deltachat-ios/Chat/ChatViewController.swift
zeitschlag: Oh, yes, you're right, good point!
Turns out: If you don't use an image in a menu element, there's no image visible.

Thank you!

zeitschlag:
For reference:

simulator_screenshot_E9A2CD0B-08CF-476E-94ED-9E6E7DCA8642

r10s:
ftr, the issue of #2048 was that we do not have control about size and position of the reaction bar. eg. it may be covered by the menu and gets unusable. or it got unusable small on longer messages

there are ideas of doing things differently, some other messengers eg. use a complete self-drawn blur-preview-menu, coming at high costs wrt maintainaince and easily look non-native and "wrong".
other messengers use preferredElementSize, having other issues (show selected, only 4 items)

therefore, and to allow is to focus on the next thing, we decided for this simple "plan b" to have sth stable in an apple way. it is one tap more - but at an expected position and it looks and works good - and we can think over in the next iteration

PR #2071, opened 2024-02-08
use core background fetch, use .default qos

r10s:
this PR aims to improve the existing background fetch further by:

- use the new `[dc_accounts_background_fetch()`](https://github.com/deltachat/deltachat-core-rust/pull/5104) api which was explicitly made for this purpose 

- changing qos ("quality of service") of background task to `.default` instead of `.background` - while the latter sounds as if it is correct, it was probably never the intention as  `.background` may be delayed by a loooong time [according to stackoverflow](https://stackoverflow.com/questions/53141829/performance-of-background-queue-in-swift) - but also by our [own observations](https://github.com/deltachat/deltachat-ios/blob/main/deltachat-ios/AppDelegate.swift#L487) (in theory, there is also `.utility`, however, finer granulation is probably not needed, i assume anyways that there is some additional per-app-weight)

the "one second sleep" is untouched for now, in fact, you'll see some core activity between "finishing fetch" and "fetch done", at least in the simulator, maybe in "real background" where re-start IO is not needed, it is really superfluous. also, in theory, swift could have started short-living threads from events need to finish. so, let's do not have too many moving targets and leave that as is for now

### DIFF
--- a/DcCore/DcCore/DC/events.swift
+++ b/DcCore/DcCore/DC/events.swift
@@ -156,15 +156,17 @@ public class DcEventHandler {
             }
 
         case DC_EVENT_CONNECTIVITY_CHANGED:
-            if let sem = dcAccounts.fetchSemaphore, dcAccounts.isAllWorkDone() {
-                sem.signal()
-            }
             if accountId != dcAccounts.getSelected().id {
                 return
             }
             logger.info("📡[\(accountId)] connectivity changed")
             DispatchQueue.main.async {
                 NotificationCenter.default.post(name: eventConnectivityChanged, object: nil)
+            }
+
+        case DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE:
+            if let sem = dcAccounts.fetchSemaphore {
+                sem.signal()
             }
 
         case DC_EVENT_WEBXDC_STATUS_UPDATE:
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -501,9 +501,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             }
             let diff = CFAbsoluteTimeGetCurrent() - start
 
-            // wait for DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE,
-            // marking the end of events needed to being processed.
-            // as IO was not started, random events caused by other reasons are not added.
+            // wait for DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE;
+            // without IO being started, more events that could interfere with shutdown are not added
             _ = self.dcAccounts.fetchSemaphore?.wait(timeout: .now() + 20)
             self.dcAccounts.fetchSemaphore = nil
             let diff2 = CFAbsoluteTimeGetCurrent() - start
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -488,30 +488,28 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         DispatchQueue.global().async { [weak self] in
             guard let self else { completionHandler?(.failed); return }
 
-            // we're in background, run IO for a little time
-            self.dcAccounts.startIo()
-            self.dcAccounts.maybeNetwork()
             self.pushToDebugArray("2")
 
             self.addDebugFetchTimestamp()
-
-            // create a new semaphore to make sure the received DC_CONNECTIVITY_CONNECTED really belongs to maybeNetwork() from above
-            // (maybeNetwork() sets connectivity to DC_CONNECTIVITY_CONNECTING, when fetch is done, we're back at DC_CONNECTIVITY_CONNECTED)
             self.dcAccounts.fetchSemaphore = DispatchSemaphore(value: 0)
+            let start = CFAbsoluteTimeGetCurrent()
+
+            // backgroundFetch() pauses IO as needed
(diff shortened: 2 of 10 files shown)


deltachat-ios/AppDelegate.swift
```diff
@@ -484,35 +484,33 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
+            // backgroundFetch() pauses IO as needed
+            if !self.dcAccounts.backgroundFetch(timeout: 20) {
+                logger.error("backgroundFetch failed")
+                self.pushToDebugArray("ERR2")
+            }
+            let diff = CFAbsoluteTimeGetCurrent() - start
+
+            // wait for DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE,
+            // marking the end of events needed to being processed.
+            // as IO was not started, random events caused by other reasons are not added.
```
Simon-Laux, line 506: maybe also mention that this event is also emitted in case of timeout, so it is clearer that we can not just do an early exit when the fetch timed-out (because the event is still emitted and we still might have a partial fetch that has some of the new message events)

Simon-Laux (approved):
Looks great.

Can't say anything about the `qos` stuff, because it's outside of my area of expertise, so maybe wait for zeitschlag's opinion or we learn by trying it out.

DcCore/DcCore/DC/events.swift
```diff
@@ -167,6 +164,11 @@ public class DcEventHandler {
                 NotificationCenter.default.post(name: eventConnectivityChanged, object: nil)
             }
 
+        case DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE:
```
zeitschlag, line 167: According to [the docs](https://c.delta.chat/classdc__accounts__t.html#a23e7f5095f52a80a605e0266f686f3b7), the [`DC_EVENT_CONNECTIVITY_CHANGED`](https://c.delta.chat/group__DC__EVENT.html#ga3ced1bbde752feba46ab089611287d71) gets emitted when work is done (but also when connectivity changes, so this sounds like a little ... workaround).

From what I understand now, we don't use the `dc_accounts_all_work_done` anymore, but replaced it with all new [`dc_accounts_background_fetch`](https://c.delta.chat/classdc__accounts__t.html#a4b735a1e51222a06aa992773e42d06b0). Would it make sense to remove `DcAccounts.isAllWorkDone()`?

Also, if I understand this correctly, as a consequence of this change, we don't wait anymore for things to update when the connectivity changes. Instead, we wait for the work to be finished as part of the iOS-background fetch, right?

deltachat-ios/AppDelegate.swift
```diff
@@ -484,35 +484,33 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         pushToDebugArray("1")
 
         // move work to non-main thread to not block UI (otherwise, in case we get suspended, the app is blocked totally)
-        // (we are using `qos: default` as `qos: .background` or `main.asyncAfter` may be delayed by tens of minutes)
+        // (we are using `qos: default` as `qos: .background` may be delayed by tens of minutes)
```
zeitschlag, line 487: From what I know, `.background`-work only gets done when the app is in background and stuff gets done with a very low priority. So it's not a background thread (as in "not main") while the app is running in foreground. So it's not only a question of time.

When we switch to iOS 13+, we could start using `Task` and `async/await` for background-stuff.

deltachat-ios/AppDelegate.swift
```diff
@@ -484,35 +484,33 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
-
-            // create a new semaphore to make sure the received DC_CONNECTIVITY_CONNECTED really belongs to maybeNetwork() from above
-            // (maybeNetwork() sets connectivity to DC_CONNECTIVITY_CONNECTING, when fetch is done, we're back at DC_CONNECTIVITY_CONNECTED)
             self.dcAccounts.fetchSemaphore = DispatchSemaphore(value: 0)
+            let start = CFAbsoluteTimeGetCurrent()
+
+            // backgroundFetch() pauses IO as needed
+            if !self.dcAccounts.backgroundFetch(timeout: 20) {
+                logger.error("backgroundFetch failed")
+                self.pushToDebugArray("ERR2")
```
zeitschlag, line 500: What does `ERR2` stand for?

zeitschlag (approved):
looks reasonable to me. I added a few remarks and questions here and there, but nothing blocking 👍

deltachat-ios/AppDelegate.swift
r10s: it js just a random short  string added to "Settings/Advaned/View Log". it makes sense only if you look at the code. once things are settled, we can remove this DebugArray completely

deltachat-ios/AppDelegate.swift
r10s: i made the comment more precise, putting it more in code context and not just repeating documentation

main purpose of the comment is anyway to know that you need to grep for DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE :)

r10s:
thanks for review and thoughts! i am also using the PR already since a day on my phone, and it seems to work in general :)

deltachat-ios/AppDelegate.swift
Simon-Laux: I would also like async/await because it would make working with jsonrpc nicer.
_Also iOS 13 brings swiftUI but that seems to be not mature enough to be used from what I heard._
But on the other hand we are a messenger and people appreciate that we support older devices as well at least on android, for iPhone users it might be more common to upgrade to a new phone regularly - I suggest we look at the usage percentage take that into account.

PR #2080, opened 2024-02-19
factor out notification token handling

r10s:
this PR moves the handling of the notification token to DcAccounts and DcContext objects using the new core API that supports **real push notifications** (correct core PR needs to be  checked out at deltachat-ios/libraries/deltachat-core-rust  manually to make this PR work)

Overview of related PRs:
- notifiers PR: https://github.com/deltachat/notifiers/pull/17 (adding endpoint to receive notifications from chatmail server)
- chatmail PR: https://github.com/deltachat/chatmail/pull/201 (adding metadata server to register device tokens for notifications and sending them to notifiers)
- core PR: https://github.com/deltachat/deltachat-core-rust/pull/5286 (adding APIs to set tokens and register them with chatmail servers or fallback heartbeat notifications)
- this iOS PR: https://github.com/deltachat/deltachat-ios/pull/2080 (using new core APIs)

EDIT:  apart from the expected issues wrt "mute chat" and "mute all" the PRs play well together and things are working nicely (idea for "mute all" is to drop that switch and just open system notifications setting where you can "mute all" as well. or, if per-account-mute is needed, a revoke api. for "mute chat", maybe this is doable by a notification service exetension as well; i started a branch with that. that could also be used for "mute all". we'll see :) both, however should not block merging this pr)

~~there are two new API that will be replaced by the following core implementation soon:~~
~~- `dc_accounts_set_notify_token(token)` - passes the token received from apple to the core. core will forward it to supported (chatmail) server or, as a fallback, use it for heartbeat. as we have only one token for all accounts, core will do that for all accounts.  the method may be called on every app start as the tokens may be short-living. the function may take a while (we should decide if UI needs to call it in a thread or core does that) there is no need for a return value, UI will check the state using the following method.~~
~~- `dc_context_get_notify_state(context)` - returns NOT_CONNECTED (0), HEARTBEAT (1) or PUSH (2) - this is a per-context method as some accounts may use chatmail while others not. the method is used by the UI mainly to show the state to the user~~
~~the PR adapts flow and connectivity view already to these new calls, however, still uses a "dummy" UI implementation. i also did this PR to make myself clear what is actually and really needed for UI :) (@hpk42 and me are first thinking that we can just use a set_config(), however, this does not work nicely as the token is same for all accounts)~~
~~@link2xt @zeitschlag if this roughly makes sense  to you, i suggest to merge this PR already in, we can then focus on the core~~ EDIT: we decided to go for changing core directly

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -376,7 +376,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         let notifyToken = tokenParts.joined()
         logger.info("Notifications: Token: \(notifyToken)")
         self.notifyToken = notifyToken
-        dcAccounts.setNotifyToken(token: notifyToken)
+        dcAccounts.setPushToken(token: notifyToken)
     }
 
     // `didFailToRegisterForRemoteNotificationsWithError` is called by iOS
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -373,33 +373,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     // we pass the received token to the app's notification server then.
     func application(_ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
         let tokenParts = deviceToken.map { data in String(format: "%02.2hhx", data) }
-        let tokenString = tokenParts.joined()
-
-        let endpoint = "https://notifications.delta.chat/register"
-
-        logger.info("Notifications: POST token: \(tokenString) to \(endpoint)")
-
-        if let url = URL(string: endpoint) {
-            var request = URLRequest(url: url)
-            request.httpMethod = "POST"
-            let body = "{ \"token\": \"\(tokenString)\" }"
-            request.httpBody = body.data(using: String.Encoding.utf8)
-            let task = URLSession.shared.dataTask(with: request) { (data, response, error) in
-                if let error = error {
-                    logger.error("Notifications: cannot POST to notification server: \(error)")
-                    return
-                }
-                if let httpStatus = response as? HTTPURLResponse, httpStatus.statusCode == 200 {
-                    logger.info("Notifications: request to notification server succeeded")
-                } else {
-                    logger.error("Notifications: request to notification server failed: \(String(describing: response)), \(String(describing: data))")
-                }
-                self.notifyToken = tokenString
-            }
-            task.resume()
-        } else {
-            logger.error("Notifications: cannot create URL for token: \(tokenString)")
-        }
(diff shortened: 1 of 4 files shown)


zeitschlag:
I'd suggest to do nothing for now and wait until Core implemented this. Then we can remove lots of code and just call the core.

r10s:
> I'd suggest to do nothing for now and wait until Core implemented this. Then we can remove lots of code and just call the core.

makes sense as well, esp. as the core part is already in the making :)

r10s:
> What do we need dc_context_get_notify_state for? It makes sense to know whether the server supports push (i.e. whether it is a recent chatmail or not), but do we really need to know for an account (single context) whether we have registered for heartbeat or not?

indeed my idea was to know for an account (single context) to know if heartbeat was registered or not and show that to the user (as "Not Connected"  - as it may also mean the token is pending or the device is really not connected or just about to connect)

i think, this is useful for debugging.

> I am still not sure how heartbeat registration should happen. On the one hand it is not per-context because if user has 2 accounts we don't need to register twice with the same device token.

yes. i'd go to the account list and register for "real PUSH" where possible. only of there are accounts left, register one time for heartbeat. 

so, we have the following pseudeo code:

```
pub struct Accounts {
   registered_for_heartbeat: bool;
   ....
}

pub struct Context {
    registered_for_real_push: bool;
    ...
}

fn get_notify_state(context: &Context) {
   if context.registered_for_real_push {
       return DC_NOTIFY_REAL_PUSH;
   }
   return accounts.registered_for_heartbeat ? DC_NOTIFY_HEARTBEAT : DC_NOTIFY_NOT_CONNECTED;
}
```

> On the other hand if user has a single account configured with SOCKS5 to use Tor, we definitely don't want to register by connecting directly to notification server bypassing Tor. How should we register for heartbeats if we have two accounts, one configured with SOCKS5 and the other not?

i think, SOCKS5 and tor is nothing we should worry about currently. there are not even the corresponding options and on apple devices, not sure, how useful in general this is.

but to the question: if somehow any account is using SOCKS5 or Tor, i'd skip the whole registeration for all contexts

Simon-Laux:
I think you can register the heartbeat even when using tor for single accounts, tor on iOS likely means a vpn for everything anyways?
also heatbeat is not that sensitive of an information anyway, you would just learn that a user uses deltachat when monitoring the non-tor traffic.

r10s:
i adapted the PR so that it uses https://github.com/deltachat/deltachat-core-rust/pull/5286 directly now

Simon-Laux (approved):
works nicely

r10s:
k, merging that in. 

"in-app muting chats" is not working for c20/nine.testrun.org when notifications are enabled in the system, however, the advantage of having reliable push notifications makes that a minor - and a fix for muting & co is already on its way https://github.com/deltachat/deltachat-ios/tree/r10s/notification-service-extension (but needs some more time so that it should not block ongoing release efforts)

PR #2082, opened 2024-02-22
[WIP] 1737 fix scrolling

zeitschlag:


r10s (approved):
thanks for diving into the issue!

indeed, it is better than before - with less code lines!

still, the initial scrolling is a bit weird, but sth. for the next iteration, there is probably more to cleanup that at some point we have a clear sight on the issue :)

r10s:
interestingly, the scrolling is more  than only the height of the message input field, as i would have expected.

PR #2085, opened 2024-02-23
Fix autoLayout for VideoInviteCell (#2083)

### LINKED ISSUE #2083: degration: video chat invitations rendered weirdly
since 1.43 or so, "video chat invitations" are rendered as if they have a height of zero, also the text is missing.

iirc, it was all okay before.

in the image below, also note that something, maybe empty title, is rendered behind the date info (see the dot at the "e" of "February")

IMG_5218

zeitschlag:
Fixes #2083 

Simulator Screenshot - iPhone 15 Pro - 2024-02-23 at 16 34 28


r10s (approved):
works like a charm! thanks for fixing!

PR #2087, opened 2024-02-25
allow sending any emoji as reaction

r10s:
i had a quick look how we can improve reactions without much effort, a big impact will have the option to send any reaction.

i used the ["first best" emoji picker](https://github.com/topics/emoji-picker?l=swift), and this PR is the result of not much more than an hour of work, with quite some impact, i'd say :)

EDIT: i spent a little more tome to make the selector a better fitting sheet, for that, it was needed to fork the repo at https://github.com/deltachat/MCEmojiPicker (which is still added via Swift Packet Manager)

  Screenshot 2024-02-27 at 23 51 17

(OT, reminder to myself, sth. for another PR: if we have sth. like the emoji picker in, i think, we can also reconsider going for the max-4-horizontal items, "👍 👎 ❤️ •••")


zeitschlag:
@r10s That's what I suggested yesterday. I wouldn't copy it, but fork their repository, do the changes there and use SPM/pods to include the forked repo. What do you think?

r10s:
@zeitschlag thanks for all your tips, i forked the repo and added it via Swift Packet Manager, still getting the emulator-cannot-build error. 

i tried adding the same via Pod - just the same error:

Screenshot 2024-02-28 at 00 21 04

the magic behind that is totally unclear to me ... i mean, the forked repo should not have some artefact, things are just compiled - why does the emulator cause issues? also, why does the emulator asks for rosetta - on an M1 arm64 (which is compiling fine for real device) should be just fine

zeitschlag (approved):
LGTM, works better than expected, I have to admit!

Issue #2084, opened 2024-02-23
VideoInviteCells don't show reactions


zeitschlag:
Either show reactions or disable reactions for video invitations

Simon-Laux:
we hide the option to react on those messages in desktop: 
- [code](https://github.com/deltachat/deltachat-desktop/blob/b466ac146f247936dac97342201968136f7ee8b9/src/renderer/components/ReactionsBar/useReactionsBar.ts#L50)
- https://github.com/deltachat/deltachat-desktop/pull/3642

PR #2094, opened 2024-02-28
show reactions on 'Video', hide reactions on 'Video Chat Invitations', cleanup types

### LINKED ISSUE #2089: why is videochat-invitation view type nil?
cmp. https://github.com/deltachat/deltachat-ios/pull/2086/files

r10s:
this PR adds the missing viewtype for 'Video Chat Invitation' and adapts the condition to hide/show reactions accordingly.

moreover, it fixes the issue that reactions for 'Videos' are not shown

EDIT: also, this PR cleans up erroneous viewtype getters - we had two (!), in most cases, `msg.type` was used, which returns correct values. in rare cases, `msg.viewtype` was used that had errors. the latter is replaced by `msg.type` by this PR so we have only one thing - and are also closer to the android wrt naming and grep'ability
 
**for review:** "viewtype" is synonym to "message type", in core, in android and mostly on ios, we prefer the shorter form, however, but to keep this PR short and draw a line, i did not refactor all variable names and so on 

successor of #2084

closes #2089

### DIFF
--- a/deltachat-ios/Chat/DraftModel.swift
+++ b/deltachat-ios/Chat/DraftModel.swift
@@ -22,7 +22,7 @@ public class DraftModel {
     }
     var viewType: Int32? {
         if let viewType = draftMsg?.type {
-            return Int32(viewType)
+            return viewType
         }
         return nil
     }
--- a/deltachat-ios/Chat/DraftModel.swift
+++ b/deltachat-ios/Chat/DraftModel.swift
@@ -21,10 +21,7 @@ public class DraftModel {
         return draftMsg?.filemime
     }
     var viewType: Int32? {
-        if let viewType = draftMsg?.type {
-            return viewType
-        }
-        return nil
+        return draftMsg?.type
     }
 
     public init(dcContext: DcContext, chatId: Int) {
--- a/deltachat-ios/Controller/ContextMenuController.swift
+++ b/deltachat-ios/Controller/ContextMenuController.swift
@@ -23,14 +23,14 @@ class ContextMenuController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        let viewType = msg.viewtype
+        let viewType = msg.type
         var thumbnailView: UIView?
         switch viewType {
-        case .image:
+        case DC_MSG_IMAGE, DC_MSG_STICKER:
             thumbnailView = makeImageView()
-        case .video:
+        case DC_MSG_VIDEO:
             thumbnailView = makeVideoView()
-        case .gif:
+        case DC_MSG_GIF:
             thumbnailView = makeGifView()
         default:
             return
--- a/deltachat-ios/DC/DcMsg.swift
+++ b/deltachat-ios/DC/DcMsg.swift
@@ -5,33 +5,6 @@ import UIKit
 /// See [dc_msg_t Class Reference](https://c.delta.chat/classdc__msg__t.html)
 public class DcMsg {
 
-    public enum MessageViewType: CustomStringConvertible {
-        case audio
-        case file
-        case gif
-        case image
-        case text
-        case video
-        case voice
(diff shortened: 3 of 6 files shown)


r10s:
> i see the intent, however,
duplicating the view type in swift instead of using it directly as elsewhere seems a bit error prone, btw

we even have the "direct" type (i was also wondering why it otherwise works, also as "viewtype" also has a bug wrt stickers), i do a subsequent commit to fix this confusion

deltachat-ios/DC/DcMsg.swift
```diff
@@ -141,33 +116,6 @@ public class DcMsg {
...
-        case DC_MSG_FILE:
-            return .file
-        case DC_MSG_GIF:
-            return .gif
-        case DC_MSG_TEXT:
-            return .text
-        case DC_MSG_IMAGE:
-            return .image
-        case DC_MSG_STICKER:
-            return .image
```
r10s, line 159: here was another hidden bug ... this only was okay, as at most placed, `type` was already used instead of `viewtype` :)

deltachat-ios/Chat/DraftModel.swift
```diff
@@ -22,7 +22,7 @@ public class DraftModel {
     }
     var viewType: Int32? {
         if let viewType = draftMsg?.type {
```
zeitschlag, line 24: Why not just `return draftMsg?.type`? And then: Why not drop `viewType`-property in general? Reading this PRs description it seems to me, that that was the intention?

deltachat-ios/Controller/ContextMenuController.swift
```diff
@@ -23,14 +23,14 @@ class ContextMenuController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        let viewType = msg.viewtype
+        let viewType = msg.type
         var thumbnailView: UIView?
         switch viewType {
-        case .image:
+        case DC_MSG_IMAGE, DC_MSG_STICKER:
```
zeitschlag, line 29: Using a `switch` on `Int` feels a bit un-swifty when there're Enums :D

deltachat-ios/DC/DcMsg.swift
```diff
@@ -283,8 +231,8 @@ public class DcMsg {
     }
 
     // DC_MSG_*
-    public var type: Int {
-        return Int(dc_msg_get_viewtype(messagePointer))
+    public var type: Int32 {
```
zeitschlag, line 234: What's the reason for `Int32` here? Why not just `Int` as before?

zeitschlag (approved):
Nice, I added some questions here and there, but no real blockers

deltachat-ios/Controller/ContextMenuController.swift
r10s: the defines vome direcly from the bindings, no effort. maintaining an additional swift enum (or generating them somehow) seems not worth the effort. worse, it led to the bugs we had as it was not maintained correctly.

we're now just doing the same for the type as for all other defines

deltachat-ios/DC/DcMsg.swift
r10s: before, type was casted on higher levels (see simplifications in this pr)

the reason "Int" is not working is that swift uses "Int32" for ffi-bindings

deltachat-ios/Chat/DraftModel.swift
r10s: removing viewType from DcMsg was the intention, i stopped when it comes to streaming other code

but yes, i also thought about these lines, will have a second look

r10s: applied your suggestion, thanks!

r10s:
nb: the WIP CI-check thing is still in somehow ...

Issue #2091, opened 2024-02-27
contact request bar not shown


r10s:
on current master, the "contact request bar" is at least sometimes not shown.

eg. when  a chat was deleted but nit accepted, it shows up again, but the  without the bar and without any accept/block/delete button.

so, you do not have a chance to accept a contact request.

maybe the issue is even broader, did not test in detail yet EDIT: yip, it is much broader, the bar seems to appear never, so also not for "sent a message from another device"

EDIT: the issue has not been there in released testflight 1.43.1

r10s:
to be clear, this is what's missing:

<img width=320 src=https://github.com/deltachat/deltachat-ios/assets/9800740/079eaab8-8a2b-45e2-a8a9-532cc70e29c3>

and, seems the issue was introduced by a collateral damage of the fix at #2082 :)

PR #2097, opened 2024-02-29
[WIP] 2088 select more inputbar

zeitschlag:


r10s:
yeah, the bugfix is fixed! we have the 🐛🦟🪲 again :)

(but let's see after a rebase again)

r10s (commented):
yeah, #2091 still seems to be fixed ✊

while playing around, one thing i noticed is that in "Device Messages" (and probably in all read-only chats), the "Reply" option is no longer hidden (compared to current `main` where the option is hidden as needed) (devices messages as well as other read-only one-to-ones must not have "reply" nor "relpy privately". read-only groups will still have "reply-privately")

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1919,6 +1923,16 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
         contextMenu(for: indexPath).performAction(action: action, indexPath: indexPath)
     }
 
+    override func tableView(_ tableView: UITableView, canEditRowAt indexPath: IndexPath) -> Bool {
```
r10s, line 1926: very nice that this  cannot be selected any longer! 👌

r10s (approved):
tested quite a bit, and code lgtm, so i think, we can merge that in :)

PR #2100, opened 2024-03-01
info-cells can be forwarded, deleted, selected

r10s:
this also fixes that webxdc-info-messages cannot be tapped to go to the webxdc.

i've overseen that at #2097


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1004,11 +1004,6 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
     }
 
     override func tableView(_ tableView: UITableView, willSelectRowAt indexPath: IndexPath) -> IndexPath? {
-
-        let messageId = messageIds[indexPath.row]
-        let message = dcContext.getMessage(id: messageId)
-        guard message.isInfo == false else { return nil }
```
r10s, line 1010: the check for `id != DC_MSG_ID_DAYMARKER && id != DC_MSG_ID_MARKER1` is done implicitly in the condition below

PR #2105, opened 2024-03-06
add Notification Service Extension

r10s:
this PR adds a ["Notification Service Extension"](https://developer.apple.com/documentation/usernotifications/unnotificationserviceextension) that is called for every push notification received, allowing to modify the notification content, which is (localized) "You have new messages" otherwise.

as we do not know anything about the message the push notification belongs to (nor does apple or others then), we do a background-fetch over all accounts and change the notification accordingly, concretely setting title to chatname and title to message text beginning.

this also filters out reactions or other messages that do not come with DC_EVENT_INCOMING_MSG

if we get more than one message via IMAP (because of races or additional accounts), we use texts as "N messages" or "N messages in M chats".

because of that and because of muting,  it may happen, that we do not want to display a notification at all. by default this is not possible (we can only downtune via sound, interruptionLevel, relevanceScore etc.), but there is a [special entitlement that makes not displaying possible](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_developer_usernotifications_filtering)

nb: the "Notification Service Extension" is an own process (similar to DcShare), that is started independently of the main deltachat-ios app (which may or may not run at the same time). group-id makes it possible that the same data are accessible, however.

bigger todo parts:

- [x] make sure, NSE-fetch and main-app-io do not run at the same time (otherwise last-seen etc. gets issues, also double messages and double notifications)  - EDIT: done with the flags mainAppRunning/nseFetching (cherry picked from [#2107](https://github.com/deltachat/deltachat-ios/pull/2107)/[#2110](https://github.com/deltachat/deltachat-ios/pull/2110)) - we did not find a  real-time IPC check better than saving flags in shared memory
     -  add some resilience when app is terminated unexpectedly and we cannot reset the flag  -> this can go to a subsequent PR, current state is quite stable already - and as long we do not have the do-not-show-entitlement, sth. will be visible in any case
- [x] set real message content
- [x] handle mute
- [x] set badge counter accordingly
- [x] if a notification spans more than one chat, add all of them to the title (iOS will truncate and the summary be sth as "N messages in M chats")
- [x] switch accounts as needed when tapping notifications
- [x] for notifications referring exactly to one chat,  open the chat instead of the chatlist (requires using eg. userInfo instead of notificationIdentifier)
- [x] remove PUSH notifications when the chat is opened; for unspecific notifications ("N messages"), remove them when the chatlist is opened 
- [x] make localisation work, probably just some string files/references missing in the extension EDIT: needed to be re-added at 'Build Phases / Copy Bundle Resources' after rebase
- [x] force update of chatlist similar to DcShare (needed in case the app is suspended to ram)
- [x] stop showing images in notifications: the current implementation is slow and would need some more love - but the effort is questionable as showing images unexpectedly is also easily compromising. better just show a text "Image" as on android (in general "📷" instead of "Image" would be cool, btw - EDIT: this is added meanwhile) 
- [x] sync threading and other data  with normal notifications
- [x]  "New Messages - No more relevant" on next background fetches and app starts

finally:

- see what happens with entitlement - however, this is no blocker. even without entitlement, this PR is an improvement as unwanted messages do no longer sound nor turn on display. and, of course, wanted notifications get the correct text

before (showing generic "You have new messages" in german) / after (showing real message content):

 

successor of https://github.com/deltachat/deltachat-ios/pull/2080


Simon-Laux:
future question can we also send/trigger local notifications from this function/extension?

Simon-Laux:
> future question: can we also send/trigger local notifications from this function/extension?

if so we could send a notification for each message that was received, but maybe that would spam too much. so the current way of telling about "you have multiple new messages" is completely fine, maybe even better 👀

r10s:
i did a small test at some point, one can get a pointer to UNUserNotificationCenter from NSE, however, basic functionality as removing notifications did not work for me (otherwise one would not need an entitlement :) 

therefore, i did not investigate further, even if some hacks may work on some OS, this seems not to be the gist of the NSE, and there may be dragons now or in the future :)

deltachat-ios/AppDelegate.swift
```diff
@@ -55,6 +55,17 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
         logger.info("➡️ didFinishLaunchingWithOptions")
 
+        // The NSE ("Notification Service Extension") must not run at the same time as the app.
+        // The other way round, the NSE is not started with the app running.
+        UserDefaults.setMainAppRunning()
+        var pollSeconds = 0
+        while UserDefaults.nseFetching && pollSeconds < 30 {
+            logger.info("➡️ wait for NSE to terminate")
```
zeitschlag, line 63: This could flood the log if it takes some seconds for the app to start. Is this intended?

DcNotificationService/NotificationService.swift
```diff
@@ -0,0 +1,98 @@
...
+            if event.id == DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE { break }
+            if event.id == DC_EVENT_INCOMING_MSG {
+                let dcContext = dcAccounts.get(id: event.accountId)
+                let chat = dcContext.getChat(chatId: event.data1Int)
+                if !UserDefaults.standard.bool(forKey: "notifications_disabled") && !chat.isMuted {
+                    let msg = dcContext.getMessage(id: event.data2Int)
+                    let sender = msg.getSenderName(dcContext.getContact(id: msg.fromContactId))
+                    bestAttemptContent.title = chat.isGroup ? chat.name : sender
+                    bestAttemptContent.body = (chat.isGroup ? "\(sender): " : "") + (msg.summary(chars: 80) ?? "")
+                    bestAttemptContent.userInfo["account_id"] = dcContext.id
```
zeitschlag, line 42: We could think about moving these strings into an enum of constants as we use these in several places

deltachat-ios/AppDelegate.swift
```diff
@@ -516,15 +537,26 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
-        if !response.notification.request.identifier.containsExact(subSequence: Constants.notificationIdentifier).isEmpty {
-            logger.info("Notifications: notification tapped")
-            let userInfo = response.notification.request.content.userInfo
-             if let chatId = userInfo["chat_id"] as? Int,
-                 let msgId = userInfo["message_id"] as? Int {
-                 if !appCoordinator.isShowingChat(chatId: chatId) {
-                     appCoordinator.showChat(chatId: chatId, msgId: msgId, animated: false, clearViewControllerStack: true)
-                 }
-             }
+        logger.info("Notifications: notification tapped")
```
zeitschlag, line 540: What is this log-statement good for?

zeitschlag (approved):
LGTM. I added some questions here and there, but consider them feedback and feel free to just mark them as completed :)

DcCore/DcCore/Extensions/UserDefaults+Extensions.swift
```diff
@@ -3,7 +3,26 @@ public extension UserDefaults {
     static var hasExtensionAttemptedToSend = "hasExtensionAttemptedToSend"
     static var hasSavedKeyToKeychain = "hasSavedKeyToKeychain"
     static var upgradedKeychainEntry = "upgradedKeychainEntry_"
+    static var mainAppRunningKey = "mainAppRunning"
+    static var nseFetchingKey = "nseFetching"
+
     static var shared: UserDefaults? {
```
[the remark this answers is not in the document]
r10s: idk, maybe it is about that "shared" is not always needed and the computation was considered slow. otoh, it is recalculated on each access.

tbh, i assume there was not much thinking about that, otherwise i would have expected a comment. we can go for what is considered "best practise" at some point

DcNotificationService/NotificationService.swift
```diff
@@ -0,0 +1,98 @@
...
+            guard let event = eventEmitter.getNextEvent() else { break }
+            if event.id == DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE { break }
+            if event.id == DC_EVENT_INCOMING_MSG {
+                let dcContext = dcAccounts.get(id: event.accountId)
+                let chat = dcContext.getChat(chatId: event.data1Int)
+                if !UserDefaults.standard.bool(forKey: "notifications_disabled") && !chat.isMuted {
+                    let msg = dcContext.getMessage(id: event.data2Int)
+                    let sender = msg.getSenderName(dcContext.getContact(id: msg.fromContactId))
+                    bestAttemptContent.title = chat.isGroup ? chat.name : sender
+                    bestAttemptContent.body = (chat.isGroup ? "\(sender): " : "") + (msg.summary(chars: 80) ?? "")
```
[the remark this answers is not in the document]
r10s: maybe this comes from C history, where one could read in-line `?` just like so, even nested as  
`a? (b? 1 : 2) : (c? 3 : 4)` :)

but i agree, checking `isGroup` before is much nicer and makes more sense. i do a commit.

wrt rtl languages: yip, this is not optimal, but we are using that pattern at some places, if translators complain, we should adapt, maybe just a by general concatenate. in general, rtl works nicely.

deltachat-ios/AppDelegate.swift
r10s: it think it is fine, and may even give additional hints.

should not happen too often anyways - flooding would be 30 lines max and would require (a) app being stated just the moment a notification coming in and (b) having lots to do in NSE

deltachat-ios/AppDelegate.swift
r10s: unclear, it was added at some point, maybe for debugging :) i remove it, usually we do not have UI buttons etc. in log

DcNotificationService/NotificationService.swift
r10s: yeah, maybe, i would not  be against it. but i also think it is also good as is. from all issues, we rarely have problems with this kind of stuff :)

PR #2107, opened 2024-03-12
let NSE-fetch and MainApp run mutually exclusive

r10s:
this draft tries to target the issue that NSE-fetch and Main-IO must not run at the same time.

sync is done by setting/clearing `mainAppRunning` using shared UserDefaults. however, this is tricky:

- as one have to take care about clearing the `mainAppRunning` flag
- it is impossible to clear the flag ~~when the user terminated the app by "swiping out" - or~~ if main app gets terminated by the system  
EDIT:  "swiping out" is detectable by applicationWillTerminate, see last commit

in both cases, users won't get PUSH notifications until the app is running again.

so, we should see if there isn't an better option to check from NSE if Main-IO is running (the other way round with `nseFetching` flag is maybe also not optimal, but has much less things that could go wrong)

EDIT: tried by observing UserDefaultChanges at https://github.com/deltachat/deltachat-ios/pull/2110 - however, this does not work between processes :/

another hack is https://stackoverflow.com/questions/27011916/how-to-communicate-between-ios-app-containing-extension-and-extension-not-host


r10s:
cherry picked the required commits to r10s/notification-service-extension , discussion can continue there

PR #2108, opened 2024-03-12
Localize NSE

zeitschlag:


r10s (approved):
lgtm, the changed settings for letting extension recognise translations are missing, though

PR #2114, opened 2024-03-19
Improve InputBar-handling (and scrolling)

zeitschlag:
- Should fix the initial scroll to bottom (070ce00)
- Doesn't show input-menu after opening the attachment-menu
- Doesn't show input-bar after visiting the chat details

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -150,9 +150,7 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
         return UIBarButtonItem.init(barButtonSystemItem: UIBarButtonItem.SystemItem.cancel, target: self, action: #selector(onCancelPressed))
     }()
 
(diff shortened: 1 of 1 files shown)


r10s:
yeah, this is so much better at a first glance! 
/me testing and running this PR on my phone, will have a look at the code tomorrow

r10s (commented):
works very nicely! dc feels so much more stable and mature with this PR - great you got a grip on that long-standing issue!

one thing that needs to be fixed  before merging: in read-only-chats the "select more" bottom-bar does not appear (see eg. contact-requests, device messages or read-only-mailinglists)

r10s:
yeah, read-only chats work as expected now.

but i found another issue: for contact requests, the ChatContactRequestBar with "Accept", "Block" etc. never shows up :) same probably when guaranteed-e2ee is broken

r10s (approved):
yeah, lgtm, thanks again!

PR #2115, opened 2024-03-21
allow pasting qr-codes without selecting 'scan'

r10s:
this PR shows 'paste QR code from clipboard' on both of the 'QR code' tabs.

i have seen it several times where ppl want to paste a qr code and were searching for the option;
it seems superfluous - and one tap more - if one is force to select 'scan' before.

@adbenitez maybe also sth. for andoid


zeitschlag (approved):
Neat!

PR #2117, opened 2024-03-22
Show MessageBar after user joined a group (and is allowed to send messages)

### LINKED ISSUE #2116: after scanning a group invite link, message-input-bar is missing
after scanning a group invite link, the handshake runs for a moment.

during this time, the message-input-bar is not shown (as the user cannot sent as eg. the member list and overall state is unknown - scanning the QR code could also be done when being totally offline)

once, handshake is completed (maybe secondds or days later :), the user can type and the message-input-bar should be shown - which is not the case currently, probably after https://github.com/deltachat/deltachat-ios/pull/2114

when leaving and re-entering the chat, message-input-bar is shown

zeitschlag:
Fixes #2116


r10s (approved):
code lgtm, and also my tests passed :)

thanks a lot for fixing!

PR #2126, opened 2024-03-26
use unified logging system, share log

r10s:
this PR switches logging from `print()` to ["unified logging system"](https://developer.apple.com/documentation/os/logging/generating_log_messages_from_your_code) with the following advantages:

- logs are persisted, at least for some time (depending on device speed and memory)
- non-devs can stream logs using Console.app on mac
- logs can be downloaded later using `log collect`
- we can show the log again at "Settings / Advanced" (requires iOS 15)

some further hints:

- instead of using apples method to redact sensitive data from the log, we aim to not add sensitive data at all (what _sensitive is_, is another discussion :)
- apple recommends to delay building final output string for performance reasons. we cannot easily make use of that as most strings come directly from the core. however, compared to what is added _by others_ to the log, the few lines from us are very probably not a performance issue. it is also not different to before, so this PR does not worsen things
- this PR also removes some log spam, eg. old debug entries (debug/printf should be used during an active debugging session only and removed afterwards)
- we keep the hearts for now, as it still makes things easier to recognize, esp. when mixed in Console.app with other things

finally, this PR adds a "share" button so that the log can be easily sent via other channels, copied etc.

 &nbsp; 

targets https://support.delta.chat/t/two-small-tweaks-to-improve-the-log-screen-in-dc-ios/2991

### DIFF
--- a/DcCore/DcCore/DC/events.swift
+++ b/DcCore/DcCore/DC/events.swift
@@ -27,17 +27,17 @@ public class DcEventHandler {
         let accountId = event.accountId
 
         if event.id >= DC_EVENT_ERROR && event.id <= 499 {
-            logger.error("📡[\(accountId)] \(event.data2String)")
+            logger.error("[\(accountId)] \(event.data2String)")
             return
         }
 
         switch event.id {
 
         case DC_EVENT_INFO:
-            logger.info("📡[\(accountId)] \(event.data2String)")
+            logger.info("[\(accountId)] \(event.data2String)")
 
         case DC_EVENT_WARNING:
-            logger.warning("📡[\(accountId)] \(event.data2String)")
+            logger.warning("[\(accountId)] \(event.data2String)")
 
         case DC_EVENT_CONFIGURE_PROGRESS:
             logger.info("📡[\(accountId)] configure: \(Int(data1))")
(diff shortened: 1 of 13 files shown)


Simon-Laux:
Very cool! Thanks for going the extra mile to bring back logging.

> apple recommends to delay building final output string for performance reasons.

I thought it was something about referencing the string in the binary log format instead of copying it, but I haven't looked into it in detail, if my understanding would be correct that would also be about disk space usage.
I think it uses sth. like ring buffers, so in the worst case it would keep less of the log when not connected to a debugger/Console app?

DcCore/DcCore/DC/events.swift
```diff
@@ -27,17 +27,17 @@ public class DcEventHandler {
...
         case DC_EVENT_INFO:
-            logger.info("📡[\(accountId)] \(event.data2String)")
+            logger.info("[\(accountId)] \(event.data2String)")
 
         case DC_EVENT_WARNING:
-            logger.warning("📡[\(accountId)] \(event.data2String)")
+            logger.warning("[\(accountId)] \(event.data2String)")
 
         case DC_EVENT_CONFIGURE_PROGRESS:
             logger.info("📡[\(accountId)] configure: \(Int(data1))")
```
Simon-Laux, line 43:
```suggestion
            logger.info("[\(accountId)] configure: \(Int(data1))")
```
if you want to remove the icon everywhere

Simon-Laux:
How can I debug log sth after this pr in iOS? doesn't it make sense to still have an .debug method on the logger?
as far as I understood the apple docs debug messages are directly discarded if no one is listening (with the console app).

r10s:
dir debugging one can still use print() as before. or or Logger().debug(). or we can add a debug() as needed, but currently it is not

Simon-Laux:
I had an additional idea: we could offer a setting to disable logging completely for activists that want to keep their forensics footprint low (the log contains chat and account ids from that you could infer that additional accounts existed). But we wouldn't encourage this by default as we need the logs to fix issues that users have.

r10s:
surely, there is no lack of ideas :)

Simon-Laux (approved):

PR #2127, opened 2024-03-26
Scroll ChatList to top if already visible (#2125)

### LINKED ISSUE #2125: subsequent taps on "Chats" should scroll up
in the chatlist, subsequent taps on the toolbar's "Chats" button currently does nothing.

in comparable apps, subsequent taps scroll up,  this would be a little, nice improvement for Delta Chat as well.

to be clear: the first tap, should keep scroll position as it is now. there is no need to count taps or so - on a tap on "Chats", just check if "Chats" is already selected (scroll up) or not (switch to tab, use old scroll position as now)

zeitschlag:
closes #2125

### DIFF
--- a/DcCore/DcCore/Extensions/UITableView+Extensions.swift
+++ b/DcCore/DcCore/Extensions/UITableView+Extensions.swift
@@ -1,14 +1,14 @@
 import UIKit
 
 extension UITableView {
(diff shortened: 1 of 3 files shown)


r10s (approved):
really nice improvement!

PR #2135, opened 2024-04-01
Menu Refactoring (#2111)

### LINKED ISSUE #2111: make context menu better manageable by devs
currently, there is a little abstraction layer for the messages' context menu, that does not help much or avoid double code, but makes "nicer" things harder -eg. having a separator, submenus, or [preferredElementSize](https://developer.apple.com/documentation/uikit/uimenu/4013313-preferredelementsize)

let's make the code there much simpler, and split only between "current iOS, having all nice menu options" (that would be ios15) and a fallback that works also on ios12 (using the old, text-edit-like menus without blurring)

for reactions, it is fine to offer just a simple "react" for ios14 and below that opens the "any emoji selector" (currently, we do not have reactions at all for ios12)

@zeitschlag please correct me if i understood sth wrong :)

zeitschlag:
- Makes `UIMenu`s easier to maintain and work with (see #2111)
- Removes `ContextMenuProvider` and puts menu-creation into the specific `UIViewController`
- Introduces `LegacyMenuItem` to make it work with <iOS 13. Adds the reactions there, too.

Looks like this (for now):

Simulator Screenshot - iPhone 15 Pro - 2024-04-02 at 14 03 26

closes #2111


r10s (approved):
thanks a lot for diving into that menu ocean 🤿

at first tests, menus appear where they should, but if not, we can now easily fix things in subsequent PRs, things are now mich simpler and more actionable. back in menu business again :)

one thing i'd like to clarify before merging, however: i could force the "legacy ios13 menu" on iphone13mini/ios17 - but the trick you told me [^1] did not work on iphone7/ios15

[^1]: commenting whole`@available(iOS 13.0, *) extension ChatViewController { ... }` and ios13 check atop of `prepareContextMenu()` - so, wondering, if the legacy menu works at all for ios12 ...

PR #2140, opened 2024-04-04
Add CI

link2xt:
I want to add some CI so there are command-line instructions I can run to test the build of deltachat-ios with core changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,15 @@
+name: CI
+on:
+  push:
+
+jobs:
+  build:
+    runs-on: macos-14
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+          show-progress: false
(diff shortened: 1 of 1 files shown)


zeitschlag:
Oh, nice! A few remarks/thoughts:

- You might want to use the `.xcworkspace`
- Also somethingsomething `git module init` (for the Rust-Core-thing)
- The iOS-version of the simulator doesn't matter as it's the Simulator

r10s:
yip, i also wanted to remark that the compile issues probably come from the swift version, there were changes in that area.

we're using "Swift 5" according to Xcode - it does not tell the minor version, but in general, we have zero warnings, and also do all minor updates

r10s (approved):
wow! this is the first time i'm seeing this ...

Screenshot 2024-04-04 at 21 25 04

... for deltachat-ios 😊 thanks a lot!

is also core somehow trying to build for ios now? so that we can see possible troubles sooner maybe?

Issue #2113, opened 2024-03-16
faster reactions


r10s:
for the messages' context menu, we stay with what the system provided. we do this to avoid rewriting too much code and to avoid issues and future maintenance costs.  even whatsapp is struggling there, eg. long-tap-and-hold is not working - we do not want to open that can of worms, at least not currently.

as a result, currently, the reactions are a  a submenu always requiring an additional tap.

however, when https://github.com/deltachat/deltachat-ios/issues/2111 is closed, we can do better by using [preferredElementSize](https://developer.apple.com/documentation/uikit/uimenu/4013313-preferredelementsize) :

- using `.small` allows up to 4 items, set then to " 👍 | 👎 | ❤️ | ..."
    - that way, for the three emoji, one tap less is needed - same for opening the "any emoji" selector
    - for the emoji 😂 and 🙁 as many taps as before are needed when we've done https://github.com/deltachat/deltachat-ios/issues/2112 
    - moreover, reactions are much more outstanding in general in the menu, esp. if we streamline the menu further, showing less items by default
- the 👍  👎  ❤️ emojis should not use SFSymbols but should be shown as rendered later (so green thumbs, red heart), we can just use PNG-or-so images for that
- to show if one of 👍  👎  ❤️ is selected we may add images with checkmarks, however, this can also be done in a subsequent PR, maybe it is not really needed, esp. as selection is already shown in the bubbles directly, might be, that is just good enough, i'd like to see how it feeks. changing order or even the selected emojis is a bad idea, simplex is doing that and it is just confusing

r10s:
> the 👍 👎 ❤️ emojis should not use SFSymbols but should be shown as rendered later (so green thumbs, red heart), we can just use PNG-or-so images for that

seems not to be needed at first tests: iOS just displays text if no image is set, therefore we can just use the emoji directly

> to show if one of 👍 👎 ❤️ is selected we may add images with checkmarks

also that seems to be easier: we can just add a unicode-checkmark to the text

(tested on iOS 17)

Issue #2112, opened 2024-03-16
init "Frequently Used" reactions with defaults


r10s:
init [MCEmojiPicker's](https://github.com/deltachat/MCEmojiPicker) "Frequently Used" reactions with our default reactions 👍 👎 ❤️ 😂 🙁 - that way, we can go for using [preferredElementSize](https://developer.apple.com/documentation/uikit/uimenu/4013313-preferredelementsize)  with fewer defaults without much drawbacks (for 😂 and 🙁 ppl more often want to chose sth different anyways)

moreover, this allows us an easy fallback for ios14 and below, see https://github.com/deltachat/deltachat-ios/issues/2111

r10s:
> that way, we can go for using [preferredElementSize](https://developer.apple.com/documentation/uikit/uimenu/4013313-preferredelementsize) with fewer defaults without much drawbacks (for 😂 and 🙁 ppl more often want to chose sth different anyways)

ftr, this was already done in https://github.com/deltachat/deltachat-ios/pull/2141

PR #2141, opened 2024-04-04
cleanup and improve chat's context menu

r10s:
as discussed before, this PR:

- adds functions canReply() and canReplyPrivately()
  for consistency and to cleanup code

- if possible, shows the reaction options in the first row;
  as this is limited to 4 options,
  we have now three default reactions plus "..." there.
  more default reactions will go to the reaction picker, see https://github.com/deltachat/deltachat-ios/issues/2112;
  this closes https://github.com/deltachat/deltachat-ios/issues/2113

- "buerocratic, dangerous" options go to "more".
  this makes the menu smaller and avoids scrolling
  as it happens even on larger screens.

- as a side-effect, iOS 12 menu is cleaned up as well

this is how it looks like on iOS 16 (available since iPhone 8):

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -756,12 +756,8 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
 
     override func tableView(_ tableView: UITableView, leadingSwipeActionsConfigurationForRowAt indexPath: IndexPath) -> UISwipeActionsConfiguration? {
         let messageId = messageIds[indexPath.row]
-        if messageId == DC_MSG_ID_MARKER1 || messageId == DC_MSG_ID_DAYMARKER || !dcChat.canSend {
-            return nil
-        }
-
         let message = dcContext.getMessage(id: messageId)
-        if message.isInfo || message.type == DC_MSG_VIDEOCHAT_INVITATION {
+        if !canReply(message: message) {
             return nil
         }
 
@@ -982,6 +978,14 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
 
         self.showEmptyStateView(self.messageIds.isEmpty)
         self.reloadData()
+    }
+
+    private func canReply(message: DcMsg) -> Bool {
+        return message.id != DC_MSG_ID_MARKER1 && message.id != DC_MSG_ID_DAYMARKER && !message.isInfo && message.type != DC_MSG_VIDEOCHAT_INVITATION && dcChat.canSend
+    }
+
+    private func canReplyPrivately(message: DcMsg) -> Bool {
+        return dcChat.isGroup && !message.isFromCurrentSender
     }
 
     private func isLastRowScrolledToBottom() -> Bool {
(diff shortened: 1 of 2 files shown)


dignifiedquire:
screenshot is looking great, looking forward to using this

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -984,6 +980,14 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
         self.reloadData()
     }
 
+    private func canReply(message: DcMsg) -> Bool {
```
zeitschlag, line 983: 👍 This and `canReplyPrivately` is good!

The only small suggestion (as in: feel free to ignore) I have is to rename the first parameter to `to` like this `canReply(to message: DcMsg)`. The idea is to have a `canReply(to: message)`, to me it's more readable :)

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1974,21 +1929,14 @@ extension ChatViewController {
         return preview
     }
 
-    private func reactionsMenu(indexPath: IndexPath) -> UIMenu {
-
+    private func reactionsMenuItems(appendTo: inout [UIMenuElement], indexPath: IndexPath) {
```
zeitschlag, line 1932: My gut feeling isn't very happy with using an `inout` and modifying a list of `UIMenuElement` instead of just returning something. It feels like a sideffect: This method builds the reactions menu _and_ oh by the way modifies a parameter. But as always: No hard feelings here :)

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1974,21 +1929,14 @@ extension ChatViewController {
...
             let sentThisReaction = myReactions.contains(where: { $0 == reaction.emoji })
-            let selectedImage: UIImage?
-            if sentThisReaction {
-                selectedImage = UIImage(systemName: "checkmark")
-            } else {
-                selectedImage = nil
-            }
-
-            return UIAction(title: reaction.emoji, image: selectedImage) { [weak self] _ in
+            let title = sentThisReaction ? (reaction.emoji + "✓") : reaction.emoji
```
zeitschlag, line 1938: Adding a checkmark to mark the sent reaction is smart!

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -2041,57 +1981,43 @@ extension ChatViewController {
...
-                    replyMenuChildren = [
-                        UIAction.menuAction(localizationKey: "notify_reply_button", systemImageName: "arrowshape.turn.up.left.fill", indexPath: indexPath, action: { self.reply(at: $0 ) }),
-                    ]
-                } else {
-                    replyMenuChildren = []
+                if canReply(message: message) {
+                    if #available(iOS 16.0, *) {
+                        // together with preferredElementSize below, this shows the reaction options in a row
+                        reactionsMenuItems(appendTo: &children, indexPath: indexPath)
+                    } else {
```
zeitschlag, line 1991: This is why we refactored this, right? To make it easier to use new APIs. It's good to see it working :)

zeitschlag (approved):
PR LGTM from a code perspective. Some remarks and suggestions here and there, but that's it.
While testing I came across this and it looked bit weird to me:

Simulator Screenshot - iPhone 15 Pro - 2024-04-05 at 09 25 35

This is a screenshot from the device messages and I'm not sure if this is intended? You know better than me :)

r10s:
> This is a screenshot from the device messages and I'm not sure if this is intended? You know better than me :)

no it is not :)  good catch!

if a chat cannot be reacted to, we should no apply `preferredElementSize` at all. i will do a fix

deltachat-ios/Chat/ChatViewController.swift
r10s: sounds good, i will add a commit

deltachat-ios/Chat/ChatViewController.swift
r10s: yes. and sth. as `inout` is the pattern that seems useful for that - otherwise, we would copy items around twice, including additional code. i am not too afraid on that :) but maybe we can change the function name to make that aspect clearer `appendReactionItems(to: ..., indexPath: ...)`

PR #2146, opened 2024-04-04
update core to current master

r10s:
works, also locally in simulator and on real device, thanks a lot @link2xt for getting all these core and ios build issues sorted.

not sure, if we should merge this PR, maybe better do a real tagged core


r10s:
waiting for https://github.com/deltachat/deltachat-core-rust/pull/5432

PR #2150, opened 2024-04-06
duplicate app-icon badge counter to chat's tab-bar-item

### LINKED ISSUE #2130: show number of unread messages on "Chats" icon in tabbar
this makes it easier to notice new messages eg. if the list is scrolled down, for whatever reason. by the improved notifications, this situation has become even more frequent for "N messages in M chats" messages.

also, this is what most other comparable apps are doing.

- there is a system option for that at `UITabBarItem.badgeValue` (thanks for the hint, @zeitschlag), we should just use that. it is shown in "red", not in blue, however, this is fine - on the homescreen, badges are also "red", this somehow fades the color to the app. seems advisable to stay with what the system provides here, and again, this is also what comparable apps are doing.  
EDIT: there is also https://developer.apple.com/documentation/uikit/uitabbaritem/1648567-badgecolor - so seems we have options wrt color :)

- the correct value is at `dcContext.getFreshMessages().count` - EDIT: or maybe repeat the value from the app icon? we should try that out in practise

- for updating, this seems similar to when `getFreshMessageCount()` is called to update the "other accounts unread" in the upper left corner

r10s:
to make it easier to notice new messages,
this PR shows the number of unread messages in the chat's tab-bar item -
and not only on the app-icon on the homescreen.

this is esp. useful as [tapping the icon scrolls up now](https://github.com/deltachat/deltachat-ios/pull/2127),
giving users and intuitive way to get to the new chats.

both badge counters will show the same number in the same layout.
if notifications are disabled, both counters are hidden
(the badges beside chats stay)

in practise, this feels totally normal and correctly,
this behavior is known from many other apps on iOS.
(i was also trying to show only the number of the active account,
that is more confusing, also in some tiny tests with some ppl.
i also tried other colors, but the default color seems the way to go here)

this also resets the counter on app start,                                      
maybe fixing some rare bugs where the counter is wrong on the app icon 

closes #2130

note, that the badge counters are not strictly event driven as they may be updates also when the app is not running by the [NSE](https://github.com/deltachat/deltachat-ios/pull/2105)

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -284,6 +284,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         logger.info("➡️ applicationDidBecomeActive")
         UserDefaults.setMainIoRunning()
         applicationInForeground = true
+        NotificationManager.updateBadgeCounters()
     }
 
     func applicationWillResignActive(_: UIApplication) {
--- a/deltachat-ios/Controller/Settings/SettingsViewController.swift
+++ b/deltachat-ios/Controller/Settings/SettingsViewController.swift
@@ -246,7 +246,7 @@ internal final class SettingsViewController: UITableViewController {
             NotificationManager.removeAllNotifications()
         }
         UserDefaults.standard.synchronize()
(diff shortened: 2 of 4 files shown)


zeitschlag (approved):
LGTM!

PR #2155, opened 2024-04-17
Second Attempt at "Long Press to ask user to copy link" / Link-"context menu"

Simon-Laux:
- detect long tap on link (I finally found out how to do it 🎉 )
- add an action sheet

This still is a draft with probably rather bad code, but it already works.

- only works on iOS 13+ but I think thats ok
- [ ] I don't really like my time-based-workaround against false triggers of the normal tap handler.
- [ ] A. Would be cool to also use the ios context menu instead of the action sheet, but there is no real link element, it's just part of the UILabel, so probably complicated to make it look good
- [ ] B. or as an alternative to A we could also copy the design of the "copy link sheet" from whatsapp, it seems to be similar to the safari tab switcher or our account switcher.
- [ ] The jumping around of the message list when the sheet opens is not nice to look at

Optional:
- [ ] add small haptic feedback when opening the action sheet, that would already be included with A.
- [ ] Add a option to copy to phone numbers, maybe also a "create phone contact" action

Disclaimer: This is just me playing around because I want this feature (https://support.delta.chat/t/copy-a-link-from-a-message-on-long-press-in-ios/2755).
It is not currently a project priority.

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -2027,7 +2027,16 @@ extension ChatViewController {
         if tableView.isEditing || messageId == DC_MSG_ID_MARKER1 || messageId == DC_MSG_ID_DAYMARKER {
             return nil
         }
-
+        // Check if the long tap is on a link (or other message text element with custom long tap behavior)
+        if let msgcell = tableView.cellForRow(at: indexPath) as? BaseMessageCell {
+            let label = msgcell.messageLabel.label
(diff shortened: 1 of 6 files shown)


Simon-Laux:
Using this already daily since I made the pr and it works good for me most of the time.
Is probably easy to implement more cleanly for someone with more fundamental knowledge of iOS tap handlers and tap-event propagation.

r10s:
i picked up esp. the last suggestion and created https://github.com/deltachat/deltachat-ios/pull/2445

PR #2158, opened 2024-04-25
try to support openpgp4fpr links inside webxdc

### LINKED ISSUE #2157: allow openpgp4fpr scheme links inside webxdc
so iOS users can discover bots and benefit from the [Public Bots Index](https://github.com/deltachat-bot/public-bots) which is also important WRT the new onboarding effort since with chatmail instances, users can not write directly to bots but need to verify them first to be able to write, with this chatmail users will be able to use bots

this is already supported by Delta Chat Desktop and will be supported by Android in https://github.com/deltachat/deltachat-android/pull/3008

adbenitez:
I am rather skeptical this is correct since couldn't find a similar case to imitate but hopefully at least the PR could serve as a base for someone to take over and tweak/fix

close #2157

### DIFF
--- a/deltachat-ios/Controller/WebxdcViewController.swift
+++ b/deltachat-ios/Controller/WebxdcViewController.swift
@@ -322,7 +322,7 @@ class WebxdcViewController: WebViewViewController {
                 openChatFor(url: url)
                 decisionHandler(.cancel)
                 return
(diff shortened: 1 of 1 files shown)


r10s:
nb: for testing, ex. write to <botsindex@nine.testrun.org> and try to click a bot in the list

r10s (approved):
like a pro!

works as expected 👌

PR #2160, opened 2024-05-04
reorder settings

r10s:
- creates a dedicated "experimental" section and stops using a button as a sort of headline and mixing the section with "not experimental, but not recommended" reading address book

- sync settings positions with android,  taking new chatmail into account

- moreover, when on that, use less error-prone settings array (copy pattern used elsewhere as well)

closes https://github.com/deltachat/deltachat-ios/issues/2162


adbenitez (approved):
I am not completely happy with the style and best practices but at least it compiles... in my mind

PR #2164, opened 2024-05-09
Make Instant Onboarding-Screen a little nicer

zeitschlag:
- Improves the "Instant Onbaording"-Page
- Adds a "read privacy policy"-button
- Adds a button to add an image
- Makes adding a name mandatory

<details>
<summary>Screenshots</summary>

Simulator Screenshot - iPhone 15 Pro - 2024-05-15 at 17 09 10

Simulator Screenshot - iPhone 15 Pro - 2024-05-15 at 17 08 27
Simulator Screenshot - iPad (10th generation) - 2024-05-15 at 17 17 22

</details>


deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingViewController.swift
```diff
@@ -1,78 +1,50 @@
...
-    }
+class InstantOnboardingViewController: UIViewController, ProgressAlertHandler {
 
     private let dcContext: DcContext
     private let dcAccounts: DcAccounts
     weak var progressAlert: UIAlertController?
     var progressObserver: NSObjectProtocol?
 
-    private lazy var mediaPicker: MediaPicker? = {
+    var contentView: InstantOnboardingView { view as! InstantOnboardingView }
```
r10s, line 12: there is a linter warning about the force cast - on github as well as in xcode.

in general, we're not using force cast because of bad experiences in the past - surely, it is useful here and there, but these things tend to sneak in to places where they really should not be used and cause real issues then.

not sure about the best graceful way, but a ...

```suggestion
    var contentView: InstantOnboardingView? { view as? InstantOnboardingView }
```

... and replacing the few usages by `contentView?` seems okay. alternatively, we could do a `guard` and crash explicitly if that fails.

deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingViewController.swift
```diff
@@ -1,78 +1,50 @@
...
+        let contentView = InstantOnboardingView(avatarImage: dcContext.getSelfAvatarImage())
+        contentView.agreeButton.addTarget(self, action: #selector(InstantOnboardingViewController.acceptAndCreateButtonPressed), for: .touchUpInside)
+        contentView.imageButton.addTarget(self, action: #selector(InstantOnboardingViewController.onAvatarTapped), for: .touchUpInside)
+        contentView.privacyButton.addTarget(self, action: #selector(InstantOnboardingViewController.showPrivacy(_:)), for: .touchUpInside)
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(InstantOnboardingViewController.textDidChangeNotification(notification:)),
+            name: UITextField.textDidChangeNotification,
+            object: contentView.nameTextField
+        )
```
r10s, line 45: don't we need to remove the observer again?

at many places, (also in code not caused by us :) ` `NotificationCenter.default.addObserver` is balanced by a`NotificationCenter.default.removeObserver()`

r10s (approved):
please have a second look if observer removal is required and avoid the force-cast.

otherwise, lgtm and can be merged then - thanks a lot!

PR #2165, opened 2024-05-11
fix showing input bar after forwarding

### LINKED ISSUE #2153: message input bar disappears after forwarding
when forwarding  a message to a normal chat (not: "Saved Messages"), the message input bar makes place for the confirmation alert, which is fine.

however, after forwarding the message input bar does not reappear, so that one can only send another message after re-entering the chat.

this bug probably came in by recent changes in that area (#2117, #2114)

r10s:
this commit shows the forward-confirmation-alert
to _after_ the view controller is really created.

otherwise, on some iOS versions,
the alert interferes with setting up the view controller in viewWillAppear

tested on iphone13/ios17.4, iphone7/ios15 and emulator/ios17.0

closes #2153


zeitschlag (approved):
Makes sense to me, nice catch!

PR #2166, opened 2024-05-14
for chatmail accounts, hide some not useful settings

r10s:
counterpart of https://github.com/deltachat/deltachat-android/pull/3050 ,   see there for discussion and reasoning

this is how the advanced settings look like now: left classicmail, right chatmail ... so much cleaner :)

### DIFF
--- a/deltachat-ios/Controller/Settings/AdvancedViewController.swift
+++ b/deltachat-ios/Controller/Settings/AdvancedViewController.swift
@@ -12,8 +12,8 @@ internal final class AdvancedViewController: UITableViewController, ProgressAler
     }
 
     private enum CellTags: Int {
+        case defaultTagValue = 0
         case showEmails
-        case autocryptPreferences
         case sendAutocryptMessage
         case manageKeys
         case videoChat
@@ -49,7 +49,6 @@ internal final class AdvancedViewController: UITableViewController, ProgressAler
 
     private lazy var autocryptPreferencesCell: UITableViewCell = {
         let cell = UITableViewCell(style: .default, reuseIdentifier: nil)
-        cell.tag = CellTags.autocryptPreferences.rawValue
         cell.textLabel?.text = String.localized("autocrypt_prefer_e2ee")
         cell.accessoryView = autocryptSwitch
         cell.selectionStyle = .none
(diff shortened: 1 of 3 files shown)


deltachat-ios/Controller/Settings/AdvancedViewController.swift
```diff
@@ -56,17 +56,17 @@ internal final class AdvancedViewController: UITableViewController, ProgressAler
         return cell
     }()
 
-    private lazy var sendAutocryptMessageCell: ActionCell = {
-        let cell = ActionCell()
+    private lazy var sendAutocryptMessageCell: UITableViewCell = {
+        let cell = UITableViewCell(style: .value1, reuseIdentifier: nil)
```
zeitschlag, line 60: What's the reason vor `.value1` here? From the docs:

> A style for a cell with a label on the left side of the cell with left-aligned and black text; _on the right side is a label that has smaller blue text and is right-aligned._

Do we need that _smaller blue text_ or would `default` suffice, also for `manageKeysCell`?

deltachat-ios/Controller/Settings/AdvancedViewController.swift
```diff
@@ -219,20 +219,32 @@ internal final class AdvancedViewController: UITableViewController, ProgressAler
...
-        let serverSection = SectionConfigs(
-            headerTitle: String.localized("pref_server"),
-            footerTitle: String.localized("pref_only_fetch_mvbox_explain"),
-            cells: [accountSettingsCell, sentboxWatchCell, sendCopyToSelfCell, mvboxMoveCell, onlyFetchMvboxCell])
-        return [viewLogSection, experimentalSection, appAccessSection, autocryptSection, serverSection]
+
+        if dcContext.isChatmail {
+            let encryptionSection = SectionConfigs(
+                headerTitle: String.localized("pref_encryption"),
+                footerTitle: nil,
```
zeitschlag, line 226: We could think about adding a custom `init` for `SectionConfigs` with `nil` as default value for `footerTitle` 🤔

zeitschlag (approved):
LGTM. From what I understand, this requires to update the core-dependency, right?

Oh, and I just saw this message in the Xcode console:

> 🧡 [1] deltachat-ffi/src/lib.rs:285: dc_get_config(): invalid key 'is_chatmail'

This probably says that I didn't update core, right?

deltachat-ios/Controller/Settings/AdvancedViewController.swift
r10s: yes, maybe that makes sense. however, the pattern is used also in many other views, so, i'll stay with it for now, it is easy enough

r10s:
> This probably says that I didn't update core, right?

core was not updated on the `main`, will do that now

PR #2168, opened 2024-05-16
Make instant onboarding even nicer and more useful

zeitschlag:
- Replaces the Logo on Welcome-screen to show the delta.chat-logo
- Adds a few buttons here and there (for provider change or QR-code-scanning)
- Speaking of which: Users can change their provider now using the "Explore other Options"-button
- Speaking of which: Users can scan a QR-Code now using the "Scan QR Code"-button (this will need some more testing
- Screenshots from an iPhone 15 Pro Max

dc_2168

successor of #2163, #2164

### DIFF
--- a/deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingViewController.swift
+++ b/deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingViewController.swift
@@ -82,7 +82,7 @@ class InstantOnboardingViewController: UIViewController, ProgressAlertHandler {
         } else {
             customProvider = nil
         }
-        let contentView = InstantOnboardingView(avatarImage: dcContext.getSelfAvatarImage(), customProvider: customProvider)
+        let contentView = InstantOnboardingView(avatarImage: dcContext.getSelfAvatarImage(), name: dcContext.displayname, customProvider: customProvider)
         contentView.agreeButton.addTarget(self, action: #selector(InstantOnboardingViewController.acceptAndCreateButtonPressed), for: .touchUpInside)
         contentView.imageButton.addTarget(self, action: #selector(InstantOnboardingViewController.onAvatarTapped), for: .touchUpInside)
         contentView.privacyButton.addTarget(self, action: #selector(InstantOnboardingViewController.showPrivacy(_:)), for: .touchUpInside)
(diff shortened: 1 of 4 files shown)


r10s:
when adding an avatar and then entering a name, the "Agree & Continue" button is not enabled.

in general, it is not obvious that the button is in disabled state, may be regarded as "not working", ppl may not read, tap the policy to do sth then etc.

 it seems better to just always enable it and to add a toast/alert/whatever, pointing to the need to enter a name then (as on android)

deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingViewController.swift
```diff
@@ -90,13 +132,35 @@ class InstantOnboardingViewController: UIViewController, ProgressAlertHandler {
...
+            UIApplication.shared.open(url)
+        }
+    }
+
+    @objc private func scanQRCode(_ sender: UIButton) {
+        let qrReader = QrCodeReaderController(title: String.localized("multidevice_receiver_title"),
+                    addHints: "➊ " + String.localized("multidevice_same_network_hint") + "\n\n"
+                        +     "➋ " + String.localized("multidevice_open_settings_on_other_device") + "\n\n"
+                        +     String.localized("multidevice_experimental_hint"),
+                    showTroubleshooting: true)
```
r10s, line 156: title/hints should be like the following:

```suggestion
        let qrReader = QrCodeReaderController(title: String.localized("scan_invitation_code"))
```

(taking from removed https://github.com/deltachat/deltachat-ios/pull/2163/files#diff-d82264e2f2825dbdcaa4e863471e22a9b8e3bb46ee19e7b22528427ab6adcc86L38 )

r10s:
the code looks good, i did some functionality test and came over the following two issues:

- when using "scan QR code", the app should land again in `InstantOnboardingViewController` - just the same as if one selects a server with "Explore Other Options". an additional question after "scan QR code" is not needed as one need to tap "Agree & Create" in `InstantOnboardingViewController` anyways. also the old "edit profile" dialog that currently pops up is not needed then (as already set in `InstantOnboardingViewController` )

- when setting name+avatar and then going to "Explore Other Options" or  "scan QR code", one is back in  `InstantOnboardingViewController` - however, the set name+avatar is reseted. we just had [the same issue on android](https://github.com/deltachat/deltachat-android/pull/3067) and fixed that by saving name+avatar always on leaving the activity/ViewController 

for testing, try eg. https://bcc.chat

r10s (approved):
things basically work now, thanks a lot!

let's merge that in an iterate in subsequent PR

PR #2172, opened 2024-05-21
update translations

r10s:
bug fixed at first commit, also a description there


adbenitez (approved):

r10s:
thanks a lot for approval, we all agree it is questionable to use regex for xml at all, but well, these are only few well-known files, things run on dev machine only and checked by hand afterwards ... it is what it is

PR #2173, opened 2024-05-21
port convert-translations script to python

adbenitez:
use a xml parser instead of relaying on regex for xml parsing which can cause several issues in the future like the one fixed in https://github.com/deltachat/deltachat-ios/pull/2172


r10s (approved):
very nice!

thanks a looooot!

this gets us more actionable, also when it comes to android maybe switching to weblate - and we would probably get the strings out of the local deltachat-android checkout

PR #2182, opened 2024-05-27
tweak onboarding

### LINKED ISSUE #2180: instant onboarding: adapt title of "Privacy Policy" link
the title needs to read "Privacy Policy for %1$s" (`instant_onboarding_agree_default2`) with %1$s set to nine.testrun.org (without quotes or so). the old "Read the Privacy Policy" string is deprecated.

this make the server the user onboards clear. also the user gets an idea what will be replaced by "Use Another Server"

i  forgot to mention that in https://github.com/deltachat/deltachat-ios/issues/2170, counterpart of https://github.com/deltachat/deltachat-desktop/issues/3860

### LINKED ISSUE #2181: new groups offers directly adding contacts manually
we should go for "New Contact" using QR codes first (and not forget to refresh the memberlist) or remove the option alltogether, for a created group, there is a  group-invite (iirc, this is the way we're going on android and desktop)

r10s:
this PR adds the domain to the "privacy policy" and tweaks colors and sizes.  
it is all tiny changes only, however, esp. in combination with the welcome screen, it looks more coherent, closes #2180

moreover, this PR removes the option to add contacts to groups by "email address" for chatmail-profiles, closes #2181

before / after:

### DIFF
--- a/deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingView.swift
+++ b/deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingView.swift
@@ -71,7 +71,7 @@ class InstantOnboardingView: UIView {
         contentStackView.setCustomSpacing(16, after: imageButton)
         contentStackView.setCustomSpacing(16, after: nameTextField)
         contentStackView.setCustomSpacing(8, after: hintLabelWrapper)
-        contentStackView.setCustomSpacing(16, after: privacyButtonWrapper)
+        contentStackView.setCustomSpacing(32, after: privacyButtonWrapper)
         contentStackView.setCustomSpacing(16, after: agreeButton)
 
         contentScrollView = UIScrollView()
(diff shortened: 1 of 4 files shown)


deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingView.swift
```diff
@@ -144,16 +147,18 @@ class InstantOnboardingView: UIView {
 
         if buttonShouldBeEnabled {
             agreeButton.backgroundColor = .systemBlue
+            otherOptionsButton.setTitleColor(.systemBlue, for: .normal)
         } else {
             agreeButton.backgroundColor = UIColor.systemGray.withAlphaComponent(0.3)
+            otherOptionsButton.setTitleColor(UIColor.systemGray.withAlphaComponent(0.5), for: .normal)
```
zeitschlag, line 153: From what I understand, `otherOptionsButton` isn't disabled here. It's just the text color that gets set to a light gray, if a user doesn't give a name. Is this intended?

zeitschlag (approved):

deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingView.swift
r10s: yeah, this is not optimal, but still better than before imo, see last commit message

PR #2185, opened 2024-05-29
import vcards

r10s:
targets #2167

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -2123,9 +2123,9 @@ extension ChatViewController {
     }
 
     func didTapVcard(msg: DcMsg) {
-        guard let file = msg.file else { return }
-        guard let vcards = dcContext.parseVcard(path: file) else { return }
-        guard let vcard = vcards.first else { return }
+        guard let file = msg.file,
+              let vcards = dcContext.parseVcard(path: file),
+              let vcard = vcards.first else { return }
         
         let alert = UIAlertController(title: String.localizedStringWithFormat(String.localized("ask_start_chat_with"), vcard.displayName), message: nil, preferredStyle: .safeActionSheet)
         alert.addAction(UIAlertAction(title: String.localized("start_chat"), style: .default, handler: { _ in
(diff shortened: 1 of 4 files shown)


deltachat-ios/DC/DcContext.swift
```diff
@@ -388,6 +388,29 @@ public class DcContext {
         }
     }
 
+    public func parseVcard(path: String) -> [DcVcardContact]? {
+        if let data = DcAccounts.shared.blockingCall(method: "parse_vcard", params: [path as AnyObject]) {
```
zeitschlag, line 392: We could put all those strings (`parse_vcard`, `import_vcard`, `send_reaction`) into constants at some point

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -2114,6 +2122,20 @@ extension ChatViewController {
         navigationController?.present(inputDlg, animated: true, completion: nil)
     }
 
+    func didTapVcard(msg: DcMsg) {
+        guard let file = msg.file else { return }
+        guard let vcards = dcContext.parseVcard(path: file) else { return }
+        guard let vcard = vcards.first else { return }
```
zeitschlag, line 2128: We could combine these three `guard` in one, like:

```suggestion
        guard let file = msg.file,
              let vcards = dcContext.parseVcard(path: file),
              let vcard = vcards.first else { return }
```

zeitschlag (approved):
Looks good! I added some minor comments here and there, but nothing blocking, so I am going to merge this for now, as discussed, and start working on that cell.

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -2114,6 +2122,20 @@ extension ChatViewController {
...
     }
 
+    func didTapVcard(msg: DcMsg) {
+        guard let file = msg.file,
+              let vcards = dcContext.parseVcard(path: file),
+              let vcard = vcards.first else { return }
+        
+        let alert = UIAlertController(title: String.localizedStringWithFormat(String.localized("ask_start_chat_with"), vcard.displayName), message: nil, preferredStyle: .safeActionSheet)
+        alert.addAction(UIAlertAction(title: String.localized("start_chat"), style: .default, handler: { _ in
+            if let ids = self.dcContext.importVcard(path: file) {
```
zeitschlag, line 2132:
```suggestion
            if let ids = self.dcContext.importVcard(path: file) {
                print("importing ids: \(ids)")
```

PR #2187, opened 2024-05-30
Show avatar for vCards (#2167)

zeitschlag:
This PR improves vCard-UI in chats by adding a dedicated view and cell. The result looks pretty similar to when sending files, but instead of the icon, you see an image (if the vCard contains an image) or the initials with a background color (aka `InitialsBadge`, also used elsewhere. For privacy reasons the screenshot below has been edited)

Simulator Screenshot - iPhone 15 Pro - 2024-06-04 at 15 06 36


r10s:
did a quick functionality test - very nice so far!

the only thing that i recognised so far, is that we do not want to have `person.crop.circle` (sorry, once again not :) or so as a fallback - but the user's color and the first letter of the displayname (both are returned by `parseVcard()`)  - see eg. the normal contact list (from which we should share the corresponding code ideally)

this is how it looks like on android:

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -655,11 +656,9 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
                 cell = audioMessageCell
             }
         case DC_MSG_VCARD:
-            cell = tableView.dequeueReusableCell(withIdentifier: "text", for: indexPath) as? TextMessageCell ?? TextMessageCell()
-            message.text = "👤 Contact"
-            if let file = message.file, let vcards = dcContext.parseVcard(path: file), let vcard = vcards.first {
-                message.text = "👤 " + vcard.displayName
-            }
+            guard let contactCell = tableView.dequeueReusableCell(withIdentifier: ContactCardCell.reuseIdentifier, for: indexPath) as? ContactCardCell else { fatalError("WTF wrong cell??") }
```
r10s, line 659: let's do the same here as for the other cases.

even if that should really not happen, some little forgiveness may be appreciated. there may be cornercases, races, bugs in ios in future versions, bugs in delta chat in future versions. crashing the app when opening a chat is really a bad thing for the user.

if we're really sure this never can't happen (i am not :) we should change all cases in another PR. 

```suggestion
            let contactCell = tableView.dequeueReusableCell(withIdentifier: ContactCardCell.reuseIdentifier, for: indexPath) as? ContactCardCell ?? ContactCardCell()
```

deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingViewController.swift
```diff
@@ -130,7 +130,7 @@ class InstantOnboardingViewController: UIViewController, ProgressAlertHandler {
 
     private func deleteProfileIconPressed(_ action: UIAlertAction) {
         dcContext.selfavatar = nil
-        contentView?.imageButton.setImage(UIImage(named: "person.crop.circle"), for: .normal)
+        contentView?.imageButton.setImage(UIImage(named: "camera"), for: .normal)
```
r10s, line 133: ftr: unrelated change, this resets the avatar in the onboarding screen correctly

r10s (approved):
very nice, good looking and does what it should :)

some minor comments only

Pods/Swifter/XCode/Sources/String+BASE64.swift
```diff
@@ -13,3 +13,11 @@ extension String {
...
     }
 }
+
+extension UIImage {
+    public static func fromBase64(string: String) -> UIImage? {
+        guard let imageData = Data(base64Encoded: string) else { return nil }
+
+        return UIImage(data: imageData)
+    }
+}
```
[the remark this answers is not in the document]
zeitschlag: Good catch! I moved it out of the pod (as it doesn't belong there) to our code base

PR #2188, opened 2024-05-30
create chat with first imported contact

r10s:
targets #2167


r10s:
moved forward to merge that bit in to have it in the testflight

PR #2194, opened 2024-06-01
hint user that rollout may take time

r10s:
seen a similar hint for other apps,
it is useful to give interested users a clue
why they have a feature but others may not.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -51,7 +51,8 @@ on https://appstoreconnect.apple.com :
 
 8. for a **Regular release**, open "My Apps/Delta Chat iOS/iOS App+ (first item)"
    a) enter the version number (without leading "v")
-   b) fill out "what's new", use CHANGELOG.md as a template
+   b) fill out "what's new", use CHANGELOG.md as a template, add the line:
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -51,7 +51,8 @@ on https://appstoreconnect.apple.com :
 
 8. for a **Regular release**, open "My Apps/Delta Chat iOS/iOS App+ (first item)"
    a) enter the version number (without leading "v")
-   b) fill out "what's new", use CHANGELOG.md as a template
+   b) fill out "what's new", use CHANGELOG.md as a template, add the line:
+      "These features will roll out over the coming days. Thanks for using Delta Chat!"
```
zeitschlag, line 55: We could add something like "If you want to get them now, please head to the App Store and download the latest (and greatest) Delta Chat manually". This is a way for users to bypass the phased release

zeitschlag (approved):
👍

RELEASE.md
r10s: hm, was also expecting it to work like that, however, last time we tried that, we did not find a  reliable way to force an update. but maybe it has changed since then - or i messed sth. up.

PR #2199, opened 2024-06-04
Switch menu entries and put them in submenus (#2193)

### LINKED ISSUE #2193: "delete" should be last in gallery context menu
corrently, "delete" is the first item, which ins unusual and breaks both, inner and outer consistency

zeitschlag:
Original idea was to switch order of entries, but I thought: Why not adding that thin line? Anyway, it looks like this now:

Simulator Screenshot - iPhone 15 - 2024-06-04 at 15 51 52

Fixes #2193


r10s (approved):
very nice!

PR #2201, opened 2024-06-07
Cleanup dequeue

zeitschlag:
- Crash if dequeued cell can't be casted (i.e. someone forgot to register that very cell)
- use static `reuseIdentifier` for cells instead of strings
- Use this pattern everywhere (and cleanup here and there)


r10s (approved):
thanks for the chore! 🧹

Issue #2128, opened 2024-03-26
pasted photos are not staged but directly sent


r10s:
while this is okay for stickers (which we assume to be an often pasted content as this ways is used when a sticker is selected from the keyboard), photos should be staged before (eg. when using copy+paste for a photo received from whatsapp)

that time, we did not figure out how to differ between stickes and non-stickers - and were happy we support that way at all - but i think we should recheck and see if we can do better

PR #2203, opened 2024-06-08
Stage images, but send stickers (#2128)

zeitschlag:
~~`isSticker` with checking on image sizes feels so wrong, if someone has a better idea how to differ between stickers and images: I'm all ears!~~

After a short debate we decided to check for size for now. If that doesn't work well enough,  we'll consider adding a check for a transparent pixel on `[0, 0]`


Simon-Laux:
>  if someone has a better idea how to differ between stickers and images: I'm all ears!

check for transparent border / edges. "real" stickers have a transparent background and a white outline to make them look like stickers, at least telegram and others have that in their guidelines for sticker creation.

r10s (approved):

Simon-Laux:
> After a short debate we decided to check for size for now. 

what if my sticker is high-res? or is this size some official number/size for sticker keyboards on iOS? like where does the number come from?

PR #2204, opened 2024-06-09
show device-message only once as unread

### LINKED ISSUE #2190: add device update messages as unread only for the first account
if subsequent accounts are added, it does not make sense to show the device message unread-counter for already read messages over and over

see https://github.com/deltachat/deltachat-android/pull/3069/files

no priority, this is mainly an issue for us devs :)

r10s:
on switching or adding profiles,
the device messages (welcome and update messages) are added, whowever, the chat is marked as being read.

this avoid annoying "unread badges" over and over
when using multiple profiles.

closes #2190


zeitschlag (approved):
:D I had to smile as this feels like a pretty down to earth-approach. 👍

r10s:
well, i was first thinking about a ChangeableState enum and then a factory to create IsStateChangeEligable objects asynchrously. event-driven, of course

PR #2208, opened 2024-06-13
Send vCards

### LINKED ISSUE #2186: Send vCards
- [x] add support for sending vCards (depends on https://github.com/deltachat/deltachat-core-rust/pull/5571)
- [x] display in draft/staging

See #2167.

zeitschlag:
Adds a way to send contacts (via vCards) to other users. Users can select one from a list of their contacts. The selected contact gets staged then and they can add a message. Looks like this:

send_vcards

Closes #2186.

### DIFF
--- a/deltachat-ios/Chat/Views/ContactCardPreview.swift
+++ b/deltachat-ios/Chat/Views/ContactCardPreview.swift
@@ -1,37 +1,35 @@
 import UIKit
 import DcCore
 
-public class DocumentPreview: DraftPreview {
+public class ContactCardPreview: DraftPreview {
 
     weak var delegate: DraftPreviewDelegate?
 
-    lazy var fileView: FileView = {
-        let view = FileView()
+    lazy var contactCardView: ContactCardView = {
+        let view = ContactCardView()
         view.translatesAutoresizingMaskIntoConstraints = false
-        view.allowLayoutChange = false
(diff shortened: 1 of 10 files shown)


deltachat-ios/Chat/Views/ContactCardPreview.swift
```diff
@@ -0,0 +1,64 @@
...
+            delegate?.onAttachmentAdded()
+            accessibilityLabel = "\(String.localized("attachment")), \(contactCardView.configureAccessibilityLabel())"
+            isHidden = false
+        } else {
+            isHidden = true
+        }
+    }
+
+    @objc func contactCardViewTapped() {
+//        delegate?.onAttachmentTapped()
```
r10s, line 62: that line can be removed
```suggestion
```

r10s (approved):
lgtm, very nice!

for the "Cyclomatic Complexity Violation" of ChatViewController::tableView(): i do not agree with the linter that the function is complex at all. it is just a little longer and contains some simple, sequencial if/switches - splitting that to more functions _adds_ complexity.

therefore, and to protect against more wrong warnings, in `.swiftlint.yml` just set `cyclomatic_complexity: warning:` to `25` (adding an exception for the concrete function looks as if we agree with the linter in general)

PR #2215, opened 2024-06-21
Remove message after five (not four) weeks (#2152)

### LINKED ISSUE #2152: Replace 4 weeks with 5 weeks at "Setting / Chats and Media / Delete Old Messages"
Similar desktop issue: https://github.com/deltachat/deltachat-desktop/issues/3509

From the chat:
> I noticed that my iPadOS 1.44 delta chat client offers a server and local device message delete interval of four weeks whereas every other client offers five weeks. Anyone else noticing this?

zeitschlag:
Update Settings-screen to match Message-Removal-Intervals. Closes #2152

### DIFF
--- a/deltachat-ios/Controller/Settings/AutodelOptionsViewController.swift
+++ b/deltachat-ios/Controller/Settings/AutodelOptionsViewController.swift
@@ -13,11 +13,11 @@ class AutodelOptionsViewController: UITableViewController {
     private static let autodelDeviceOptions: [Options] = {
         return [
             Options(value: 0, descr: "never"),
-            Options(value: 3600, descr: "autodel_after_1_hour"),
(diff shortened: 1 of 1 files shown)


r10s (commented):
lgtm.

i was first a bit sceptical, what happens if the user has selected "4 weeks" before - but actually, it is not that bad.

- just no option is shown, the user can select a new one - if they just hit "OK" without doing a selection, the original value is preserved

- not super-nice is the summary shown one level up: that says "Err" (comes from AutodelOptionsViewController::getSummary()) - i suggest to do the same as on desktop here and just unexitedly show the value in seconds.  
so do a `return "After \(val) seconds"` instead of `return "Err"` - a translation is not needed as this is a cornercase that will disappear over time (no one ever complained on desktop)

r10s (approved):

PR #2218, opened 2024-06-21
Add search to send-contact-list (#2209)

### LINKED ISSUE #2209: Add Search to "Send Contact"-screen
TBD

- We have to use Core to filter
- Something something use a shared component with `GroupMembers`/`NewChat`/...

zeitschlag:
Steals some of stuff from `GroupMemberViewController`, but I don't think a new component makes a lot of sense here. 
It uses the Core(tm) for search and to the user it looks like this:

Closes #2209.


r10s:
two bugs i encountered during testing:

- when searching for a contact and tapping the result, not the tapped result is staged as vcard - but sth. else (maybe the contact at the given position without search)

- sometimes, the input bar with the staged vcard is not shown - it is still staged (so if you leave and re-enter the chat, you see the staged vcard)  
this happens every few times for me on iphone13 (i also have video with lots of personal data, however, can send that in private if needed)

r10s (approved):

PR #2219, opened 2024-06-24
tint 'delete' buttons red consistently

r10s:
pr #2211 makes the trash cans red for the chat view, the menu entries are red as well,
so use the same for the chatlist view.


zeitschlag (approved):
👍

PR #2223, opened 2024-06-28
Consider keyboard-frame when keyboard shows up (#2122)

### LINKED ISSUE #2122: log view: keyboard covers the last lines of text, making selecting/editing hard in that area
at best, the input area should shrink, maybe just a flag (ahh, input-accessory hell again :) if in doubt, a dirty hack would be to add a number of empty lines to the end, for a log, probably good enough.

for details see: https://support.delta.chat/t/two-small-tweaks-to-improve-the-log-screen-in-dc-ios/2991

zeitschlag:
Also: Cleanup here and there.

Closes #2122


r10s (approved):
thanks a lot!

tested + code lgtm!

Issue #2221, opened 2024-06-27
InputBar-Issues in ChatViewController


zeitschlag:
## InputBar visible when it shouldn't

- In a chat, stage a Contact to send, but don't send it
- Remove that Contact from Staging Area
- Open Contact-Select-screen from attachment

-------

## InputBar disappears when it shouldn't

- Open Attach-Contact-Screen
- Search for a contact
- Select and stage a contact
- Remove that contact from Staging Area
- Open Attach-Contact-Screen
- Search for a contact
- Select and stage a contact

--------

Somethingsomething `firstResponder`, `shouldBecomeFirstResponder` on `ChatViewController`

PR #2224, opened 2024-06-28
Set First Responder (or don't, #2221)

zeitschlag:
Addresses both issues of #2221. At least on my machine™️ both issues are fixed. It required some\* manual testing

\* = lots of


r10s (approved):
code lgtm - but that does not mean much on such issues :)

did some testing on my machine™ - also fine :)

no doing lots™ of testing the next days on my device, but i am positive that we can merge already :)

PR #2231, opened 2024-07-11
Scroll to bottom when scrolled to bottom

### LINKED ISSUE #2129: scroll down on-sent missing sometimes
when the message view is already scrolled down completel, sending a new message does not scroll down.

cave: for incoming messages we want to scroll down only if already scrolled down - otherwise the scrolling position should stay as is (otherwise it is hard to read old messages in a chatty chat)

Seems to happen with stickers.

zeitschlag:
Fixes #2129

### DIFF
--- a/DcCore/DcCore/DC/events.swift
+++ b/DcCore/DcCore/DC/events.swift
@@ -72,8 +72,8 @@ public class DcEventHandler {
                 ])
             }
         case DC_EVENT_MSGS_CHANGED:
-            guard accountId == dcAccounts.getSelected().id { return }
-            
+            guard accountId == dcAccounts.getSelected().id else { return }
+
             logger.info("📡[\(accountId)] msgs changed: \(data1), \(data2)")
             DispatchQueue.main.async {
                 NotificationCenter.default.post(name: .messageChanged, object: nil, userInfo: [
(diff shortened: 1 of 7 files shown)


r10s (approved):
yeah, works great - and fixes the reproduce-way i just added to #2129

the "scroll down" button shortly appears, however, that's not too bad, shows a but what happens :) i would change that only if it is really easy to do with a line of code and no other issues opened.

DcCore/DcCore/DC/events.swift
```diff
@@ -1,7 +1,11 @@
 import UIKit
 import UserNotifications
 
-public let eventMsgsChangedReadDeliveredFailed = Notification.Name(rawValue: "eventMsgsChangedReadDeliveredFailed")
+extension Notification.Name {
+    public static let messageChanged = Notification.Name(rawValue: "eventMsgsChanged")
```
r10s, line 5: it's plural [as multiple messages may be affected](https://c.delta.chat/group__DC__EVENT.html#ga0f52cdaad70dd24f7540abda6193cc2d). we should stay close to the original meaning here - so either `msgsChanged or messagesChanged`.

(read/delivered/failed/reactionsChanged otoh affect only a single message and singular is fine therefore)

r10s (approved):
lgtm, so far. thanks a lot, not testing ... but it can already be merged :)

DcCore/DcCore/DC/events.swift
zeitschlag: 🫨😵‍💫😅👍

PR #2235, opened 2024-07-12
ChatViewController never got deallocated

zeitschlag:
This came up during Notifications-refactoring: Instances of `ChatViewController` were never deallocated. An easy way to find out is a [Symbolic Breakpoint](https://sarunw.com/posts/easy-way-to-detect-retain-cycle-in-view-controller/) and later Memory Graph.

The reason for this was that the `ChatViewController` references an `InputBarAccessoryView` (the `messageInputBar`) which itself has a closure `onScrollDownButtonPressed` which itself hold a strong reference to the `scrollToBottom`-function of in instance of `ChatViewController` (that although disappearing, never deallocated).

I changed `onScrollDownButtonPressed` to have its own closure that holds a `weak` reference to `self` and calls `scrollToBottom`.

This was a nice, fun little puzzle to solve.


r10s (approved):
woah! good catch! excellent 👌

PR #2236, opened 2024-07-12
Remove observers for events

zeitschlag:
- Fix some leftovers from #2234
- Migrate over everything but Progress-observers from closures to methods. This way there's no need to remove the observers manually.
- Clean up `DispatchQueue.Main`-handling (at least in `ChatViewController`) (Caller site decides)

------

- As discussed: Observers in `ProgressAlertHandler` are not part of this PR as I need to come up with an alternative approach to that Protocol-extension and closures. Let me think about it a little longer.

### DIFF
--- a/DcCore/DcCore/DC/events.swift
+++ b/DcCore/DcCore/DC/events.swift
@@ -1,31 +1,31 @@
 import UIKit
 import UserNotifications
 
-extension Notification.Name {
+public enum Event {
     // Messages
-    public static let messagesChanged = Notification.Name(rawValue: "eventMsgsChanged")
+    public static let messagesChanged = Notification.Name(rawValue: "messagesChanged")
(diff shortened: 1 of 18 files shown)


r10s:
yeah:  Screenshot 2024-07-19 at 14 55 08 is the harvest of the [previous](https://github.com/deltachat/deltachat-ios/pull/2235) [PRs](https://github.com/deltachat/deltachat-ios/pull/2234) 👌

r10s:
> Clean up DispatchQueue.Main-handling (at least in ChatViewController) (Caller site decides)

we're often calling `DispatchQueue.main.async` two times now. probably no big deal, however, in the future, we can consider to say that DcEventHandler does not dispatch to main and it is up to the consumer to do that as needed. that step would require careful double checking, of course.

but that is no blocker of this PR with excellent improvements together with the other efforts. this will make maintenance much easier 👌

r10s (approved):

PR #2239, opened 2024-07-16
when tapping 'Chats' multiple times, end search

### LINKED ISSUE #2171: subsequent taps on "Chats" should end search
with https://github.com/deltachat/deltachat-ios/issues/2125, multiple taps on the "Chats" scroll up; the gist is esp. to get to the new messages if a badge counter is shown there.

in case one is in search mode, however, if you see a badge counter on the "Chats" tab, the layout scrolls up, but stays in search mode, so that the new messages are not visible most times.
(EDIT: to see the tabbar, you have to hide the keyboard, this can be done eg. by entering and leaving any chat)

- whatsapp solves that by hiding the tabbar in searchmode - i do not think, this is a good idea, ux wise

- signal has the same issue as we

i suggest to end "search mode" on subsequent taps on "Chats", and then scroll up

r10s:
when tapping 'Chats' when chats is already selected, the gist is to bring the view into kind of default - so that you see eg. the messages annonced by a badge counter.

PR #2127 did a scroll-to-top for this purpose,
this PR additionally ends search.

closes #2171

### DIFF
--- a/deltachat-ios/Controller/ChatListViewController.swift
+++ b/deltachat-ios/Controller/ChatListViewController.swift
@@ -412,7 +412,8 @@ class ChatListViewController: UITableViewController {
             tableView.rowHeight = ContactCell.cellHeight
         }
     }
-    private func quitSearch(animated: Bool) {
+
+    func quitSearch(animated: Bool) {
         searchController.searchBar.text = nil
         self.viewModel?.endSearch()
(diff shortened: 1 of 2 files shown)


zeitschlag (approved):
The "quit"-animation feels a bit clunky but it does what it should do: End the search 👍

PR #2240, opened 2024-07-19
Show default-reactions as frequently used emojis

zeitschlag:


zeitschlag:
Will close this, it was based on a misunderstanding, better PR is on its way 🤙🤙

PR #2245, opened 2024-07-22
mute profiles

### LINKED ISSUE #2238: make use of core's new `is_muted` config
- "Settings / Notification" should become a per-profile setting using [core's `is_muted` config](https://c.delta.chat/classdc__context__t.html#aff3b894f6cfca46cab5248fdffdf083d)
- the profile switcher should show the "muted" state
- a migration should be done, if notifications were muted previously, all profiles should be set to muted as well

needs core 1.141.2, cmp https://github.com/deltachat/deltachat-android/pull/3143, https://github.com/deltachat/deltachat-desktop/issues/4024

r10s:
this PR allows to mute notifications for some profiles while leaving them on for others. 

muted profiles are indicated by a "mute" symbol and "grey badges" (instead of "blue") in the profile switcher, this is what is also done on desktop/android.

moreover, the "grey badges" are also used in the chatlist, to make clearer what is going on (however, without all chats having the "mute" symbol), this is a minor change, but overall, things appear more logical that away (blue badges summing up). let's see :)

the PR makes use of the new `isMuted()` API. if in doubt, we still register for notifications - a muted profile does not mean one does not want things being available instantly when opening the app. if one wants to disable notifications at all, one can still do this in the system settings, but usual iOS users won't care as push notifications are required virtually everywhere anyway. for saving battery, push notifications are anyways disabled.
also, a muted chatmail profile can still help a non-chatmail profile being updated faster this way.

however, we keep not registering for notifications for unconfigured profiles. reason is that the registration may result in a dialog-popup, we want that delayed to when the profile is really created.

closes #2238

### DIFF
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -147,9 +147,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
                 DispatchQueue.global().async { [weak self] in
                     guard let self else { return }
                     self.dcAccounts.maybeNetwork()
-                    if self.notifyToken == nil &&
-                        self.dcAccounts.getSelected().isConfigured() &&
-                        !UserDefaults.standard.bool(forKey: "notifications_disabled") {
+                    if self.notifyToken == nil && self.dcAccounts.getSelected().isConfigured() {
                         self.registerForNotifications()
                     }
                 }
@@ -175,7 +173,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             performFetch()
         }
 
-        if dcAccounts.getSelected().isConfigured() && !UserDefaults.standard.bool(forKey: "notifications_disabled") {
+        if dcAccounts.getSelected().isConfigured() {
             registerForNotifications()
         }
 
--- a/deltachat-ios/AppDelegate.swift
+++ b/deltachat-ios/AppDelegate.swift
@@ -75,6 +75,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         dcAccounts.openDatabase(writeable: true)
         migrateToDcAccounts()
 
+        // migrating global notifications pref. to per-account config, added 2024-07, can be removed after some time
+        if UserDefaults.standard.bool(forKey: "notifications_disabled") {
+            for accountId in dcAccounts.getAll() {
+                dcAccounts.get(id: accountId).setMuted(true)
+            }
+            UserDefaults.standard.removeObject(forKey: "notifications_disabled")
+        }
+        // /migrating global notifications
+
         self.launchOptions = launchOptions
         continueDidFinishLaunchingWithOptions()
         return true
--- a/deltachat-ios/Controller/AccountSwitchViewController.swift
+++ b/deltachat-ios/Controller/AccountSwitchViewController.swift
@@ -327,7 +327,7 @@ class AccountCell: UITableViewCell {
         }
 
         let unreadMessages = dcContext.getFreshMessages().count
-        accountAvatar.setUnreadMessageCount(unreadMessages)
+        accountAvatar.setUnreadMessageCount(unreadMessages, isMuted: dcContext.isMuted())
 
         accountName.text = encrypted + title
         if unreadMessages > 0 {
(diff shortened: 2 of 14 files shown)


deltachat-ios/View/ContactCell.swift
```diff
@@ -327,7 +327,7 @@ class ContactCell: UITableViewCell {
         }
 
         contactRequest.isHidden = !isContactRequest
-        mutedIndicator.isHidden = !isMuted
+        mutedIndicator.isHidden = !isChatMuted
```
zeitschlag, line 330: What about `isAccountMuted` here?

deltachat-ios/Controller/AccountSwitchViewController.swift
```diff
@@ -246,27 +246,38 @@ class AccountCell: UITableViewCell {
...
         return view
     }()
 
-    lazy var accountName: UILabel = {
+    private lazy var mutedIndicator: UIImageView = {
+        let view = UIImageView()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.widthAnchor.constraint(equalToConstant: 16).isActive = true
+        view.tintColor = DcColors.middleGray
+        view.image = #imageLiteral(resourceName: "volume_off").withRenderingMode(.alwaysTemplate)
```
zeitschlag, line 274: Why `#imageLiteral` and not `UIImage(named:...)`?

zeitschlag (approved):
LGTM! The usual questions

deltachat-ios/AppDelegate.swift
```diff
@@ -75,6 +75,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         dcAccounts.openDatabase(writeable: true)
         migrateToDcAccounts()
 
+        // migrating global notifications pref. to per-account config, added 2024-07, can be removed after some time
```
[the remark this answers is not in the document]
r10s: i think, we will stumble upon that. we had that before and removed migrations about 6~12 months later - if in doubt, when another migration is needed, the old can be replaced (happend on android :)

deltachat-ios/Controller/Settings/SettingsViewController.swift
```diff
@@ -243,7 +243,8 @@ internal final class SettingsViewController: UITableViewController {
             NotificationManager.removeAllNotifications()
         }
         UserDefaults.standard.synchronize()
-        NotificationManager.updateBadgeCounters(forceZero: !sender.isOn)
+        NotificationManager.updateBadgeCounters()
+        NotificationCenter.default.post(name: Event.messagesChanged, object: nil, userInfo: ["message_id": Int(0), "chat_id": Int(0)])
```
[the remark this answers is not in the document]
r10s: as we show muted profile bages grey (instead of blue), we need to update the chatlist on switching notifications

deltachat-ios/DC/DcAccount.swift
```diff
@@ -120,7 +120,10 @@ public class DcAccounts {
         let skipId = skipCurrent ? getSelected().id : -1
         for accountId in getAll() {
             if accountId != skipId {
-                freshCount += get(id: accountId).getFreshMessages().count
+                let dcContext = get(id: accountId)
+                if !dcContext.isMuted() {
```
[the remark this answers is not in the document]
r10s: feel free to adapt :) i am always more happy with things that do not make me think :)

deltachat-ios/View/ContactCell.swift
r10s: idea is to show the "mute icon"  for explicit action the user did - otherwise just all chats would get the icon. this is also consisten with android/desktop

the greyed badges otoh are a more subtile, alos know  not to result from user action. 

i used them here as it otherwise feels wrong to have "blue badge" on the chats that are not summed up on account switcher icon, chats icon, app icon as normally.

i an using this approch since some days and it feels quite intuitive, also what j got from my  testers nearby :)

(we can consider doing the same on android/desktop, however, there the "sum up badges" are leas of a thing, so, let's test out on ios)

deltachat-ios/Controller/AccountSwitchViewController.swift
r10s: idk - copy+paste :)

i find the representation in xcode weird as well. maybe create an issue to make that more ios-like and consistent for all places, so that next copy+paste is "better"

r10s: i changed the thing to UIImage(named:...) and created an issue for general removal: https://github.com/deltachat/deltachat-ios/issues/2248

PR #2247, opened 2024-07-24
use 'Cancel' instead of 'Back arrow' for server settings

### LINKED ISSUE #2246: server settings need a "cancel" button
sever settings currently have a "OK" button labeled "Log In" - and a "Back arrow".

UX wise, this is wrong - dialogs should _either_ have a "back arrow" in which case things are updated immediately _or_ "Cancel / OK", so "Cancel  / Log In" in this case.

r10s:
this PR fixes the issue that selecting "IMAP Security", "SMTP Security" and "Certificate Checks" are persisted in core even if no "Log In" was tried. this leads to inconsistent views. even more inconsistency was added as the text fields are persisted correctly on "Log In", so, the user never knows which field was cancelled and which not.

the reason was, that `SecuritySettingsController` and `CertificateChecksController` persisted the value too soon in core when hitting "OK" in the controller.

this PR calls a delegate instead which sets a value in the `AccountSetupController`. only if finally "Log In" is hit, the values are persisted in core together with the value from the text fields.

the code of the whole controller is a little bit messy, this PR does not clean up there as things actually are working so far. instead, the PR fixes the issue on point, keeping other flows and conventions.

finally, the PR removes some unimportant summary which is available better in "view log", adds simple "back" buttons to subsequent dialogs and renames the main "back" button to cancel (which was the original reason of #2246, i thought the rename is easy :)

closes #2246

### DIFF
--- a/deltachat-ios/Controller/AccountSetup/AccountSetupController.swift
+++ b/deltachat-ios/Controller/AccountSetup/AccountSetupController.swift
@@ -230,7 +230,7 @@ class AccountSetupController: UITableViewController, ProgressAlertHandler {
     }()
 
     lazy var certCheckCell: UITableViewCell = {
-        let certCheckType = CertificateCheckController.ValueConverter.convertHexToString(value: dcContext.certificateChecks)
+        let certCheckType = CertificateCheckController.ValueConverter.convertHexToString(value: certValue)
         let cell = UITableViewCell(style: .value1, reuseIdentifier: nil)
         cell.textLabel?.text = String.localized("login_certificate_checks")
         cell.detailTextLabel?.text = certCheckType
@@ -238,6 +238,8 @@ class AccountSetupController: UITableViewController, ProgressAlertHandler {
         cell.accessoryType = .disclosureIndicator
         return cell
     }()
+
+    lazy var certValue: Int = dcContext.certificateChecks
 
     lazy var viewLogCell: UITableViewCell = {
         let cell = UITableViewCell(style: .value1, reuseIdentifier: nil)
@@ -589,6 +591,7 @@ class AccountSetupController: UITableViewController, ProgressAlertHandler {
                 }
             }
         }
+        dcContext.certificateChecks = certValue
     }
 
     private func handleLoginSuccess() {
@@ -607,7 +610,7 @@ class AccountSetupController: UITableViewController, ProgressAlertHandler {
     private func initSelectionCells() {
         imapSecurityCell.detailTextLabel?.text = SecurityConverter.getSocketName(value: Int32(dcContext.getConfigInt("mail_security")))
         smtpSecurityCell.detailTextLabel?.text = SecurityConverter.getSocketName(value: Int32(dcContext.getConfigInt("send_security")))
-        certCheckCell.detailTextLabel?.text = CertificateCheckController.ValueConverter.convertHexToString(value: dcContext.certificateChecks)
+        certCheckCell.detailTextLabel?.text = CertificateCheckController.ValueConverter.convertHexToString(value: certValue)
     }
 
     private func resignFirstResponderOnAllCells() {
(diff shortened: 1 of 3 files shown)


deltachat-ios/Controller/AccountSetup/SecuritySettingsController.swift
```diff
@@ -1,24 +1,16 @@
 import UIKit
 import DcCore
 
+protocol SecuritySettingsDelegate: AnyObject {
+    func onSecuritySettingsChanged(newValue: Int)
+}
+
 class SecuritySettingsController: UITableViewController {
 
     private var options: [Int32] = [DC_SOCKET_AUTO, DC_SOCKET_SSL, DC_SOCKET_STARTTLS, DC_SOCKET_PLAIN]
```
zeitschlag, line 10: Why is `options` `var`? Or is this part of the "this PR does not clean up"-part?

deltachat-ios/Controller/AccountSetup/SecuritySettingsController.swift
```diff
@@ -1,24 +1,16 @@
 import UIKit
 import DcCore
 
+protocol SecuritySettingsDelegate: AnyObject {
+    func onSecuritySettingsChanged(newValue: Int)
```
zeitschlag, line 5: `options` is a list of `Int32`, what's the reason for this `Int32` -> `Int`-cast-dance? In other words: What's the reason, this parameter is of type `Int` with `options`-values being of type `Int32`?

deltachat-ios/Controller/AccountSetup/SecuritySettingsController.swift
```diff
@@ -28,15 +20,8 @@ class SecuritySettingsController: UITableViewController {
...
-        self.securityType = type
-        self.dcContext = dcContext
-        switch securityType {
-        case .IMAPSecurity:
-            selectedIndex = options.firstIndex(of: Int32(dcContext.getConfigInt("mail_security"))) ?? 0
-        case .SMTPSecurity:
-            selectedIndex = options.firstIndex(of: Int32(dcContext.getConfigInt("send_security"))) ?? 0
-        }
+    init(initValue: Int, title: String) {
+        selectedIndex = options.firstIndex(of: Int32(initValue)) ?? 0
```
zeitschlag, line 24: Setting `0` as default feels kind of weird, but I suspect it being another "this PR does not clean up"?

deltachat-ios/Controller/AccountSetup/AccountSetupController.swift
```diff
@@ -729,3 +727,21 @@ extension AccountSetupController: UITextFieldDelegate {
...
+    var value: Int
+
+    init(initValue: Int) {
+        value = initValue
+    }
+
+    func onSecuritySettingsChanged(newValue: Int) {
+        value = newValue
+    }
+}
```
zeitschlag, line 747: What's the reason this implementation of `SecuritySettingsDelegate` is a dedicated class, while `CertificateCheckDelegate` is implemented as extension of `AccountSetupController`?

zeitschlag (approved):
LGTM! I added a few questions, but feel free to ignore them :)

deltachat-ios/Controller/AccountSetup/SecuritySettingsController.swift
r10s: > Or is this part of the "this PR does not clean up"-part?

no longer :) thanks for the hint, i took the chance to tweak visibility and mutability

deltachat-ios/Controller/AccountSetup/SecuritySettingsController.swift
r10s: i added a commit that lets security-option be `Int32` throughout `SecuritySettingsController`; a cast is then needed when it comes to set/get-config then, however, that seems to be a more logical stage

deltachat-ios/Controller/AccountSetup/SecuritySettingsController.swift
r10s: that :)

but it also makes sense: when the database holds some weird value that does not map to an existing option (maybe it was sth. removed - or sth. comes from a backup from eg. android with a newer corer) - we select just the first item - which is the default resp. "Automatic"

deltachat-ios/Controller/AccountSetup/AccountSetupController.swift
r10s: we need two different delegates for `SecuritySettingsDelegate` - i did not find out a smart way to do that as an extension. one could add some state to the delegate - but that seems to work against the idea of the delegate

r10s:
thanks a lot for approval - and also, let me note, that the notes and comments are very welcome and on point, thanks for that as well 🙏

PR #2249, opened 2024-07-25
Refactor Alert-handling for events

zeitschlag:
Refactor Progress-alert-handling.

The original idea was to get rid of the `removeObservers` in `viewDidDisappear`-dance. This PR addresses the not-addressed leftovers from #2236 which is the Progress-handling. I tested it but a Testflight woulnd't hurt.

On another note: I gave up on the `WelcomeViewController` as it has a custom implementation of the alert-handling: Everywhere else the `ProgressAlertHandler` takes care of everything: Listening to the events, presenting the alert, updating the alert, whereas in `WelcomeViewController` multiple alerts are managed. So here my approach failed.


deltachat-ios/Handler/ProgressAlertHandler.swift
```diff
@@ -1,94 +1,119 @@
...
-        let progressAlert = makeProgressAlert(dcContext: dcContext)
-        progressAlert.actions[0].isEnabled = true
-        progressAlert.title = title
-        progressAlert.message = String.localized("one_moment")
-        self.present(progressAlert, animated: true, completion: nil)
-        self.progressAlert = progressAlert
+    let dcAccounts: DcAccounts
+    weak var dataSource: ProgressAlertHandlerDataSource?
+    var onSuccess: (() -> Void)?
+    let checkForInternetConnectivity: Bool
```
r10s, line 19:
```suggestion
    private var onSuccess: (() -> Void)?
    private let checkForInternetConnectivity: Bool
```

deltachat-ios/Handler/ProgressAlertHandler.swift
```diff
@@ -1,94 +1,119 @@
...
+class ProgressAlertHandler {
 
-    func showProgressAlert(title: String, dcContext: DcContext) {
-        let progressAlert = makeProgressAlert(dcContext: dcContext)
-        progressAlert.actions[0].isEnabled = true
-        progressAlert.title = title
-        progressAlert.message = String.localized("one_moment")
-        self.present(progressAlert, animated: true, completion: nil)
-        self.progressAlert = progressAlert
+    let dcAccounts: DcAccounts
```
r10s, line 16:
```suggestion
    private let dcAccounts: DcAccounts
```

r10s (approved):
code lgtm, also tested a bit :)

PR #2258, opened 2024-08-02
send images as sticker not only when they have the 140x140 dimensions, but also when they have transparent corner(s)

Simon-Laux:
also add the send as sticker logic to darg and droping an image into a chat, this makes it super convinient to create stickers from images with iOS's AI selection feature.
(long press to select your favorite pet or person then hold the object and use a second finger to navigate to the chat you want to send it to.)

### DIFF
--- a/DcCore/DcCore/Extensions/UIImage+Extensions.swift
+++ b/DcCore/DcCore/Extensions/UIImage+Extensions.swift
@@ -53,5 +53,40 @@ public extension UIImage {
       guard let alpha: CGImageAlphaInfo = self.cgImage?.alphaInfo else { return false }
       return alpha == .first || alpha == .last || alpha == .premultipliedFirst || alpha == .premultipliedLast
     }
+    
+    func hasTransparentCorner() -> Bool {
+        if !self.isTransparent() {
+            return false
+        }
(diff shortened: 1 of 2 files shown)


Simon-Laux:
dropping the image to send as sticker is still unreliable and for some images there is still a black background. so not ready to be merged.

Simon-Laux:
> I tested 2 images, the bigger one was correctly converted to png, the smaller one was converted to jpg and lost transparency

meanwhile I tested more images (create a sticker with iOS AI selection), and I haven't found another image yet where the black background appeared. It she not seem to be realtime to image dimensions, because I tried smaller images and they worked just fine.

Furthermore the transparency is preserved when I drop the image in question (with the black bg) into other apple apps, like Notes.
I suspect there is some conversion going on either in core or in deltachat-ios code before giving core the image, I haven't looked at the whole code path/flow yet (how do images get to core and treated there).
The files app says it's HEIF files, but when I send them to deltachat desktop the good ones arrive as png, while the bad one is a jpg. So maybe this HEIF issue in desktop is related? https://github.com/deltachat/deltachat-desktop/issues/3879

r10s:
i referenced this PR from the corresponding issues

PR #2261, opened 2024-08-09
Scroll to bottom (#2260)

### LINKED ISSUE #2260: Chat doesn't scroll to bottom when I send a message
> another bug: When I sent a message it does not scroll down to it when i have the message list scrolled down to the bottom .

-------

Most probably somethingsomething with threads

zeitschlag:
The culprit was `let numberOfRows = self.tableView.numberOfRows(inSection: 0)` it seems. (My educated guess is that `tableView` wasn't ready.) From now on it checks for `messageIds.count`

Also: Cleanup (some `self` here, but not there, but for incoming message the same happens as for outgoing messages)

Closes #2260.

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -470,12 +470,13 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


r10s (approved):
works for me, the bug is gone!

r10s (approved):

PR #2264, opened 2024-08-09
mark bots as such, improve chat-deletion confirmation

r10s:
this PR brings two recent, minor android tweak to iOS:

- mark bots in chat titles and profiles as such (see image 1+2, https://github.com/deltachat/deltachat-android/pull/3220 on android)

- if only one chat is about to being deleted, ask for confirmation using the chat title, not the number (for multiple selection, we things are unchanged) (see image 3, https://github.com/deltachat/deltachat-android/pull/3219 on android)

### DIFF
--- a/deltachat-ios/Controller/ChatListViewController.swift
+++ b/deltachat-ios/Controller/ChatListViewController.swift
@@ -746,11 +746,17 @@ class ChatListViewController: UITableViewController {
         if selected == 0 {
             return
         }
-        let alert = UIAlertController(
-            title: nil,
-            message: String.localized(stringID: "ask_delete_chat", parameter: selected),
-            preferredStyle: .safeActionSheet
-        )
+
+        let message: String
+        if selected == 1,
+           let chatIds = viewModel?.chatIdsFor(indexPaths: tableView.indexPathsForSelectedRows),
+           let chatId = chatIds.first {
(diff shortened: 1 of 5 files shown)


deltachat-ios/Controller/ChatListViewController.swift
```diff
@@ -746,11 +746,17 @@ class ChatListViewController: UITableViewController {
...
             return
         }
-        let alert = UIAlertController(
-            title: nil,
-            message: String.localized(stringID: "ask_delete_chat", parameter: selected),
-            preferredStyle: .safeActionSheet
-        )
+
+        let message: String
+        if selected == 1,
```
zeitschlag, line 751: Maybe it'd make sense to put these magic numbers into more verbose constants?

zeitschlag (approved):
LGTM.

What's the reason for these two changes (that, at least from my perspective, are independent from each other) come in one PR?
Usually I'd have expected two changes to be two seperate PRs.

deltachat-ios/Controller/ChatListViewController.swift
r10s: it is not a magic number, it is just a counter. but the variable name could be more verbose

zeitschlag: Ah! Now it makes sense 🙈

r10s: i changed the variable name from `selected` to `selectedCount`, that should make things clearer

PR #2265, opened 2024-08-10
v1.46.9 (Second Try)

zeitschlag:
- Bumped build to `109` and updated Changelog
- Added Markdown-files to Xcode-project (for convenience)
- Moved Build-number to Project-level (in Build settings), so that we only have to change it once. We could even go a step further and use xcconfig-files.


r10s (commented):
> Moved Build-number to Project-level (in Build settings), so that we only have to change it once

cool! we should update RELEASE.md at some point as well then

for the version number: let's go for 1.46.9, this makes communicating with testers easier as 1.46.8 has bugs

PR #2266, opened 2024-08-10
1 46 9

zeitschlag:
#2265 but with new version number

PR #2267, opened 2024-08-10
version numbers are on project-level now

r10s:
@zeitschlag did i got that correctly? that with #2265 resp. #2266 the version number needs to be changed only once in the project settings?

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -32,12 +32,9 @@ then, create a "bump-to-VERSION" PR:
 
 in Xcode:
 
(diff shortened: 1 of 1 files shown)


zeitschlag (approved):
LGTM. Maybe it'd make sense to add a little screenshot?

Screenshot 2024-08-11 at 10 58 17

r10s:
i personally always use "vim" for going through RELEASE.md, so i would not push for a screenshot :)

but i would not be against it  :)

PR #2268, opened 2024-08-11
Show contact when tapping on their name in Reactions-overview

### LINKED ISSUE #2259: make reactions details tappable
when tapping a contact in the reaction details, the contact's profile should open as on android.

cmp. https://github.com/deltachat/deltachat-android/pull/3207

zeitschlag:
When a user taps on the reactions next to a message, they see how (other) users reacted. When they tap on one of the user names, that user's profile opens (except it's me, than nothing happens)

Closes #2259

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -2842,8 +2842,6 @@ extension ChatViewController: SendContactViewControllerDelegate {
 
 extension ChatViewController: ReactionsOverviewViewControllerDelegate {
     func showContact(_ viewController: UIViewController, with contactId: Int) {
-        // dismiss, but push contact-screen
-
         viewController.dismiss(animated: true)
 
         let contactDetailController = ContactDetailViewController(dcContext: dcContext, contactId: contactId)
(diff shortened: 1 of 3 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -2836,3 +2837,16 @@ extension ChatViewController: SendContactViewControllerDelegate {
         stageVCard(url: vcardURL)
     }
 }
+
+// MARK: - ReactionsOverviewViewControllerDelegate
+
+extension ChatViewController: ReactionsOverviewViewControllerDelegate {
+    func showContact(_ viewController: UIViewController, with contactId: Int) {
+        // dismiss, but push contact-screen
+
```
r10s, line 2846: comments, that just repeat _what_ happens in the next lines, are usually superfluous, as this is obvious.

commenting the _why_ is usually better, but well, in this case it seems not needed as well, we're not doing that at similar places as well.

therefore i would just remove the comment. so it also cannot get stale on code changes.

```suggestion
```

r10s (approved):
very nice! code lgtm and also tested already a bit.

i would only remove the comment

deltachat-ios/Chat/ChatViewController.swift
zeitschlag: ah, damn, it's a leftover from development, I just didn't remove it. Busted!

Good, that you caught that.

PR #2270, opened 2024-08-19
Update Message-timestamp-updating

### LINKED ISSUE #2220: message bubble timestamp sometimes updated too late
sometimes, message bubble timestamp are updated too late:

- the timestamps in the message bubbles are updated every minute by a timer
- the timer seems to be stopped when the app goes to background and the app gets suspended; this stopping is probably okay
- when the app comes back from suspension, the timer is started - however not executed immediately, but only after a minute

-> this leads to delayed update; the app shows outdated timestamps for a minute.

things are fine when the app gets exited in between - or if the chat is opened.

things are also fine for all cases for the chatlist, the issue at hand seems to be fixed explicitly at https://github.com/deltachat/deltachat-ios/pull/1554

zeitschlag:
- Timer moved from `ChatViewController` to `BaseMessageCell`: There's no need to reload the entire table just to update a label.
- Timer doesn't get stopped when app gets sent to background or restarted when app returns to foreground, they just continue to run once per minute updating the labels
- Closes #2220.

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -101,8 +101,6 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
         return false
     }
 
-    private weak var timer: Timer?
-
     private lazy var navBarTap: UITapGestureRecognizer = {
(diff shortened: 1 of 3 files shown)


r10s (approved):
very elegant, removed lines, less complexity and even faster 👌

PR #2271, opened 2024-08-20
Share log as file instead of just text

### LINKED ISSUE #2205: log should better be shared as a file
with https://github.com/deltachat/deltachat-ios/pull/2126 we have the option to share a log, which is great!

however, the log is shared as text, which is unwieldy and error prone. we went through that on android recently :)

better share the log as a file, this allows the user also to take some notes - wrt editing: this is possible sufficiently beforehand

zeitschlag:
When sharing, log-text and notes get written into a file `log.txt` which gets shared then.

Closes #2205.

### DIFF
--- a/DcCore/DcCore/Helper/DatabaseHelper.swift
+++ b/DcCore/DcCore/Helper/DatabaseHelper.swift
@@ -16,7 +16,12 @@ public class DatabaseHelper {
     }
 
     var localDbFile: String {
-        return FileManager.default.localDocumentsDir.appendingPathComponent("messenger.db").path
+        return localDocumentsDir.appendingPathComponent("messenger.db").path
+    }
+
+    var localDocumentsDir: URL {
+        let paths = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true)
+        return URL(fileURLWithPath: paths[0], isDirectory: true)
     }
 
     public var unmanagedDatabaseLocation: String? {
--- a/DcCore/DcCore/Helper/DatabaseHelper.swift
+++ b/DcCore/DcCore/Helper/DatabaseHelper.swift
@@ -5,6 +5,8 @@ public class DatabaseHelper {
     /// The ID is created in the apple developer portal and can be changed there.
     static let applicationGroupIdentifier = "group.chat.delta.ios"
 
+    public init() {}
+
     public var sharedDbFile: String {
         guard let fileContainer = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: DatabaseHelper.applicationGroupIdentifier) else {
             return ""
@@ -14,12 +16,7 @@ public class DatabaseHelper {
     }
 
     var localDbFile: String {
-        return localDocumentsDir.appendingPathComponent("messenger.db").path
-    }
-
-    var localDocumentsDir: URL {
-        let paths = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true)
-        return URL(fileURLWithPath: paths[0], isDirectory: true)
+        return FileManager.default.localDocumentsDir.appendingPathComponent("messenger.db").path
     }
 
     public var unmanagedDatabaseLocation: String? {
@@ -31,7 +28,4 @@ public class DatabaseHelper {
         }
         return nil
     }
-
-    public init() {}
-
 }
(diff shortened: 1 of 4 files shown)


deltachat-ios/Helper/Utils.swift
```diff
@@ -82,16 +82,25 @@ struct Utils {
...
+
+        Utils.share(url: url, parentViewController: parentViewController, sourceItem: sourceItem)
     }
 
     public static func share(text: String, parentViewController: UIViewController, sourceItem: UIBarButtonItem) {
-        let activityVC = UIActivityViewController(activityItems: [text], applicationActivities: nil)
-        activityVC.popoverPresentationController?.barButtonItem = sourceItem // iPad crashes without a source
+        guard let data = text.data(using: .utf8) else { return }
+
+        let tempLogfileURL = FileManager.default.localDocumentsDir.appendingPathComponent("log.txt")
```
r10s, line 93: why not use `temporaryDirectory`, that seems to be a better fit that saving the file "forever" in the app library directory beside the database. 

```suggestion
        let tempLogfileURL = FileManager.default. temporaryDirectory.appendingPathComponent("deltachat-log.txt")
```

i also made the file name a bit more verbose (this is also the prefix we use on android)

r10s (commented):
sth. is wrong with the encoding, when trying to save log.txt to the "Notes App", all unicode characters are broken:



which is weird, as `.utf8` is stated explicitly

deltachat-ios/Helper/Utils.swift
zeitschlag: Good points, thank you! 

Using the temporary directory for that just didn't come to my mind, but you're right: It's a better fit.

zeitschlag:
This is interesting, as when I open `deltachat-log.txt` on a Mac with Preview.app or BBEdit or TextEdit.app, I see the emojis:

Screenshot 2024-08-21 at 10 07 34

So at this point I'd blame Notes or Safari on iOS?

deltachat-ios/Helper/Utils.swift
```diff
@@ -82,16 +82,25 @@ struct Utils {
...
-        }
+        guard let url = URL(string: url) else { return }
+
+        Utils.share(url: url, parentViewController: parentViewController, sourceItem: sourceItem)
     }
 
     public static func share(text: String, parentViewController: UIViewController, sourceItem: UIBarButtonItem) {
-        let activityVC = UIActivityViewController(activityItems: [text], applicationActivities: nil)
-        activityVC.popoverPresentationController?.barButtonItem = sourceItem // iPad crashes without a source
+        guard let data = text.data(using: .unicode) else { return }
```
r10s, line 91: `.unicode` works in general, however, generates UTF-16/USC-2 instead of UTF-8. this can result in issues on the receiving side, as UTF-16/USC-2 is not as widely supported as UTF-8.

however, this is no longer a blocker, i'll have a rought look if there isn't just a BOM missing or so.

DcCore/DcCore/Helper/DatabaseHelper.swift
```diff
@@ -14,12 +16,7 @@ public class DatabaseHelper {
     }
 
     var localDbFile: String {
-        return localDocumentsDir.appendingPathComponent("messenger.db").path
-    }
-
-    var localDocumentsDir: URL {
-        let paths = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true)
-        return URL(fileURLWithPath: paths[0], isDirectory: true)
+        return FileManager.default.localDocumentsDir.appendingPathComponent("messenger.db").path
```
r10s, line 19: this unrelated change,  touching a _very_ sensible area,  is no longer needed. i think, the additional abstraction is not really helpful without code being reused.

but no blocker.

r10s (approved):
please double check if we still want to have the unrelated changes in opening main database in.

otherwise lgtm

deltachat-ios/Helper/Utils.swift
r10s: indeed, when adding a BOM, UTF-8 works as expected.

background: BOM (Byte-Order-Mark) is not _needed_ for an UTF-8 file, however, many apps and libs need that at the beginning of a UTF-8 file to identify it as such (otherwise, a UTF-8 file content looks same to an ASCII/ANSI file ... and, without having metadata, can only be distinguished by some heuristics - which some libs do and others do not)

```suggestion
        guard var data = text.data(using: .utf8) else { return }
        let bom = Data([0xEF, 0xBB, 0xBF])
        data.insert(contentsOf: bom, at: 0)
```

with this change, UTF-8 works in "Notes" pad as well as in "WhatsApp"

PR #2282, opened 2024-08-21
Hide email in "New chat"-list for trusted contacts

### LINKED ISSUE #2272: remove email address from "new chat" screen
the email address should only be shown when the contact is not verified (has no green checkmark), same as in chat titles

zeitschlag:
Subtitle-Label gets hidden if `chat.isProtected`.
I'm not super happy with hiding the `subtitleLabel` but it's one of two reasonable approaches:

1. Hide the subtitleLabel
2. Override the `subtitle` of the ViewModel to be `nil`

Another option would have been to cleanup and declutter the `ViewModel`-stuff but this wouldn't have been pragmatic or reasonable.

Closes #2272.

### DIFF
--- a/deltachat-ios/View/ContactCell.swift
+++ b/deltachat-ios/View/ContactCell.swift
@@ -209,6 +209,12 @@ class ContactCell: UITableViewCell {
         fatalError("init(coder:) has not been implemented")
     }
 
+    override func prepareForReuse() {
+        super.prepareForReuse()
+
+        subtitleLabel.isHidden = false
+    }
+
     private func setupSubviews() {
         let margin: CGFloat = 10
         isAccessibilityElement = true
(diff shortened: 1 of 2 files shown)


deltachat-ios/View/ContactCell.swift
```diff
@@ -401,6 +407,14 @@ class ContactCell: UITableViewCell {
         case .contact(let contactData):
             let contact = cellViewModel.dcContext.getContact(id: contactData.contactId)
             titleLabel.attributedText = cellViewModel.title.boldAt(indexes: cellViewModel.titleHighlightIndexes, fontSize: titleLabel.font.pointSize)
+
+            if let chatId = contactData.chatId {
+                let chat = cellViewModel.dcContext.getChat(chatId: chatId)
+                if chat.isProtected {
+                    subtitleLabel.isHidden = true
+                }
```
r10s, line 415: contacts may have no chat assigned yet (the "New Chat" dialog is to create a chat in exact that case).

therefore, we need to check if a contact is verified - the same way, we show the green checkmark. this also results in email and green-checkmark being mutually exclusive.

```suggestion
            if contact.isVerified {
                subtitleLabel.isHidden = true
```

this makes things even easier and saves one db read :)

sorry to be unclear here, while writing the issue i thought it is the same condition as for the chat-titlebar (there, we know the chat exist - but the chat may be sth. else than a contact)

r10s (approved):
lgtm - and hiding the subtitle makes sense to me as well :)

for the condition, see below, that needs to be tweaked.

deltachat-ios/View/ContactCell.swift
zeitschlag: Thanks a lot for pointing this out! I didn't really think about it as you were so convinced and it didn't _feel_ wrong to show the email even for contacts I don't have a verified chat with. Cause you know why should I trust people just cause we're both in a shared group?

I fixed it!

PR #2284, opened 2024-08-21
"Invite friends" from settings and "New Chat"

### LINKED ISSUE #2277: add "invite friends" option
add an "invite friends" option to the "settings" screen, probably above "help", as a separate section.

icon could be a heart.

function would be similar to "share invite link", but with a little text prepended (use `invite_friends_text`)

compare https://github.com/deltachat/deltachat-android/pull/3196

zeitschlag:
Add options to invite your friends to Settings and when you want to start a new chat. Looks like this and closes #2277

### DIFF
--- a/deltachat-ios/Controller/NewChatViewController.swift
+++ b/deltachat-ios/Controller/NewChatViewController.swift
@@ -15,8 +15,9 @@ class NewChatViewController: UITableViewController {
     private let newOptions: [NewOption]
 
     private let sectionNew = 0
-    private let sectionContacts = 1
-    private let sectionsCount = 2
+    private let sectionInviteFriends = 1
+    private let sectionContacts = 2
+    private let sectionsCount = 3
 
     private lazy var searchController: UISearchController = {
         let searchController = UISearchController(searchResultsController: nil)
@@ -100,6 +101,14 @@ class NewChatViewController: UITableViewController {
     }
 
     // MARK: - actions
+
+    private func inviteFriends(cell: UITableViewCell) {
+        guard let inviteLink = Utils.getInviteLink(context: dcContext, chatId: 0) else { return }
+
+        let invitationText = String.localized(stringID: "invite_friends_text", parameter: inviteLink)
+        Utils.share(text: invitationText, parentViewController: self, sourceView: cell)
+    }
+
     @objc func cancelButtonPressed() {
         dismiss(animated: true, completion: nil)
     }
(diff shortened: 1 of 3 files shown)


deltachat-ios/Controller/Settings/SettingsViewController.swift
```diff
@@ -38,7 +45,7 @@ internal final class SettingsViewController: UITableViewController {
         let cell = UITableViewCell(style: .value1, reuseIdentifier: nil)
         cell.tag = CellTags.chatsAndMedia.rawValue
         cell.textLabel?.text = String.localized("pref_chats_and_media")
-        if #available(iOS 16.0, *) {
+        if #available(iOS 13.0, *) {
```
r10s, line 48: we're using the "minimum required os version fitting all icons"  on purpose. otherwise, eg. on iOS 13 the dialog would end up in having _some_ icons, which looks like a bug. better just show no icons - which is not even noticeable as this [was the default in older Delta Chat versions](https://github.com/deltachat/deltachat-ios/pull/1827) :)

but indeed, the "minimum required os version fitting all icons" would be 16.1

and maybe the comment can be clearer.

deltachat-ios/Controller/NewChatViewController.swift
```diff
@@ -15,8 +15,9 @@ class NewChatViewController: UITableViewController {
     private let newOptions: [NewOption]
 
     private let sectionNew = 0
-    private let sectionContacts = 1
-    private let sectionsCount = 2
+    private let sectionInviteFriends = 1
```
r10s, line 18: i am wondering if it isn't better to have "Invite Friends" at the end of the list:

- for new user without contacts or with few contacts it is as easy accessible as before

- however, when the list gets filled, it gets out of the way - the same time, the list is filled and "Invite Friends" obviously works. and, it is still clear where the option is: at the end of the list

r10s (approved):
tested a bit, things work as expected.

the needs to be embraced by `if #available(iOS 16.1, *)` all together. otherwise lgtm

deltachat-ios/Controller/NewChatViewController.swift
zeitschlag: Thanks for your feedback, I thought about that, too, and will change it to be that way.

Originally I thought it might be a good idea to nudge people to invite their friends, but you're totally right: What happens if all their friends already use Deltachat? So I'll put it at the end :)

deltachat-ios/Controller/Settings/SettingsViewController.swift
zeitschlag: We could think about adding the icons from SF Symbols to the assets so that everyone can enjoy them?

Until then, icons are shown for iOS 16+ only 👍

PR #2288, opened 2024-08-22
improve "Edit Name" dialog

### LINKED ISSUE #2286: enhance "edit name" dialog
in the "edit contact name" dialog (accessible from "contact profile"), it is a bit unclear, how "Name" us used:

we prepared the following strings for enhancement (this requires the strings to be updated first):

- use string `edit_name_placeholder` as a placeholder for the input field ("test" in the screenshot)
- use `edit_name_explain` as text below or above the input field
- both strings take the result of [`dc_contact_get_auth_name()`](https://c.delta.chat/classdc__contact__t.html#a7778540e0592d4c7a480222ea30e9ac8) as a parameter. if empty, use [`dc_contact_get_addr()`](https://c.delta.chat/classdc__contact__t.html#a017fc5f3c27f52547c515fadc5658e01). do not use the "display name" [^1]

the email address and layout can stay as is.

cave: existing strings and placeholders should still be used when the dialog is used as "add contact manually" for non-chatmail accounts

compare https://github.com/deltachat/deltachat-desktop/issues/4090 and https://github.com/deltachat/deltachat-android/pull/3254

[^1]: the "display name" may be the nickname, this is fine nearly everywhere else and indeed the preferred method to display a name - but not in this special case :)

r10s:
this PR sets placeholder and footer according to suggestions at https://github.com/deltachat/deltachat-ios/issues/2286

moreover, the title is changed to "Edit Name" and the email address is hidden (as on android/desktop - the dialog focuses on the nickname)

i first tried to follow the way of subclassing NewContactController, however that became even more hacky as before (cells would needed to be var, first reponder check in base class would needed to be changed etc.). not worth few lines of saved code. 

therefore, i made `EditContactController` self-contained and independet. it is 20 lines more, but straight-forward and no hacks so one does not need to check what in the super-class happens all the time.

closes #2286

before / after - note the improved placeholder and footer and the general cleanup :)

### DIFF
--- a/deltachat-ios/Controller/EditContactController.swift
+++ b/deltachat-ios/Controller/EditContactController.swift
@@ -19,6 +19,7 @@ class EditContactController: UITableViewController {
         nameCell.textField.text = dcContact.editedName
         nameCell.textField.enablesReturnKeyAutomatically = false
         nameCell.textField.returnKeyType = .done
+        nameCell.useFullWidth()
         nameCell.placeholder = String.localizedStringWithFormat(String.localized("edit_name_placeholder"), authNameOrAddr)
 
         title = String.localized("menu_edit_name")
--- a/deltachat-ios/Controller/EditContactController.swift
+++ b/deltachat-ios/Controller/EditContactController.swift
@@ -1,37 +1,64 @@
 import UIKit
 import DcCore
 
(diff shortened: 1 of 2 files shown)


zeitschlag (approved):
Nice!

PR #2295, opened 2024-09-03
Open i.delta.chat-links in App (#2134)

### LINKED ISSUE #2134: handle invite-links in-app
instead of opening the browser where the user has to tap "open chat", which brings you then back to delta chat, we should be able to handle the invite-links directly

for testing: eg. "QR Code / QR Invite Code / Copy to Clipboard", send the invite link to another account and tap it there.

corresponding android issue: https://github.com/deltachat/deltachat-android/pull/2949

zeitschlag:
When users tap on a `i.delta.chat` link in a chat (or in a user's signature), they can start a chat with that very user right away (instead of that Safari opens). Just like this:

IMG_F59507DB384E-1

Pressing on *Cancel* in that alert just closes the alert. No surprises here.
(The large number of lines comes from moving the QR-handling-code from `QrPageController` to `AppCoordinator`)

Closes #2134.


deltachat-ios/Helper/Utils.swift
```diff
@@ -42,12 +53,31 @@ struct Utils {
...
+        var qrcodeString = "OPENPGP4FPR:"
+        qrcodeString.append(String(fingerprint))
+        qrcodeString.append("#")
+
+        let queryParams = fragments.dropFirst().joined(separator: "&")
+        qrcodeString.append(queryParams)
+
+        return qrcodeString
+    }
+
```
r10s, line 74: this conversion is not needed - you can pass an i.delta.chat invite-link directly to handleQRCode() (as core already handles `OPENPGP4FPR:` and `https://i.delta.chat` equally; sorry for the additional work, i could have pointed that out clearer)

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -2411,6 +2411,11 @@ extension ChatViewController: BaseMessageCellDelegate {
         if Utils.isEmail(url: url) {
             let email = Utils.getEmailFrom(url)
             self.askToChatWith(email: email)
+        } else if url.isDeltaChatInvitation,
+                  let qrCode = Utils.makeDeltaChatInvitationQRCode(from: url),
+                  let appDelegate = UIApplication.shared.delegate as? AppDelegate,
+                  let appCoordinator = appDelegate.appCoordinator {
+            appCoordinator.handleQRCode(qrCode)
```
r10s, line 2418: when hitting "Cancel" in the question, the keyboard will be gone:

https://github.com/user-attachments/assets/f8e1e7af-027d-48ab-89b7-f7a7d5032d7d

r10s (commented):
beside the two comments that need targeting, works as expected!

thanks a lot for taking care!

deltachat-ios/Helper/Utils.swift
zeitschlag: That's good to know, thank you!

deltachat-ios/Chat/ChatViewController.swift
zeitschlag: The first thing that comes to mind is to inject for an optional closure to `handleQRCode` of `QrPageController` what to do when tapping the cancel-button.
This also means forwarding it everywhere

r10s (approved):

r10s (approved):
lgtm, thanks a lot for diving in that!

Issue #2280, opened 2024-08-20
Show number of unread messages in Back-button on ChatViewController


r10s:
... if > 0

when Alices chats in "Chat with Bob" for hours, and a new message arrives in "Chat with Claire", Alice won't be aware of Claire's messages until she goes back to the chatlist at some point.

on android, we solved this issue by showing the "normal" notifications in case a message arrives in a foreign chat while being in another chat. this probably requires to track the chat_id in-app (the "Notification Service Extension" (NSE)  is not envolved when the app is in foreground iirc, but not sure) (~~ftr~~ for the record: adding flags to `completionHandler([.badge])` sounds promising)

these additional notification must not be shown when already being in the chatlist (we have the badge counters there) and are also not needed from inside the settings or other screens (either the badge counter abottom is visible - and, usually, one is not that often and for hours in these screens)

if this is not possible at all, an alternative is to show the unread-badges atop of the in-chat back-button

zeitschlag:
If we'd show a TabBar-item indicator, we show the number of unread messages in the back-button of a chat

PR #2303, opened 2024-09-20
Show number of unread messages in back-button (#2280)

zeitschlag:
- These changes are iOS 13+ only.
- Adds an indicator with the number of unread messages to the back-button in chat if there are unread messages.

Simulator Screenshot - iPhone 16 Pro - 2024-09-23 at 12 24 58

- If you have more than 50 unread messages, you'll just see a dot.
- ~~Also: Fix build Xcode 18~~

### DIFF
--- a/DcCore/DcCore/DC/events.swift
+++ b/DcCore/DcCore/DC/events.swift
@@ -105,6 +105,7 @@ public class DcEventHandler {
             
             NotificationCenter.default.post(name: Event.messagesNoticed, object: nil, userInfo: [
                 "chat_id": Int(data1),
+                "account_id": accountId
             ])
 
         case DC_EVENT_CHAT_MODIFIED:
@@ -129,7 +130,11 @@ public class DcEventHandler {
 
         case DC_EVENT_INCOMING_MSG:
             
-            NotificationCenter.default.post(name: Event.incomingMessageOnAnyAccount, object: nil)
+            NotificationCenter.default.post(name: Event.incomingMessageOnAnyAccount, object: nil, userInfo: [
+                "chat_id": Int(data1),
+                "account_id": accountId
+            ])
+            
             if accountId != dcAccounts.getSelected().id {
                 return
             }
@@ -138,6 +143,7 @@ public class DcEventHandler {
             NotificationCenter.default.post(name: Event.incomingMessage, object: nil, userInfo: [
                 "message_id": Int(data2),
                 "chat_id": Int(data1),
+                "account_id": accountId
             ])
 
         case DC_EVENT_CONTACTS_CHANGED:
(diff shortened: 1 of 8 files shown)


zeitschlag:
The current implementation doesn't show the number in case the account is muted. This kind of feels like a bug? So I was thinking about adding this additional piece. WDYT @r10s?

```patch
diff --git a/deltachat-ios/Controller/ChatListViewController.swift b/deltachat-ios/Controller/ChatListViewController.swift
index 64eadc64..0bfa5b03 100644
--- a/deltachat-ios/Controller/ChatListViewController.swift
+++ b/deltachat-ios/Controller/ChatListViewController.swift
@@ -253,7 +253,7 @@ class ChatListViewController: UITableViewController {
     }
 
     private func updateNextScreensBackButton(accountId: Int? = nil, chatId: Int? = nil) {
-        let numberOfUnreadMessages = DcAccounts.shared.getFreshMessageCount()
+        let numberOfUnreadMessages = DcAccounts.shared.getAllFreshMessages()
 
         if isArchive {
             navigationItem.backBarButtonItem = nil
diff --git a/deltachat-ios/DC/DcAccount.swift b/deltachat-ios/DC/DcAccount.swift
index ff1cbe9a..fca72f7f 100644
--- a/deltachat-ios/DC/DcAccount.swift
+++ b/deltachat-ios/DC/DcAccount.swift
@@ -129,6 +129,17 @@ public class DcAccounts {
         return freshCount
     }
 
+    public func getAllFreshMessages() -> Int {
+
+        let accountIDs = getAll()
+        let numberOfFreshMessages = accountIDs
+            .compactMap { get(id: $0) }
+            .compactMap { $0.getFreshMessages().count }
+            .reduce(0) { $0 + $1 }
+
+        return numberOfFreshMessages
+    }
+
     @discardableResult
     public func blockingCall(method: String, params: [AnyObject]) -> Data? {
         if let paramsData = try? JSONSerialization.data(withJSONObject: params),
```

Amzd:
> The current implementation doesn't show the number in case the account is muted

I think that would be expected. If you mute eg a group chat you don't want it still to show up in the back button because that would hide wether there is actual messages that you need to look at.

r10s (approved):
i am using this PR since quite some time, and it is really great!

i am still a bit unsure about the displaying of "50 max" - discussed alternatives are
(1) using a normal number (this is what whatsapp is doing) or 
(2) do not show a number at all but show a "unread dot" only if sth. new comes in or 
(3) use a "real" badge (this is more effort, and we should really know that we want the badge to stay

however, let's stay with the current state, maybe it is just fine and good enough :)

wondering, btw about the reasoning of using updateNextScreensBackButton() to pass the event from ChatListViewController to ChatViewController. is that because of improved performance as when using an event in ChatViewController, one has to calculate things again?

PR #2309, opened 2024-10-02
add 'learn more' button for managing keys

r10s:
this PR makes the help - that describes key format, supported keys etc. - accessible directly from the "manage keys" section.

before, these information were harder to find.

Screenshot 2024-10-02 at 19 44 52

the learn more buttons links to https://delta.chat/en/help#can-i-reuse-my-existing-private-key in the offline help (which is shipped with the app, no internet required)

came over this by the desktop issue  https://github.com/deltachat/deltachat-desktop/issues/4155 ,  counterpart of https://github.com/deltachat/deltachat-android/pull/3328

### DIFF
--- a/deltachat-ios/Controller/Settings/AdvancedViewController.swift
+++ b/deltachat-ios/Controller/Settings/AdvancedViewController.swift
@@ -398,7 +398,8 @@ internal final class AdvancedViewController: UITableViewController {
             self.present(alert, animated: true, completion: nil)
         }))
 
-        alert.addAction(UIAlertAction(title: String.localized("learn_more"), style: .default, handler: { _ in
+        alert.addAction(UIAlertAction(title: String.localized("learn_more"), style: .default, handler: { [weak self] _ in
+            guard let self else { return }
             self.navigationController?.pushViewController(HelpViewController(dcContext: self.dcContext, fragment: "#importkey"), animated: true)
         }))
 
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/Settings/AdvancedViewController.swift
```diff
@@ -398,6 +398,10 @@ internal final class AdvancedViewController: UITableViewController {
             self.present(alert, animated: true, completion: nil)
         }))
 
+        alert.addAction(UIAlertAction(title: String.localized("learn_more"), style: .default, handler: { _ in
+            self.navigationController?.pushViewController(HelpViewController(dcContext: self.dcContext, fragment: "#importkey"), animated: true)
```
zeitschlag, line 402: What we could do prevent a retain cycle is to use a `[weak self]` and some `guard let` here. I don't consider this a blocker, just had to think about #2235

zeitschlag (approved):
LGTM!

adbenitez (approved):

deltachat-ios/Controller/Settings/AdvancedViewController.swift
r10s: i did a commit, thanks for the pointer.

it seems wise in general to prefer weak self for "back pointers" if in doubt, for the other occurances, however, that should go to a separate PR

PR #2310, opened 2024-10-05
 do not expect special value for backup transfer progress

r10s:
'backup' and 'add second device' do not longer have special values but use the whole range smoothly

counterpart of https://github.com/deltachat/deltachat-android/pull/3333

### DIFF
--- a/deltachat-ios/Controller/AccountSetup/WelcomeViewController.swift
+++ b/deltachat-ios/Controller/AccountSetup/WelcomeViewController.swift
@@ -216,13 +216,9 @@ class WelcomeViewController: UIViewController {
         } else {
             guard let permille = ui["progress"] as? Int else { return }
             var statusLineText = ""
-            if permille <= 100 {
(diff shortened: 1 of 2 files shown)


deltachat-ios/Controller/AccountSetup/WelcomeViewController.swift
```diff
@@ -216,13 +216,9 @@ class WelcomeViewController: UIViewController {
         } else {
             guard let permille = ui["progress"] as? Int else { return }
             var statusLineText = ""
-            if permille <= 100 {
-                statusLineText = String.localized("preparing_account")
-            } else if permille <= 950 {
-                let percent = ((permille-100)*100)/850
+            if permille < 1000 {
+                let percent: Int = permille/10
```
r10s, line 220: here, permille 0 and 1000 are handled by special booleans 'done' and 'error' above; that could get some streamlining as well, but i wanted to keep this PR on point

Issue #2311, opened 2024-10-09
Input-textfield shows up in chat when tapping on a vCard


zeitschlag:
1. Go to a chat with a sent contact (vCard)
2. Tap on the input-textfield to enter a chat message
3. Tap on that vCard.

**Expectation**

I see the alert to start a chat with that person

**Reality**

I see the alert to start a chat with that person but also the Input-textfield:

![IMG_4942](https://github.com/user-attachments/assets/02f100a1-acc5-44af-837a-d579763d48d6)

**Notes**

Somethingsomething `firstResponder` and `accessoryView`.

Issue #2307, opened 2024-09-29
After opening a chat, the beginning of the chat history is shown and not the most recent part.


Raiden-GH:
**Expected behavior:**

After opening a chat, the most recent part of the chat history is shown. 

**Actual behavior:**

After opening a chat, the beginning of the chat history is shown.

This problem suddenly appeared and affected every profile. After I added an additional one in DC-iOS, the bug occurred there too. Then I deleted all profiles and added them again. But the problem still persisted. After deleting the app, restarting iOS and reinstalling DC-iOS, the problem no longer occurred.

iOS 18
iPhone 14
DC-iOS 1.46.10

Issue #2305, opened 2024-09-27
Unread messages not (always) on top?


zeitschlag:
When receiving a new message in `ChatListViewController` it occasionally seems to happen, that the unread chats don't move to the top. Need further investigation. I'm not sure if it's reproducible, could be connected to some Notification-refactoring in the near past?

Also (relative) timestamps are not updated.

r10s:
as this might have been fixed with #2315, this needs a retry

Issue #2233, opened 2024-07-11
Scroll-button is visible when scrolled to bottom


zeitschlag:
Scroll-bottom is visible for some moments when it shouldn't be.

1. Go to a long chat and scroll to the bottom
2. Receive a message

## Expectation

"Scroll to bottom"-button doesn't show up

## Result

"Scroll to bottom"-button shows up

## Notes

- `updateScrollDownButtonVisibility()` is invoked in several places, also in `UIScrollViewDelegate`-code

PR #2315, opened 2024-10-12
ScrollView jumping improvements

Amzd:
In this PR I flipped the scrollView so now new messages automatically cause the rest of the messages to move up so we no longer need to programmatically scroll down when messages are sent and/or received.

### Fixes:
- #2232
- #2307 - since the scrollview in this PR starts with the newest message at the top (virtual top, actual bottom, because the tableView is flipped) this issue will never occur because even if the vc doesn't scroll anywhere, it will be showing your latest messages.
- Probably: #2305
- #2311
- When sending a long message, the chat doesn't scroll down to bottom of last message.
- #2233
- Maybe: #2161

### Intended changes:
- Less twitching when receiving/sending messages
- Smooth scrolling when searching
- No longer show inputAccessoryView when presenting other views (without having to resign firstResponder)

### Side effects:
**Messages are not refreshed until user stops editing** 
This is because we rely on selectedIndexPaths to keep which messages are selected, which can get out of sync with the actual message list when new messages are received or when they are automatically deleted. This was already the case before with deleted messages (eg through auto deletion) but is now also the case for new messages because messages are now pushed at the beginning of the indexPaths instead of at the end (because tableView is flipped).

This can be mitigated by tracking selected messages instead of relying on selectedIndexPaths.

**Messages don't float to the top when you have only few messages**
Since tableView now starts at the bottom, when you only have like 2 messages they are at the bottom instead of at the top.

This can totally be mitigated with a bit more effort.

**Empty state view is a bit higher up**
I think this is due to safeArea change to tableView but it is not a bad change because now it is still visible when keyboard is up (on iPhone SE anyway)

### Notes:

- I feel like it would be better if we did not use inputAccessoryViews that switch between keyboard and view controller attachment but instead just a normal view which we animate to always stay above the keyboard. This would eliminate the need for much of the first responder juggling we do right now and also eliminate the edge case where the input view is hidden when you don't expect it.

- Would be really nice to have tests that can confirm if the keyboard comes back after actions when expected.

### DIFF
--- a/deltachat-ios.xcodeproj/project.pbxproj
+++ b/deltachat-ios.xcodeproj/project.pbxproj
@@ -131,6 +131,7 @@
 		30FDB70524D1C1000066C48D /* ChatViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 30FDB6F824D1C1000066C48D /* ChatViewController.swift */; };
 		30FDB71F24D8170E0066C48D /* TextMessageCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 30FDB71E24D8170E0066C48D /* TextMessageCell.swift */; };
 		30FDB72124D838240066C48D /* BaseMessageCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 30FDB72024D838240066C48D /* BaseMessageCell.swift */; };
+		5F785F6E2CB9344F003FFFB9 /* ReusableCellProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5F785F6D2CB9344F003FFFB9 /* ReusableCellProtocol.swift */; };
 		7070FB9B2101ECBB000DC258 /* NewGroupController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7070FB9A2101ECBB000DC258 /* NewGroupController.swift */; };
 		7092474120B3869500AF8799 /* ContactDetailViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7092474020B3869500AF8799 /* ContactDetailViewController.swift */; };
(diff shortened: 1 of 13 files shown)


Amzd:
I think this PR already fixed #2312 by disallowing first responder when a the chatvc is presenting another vc, which is the conflict.

r10s:
thanks a lot for diving into all that!

i already run the draft in its current state, and i have to say, that it improves things significantly already. wow!

zeitschlag:
> This PR already fixed #2312 by disallowing first responder when a the chatvc is presenting another vc, which is the conflict.

lmk if you want me to revert it.

Amzd:
> i am not sure what you mean here - i just tested receiving messages while using the input field - and that works as expected

I mean while selecting messages (long press message > more options)

> this is totally fine, and also how it was intended to work.

OK that's easy then. My reference (iMessage) keeps it at the top so I thought that was intended.

> please let us know, when it is ready-to-review - or if you have any questions

It is basically all I wanted to do for this PR, I opened it in draft because I wasn't sure if it was OK to create PRs of this size.

Amzd:
The image from the read me showing the chat screen is no longer accurate since this PR (since messages float down instead of up here)

Amzd:
[af65d6f](https://github.com/deltachat/deltachat-ios/pull/2315/commits/af65d6fabbfd12237491d9f2458586968386a296) broke first message scrolling while trying to prevent the empty state view from jumping up and down when scrollview appears.

Amzd:
Resolved

Amzd:
https://github.com/user-attachments/assets/0d8e585e-872a-4e5a-a416-c25d43050634

Amzd:
@r10s I don't have permissions to add reviewers

adbenitez:
I am **amazed** thanks a lot!!! :star_struck:

Amzd:
> I am **amazed** thanks a lot!!! :star_struck: 

Wait then who am I?

PR #2318, opened 2024-10-14
relicense to MPL and be a bit more specific which files are MPLed

hpk42:
anyone who provides code in PRs needs to be fine with this license
(so far all code except minor details was developed from merlinux-contracted or -affiliated people). 
If we kept the GPL, then we would need to ask contributors for a different additional license 
because Apple does not allow GPLed licensed code into the store. 

For a discussion of MPL versus GPL with apple app store see https://opensource.stackexchange.com/questions/1762/is-the-mozilla-public-license-compatible-with-apples-app-store

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,636 +1,378 @@
-# GNU GENERAL PUBLIC LICENSE
-Version 3, 29 June 2007
-
-Copyright (C) 2007 [Free Software Foundation, Inc.](http://fsf.org/)
-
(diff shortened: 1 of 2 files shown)


Amzd:
Interesting read. Had no idea.

r10s (approved):

Amzd (approved):

Simon-Laux (approved):

adbenitez (approved):

PR #2319, opened 2024-10-15
Fix background notifications

Amzd:
broken in #2245

to test:
- close app
- send message to account
- now: receive notification
- previously: only badge number updated


r10s (approved):
> send message to account

this has to happen within ~30 seconds. after that,  `deltachat-ios` is terminated and notifications are handled by `DcNotificationService` (out notification service extension), that does not have this bug

still, bad enough to not get banners but only have badge counters updated for 30 seconds.

very good catch, thanks a lot ❤️

PR #2322, opened 2024-10-15
Cached webxdc and enabled back swipe in webxdc apps

Amzd:


Amzd:
Well, since this PR was created there is now real-time by default. This PR doesn't deinit the webxdc when swipe back, only when you start a new webxdc or exit the chat is the webxdc closed. If you could test if this works fine then I think it can be merged.

This was created to forward the discussion on the issue but no one responded on it that's why it stayed in draft.

r10s:
k, then let's postpone this PR, shutting down is hard enough, it is unclear if we are fully there in the current state; adding a "pending webxdc" is then sth. later, when realtime shutdown has settled

putting to resurrection

PR #2325, opened 2024-10-16
clear chat: do not count or delete daymarker

### LINKED ISSUE #2324: clear chat: DC always indicates a wrong number of messages
### Expected behavior:

When clearing a chat history, DC displays the correct number of messages in the security prompt, which will be deleted after confirmation.

### Actual behavior:

When clearing a chat history, DC always displays one message too much in the security prompt.

### Steps to reproduce the problem:

- Start a new chat
- Write and send some messages, 
- Clear chat history

iOS 18.01
iPhone 14
DC-iOS 1.46.10

r10s:
this PR fixes the calculation of the number of messages to delete on "clear chat".

unconditionally adding the flag DC_GCM_ADD_DAYMARKER was just wrong.

closes #2324

### DIFF
--- a/deltachat-ios/DC/DcContext.swift
+++ b/deltachat-ios/DC/DcContext.swift
@@ -102,9 +102,9 @@ public class DcContext {
         return Int(dc_send_videochat_invitation(contextPointer, UInt32(chatId)))
     }
 
-    public func getChatMsgs(chatId: Int) -> [Int] {
+    public func getChatMsgs(chatId: Int, flags: Int32) -> [Int] {
         let start = CFAbsoluteTimeGetCurrent()
-        let cMessageIds = dc_get_chat_msgs(contextPointer, UInt32(chatId), UInt32(DC_GCM_ADDDAYMARKER), 0)
+        let cMessageIds = dc_get_chat_msgs(contextPointer, UInt32(chatId), UInt32(flags), 0)
         let diff = CFAbsoluteTimeGetCurrent() - start
         logger.info("⏰ getChatMsgs: \(diff) s")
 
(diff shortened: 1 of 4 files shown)


deltachat-ios/DC/DcContext.swift
```diff
@@ -102,9 +102,9 @@ public class DcContext {
         return Int(dc_send_videochat_invitation(contextPointer, UInt32(chatId)))
     }
 
-    public func getChatMsgs(chatId: Int) -> [Int] {
+    public func getChatMsgs(chatId: Int, flags: Int32) -> [Int] {
```
Amzd, line 105:
```suggestion
    public func getChatMsgs(chatId: Int, includingDayMarker: Bool = true) -> [Int] {
```
or 
```suggestion
    public func getChatMsgs(chatId: Int, options: GetChatMsgOptions = .includeDayMarker) -> [Int] {
```
I think the Swift API should be easily understandable from call site

Amzd (changes requested):

deltachat-ios/DC/DcContext.swift
r10s: using a boolean seems fine. 

for duplicating "everything" to a swift enum or sth - not sure, at least we did not do that in the past. unless generated automatically, that's a error-prone and/or need quite some maintenance - we're doing that for Java, as we have to - and are quite happy that you can use C from Swift as a first-class thing. in fact, iirc, not so far away we fixed bugs as the swift-enum differ from the C-ones ...

so, if in doubt, i can happily life with some C-constants - esp. as the writer of the glue code is the same as the caller of the glue code for the foreseeable future :)

also, the more far we go away from the C-bindings, the more there is a need for a dedicated documentation - up to now, we can just work with https://c.delta.chat

Amzd: Sure, I personally found the `flags: 0` not very self explaining which is why I commented, but you're the boss.

Amzd (approved):

PR #2331, opened 2024-10-20
notify reactions added

r10s:
this PR shows notifications for reactions to one's own messages.

requires a core with https://github.com/deltachat/deltachat-core-rust/pull/6072

this is how it looks like on the lock screen:

the accumulation is done when we discover multiple messages/reaction when processing a PUSH notification in the "notification service extension", where we can only modify the single, given notification, but not add additional ones (often, this accumulation is not needed, as PUSH notifications receive in time and one-per-one)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - offer to open http-links detected as proxy also in the browser (#6237)
+- Show notifications for reactions on own messages (#2331)
 
 
 ## v1.48.4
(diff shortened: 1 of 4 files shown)


r10s:
at the latest, when this gets merged, file issues for android/desktop

r10s:
we should not merge this PR until we have an answer to the multi-device question at https://github.com/deltachat/deltachat-android/pull/3377#pullrequestreview-2389859336

adbenitez (approved):

adbenitez (approved):

PR #2343, opened 2024-10-26
Input field missing after adding a reaction

### LINKED ISSUE #2342: input field missing after adding a reaction
input field is missing after adding a _reaction using "..."_ if the _keyboard was expanded_ before.

if not both conditions are met, things seem to be fine.

this sounds very similar to the "input field is missing after using camera if the _keyboard was not expanded_ before" fixed in #2323.

might still be a side effect of the ["ScrollView jumping improvements" PR](https://github.com/deltachat/deltachat-ios/pull/2315).

cc @Amzd

Amzd:
Fixes #2342 

Now the keyboard also comes back up after using context menu which I think is preferred but not how it worked in last version.

Note: I think this works differently on < iOS 13 but I can't test that as the lowest simulator I can use is iOS 15.


r10s (approved):
thanks a lot for fixing ❤️

PR #2346, opened 2024-10-26
Action bar missing in read only chat

### LINKED ISSUE #2345: action bar missing in read-only-chats and if keyboard is expanded
_this issue was introduced with <https://github.com/deltachat/deltachat-ios/issues/2342>_

in read-only-chats [^1], the action bar to delete/forward/copy/etc single or multiple messages is missing. usually, it appears when long-tapping a message and selecting "More Options".

same, when the keyboard is expanded before the message is long-tapped and "More Options" is selected.

this is how it should look:

this is how it actually looks:

 

that used to work in 1.46, and without #2342 so, another side effect of ["ScrollView jumping improvements" PR](https://github.com/deltachat/deltachat-ios/pull/2315) :)

cc @Amzd 

[^1]: read-only-chats are eg. some mailing lists, left groups and the "Device Chat" -- the latter is probably the easiest to try out as always there when a new test profile is created

Amzd:
Fixes #2345

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -158,14 +158,10 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
         set { customInputAccessoryView = newValue }
     }
 
-    /// Set this to false if you are doing something with the UI inside this view controller that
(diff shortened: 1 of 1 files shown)


r10s (approved):
thanks a lot for fixing! it works in my test ... :)

PR #2347, opened 2024-10-27
add 'select all' and 'mute' to chatlist

r10s:
this PR adds some features recently added to Android: 

-  'Mute/Unmute' chats directly from the chatlist; also in multi-select mode

-  'Select All' chats - so that you can eg. archive, mute, unmute or even delete all chats (deleting all chats sounds more dangerous as it is, there are quite some taps needed - it is easier to delete the whole app accidentally :)

moreover, the PR cleans up the old "mute dialogs", using the new approach.

of course, among other things, there are also ideas to a mute to the swipe-action as well, but this is another thing :)


Amzd:
Maybe this is a good PR to add two finger swipe selection? I was missing it already and it's basically just a more native way to select many items that many other apps already have. https://developer.apple.com/documentation/uikit/uitableviewdelegate/selecting_multiple_items_with_a_two-finger_pan_gesture

Amzd:
If you think this is the correct place for it you can cherry pick https://github.com/deltachat/deltachat-ios/commit/fbe791a68bcc320619791de95ffbf108fa821fd9

otherwise I will open a separate PR

r10s:
> otherwise I will open a separate PR

let's keep things on point, this is better for review and following.

so, i'd merge this PR as is, and we can do a separate PR for the two-finger-swipe (which, btw, i was not aware of at all - i was thinking all the time, signal does not even have a multi-select-mode! from known messengers, it seems to be supported by signal and iMessage only, so, i would not expect users to know about that - in contrast to long tap. but we can just do both, i think, this is also the idea)

Amzd:
The two finger swipe is supported by many of the native apps (Mail, Notes, Reminders, Photos). GitHub app supports it (when selecting lines to comment on in review). My recipe app supports it.

So it's not that rare and would be nice to have, and yea probably only used by people that know these kinds of hidden gestures but very useful.

I will open a separate PR

r10s:
let'smerge this one in then, to avoid conflicts.

PR #2349, opened 2024-10-27
implement multiple selection

Amzd:
Implemented multiple selection interaction for both chat list and chat messages 

https://github.com/user-attachments/assets/cb25546c-2703-4210-8582-798ce626f2bd


r10s (approved):
very nice improvement, with very few code!

great, thanks a lot for doing that and pointing to that - i was not aware that this exist (i also asked some young ppl, very much more used to  iphone, they were not aware - but after they learned it, they loved it :)

r10s:
cherry-pick and merge via https://github.com/deltachat/deltachat-ios/pull/2352

PR #2351, opened 2024-10-28
tweak chatlist's swipe gestures

r10s:
this PR makes **"mute chat" easier accessible**, by adding it directly to a swipe gesture.

this is also what many other messengers are doing.

as more than three swipe gestures per side are questionable, we moved "pin" to the left side, together with new **mark read** - also that is known by many other messengers and apps.

moreover, the PR sets **images** - ~~iOS _only_ renders the images if title+images are set, quite some other apps are adding the title via a special image (see eg. [here](https://forzfilm.medium.com/how-to-set-image-and-label-in-trailing-swipe-action-with-custom-font-ios11-4a49d4794669)), however, i think, it is fine to stay with what the system thinks is useful here - the images are quite self-explaining and "easy", using system-images helps on that. and in the action bar, we're using the same images. we can iterate on that, surely :)~~ EDIT: we decided to go for image+text 

this is how the swipe actions look like now:

what is also new, is that **context is preserved** when tapping an action that is followed by a dialog:

<details>
<summary>old view for comparison</summary>

what's also gone is the pretty bad green-red contrast :)

</details>

### DIFF
--- a/deltachat-ios/Helper/Utils.swift
+++ b/deltachat-ios/Helper/Utils.swift
@@ -53,6 +53,27 @@ struct Utils {
         return window?.safeAreaInsets.bottom ?? 0
     }
 
+    @available(iOS 13.0, *)
+    static func makeImageWithText(image: UIImage?, text: String) -> UIImage? {
+        guard let image = image?.withTintColor(UIColor.white) else { return nil }
+
+        let spacing: CGFloat = 4
+        let textAttributes: [NSAttributedString.Key: Any] = [.font: UIFont.systemFont(ofSize: 14), .foregroundColor: UIColor.white]
+
+        let textSize = text.size(withAttributes: textAttributes)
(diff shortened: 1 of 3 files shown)


r10s:
IMG_7065
IMG_7064

r10s:
nb: when this PR is merged, for inner consistency, we should change the few remaining swipe-gestures so that they also get the icons

deltachat-ios/Helper/Utils.swift
```diff
@@ -53,6 +53,27 @@ struct Utils {
         return window?.safeAreaInsets.bottom ?? 0
     }
 
+    @available(iOS 13.0, *)
+    static func makeImageWithText(image: UIImage?, text: String) -> UIImage? {
```
Amzd, line 57: Cool!

PR #2354, opened 2024-10-29
use new swipe-layout also in new chat

r10s:
as discussed, this uses the new swipe layout from https://github.com/deltachat/deltachat-ios/pull/2351 also for the few remaining lists

### DIFF
--- a/deltachat-ios/Controller/NewChatViewController.swift
+++ b/deltachat-ios/Controller/NewChatViewController.swift
@@ -348,7 +348,7 @@ extension NewChatViewController: ContactListDelegate {
 
 // MARK: - alerts
 extension NewChatViewController {
-    private func askToDeleteContact(contactId: Int, indexPath: IndexPath, callback: (() -> Void)? = nil) {
+    private func askToDeleteContact(contactId: Int, indexPath: IndexPath, didDelete: (() -> Void)? = nil) {
         let contact = dcContext.getContact(id: contactId)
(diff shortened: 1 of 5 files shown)


deltachat-ios/Controller/NewChatViewController.swift
```diff
@@ -349,6 +357,7 @@ extension NewChatViewController {
         )
         alert.addAction(UIAlertAction(title: String.localized("delete"), style: .destructive, handler: { [weak self] _ in
             self?.deleteContact(contactId: contactId, indexPath: indexPath)
+            callback?()
```
Amzd, line 360: callback is not called on all paths which means the UIContextualAction's completionHandler won't be called when you cancel.

r10s: is it really needed that the completion handler is always called. 

how does eg. WhatsApp keep a swipe-action open cancelling a subsequent dialog (eg. at "more" there)

Amzd: AH so it is intentional, mb

r10s: still, the name `callback()` is not-so-great, maybe better `confirmed()`

Amzd: Ah yeah or `didDelete()`

Issue #2356, opened 2024-10-29
ContactRequestBar missing


r10s:
when receiving a message from someone not  being in contact before, the chat is shown as a so called "Contact Request" - that means, you first have to accept the chat, then you can write etc. 

Acceptance or Blocking is done using `ChatContactRequestBar`, that is shown _instead_ of the message input field.

however, probably as a side effect of #2315 or subsequent changes,  the `ChatContactRequestBar` is never shown and the chat looks like:

<img width=320 src=https://github.com/user-attachments/assets/b47c5ec2-5ecd-4c9e-b39e-32404e7f6ee4>

note, that the ContactRequestBar is also use to show a warning if the "guaranteed e2ee" is broken - also then, the user must not type a message, but deal with the state first.

cc @Amzd

r10s:
@Amzd i played around a bit and came over https://github.com/deltachat/deltachat-ios/pull/2357

Amzd:
You can also see the issue of no bottom safe area when missing the inputAccessoryView in that screenshot. I'll fix that coming weekend.

r10s:
closed by #2357

PR #2357, opened 2024-10-30
fix contact-request-bar

r10s:
see https://github.com/deltachat/deltachat-ios/issues/2356

@Amzd i do not know much about first responders, so not sure if that makes sense  - i just lurked at your previous PR and the things changed there - and notices that "canSend()" is not sufficient from the core logic. at a first glance, the PR works as expected.

feel free to commit to this PR, or to replace it completely, if you think, another approach is more fitting.

unfortunately, i did not manage to delete lines in this PR 😭

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -169,7 +169,7 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
             // Don't show inputAccessoryView when context menu is visible
             return false
         } else {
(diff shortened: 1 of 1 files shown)


Amzd:
Exactly what I was thinking. Maybe check if there is an `if canSend { becomeFirstResponder() }` somewhere or if I removed that (can't remember)

Amzd (approved):

r10s:
> Maybe check if there is an if canSend { becomeFirstResponder() } somewhere or if I removed that (can't remember)

i had a second look, did not find any more suspicious in the code ... we'll see what testing tells :)

PR #2367, opened 2024-11-01
webxdc vc memory leak fix

Amzd:
previously every time you opened then closed a webxdc it leaked the whole vc with the WKWebView which is quite big

this also fixes realtime never being stopped

### DIFF
--- a/deltachat-ios/Controller/WebxdcViewController.swift
+++ b/deltachat-ios/Controller/WebxdcViewController.swift
@@ -233,12 +233,12 @@ class WebxdcViewController: WebViewViewController {
         let preferences = WKPreferences()
         let contentController = WKUserContentController()
         
-        contentController.add(self, name: WebxdcHandler.sendStatusUpdate.rawValue)
-        contentController.add(self, name: WebxdcHandler.log.rawValue)
-        contentController.add(self, name: WebxdcHandler.sendToChat.rawValue)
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -591,3 +587,45 @@ extension WebxdcViewController: WKURLSchemeHandler {
...
 }
+
+// MARK: Memory Leak Prevention
+
+extension WKUserContentController {
+    func add(weak scriptMessageHandler: any WKScriptMessageHandler, name: String) {
+        add(WeakScriptMessageHandler(scriptMessageHandler), name: name)
+    }
+
+    private class WeakScriptMessageHandler: NSObject, WKScriptMessageHandler {
```
r10s, line 598: cool, this `WKScriptMessageHandler` is a protocol - i was first wondering how we can know if we've overwritten/implemented all methods required.

r10s (approved):
respect.

thanks a lot for finding the issue - and then for the fix, which even is pretty elegant!

also, this fits nicely in the planned shipping of webxdc realtime

you earned the black belt of swift 🥋:)

deltachat-ios/Controller/WebxdcViewController.swift
Amzd: I'm still always paranoid about that because protocols can have optional requirements

PR #2368, opened 2024-11-01
dont ask for image access when opening gallery because it is not needed

Amzd:
> Apps don’t need to request photo library permission when using either [PHPickerViewController or UIImagePickerController]

https://developer.apple.com/documentation/photokit/selecting_photos_and_videos_in_ios


r10s (approved):
make sense.

maybe the permission query came from android, where the picker does not require a permission, however, as we want to show a camera roll, we query the permission there (asking out of context for gallery permission, when one wants to attach eg a contact was regarded as weird that time)

however, on iOS we do not have a camera roll currently, so it is good not to ask for permission we do not need.

thanks a lot for fixing!

Issue #2358, opened 2024-10-30
 bottom safe area missing if there is no input-bar


r10s:
on phones with safe area eg. due to rounded screen there is no safe area calculated added at the bottom when there is no input-view (so eg. in the device messages chat the cells are too close to the bottom)

see screenshot at https://github.com/deltachat/deltachat-ios/issues/2356

cc @Amzd

Issue #2323, opened 2024-10-15
minor side effects of "ScrollView jumping improvements"


r10s:
["ScrollView jumping improvements" PR](https://github.com/deltachat/deltachat-ios/pull/2315) introduced some side effects that need to be targeted:

- the icon shown on swipe-to-reply is flipped
- ~~tapping the system clock does not scroll _up_ as usual, but scrolls _down_~~ EDIT: might be okay, may be considered even as a fix as we no scroll to the most recent information; the scroll-up in eg. whatsapp is also more random as it scrolls just somewhere to the middle (as more messages are loaded there immediated)
- the scrollbar is gone
- input field is missing after using camera if the _keyboard was not expanded_ before

cc @Amzd

Amzd:
Did not fix the new behavior of status bar scrolling to the most recent message yet as other messengers (like discord) also do this. If users really want this we can always look at this again later. We will need to add a scrollview at the same level as the tableview which can intercept the status bar tap. So we'll need to use a UIViewController instead of UITableViewController.

Other issues are fixed in PR

PR #2369, opened 2024-11-01
fix for safe area

Amzd:
fix for safe area #2358


r10s (approved):
code lgtm, as far as i can tell.

but i also tested a bit, things work as expected - i also double checked all previous issues we closed recently after ["jumping improvements"](https://github.com/deltachat/deltachat-ios/pull/2315) (namely #2356, #2345, #2342, #2323)

thanks a lot for fixing ⚡️

PR #2370, opened 2024-11-01
relative dates in chat date markers

Amzd:
show "Today" instead of current date in chat

image


r10s (approved):
very nice!

played around a bit,  respects even the language set per-app - excellent 👌

PR #2372, opened 2024-11-02
UITests

Amzd:
This PR adds a UITest for the chat view controller. There is a couple TODO comments that can maybe be fixed but I don't think they are blocking.

I don't know how to handle xcproj changes, they are a mystery to me.


deltachat-ios/DC/DcContext.swift
```diff
@@ -61,6 +61,14 @@ public class DcContext {
...
     public func sendMessage(chatId: Int, message: DcMsg) {
         dc_send_msg(contextPointer, UInt32(chatId), message.messagePointer)
+        if CommandLine.arguments.contains("--UITests") {
+            // TODO: FIX
+            // FIXME: I think this is a bug but sendMessage does not publish the DC_EVENT_MSGS_CHANGED event in UITests. Probably because the message is not actually sent due to it not being a real, functioning account?
+            NotificationCenter.default.post(name: Event.messagesChanged, object: nil, userInfo: [
+                "message_id": message.id,
+                "chat_id": chatId,
+            ])
+        }
```
Amzd, line 71: Not sure where to report this but I think this is a big in core but not one that would affect any real users, just UITests.

r10s:
it is a bit unclear, how UITests work at the end, and if we want to go for that at the end. that should to be discussed among maintainers first.

putting this PR to resurrection

PR #2374, opened 2024-11-02
disable two-finger pan gesture during search

r10s:
multi-select is not working in search currently, we even disable it when entering search. so, do not allow entering search using two-finger pan gesture.

we can consider to enable multi-select also during search, but that is another effort i am not sure if it is worth going for (depends on effort and new code needed, of course)

cc @Amzd


Amzd (approved):
Good catch.

r10s:
will do another PR for the message search, this was even always wrong :)

PR #2376, opened 2024-11-02
Add webxdc to homescreen for iOS 16+

Amzd:
TODO: Check if steps are clear

### DIFF
--- a/deltachat-ios/Intents/OpenWebXDCAppIntent.swift
+++ b/deltachat-ios/Intents/OpenWebXDCAppIntent.swift
@@ -19,7 +19,7 @@ struct OpenWebXDCAppShortcuts: AppShortcutsProvider {
 @available(iOS 16, *)
 struct OpenWebXDCAppIntent: AppIntent {
     @MainActor
-    static var title: LocalizedStringResource = "Open webxcd app"
+    static var title: LocalizedStringResource = "Open webxdc app"
 
     @Parameter(title: "webxdc app", requestValueDialog: IntentDialog("Which webxdc app would you like to open?"))
     var webxdcApp: WebXDCAppEntity
--- a/deltachat-ios/Intents/OpenWebXDCAppIntent.swift
+++ b/deltachat-ios/Intents/OpenWebXDCAppIntent.swift
@@ -1,0 +1,130 @@
+import AppIntents
+import DcCore
+import UIKit
+
+@available(iOS 16, *) // if you set this to 17, shortcuts crash delta chat
+struct OpenWebXDCAppShortcuts: AppShortcutsProvider {
+    static var appShortcuts: [AppShortcut] {
+        AppShortcut(
+            intent: OpenWebXDCAppIntent(),
+            phrases: [
+                "Open \(\.$webxdcApp) in \(.applicationName)",
+                "Open a webxdc app in \(.applicationName)",
+            ],
+            shortTitle: "Open a webxcd app"
+        )
+    }
+}
+
+@available(iOS 16, *)
+struct OpenWebXDCAppIntent: AppIntent {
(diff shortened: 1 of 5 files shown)


r10s:
thanks a lot for these attempts! this was a very useful prerequisite to finally decide that we'll go the widget-way, widgets are started at #2418 , and that seems to be far easier by the normal user to handle, eg. you only need to long-tap the app icon - rough idea is to have a menu entry "Add To Widget" from the app menu then

Amzd:
Yeah maybe we can remove all the add to home screen steps and just merge the shortcut integrations so that power users that already know how to do this can either add it to their home screen or use voice activated shorcuts ("hey siri") to open webxdc apps.

deltachat-ios/Intents/OpenWebXDCAppIntent.swift
```diff
@@ -0,0 +1,130 @@
...
+            ],
+            shortTitle: "Open a webxcd app"
+        )
+    }
+}
+
+@available(iOS 16, *)
+struct OpenWebXDCAppIntent: AppIntent {
+    @MainActor
+    static var title: LocalizedStringResource = "Open webxcd app"
```
Simon-Laux, line 22:
```suggestion
    static var title: LocalizedStringResource = "Open webxdc app"
```

deltachat-ios/Intents/OpenWebXDCAppIntent.swift
```diff
@@ -0,0 +1,130 @@
...
+                return DcAccounts.shared.getAll().flatMap { accountId in
+                    let dcContext = DcAccounts.shared.get(id: accountId)
+                    let webxdcMsgIds = dcContext.getChatMedia(chatId: 0, messageType: DC_MSG_WEBXDC, messageType2: 0, messageType3: 0)
+                    return webxdcMsgIds.map { webxdcMsgId in
+                        let msg = dcContext.getMessage(id: webxdcMsgId)
+                        let chat = dcContext.getChat(chatId: msg.chatId)
+                        return WebXDCAppEntity(accountId: accountId, chat: chat, msg: msg)
+                    }
+                }
+            }
```
Simon-Laux, line 127: If you have multiple accounts with many xdcs it is not easy to find the one you are looking for.
Atleast sorted by recency would only partially work with how I imagine .flatMap to work. Like for first account it is fine, but for the second account you would need to scroll past all entries for the first account, wouldn't you? That is if there is no limit on the amount of suggestions shown?

I would suggest either making custom chooser ui if thats possible (select account, then webxdc from media list or even chatlist), or giving power users a way to find out account id and message id to construct the deep link themselves (or give them a button to "copy deep link for shortcut" in the 3dot menu of the webxdc or in the message info of the webxdc's message).

Amzd: This is the UI you get in Shortcuts app:

image

r10s:
i am personally not using Siri much, but maybe i should at least test :)

and yes, we can consider to improve Siri support. iirc, we are using Siri currently for the "share activity" -- ftr, that resulted in rejected reviews and some compliance things with apple that time. as compliance-wise we want to get "Apps" officially in now, this is something to consider early 2025 then.

Amzd:
Yep, so only `deltachat-ios/Intents/OpenWebXDCAppIntent.swift` should be retained and maybe remove the logic that only shows specific webxdc (which was used on iOS17+ to make add to home screen easier by only showing a shortcut recommendation for the currently active webxdc) and maybe limit the amount of recommendations to like the last 3 webxdcs you've used? Because a Shortcuts user can just select whatever webxdc they want inside the shortcut so the recommendations are not required.

PR #2388, opened 2024-11-05
Add Proxy-Screen

zeitschlag:
- Advanced Settings gained a new entry (Proxy) under _Server_. This settings lives behind the Passcode/Biometrics-protection.
- Adds a new screen to see, add, and delete proxy-servers (_Swipe to delete_ only works on device, not in the simulator, but trust me, it's there!). Looks like this:

<details><summary>Screenshots</summary>

Simulator Screenshot - iPhone 16 Pro - 2024-11-12 at 13 01 14

Simulator Screenshot - iPhone 16 Pro - 2024-11-12 at 13 02 25

</details>

- Code follows the [Android-implementation](https://github.com/deltachat/deltachat-android/tree/main/src/main/java/org/thoughtcrime/securesms/proxy), I didn't reinvent the wheel

### DIFF
--- a/deltachat-ios/Controller/Settings/Proxy/ProxySettingsViewController.swift
+++ b/deltachat-ios/Controller/Settings/Proxy/ProxySettingsViewController.swift
@@ -269,13 +269,20 @@ extension ProxySettingsViewController {
             indexPath.section == ProxySettingsSection.proxies.rawValue
         else { return nil }
 
-        let deleteAction = UIContextualAction(style: .destructive, title: String.localized("delete")) { [weak self] _, _, completion in
+        let deleteAction = UIContextualAction(style: .destructive, title: nil) { [weak self] _, _, completion in
             DispatchQueue.main.async {
                 self?.deleteProxy(at: indexPath)
                 completion(true)
             }
         }
         deleteAction.backgroundColor = .systemRed
+        deleteAction.accessibilityLabel = String.localized("delete")
+        if #available(iOS 13.0, *) {
+            deleteAction.image = Utils.makeImageWithText(image: UIImage(systemName: "trash"), text: String.localized("delete"))
+        } else {
+            deleteAction.title = String.localized("delete")
+        }
+
 
         let configuration = UISwipeActionsConfiguration(actions: [deleteAction])
         configuration.performsFirstActionWithFullSwipe = true


deltachat-ios/Helper/Utils.swift
```diff
@@ -165,6 +165,11 @@ struct Utils {
     }
 
     public static func authenticateDeviceOwner(reason: String, callback: @escaping () -> Void) {
+
+#if DEBUG
+        return callback()
+#endif
+
```
r10s, line 172: let's get that out, it is unrelated and there is also quite some worth in devs using the same dogfood as the normal users, esp. for some potentially breaking stuff, it is good to have that tested early and by the devs

deltachat-ios/Controller/Settings/AdvancedViewController.swift
```diff
@@ -320,7 +329,10 @@ internal final class AdvancedViewController: UITableViewController {
             Utils.authenticateDeviceOwner(reason: String.localized("pref_password_and_account_settings")) { [weak self] in
                 self?.showAccountSettingsController()
             }
-
+        case .proxySettings:
+            Utils.authenticateDeviceOwner(reason: String.localized("proxy_settings")) { [weak self] in
+                self?.showProxySettings()
+            }
```
r10s, line 335: no need to do a device-owner-verification here, at least, we're also not doing that on android and also telegram does not do it.

it seems mostly important to have that dialog fast and flawlessly working, as proxy change, ppl that really need them often try out things in that dialog. all device-owner-verification will slow that down (some more (pin) and slome less (face-id) :)  

also, if someone adds a proxy without you to know, you will be able to notice that by the (upcoming) icon).

cc @adbenitez

r10s (commented):
very nice! 👍 i did not yet manage to fine a today working proxy (cc @adbenitez do you have one :), some nitpicking below, and some comments on the UI:

1. the "connecting..." and connected information are missing, afaik, there is no special "proxy connectivity", but android just repeats what is otherwise written beside "connectivity" (so the "global state"), however, this is an important information to have in that dialog when testing out if a proxy works
 
2. the "protocol" should not just be a string, let's put that into a square box as on android, we can use the same style as 

3. when there are no proxies, android does not disable the switch, but allows to turn it on and shows the "Add Proxy" dialog - i think, this is a bit more user friendly

point 1. seems quite important, i would add that to this pr, so the user gets some  some feedback, otherwise one is quite blind. also, disabling owner-verification should go to this PR

the other points can also come later, the dialog is also a huge things without 2.+3. - and there are more things to do for proxy beside polishing :)

r10s (commented):
yeah, it starts working :)

first times, we have a proxy connection on iOS :)

a bug i encountered during testing:

- add a proxy, tap a second time to activate (that should not be needed, added proxies should become automatically the active one)
-> the proxy is shown twice in the list

usability-wise, in general, a simple tap on a proxy can start using the proxy already, no need for a confirmation, this is easily undoable. 

both are shown issues are shown in the following video:

https://github.com/user-attachments/assets/f34c1ad5-3250-4597-90a1-84bc2fbd41d8

with the "not-disable-switch but show add-screen" i mentioned on first review: we can leave things as is, the "add proxy" is easy enough, even though it is different on android

adbenitez:
> i did not yet manage to fine a today working proxy

ss://YWVzLTEyOC1nY206c2hhZG93c29ja3M=@156.146.62.163:443#Switzerland%20#1270%20/%20OutlineKeys.com

ss://YWVzLTEyOC1nY206c2hhZG93c29ja3M@156.146.62.164:443#Switzerland%20%231309%20%2F%20OutlineKeys%2Ecom

r10s (commented):
when adding a bad proxy after a working one, the state is still shown as "connected":

https://github.com/user-attachments/assets/5e1b63f2-48eb-4c88-a440-a00cee8e039c

the dialog logic seems right here, as also the chatlist shows connected - after restarting the app, however, things are shown as not-connected everywhere. maybe there is a stop/start IO missing?

also, items are still duplicated if you have more than two in the list and select another one:

https://github.com/user-attachments/assets/dbd6a61f-28f6-4b94-955b-d2c63ee2c6c2

zeitschlag:
I'll have another look at it.

deltachat-ios/Controller/Settings/Proxy/ProxyTableViewCell.swift
```diff
@@ -0,0 +1,83 @@
...
+
+        setupConstraints()
+    }
+    
+    required init?(coder: NSCoder) { fatalError("init(coder:) has not been implemented") }
+
+    private func setupConstraints() {
+        let constraints = [
+            contentStackView.topAnchor.constraint(equalTo: contentView.topAnchor, constant: 8),
+            contentStackView.leadingAnchor.constraint(equalTo: contentView.leadingAnchor, constant: 8),
```
r10s, line 59: design: use whitespace aligning with the other elements on the left - as well as having the same whitespace left/right

```suggestion
            contentStackView.leadingAnchor.constraint(equalTo: contentView.leadingAnchor, constant: 16),
```

<details>
<summary>before / after</summary>

 &nbsp;  &nbsp;  &nbsp;  


(including the tag's whitespace changes)

</details>

r10s (commented):
some design tweaks

deltachat-ios/Controller/Settings/Proxy/ProxySettingsViewController.swift
```diff
@@ -0,0 +1,285 @@
...
+        tableView.deselectRow(at: indexPath, animated: true)
+    }
+
+    override func tableView(_ tableView: UITableView, trailingSwipeActionsConfigurationForRowAt indexPath: IndexPath) -> UISwipeActionsConfiguration? {
+        guard
+            proxies.isEmpty == false,
+            indexPath.section == ProxySettingsSection.proxies.rawValue
+        else { return nil }
+
+        let deleteAction = UIContextualAction(style: .destructive, title: String.localized("proxy_delete")) { [weak self] _, _, completion in
```
r10s, line 272: context is clear here:

```suggestion
        let deleteAction = UIContextualAction(style: .destructive, title: String.localized("delete")) { [weak self] _, _, completion in
```

r10s (approved):

PR #2391, opened 2024-11-08
Add Proxy-button to Chatlist

zeitschlag:
- Add a button to Chatlist to easily see Proxy-status (iOS 13+) and access Proxy-settings
- On iOS 12, it says `Proxy`, on iOS 13+, you have a nice icon:

proxy-shield

### DIFF
--- a/deltachat-ios/Controller/ChatListViewController.swift
+++ b/deltachat-ios/Controller/ChatListViewController.swift
@@ -762,10 +762,13 @@ class ChatListViewController: UITableViewController {
         if RelayHelper.shared.isForwarding() {
             // multi-select is not allowed during forwarding
             titleView.text = RelayHelper.shared.dialogTitle
-            if !isArchive {
+            if isArchive {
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/ChatListViewController.swift
```diff
@@ -737,6 +765,7 @@ class ChatListViewController: UITableViewController {
             if !isArchive {
                 navigationItem.setLeftBarButton(cancelButton, animated: true)
             }
+            navigationItem.setRightBarButtonItems([newButton], animated: true)
```
r10s, line 768: there is no "new chat" option in the archive when _not_ forwarding, so we should not have one in special forwarding-mode as well:

```suggestion
            navigationItem.setRightBarButtonItems(isArchive ? nil : [newButton], animated: true)
```

r10s (approved):
some minor things, otherwise nice!

r10s (approved):

PR #2392, opened 2024-11-10
Returns first responder after UIImagePickerController search bar selection on iOS 16

Amzd:
This could use some testing but I am hyper-focussed on something else rn so I'm parking this here for anyone to pick up if they want to.


r10s:
what is the concrete bug that this aims to fix? missing input bar after attaching sth. on iOS 16?

Amzd:
bug: on iOS 16: Open chat > attachment > gallery > search > cancel > now there is no keyboard.

Caused by the previous implementation of getting the currentFirstResponder returning nil when you init a UIImagePickerController

r10s (approved):

PR #2393, opened 2024-11-12
when NSE is fetching, do not delay app start

r10s:
this PR avoids the app being stale on startup for up to 30 seconds in case the NSE ("notification service extension")  is not yet terminated.
EDIT: the system may even terminate the app during that timeframe

instead, the app is started, but startIo() is delayed until the NSE has finished (which takes at max 30 seconds)

### some background: 

the core cannot handle multiple processes fetching concurrently, reason is that some IMAP fetch state (last seen UID?)  is hold in RAM and not synced between processes (cc @link2xt is this still correct?)

Delta Chat iOS has currently up to three processes: main, share-extension, notification-service-extension. 

share extension is fine as nothing is read from IMAP. 

for the  notification-service-extension we do a fetch using `dc_accounts_background_fetch()` - and if the user starts the mainapp in that time (max. 30 seconds), cannot start_io() - before this PR we delayed the app start - and now we delay start_io() only.

this is not great program-wise, but from the view if the user quite okay

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Show on Chatlist if there are Proxy-servers (#2383)
 - Detect Proxy-servers in Chat-messages (#2389)
 - Share Proxy-servers with your friends, family and allies (#2394)
+- fix: no startup delay when processing PUSH notifications in parallel, just show "Updating..." in title bar
 
 ## v1.48.3
 2024-11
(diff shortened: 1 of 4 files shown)


r10s:
ah i see. would that really result in the message being added as two database rows practically? usually, there is the check for already-added-message-ids, or?

wondering, if the effort we're doing here is really needed - what would be the worse case? can we exclude database corruption (meaning, having the tables in a state that we eg. end in an endless loop)? if there are some rare races, that might be acceptable - also this complex code may have races and side effects ...

r10s:
thanks for the detailed info!

then we will stay withe the existing  approach that is improved by this PR, np :)

zeitschlag (approved):
LGTM 🚀

PR #2395, opened 2024-11-13
show 'proxy on' in advanved overview

r10s:
as for the other items,
a little overview is nice

### DIFF
--- a/deltachat-ios/Controller/Settings/AdvancedViewController.swift
+++ b/deltachat-ios/Controller/Settings/AdvancedViewController.swift
@@ -77,7 +77,7 @@ internal final class AdvancedViewController: UITableViewController {
     }()
 
     private lazy var proxySettingsCell: UITableViewCell = {
-        let cell = UITableViewCell(style: .default, reuseIdentifier: nil)
+        let cell = UITableViewCell(style: .value1, reuseIdentifier: nil)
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/Settings/AdvancedViewController.swift
```diff
@@ -462,6 +462,7 @@ internal final class AdvancedViewController: UITableViewController {
     private func updateCells() {
         showEmailsCell.detailTextLabel?.text = EmailOptionsViewController.getValString(val: dcContext.showEmails)
         videoChatInstanceCell.detailTextLabel?.text = VideoChatInstanceViewController.getValString(val: dcContext.getConfig("webrtc_instance") ?? "")
+        proxySettingsCell.detailTextLabel?.text = dcContext.isProxyEnabled ? String.localized("on") : nil
```
Simon-Laux, line 465: There is probably a reason for showing a generic "On" instead of the used proxy?

Simon-Laux (approved):

deltachat-ios/Controller/Settings/AdvancedViewController.swift
r10s: it is good enough, just a hint that jt is on.

the convretely used proxy is anyways mostly a big number zhat gets truncated

PR #2396, opened 2024-11-14
2389 detect proxy urls

zeitschlag:


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -2390,6 +2392,31 @@ extension ChatViewController: BaseMessageCellDelegate {
         }
     }
 
+    private func selectProxy(url proxyURL: String) {
+        let host = dcContext.checkQR(qrCode: proxyURL).text1 ?? ""
+        let dcAccounts = DcAccounts.shared
```
zeitschlag, line 2397: I'm not a 100% happy with using the singleton here, the alternative would've been to inject a `DcAccounts`-object and when I tried to, this kind of blew up in size and complexity.

r10s: using `DcAccounts.shared` is totally fine, no need to pass that around, there really should be only one. if that changes, we need some refactoring anyways (but, really, having multiple DcAccounts objects  is not planned nor forseeable)

deltachat-ios/Controller/ContactDetailViewController.swift
```diff
@@ -608,4 +610,30 @@ extension ContactDetailViewController: MultilineLabelCellDelegate {
             UIApplication.shared.open(url)
         }
     }
+
+    private func selectProxy(url proxyURL: String) {
```
r10s, line 614: let's unduplicate that code, eg. move it beside handleDeltaChatInvitation().

this makes maintenance easier, esp. as the code from the profile is probably used far less often (still, i think, it is nice to allow that also there)

r10s (approved):
only a minor, in general, works as expected

deltachat-ios/Controller/ContactDetailViewController.swift
zeitschlag: I thought about that, my rule of thumb is: If it appears more than twice, it's worth it to unduplicate it. In this case we'd probably need to unduplicate the entire `urlTapped`-thing, right?

PR #2398, opened 2024-11-14
tweak proxy icons

r10s:
this changes the proxy icon above the chatlist
to checked-shield (ON) and emtpy-shield (OFF)

- the striked-shield for OFF is confusing, as at other places, the strike through is used to show an action, so what would happen when the button is pressed (mute). analoguely, the strike-shield, one could think, the button removes the proxy.

- the checked-shield is better in that regard: while still showing state, it is still true that the action is to check another proxy

- empty-shield as a button also works better, action is to show the proxy list

- it simply looks better when a proxy is ON

state vs. action on buttons is tricky :)

moreover, the change is analogue to Telegram-iOS,
and also to deltachat-android, at least for the empty-shield :)

on / off:

 &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;

### DIFF
--- a/deltachat-ios/Controller/ChatListViewController.swift
+++ b/deltachat-ios/Controller/ChatListViewController.swift
@@ -45,7 +45,7 @@ class ChatListViewController: UITableViewController {
         let button: UIBarButtonItem
 
         if #available(iOS 13, *) {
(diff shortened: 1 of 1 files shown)


zeitschlag (approved):
looks nice(r), thank you!

funnily I also thought about using `checkmark.shield`, but decided against as we already agreed upon the `shield` and `shield.slash`.

PR #2399, opened 2024-11-15
Share Proxy as QR-code

zeitschlag:
- Swipe a proxy to the right to share it
- QR-Code gets generated and users can either share the link using the button or show other people that QR-code. If they scan it, they get asked on their device.
- Looks like this:

Simulator Screenshot - iPhone 16 Pro - 2024-11-15 at 13 08 35

### DIFF
--- a/deltachat-ios/Controller/Settings/Proxy/ProxySettingsViewController.swift
+++ b/deltachat-ios/Controller/Settings/Proxy/ProxySettingsViewController.swift
@@ -145,6 +145,13 @@ class ProxySettingsViewController: UITableViewController {
         present(deleteAlert, animated: true)
     }
 
+    private func shareProxy(at indexPath: IndexPath) {
+        let proxyToShare = proxies[indexPath.row]
+        let shareProxyViewController = ShareProxyViewController(dcContext: dcContext, proxyUrlString: proxyToShare)
+
+        show(UINavigationController(rootViewController: shareProxyViewController), sender: self)
+    }
+
     // MARK: - Notifications
 
     @objc private func handleConnectivityChanged(_ notification: Notification) {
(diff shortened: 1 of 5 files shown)


deltachat-ios/Controller/Settings/Proxy/ShareProxyViewController.swift
```diff
@@ -0,0 +1,142 @@
+import UIKit
+import DcCore
+import SDWebImageSVGKitPlugin
+
+class ShareProxyViewController: UIViewController {
```
zeitschlag, line 5: Maybe it would be worth to extract shared code from `ShareProxyViewController` and `QRCodeViewController` into a common component?

deltachat-ios/Controller/Settings/Proxy/ProxySettingsViewController.swift
```diff
@@ -262,6 +277,33 @@ extension ProxySettingsViewController {
...
+                self?.shareProxy(at: indexPath)
+                completion(true)
+            }
+        }
+        shareAction.backgroundColor = .systemGreen
+        shareAction.accessibilityLabel = String.localized("proxy_share_link")
+        if #available(iOS 13.0, *) {
+            shareAction.image = Utils.makeImageWithText(image: UIImage(systemName: "square.and.arrow.up"), text: String.localized("proxy_share_link"))
+        } else {
+            shareAction.title = String.localized("proxy_share_link")
```
r10s, line 297: context is clear, we can get shorter:

```suggestion
            shareAction.image = Utils.makeImageWithText(image: UIImage(systemName: "square.and.arrow.up"), text: String.localized("menu_share"))
        } else {
            shareAction.title = String.localized("menu_share")
```

deltachat-ios/Controller/Settings/Proxy/ShareProxyViewController.swift
```diff
@@ -0,0 +1,142 @@
...
+            view.backgroundColor = DcColors.defaultBackgroundColor
+        }
+
+        shareLinkButton.addTarget(self, action: #selector(ShareProxyViewController.shareInviteLink(_:)), for: .touchUpInside)
+        shareLinkButton.setTitleColor(DcColors.primary, for: .normal)
+
+        let svg = dcContext.createQRSVG(for: proxyUrlString)
+        qrContentView.image = getQrImage(svg: svg)
+
+        navigationItem.leftBarButtonItem = UIBarButtonItem(title: String.localized("done"), style: .done, target: self, action: #selector(ShareProxyViewController.done(_:)))
```
r10s, line 61: a single button needs to go to the right:

```suggestion
        navigationItem.rightBarButtonItem = UIBarButtonItem(title: String.localized("done"), style: .done, target: self, action: #selector(ShareProxyViewController.done(_:)))
```

r10s (approved):
very nice! only some minors wrt ui

deltachat-ios/Controller/Settings/Proxy/ShareProxyViewController.swift
r10s: there are some more changes done on android's contact-qr-controller anyways recently, let's see if and when we want that on iOS, then we can reconsider

PR #2403, opened 2024-11-18
Hide proxy-shield in multi-select

### LINKED ISSUE #2401: Proxy: shield should be hidden in multi-select mode
in multi-select mode, it is usual to hide everything unrelated.

we should also hide the proxy-button above the chatlist during multi-select therefore

zeitschlag:
`setRightBarButton` seems to do something differently than `setBarButonItems`. I fixed it in the other places as well. And as a proof, please enjoy this screenshot:

Screenshot 2024-11-18 at 08 57 34

Fixes #2401 (I didn't include a changelog-entry as this is a fix to an unreleased thing)

### DIFF
--- a/deltachat-ios/Controller/ChatListViewController.swift
+++ b/deltachat-ios/Controller/ChatListViewController.swift
@@ -773,7 +773,7 @@ class ChatListViewController: UITableViewController {
             titleView.text = String.localized("chat_archived_label")
             if !handleMultiSelectionTitle() {
                 navigationItem.setLeftBarButton(nil, animated: true)
-                navigationItem.setRightBarButton(markReadButton, animated: true)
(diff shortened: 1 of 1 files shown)


r10s:
that set rightButton and rightButtons do different things might also be a bug in iOS ... at least apple should have documented it, if that is really the expected behavior

r10s (approved):

Amzd:
Not a bug in iOS as it is documented here: https://developer.apple.com/documentation/uikit/uinavigationitem/1624956-rightbarbuttonitems

> The first item in the array can also be set using the [rightBarButtonItem](https://developer.apple.com/documentation/uikit/uinavigationitem/1624957-rightbarbuttonitem) property.

but definitely unintuitive

PR #2404, opened 2024-11-18
Scan Proxy-QR-codes (#2400)

### LINKED ISSUE #2400: Proxy: Scan and handle Proxy-QR-Codes
- Show alert to ask people if they want to add that proxy

zeitschlag:
- QR-Reader learned a new trick: It can scan Proxy-QR-codes now.
- It's pretty simple: You scan a Proxy-QR-code with the QR-code-reader and you see this alter:

Screenshot 2024-11-18 at 15 14 03

- Fixes #2400

### DIFF
--- a/deltachat-ios/Helper/Protocols.swift
+++ b/deltachat-ios/Helper/Protocols.swift
@@ -1,7 +1,7 @@
 import UIKit
 
 protocol QrCodeReaderDelegate: AnyObject {
-    func handleQrCode(_ viewController: UIViewController, qrCode: String)
+    func handleQrCode(_ qrCode: String)
 }
 
 protocol ContactListDelegate: AnyObject {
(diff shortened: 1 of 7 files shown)


deltachat-ios/Helper/Protocols.swift
```diff
@@ -1,7 +1,7 @@
 import UIKit
 
 protocol QrCodeReaderDelegate: AnyObject {
-    func handleQrCode(_ code: String)
+    func handleQrCode(_ viewController: UIViewController, qrCode: String)
```
r10s, line 4: the added viewController seems not be be used anywhere at a first glance

r10s (approved):
works for me :)

nb: this does not support scanning from the system camera, right? wondering if we want to support that, maybe file an issue to not forget about it

deltachat-ios/Helper/Protocols.swift
zeitschlag: Good catch, thank you!

PR #2412, opened 2024-11-23
add new webxdc APIs

### LINKED ISSUE #2411: add webxdc selfAddr/notify/href APIs (M2/M3)
see and  https://github.com/deltachat/deltachat-core-rust/pull/6230,  https://github.com/deltachat/deltachat-core-rust/pull/6237 and   https://github.com/deltachat/deltachat-core-rust/pull/6248

r10s:
closes #2411

### DIFF
--- a/deltachat-ios/Controller/WebxdcViewController.swift
+++ b/deltachat-ios/Controller/WebxdcViewController.swift
@@ -430,8 +430,10 @@ class WebxdcViewController: WebViewViewController {
         DispatchQueue.global(qos: .userInitiated).async { [weak self] in
             guard let self else { return }
 
-            let url = URL(string: "\(self.INTERNALSCHEMA)://acc\(self.dcContext.id)-msg\(self.messageId).localhost/" + (href ?? "index.html"))
-            let urlRequest = URLRequest(url: url!)
+            let base ="\(self.INTERNALSCHEMA)://acc\(self.dcContext.id)-msg\(self.messageId).localhost/"
+            let url = URL(string:  base + href ?? "index.html")
+            let fallbackUrl = URL(string: base + "index.html")!
+            let urlRequest = URLRequest(url: url ?? fallbackUrl)
             DispatchQueue.main.async {
                 self.webView.load(urlRequest)
             }
--- a/deltachat-ios/Controller/WebxdcViewController.swift
+++ b/deltachat-ios/Controller/WebxdcViewController.swift
@@ -426,11 +426,7 @@ class WebxdcViewController: WebViewViewController {
             guard let self else { return }
 
             let file = if let href {
-                if href.starts(with: "#") {
-                    "index.html" + href
-                } else {
-                    href
-                }
+                href
             } else {
                 "index.html"
             }
--- a/deltachat-ios/Controller/WebxdcViewController.swift
+++ b/deltachat-ios/Controller/WebxdcViewController.swift
@@ -425,13 +425,7 @@ class WebxdcViewController: WebViewViewController {
         DispatchQueue.global(qos: .userInitiated).async { [weak self] in
             guard let self else { return }
 
-            let file = if let href {
-                href
-            } else {
-                "index.html"
-            }
-
-            let url = URL(string: "\(self.INTERNALSCHEMA)://acc\(self.dcContext.id)-msg\(self.messageId).localhost/" + file)
+            let url = URL(string: "\(self.INTERNALSCHEMA)://acc\(self.dcContext.id)-msg\(self.messageId).localhost/" + (href ?? "index.html"))
             let urlRequest = URLRequest(url: url!)
             DispatchQueue.main.async {
                 self.webView.load(urlRequest)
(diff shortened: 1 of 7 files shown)


deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -417,7 +424,18 @@ class WebxdcViewController: WebViewViewController {
...
+
+            let file = if let href {
+                if href.starts(with: "#") {
+                    "index.html" + href
+                } else {
+                    href
+                }
+            } else {
+                "index.html"
+            }
```
Amzd, line 436: I think this would be quite unintuitive to webxdc app developers reading the window.location.href. Is index.html strictly needed here? Or is it in the spec that it works like this?

r10s: index.html  is the starting point of any webxdc, this is also documented like that at https://webxdc.org/docs/

so,  `update.href` can be `#sth` in case of single-page apps (which is the vast majority) - for multi-page apps, that needs to be `page1.html#sth`, `page2.html#sth`, `index.html#sth`

Amzd: I might be missing a bit of context but since a webserver would serve the index.html with a `window.location.href` path of just `/` (`/index.html` not needed), an implementation of webxdc using web server instead of would have a different `window.location.href` path.

So for example if you send "#sth" in a webserver implementation the path part of `window.location.href` would become "#sth" while with this implementation it would become "index.html#sth". 

https://github.com/webxdc/website/pull/90/files/bf284a1901cc5edd7f04798e4261c16fb5555600#diff-6ecb9b561809502e7de51b357752efed1f95cd22736a885e19e5eff2012cc77aR11

The new doc says "you can't assume anything about the scheme or domain part of the url" which implies you can assume the path part of the url which is not true if my example here follows the spec? Tell me if I am wrong, you have much more knowledge of webxdc.

adbenitez: it is about a link to open the app, ex. if you are in a ./subpage.html file and send #foobar this doesn't mean that the link is to the current page, you have to be explicit about what page you want to open with the #fragment, because when you start a webxdc it always starts at index.html, the only situation when you can send just the #fragment is when it is for index.html, if it were always explicit it would be more obvious, but less convenient for single page apps

adbenitez: e.i. the #fragment is not relative to the current open page, otherwise the behavior would be arbitrary receiving a #fragment in the current open page when it was intended for another subpage

hpk42:
Caspar, thanks joining the discussion from a general web-app/server angle!

What is different with webxdc apps and how clicking an info message opens a href
is that there is not neccessarily a "current" page as the info message might be
a historic message and you have since navigated the app to a different page. 
So if the href is simply "#something" then the messenger would have a hard
time providing the implicit full location when starting the webview. 

This is different with classic web-apps where a href is always used 
within some current page/context and resolved relative to it.

deltachat-ios/Controller/WebxdcViewController.swift
r10s: i changed this PR to not look into `href`, as discussed. thanks a lot

deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -417,7 +424,14 @@ class WebxdcViewController: WebViewViewController {
     private func loadHtml() {
         DispatchQueue.global(qos: .userInitiated).async { [weak self] in
             guard let self else { return }
-            let url = URL(string: "\(self.INTERNALSCHEMA)://acc\(self.dcContext.id)-msg\(self.messageId).localhost/index.html")
+
+            let file = if let href {
+                href
+            } else {
+                "index.html"
+            }
```
Amzd, line 432: This is the same as 

```swift
let file = href ?? "index.html"
```

r10s: yeah, soo much switching of programming languages these days :) thanks for the pointer, i simplified that futher

deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -417,7 +424,8 @@ class WebxdcViewController: WebViewViewController {
     private func loadHtml() {
         DispatchQueue.global(qos: .userInitiated).async { [weak self] in
             guard let self else { return }
-            let url = URL(string: "\(self.INTERNALSCHEMA)://acc\(self.dcContext.id)-msg\(self.messageId).localhost/index.html")
+
+            let url = URL(string: "\(self.INTERNALSCHEMA)://acc\(self.dcContext.id)-msg\(self.messageId).localhost/" + (href ?? "index.html"))
             let urlRequest = URLRequest(url: url!)
```
Amzd, line 434:
```suggestion
            let base ="\(self.INTERNALSCHEMA)://acc\(self.dcContext.id)-msg\(self.messageId).localhost/"
            let url = URL(string:  base + href ?? "index.html")
            let fallbackUrl = URL(string: base + "index.html")!
            let urlRequest = URLRequest(url: url ?? fallbackUrl)
```

Maybe prevent crashing if href creates an invalid url? I don't know if you can make a URL invalid with only the path part but it feels unsafe to force unwrap there. Sorry for all the nitpicky comments haha

Amzd (approved):

PR #2418, opened 2024-11-27
Most Recent Webxdc apps but make it a widget

### LINKED ISSUE #2406: Show Webxdc-apps in Widget
- Select up to `n` Webxdc-apps you want to display in the Widget
- Webxdc-apps are just messages in chats (See: `DocumentGalleryFileCell.updateWebxdcMsg`)
- When you tap one of these Webdxc-apps, the app should be opened, if possible the selected Webxdc-app should be opened
- `AppCoordinator.handleDeepLinkURL`

zeitschlag:
This PR adds a Widget to Deltachat. The widget shows the most recent Webxdc-apps on the Homescreen so that users can open them right away from the Homescreen. It comes in two flavours/sizes:

- small. Shows up to 4 apps.
- medium. Shows up to 8 apps.

They look like this:

When users tap on an icon, the selected app opens right away in the right chat. The list gets updated every 15 minutes.

When adding them to the Homescreen, users see this:

2406_add

The icons for this placeholder come from various Webxdc-app-repos.

Also: This PR makes CI use Xcode 16. And closes #2406 
Sidenote: Widgets are more fun on an actual device than in the Simulator.
Sidenote: Widget is not localized yet.


DcWebxdcWidget/DcWebxdcWidget.swift
```diff
@@ -0,0 +1,177 @@
...
+
+    func getSnapshot(in context: Context, completion: @escaping (UsedWebxdcEntry) -> Void) {
+        let entry = placeholder(in: context)
+        completion(entry)
+    }
+
+    func getTimeline(in context: Context, completion: @escaping (Timeline<Entry>) -> Void) {
+        /// ----------------------------------
+        /// **DO NOT** START dcAccounts.startIo() IN WIDGET
+        /// ----------------------------------
```
r10s, line 39: i'd be a bit less broad and also try to answer "why":

```suggestion
        // no startIo() as this works only for one process, and widgets may run concurrently to the main app
```

DcWebxdcWidget/DcWebxdcWidget.swift
```diff
@@ -0,0 +1,177 @@
...
+                title: name
+            )
+        }
+
+        let currentDate = Date()
+        let entry = UsedWebxdcEntry(date: currentDate, apps: apps)
+        let nextDate = Calendar.current.date(byAdding: .minute, value: 15, to: currentDate)!
+
+        let timeline = Timeline(entries: [entry], policy: .after(nextDate))
+        completion(timeline)
```
r10s, line 89: i am wondering, how other apps create "shortcut widgets" - are there other providers than "TimelineProvider"? just a general question for understanding, for the current implementation, things are fine

DcWebxdcWidget/DcWebxdcWidget.swift
```diff
@@ -0,0 +1,177 @@
...
+            let dcContext = dcAccounts.get(id: accountId)
+            if dcContext.isOpen() == false {
+                do {
+                    let secret = try KeychainManager.getAccountSecret(accountID: accountId)
+                    _ = dcContext.open(passphrase: secret)
+                } catch {
+                    debugPrint("Couldn't open \(error.localizedDescription)")
+                }
+            }
+        }
```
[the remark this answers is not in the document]
zeitschlag: Ah, that's good to know, thank you!
I basically just copied this from `AppDelegate.continueDidFinishLaunchingWithOptions`.  Should we remove it there as well?

DcWebxdcWidget/DcWebxdcWidget.swift
zeitschlag: From what I know, there's just the Timeline-provider. Basically you can't just update the widget itself [whenever you want](https://developer.apple.com/documentation/WidgetKit/Keeping-a-Widget-Up-To-Date), just like with background activity, you have a limited reload-budget and the system is in control to allocate your widget some time to reload here and there throughout the day.

That's why you build a "Timeline" with entries and give a "Please update then"-date. Does this answer your question?

scripts/untranslated.xml
```diff
@@ -7,4 +7,7 @@
     <string name="backup_successful_explain_ios">You can find the backup in the \"Delta Chat\" folder using the \"Files\" app.\n\nMove the backup out of this folder to keep it when deleting Delta Chat.</string>
     <string name="location_denied">Location access denied</string>
     <string name="location_denied_explain_ios">In the system settings, enable \"Privacy/Location Services\" and set \"Delta Chat/Location\" to \"Always\" and \"Precise\".</string>
+    <string name="widget_no_apps">No apps.</string>
+    <string name="widget_most_recent_apps_title">Recent Webxdc-apps</string>
+    <string name="widget_most_recent_apps_description">Shows the most recent Webxdc-apps.</string>
```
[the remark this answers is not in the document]
zeitschlag: Regarding the naming, I'd like to point to the [Shortcuts-app](https://en.wikipedia.org/wiki/Shortcuts_(Apple)) from Apple.

DcWebxdcWidget/DcWebxdcWidget.swift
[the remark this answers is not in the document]
r10s: at some point, yes, then together with all the other stuff that was deprecated at #2011 - that would, however, prevent ppl from opening these accounts completely (maybe we should give another warning before)

DcWebxdcWidget/DcWebxdcWidget.swift
r10s: yes, so i learned it is always  a "timeline", thanks  for clarifying that, this was the part i was unsure about.

however,  as it is possible to let ["WidgetKit request a new timeline when something affects a widget’s current timeline"](https://developer.apple.com/documentation/WidgetKit/Keeping-a-Widget-Up-To-Date#Inform-WidgetKit-when-a-timeline-changes) it should be fine also for the case the user adds/removes widgets manually

scripts/untranslated.xml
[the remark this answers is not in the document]
r10s: yes, i had that in mind as well. seems still be better as it is not the exactly same screen (so us adding "apps" in the screen that shows "apps" :), also at least "telegram" uses the same term.

so, i am not that worried about it, but i also do not think, it is perfect. if there are other suggestions, i am of course open to that

r10s:
i think, it is okay for being merged to move forward; for a release it is already bespoken that subsequent steps are needed (maybe also for a testflight)

subsequent steps are  at least https://github.com/deltachat/deltachat-ios/pull/2440, however, as we meanwhile know by 99% that the widget cannot be removed from "Quick Actions" [^1], we can also consider to make the widgets more useful also in case the user has no apps yet or does not even know about them (we do not have an app-picker yet) - that could be the option to add shortcuts also to normal chats, as virtually all other messengers allow.

that would probably also make apple review much easier - we would not longer hide out "app" feature, but it would also be "far less in the face" (which is totally fine and UX wise even better)

depends a bit on how complicated that would become, and also about the timing wrt the added app-picker

lots of moving targets

[^1]: "Quick Actions" are shown when long taping the app-icon, here the first row allows to turn the app-icon into a widget:

r10s (approved):

DcWebxdcWidget/DcWebxdcWidget.swift
```diff
@@ -0,0 +1,166 @@
...
+struct Provider: TimelineProvider {
+    func placeholder(in context: Context) -> UsedWebxdcEntry {
+        let limit: Int
+        switch context.family {
+        case .systemSmall:
+            limit = 4
+        case .systemMedium:
+            limit = 7
+        default:
+            limit = 7
```
r10s, line 14: wondering, btw, where the "7" comes from - why not "8"?

zeitschlag: It comes from my confused brain. `8` is totally right, of course, let me fix that! 🫠

PR #2433, opened 2024-12-09
Update Rust to 1.83.0

link2xt:


r10s (approved):
builds and runs for me as expected :)

Issue #2429, opened 2024-12-07
add "mention" notification switch


r10s:
android got a recent ["mention" switch](https://github.com/deltachat/deltachat-android/pull/3456#discussion_r1866087382) for notifications.

to do the same on iOS:

- we should move the "notifications switch" from top-level settings page to a subpage (as all other settings)
this makes room for the "mention" switch - as well as maybe a button to open the sound- and banner-system-settings directly
the main settings dialog should show "on" and "off" then (any maybe hints as "disabled by low power mode" as the connectivity view, however, not sure how that can fit to the limited room, this is also not needed to close this issue)

- respect the switch using the same logic as the linked android-pr

PR #2434, opened 2024-12-09
add mentions toggle

r10s:
_review is done best with ignoring whitespace; the logic from the flowchart is in the last two commits_

this PR adds a "Mentions" toggle to the settings; to keep notification things together, a new NotificationsViewController is created.

as a bonus, a link for the system settings is added as well, making all relevant notification settings directly accessible:

counterpart of https://github.com/deltachat/deltachat-android/pull/3456

closes https://github.com/deltachat/deltachat-ios/issues/2429

### DIFF
--- a/deltachat-ios/Controller/Settings/NotificationsViewController.swift
+++ b/deltachat-ios/Controller/Settings/NotificationsViewController.swift
@@ -1,0 +1,145 @@
+import UIKit
+import DcCore
+import Intents
+
+internal final class NotificationsViewController: UITableViewController {
+
+    private struct SectionConfigs {
+        let headerTitle: String?
+        let footerTitle: String?
+        let cells: [UITableViewCell]
+    }
+
+    private enum CellTags: Int {
+        case defaultTagValue = 0
+        case systemSettings
+    }
+
+    private var dcContext: DcContext
+    internal let dcAccounts: DcAccounts
+
+    // MARK: - cells
+    private lazy var notificationsCell: SwitchCell = {
+        return SwitchCell(
+            textLabel: String.localized("pref_notifications"),
+            on: !dcContext.isMuted(),
+            action: { [weak self] cell in
+                guard let self else { return }
+
+                dcContext.setMuted(!cell.isOn)
+                if cell.isOn {
+                    if let appDelegate = UIApplication.shared.delegate as? AppDelegate {
+                        appDelegate.registerForNotifications()
+                    }
+                } else {
+                    NotificationManager.removeAllNotifications()
+                }
+
+                updateCells()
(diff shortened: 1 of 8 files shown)


r10s:
ftr, required notification flow is drafted at https://github.com/deltachat/interface/blob/main/notification-flow.md  and looks like:

394862831-5abc0889-130a-46e9-9670-dca8340c38a9

deltachat-ios/Controller/Settings/NotificationsViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+
+        switch cellTag {
+        case .systemSettings:
+            let urlString = if #available(iOS 16, *) {
+                UIApplication.openNotificationSettingsURLString
+            } else if #available(iOS 15.4, *) {
+                UIApplicationOpenNotificationSettingsURLString
+            } else {
+                UIApplication.openSettingsURLString
+            }
```
zeitschlag, line 131: I find this if-assignment-thinkie kind of ... confusing but maybe I'm an old man yelling at clouds

deltachat-ios/Controller/Settings/NotificationsViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+
+            if let url = URL(string: urlString), UIApplication.shared.canOpenURL(url) {
+                UIApplication.shared.open(url)
+            }
+        case .defaultTagValue:
```
zeitschlag, line 136: Is there a reason to not toggle the settings when tapping on one of the cells?
Right now you have to toggle the switch, but in other places (I think), we also toggle when tapping the cell

zeitschlag (approved):

deltachat-ios/Controller/Settings/NotificationsViewController.swift
r10s: it is pretty common nowadays, known eg. from rust. i think, it is a handy tool - and avoids having mutable vars

zeitschlag: As I said: I'm an old man yelling at clouds

deltachat-ios/Controller/Settings/NotificationsViewController.swift
r10s: nope.

we never toggle anywhere when tapping outside the concrete switch.

this is also the behavior, eg. apple is using in its "system settings" - and i think, it makes a lot of sense as it avoids accidental changes gracefully (on android, default behaviour is to toggle on tapping on the whole cell, with results in exactly that issue of accidental changes)

PR #2440, opened 2024-12-12
Add/remove apps from/to Shortcuts-widget

### LINKED ISSUE #2426: Make apps on widget configurable
- Menu-Entry: "Add to Widget"/"Remove from Widget" in both App-list and Webxdc-Screen
- Store `MessageID`/`AccountID` somewhere (`UserDefaults`? `Core` aka `set_config`?), aka make a struct with two values and serialize/deserialize the JSON.

zeitschlag:
This PR makes _the_ Widget (#2418) configurable. The widget doesn't show the most recent apps anymore but only those handpicked by the user. To handpick the Webxdc-apps, this PR introduces some new menu-entries to the App-gallery and the menu on the Webxdc-Screen:

The ids of the selected Webxdc-apps (read: messages) and the user-account are stored as JSON in core-config using `ui.ios.selected_apps_for_widget` as key.

(On another note: An alternative would've been to make the widget _really_ configurable, so that users can pick the apps when editing the widget itself, but this approach made way more sense for our usecase, well, it made more sense to me and it was easier and more straight forward to implement than making a configurable widget)

Closes #2426.


Simon-Laux (commented):
Works, didn't look at the code.

probably not directly related to this pr: While testing usage with multiple accounts it was a bit unclear what happens. Like pinning from different account removed the previously pinned items (my user expectation would have been that it just adds it to the list of apps).

deltachat-ios.xcodeproj/project.pbxproj
```diff
@@ -2513,7 +2527,7 @@
 				INFOPLIST_FILE = DcWebxdcWidget/Info.plist;
 				INFOPLIST_KEY_CFBundleDisplayName = DcWebxdcWidget;
 				INFOPLIST_KEY_NSHumanReadableCopyright = "Copyright © 2024 merlinux GmbH. All rights reserved.";
-				IPHONEOS_DEPLOYMENT_TARGET = 18.1;
+				IPHONEOS_DEPLOYMENT_TARGET = 15.6;
```
r10s, line 2530: great - so widgets will work on all phones but the ones only having ios12 (which is about being deprecated anyways soon)

deltachat-ios/DC/DcContext+Extension.swift
```diff
@@ -0,0 +1,44 @@
+import Foundation
+import DcCore
+import WidgetKit
+
+@available(iOS 15, *)
+extension DcContext {
+    private static let key = "ui.ios.selected_apps_for_widget"
+    func shownWidgets() -> [WidgetEntry] {
+        guard let jsonData = getConfig(Self.key)?.data(using: .utf8) else { return [] }
```
r10s, line 9: as discussed in the call, this goes to UserDefaults, but that can be done in a subsequent PR as well, when targeting #2449 

ftr, instead of the function name `shownWidgets()` i would use `getAllWidgetEntries()` and then `storeWidgetEntry()`

r10s (approved):
lgtm, just a minor suggestion (feel free to ignore :) beside the things we discussed already.

thanks a lot for pushing that forward!

deltachat-ios/DC/DcContext+Extension.swift
zeitschlag: Thanks for your feedback, adressed it (see below)

PR #2442, opened 2024-12-12
prepare 1.50.3

r10s:
this is a bugfix release (using .3 so that core-version is likely to be in sync with android)

idea is to have sth more stable, as larger changes are up to come :) stay tuned :)


zeitschlag (approved):
I approve (publicly)!

r10s:
/me building and then uploading to testflight ...

PR #2443, opened 2024-12-13
Hide message-bar when forwarding

### LINKED ISSUE #2435: Forwarding shows "message input" as well as alert
when forwarding messages and selecting a chat, the chat shows an alert, asking for confirmation.

however, while the alert is shown, additionally, the "message input" is shown accidentally:

this additionally results in the buttons jumping as the "message input" appears a second later.

it would be better, to just not show the "message input" - compared to other dialog shown, eg. attachment selector or the "guaranteed info"

@Amzd not sure if this directly related to https://github.com/deltachat/deltachat-ios/pull/2315 or subsequent PR - but maybe you have a direct idea :)

zeitschlag:
closes #2435

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Delta Chat iOS Changelog
 
+## Unreleased
+
+- Don't show message-input when forwarding (#2435)
+
 ## v1.50.3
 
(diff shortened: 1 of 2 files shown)


r10s (approved):
code lgtm, and in tests, it works as expected.

also, mich smoother by the removal of the dispating to main from main

PR #2444, opened 2024-12-17
fix warning: handle new authorization status 'limited'

r10s:
update to xcode 16 raises some new warnings; this PR fixes one of them:

- if we get the authorization status 'limited' for accessing contacts, we're done - it is totally fine if the user does not give us access to all contacts

to not let CI fail with "unknown identified", this PR lets CI use Xcode 16 as well


zeitschlag (commented):
Have you noticed that #2418 also [updates the CI to use Xcode 16](https://github.com/deltachat/deltachat-ios/pull/2418/commits/799ecccb88cd0e2580d9ff96287fc910d27d641e)? So this will be a merge conflict. That's the price for PRs that have been open for quite some time 🙃

r10s:
ah, then let's see who comes first in :)

no big deal, that is fixable by a small rebase

zeitschlag:
> that is fixable by a small rebase

sure it is, but we have kept that PR open for quite some time now and the longer we keep it open, the more "fixable by a small rebase" will be necessary.

r10s:
oh, in other repos, PR are open and rebased here and then over months until we find the correct merge window :)

but yes, i admit that it is a little annoying and in this case also not needed as we want to get things in, i approved #2418 and we can merge it in

zeitschlag (approved):
Thanks for fixing this, I missed this 🙈

PR #2445, opened 2024-12-17
add option to copy links instead opening

r10s:
this PR adds detects if a link was long-tapped and offers the option "Copy Link" (instead of "Copy Text").

if was also playing around with offering both options and even with a totally separate menu, however, this seems both not better UX wise and is not worth additional effort. it is easy _not_ to tap the link - and the other way round, if the links was accidentally tapped (eg. cause it is long), it is nice to have all options at hand (reactions :)

this is how it looks like:

successor of https://github.com/deltachat/deltachat-ios/pull/2155  
feature request at https://support.delta.chat/t/copy-a-link-from-a-message-on-long-press-in-ios/2755

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -2028,6 +2028,15 @@ extension ChatViewController {
             return nil
         }
 
+        // Check if the long tap is on a link (or other message text element with custom long tap behavior)
+        if let msgcell = tableView.cellForRow(at: indexPath) as? BaseMessageCell {
+            let label = msgcell.messageLabel.label
+            let localTouchLocation = tableView.convert(point, to: label)
+            if let (detectorType, value) = label.detectGesture(localTouchLocation) {
+                print("url: \(detectorType) -- \(value) -- ")
+            }
+        }
+
         return UIContextMenuConfiguration(
             identifier: NSString(string: "\(messageId)"),
             previewProvider: nil,
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -2021,20 +2021,20 @@ extension ChatViewController {
         menuElements.append(action)
     }
 
+    private func isLinkTapped(indexPath: IndexPath, point: CGPoint) -> String? {
+        if let cell = tableView.cellForRow(at: indexPath) as? BaseMessageCell {
+            let label = cell.messageLabel.label
+            let localTouchLocation = tableView.convert(point, to: label)
+            return label.getCopyableLinkText(localTouchLocation)
+        }
+        return nil
+    }
+
     // context menu for iOS 13+
     override func tableView(_ tableView: UITableView, contextMenuConfigurationForRowAt indexPath: IndexPath, point: CGPoint) -> UIContextMenuConfiguration? {
         let messageId = messageIds[indexPath.row]
         if tableView.isEditing || messageId == DC_MSG_ID_MARKER1 || messageId == DC_MSG_ID_DAYMARKER {
             return nil
-        }
-
-        // Check if the long tap is on a link (or other message text element with custom long tap behavior)
-        if let msgcell = tableView.cellForRow(at: indexPath) as? BaseMessageCell {
-            let label = msgcell.messageLabel.label
-            let localTouchLocation = tableView.convert(point, to: label)
-            if let (detectorType, value) = label.detectGesture(localTouchLocation) {
-                print("url: \(detectorType) -- \(value) -- ")
-            }
         }
 
         return UIContextMenuConfiguration(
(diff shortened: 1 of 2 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -2021,6 +2021,15 @@ extension ChatViewController {
         menuElements.append(action)
     }
 
+    private func isLinkTapped(indexPath: IndexPath, point: CGPoint) -> String? {
+        if let cell = tableView.cellForRow(at: indexPath) as? BaseMessageCell {
```
zeitschlag, line 2025: Why `if` instead of `guard`?

deltachat-ios/Chat/Views/MessageLabel.swift
```diff
@@ -452,49 +452,65 @@ open class MessageLabel: UILabel {
...
     }
 
-    /// swiftlint:disable cyclomatic_complexity
-    private func handleGesture(for detectorType: DetectorType, value: NewMessageTextCheckingType) {
+    internal func getCopyableLinkText(_ touchLocation: CGPoint) -> String? {
+        guard let (_, value) = detectLink(touchLocation) else { return nil }
+
+        switch value {
+        case let .link(url):
+            guard let url = url else { return nil }
```
zeitschlag, line 474:
```suggestion
            guard let url else { return nil }
```

deltachat-ios/Chat/Views/MessageLabel.swift
```diff
@@ -452,49 +452,65 @@ open class MessageLabel: UILabel {
...
+    }
+
+    internal func handleGesture(_ touchLocation: CGPoint) -> Bool {
+        guard let (detectorType, value) = detectLink(touchLocation) else { return false }
+
         switch value {
         case let .addressComponents(addressComponents):
             var transformedAddressComponents = [String: String]()
-            guard let addressComponents = addressComponents else { return }
+            guard let addressComponents = addressComponents else { return false }
```
zeitschlag, line 491:
```suggestion
            guard let addressComponents else { return false }
```

deltachat-ios/Chat/Views/MessageLabel.swift
```diff
@@ -452,49 +452,65 @@ open class MessageLabel: UILabel {
...
             addressComponents.forEach { (key, value) in
                 transformedAddressComponents[key.rawValue] = value
             }
             handleAddress(transformedAddressComponents)
         case let .phoneNumber(phoneNumber):
-            guard let phoneNumber = phoneNumber else { return }
+            guard let phoneNumber = phoneNumber else { return false }
```
zeitschlag, line 497:
```suggestion
            guard let phoneNumber else { return false }
```

deltachat-ios/Chat/Views/MessageLabel.swift
```diff
@@ -452,49 +452,65 @@ open class MessageLabel: UILabel {
...
             handlePhoneNumber(phoneNumber)
         case let .date(date):
-            guard let date = date else { return }
+            guard let date = date else { return false }
```
zeitschlag, line 500:
```suggestion
            guard let date else { return false }
```

deltachat-ios/Chat/Views/MessageLabel.swift
```diff
@@ -452,49 +452,65 @@ open class MessageLabel: UILabel {
...
             var transformedTransitInformation = [String: String]()
-            guard let transitInformation = transitInformation else { return }
+            guard let transitInformation = transitInformation else { return false }
             transitInformation.forEach { (key, value) in
                 transformedTransitInformation[key.rawValue] = value
             }
             handleTransitInformation(transformedTransitInformation)
         case let .custom(pattern, match):
-            guard let match = match else { return }
+            guard let match = match else { return false }
```
zeitschlag, line 513:
```suggestion
            guard let match else { return false }
```

deltachat-ios/Chat/Views/MessageLabel.swift
```diff
@@ -452,49 +452,65 @@ open class MessageLabel: UILabel {
...
+            guard let date = date else { return false }
             handleDate(date)
         case let .link(url):
-            guard let url = url else { return }
+            guard let url = url else { return false }
             handleURL(url)
         case let .transitInfoComponents(transitInformation):
             var transformedTransitInformation = [String: String]()
-            guard let transitInformation = transitInformation else { return }
+            guard let transitInformation = transitInformation else { return false }
```
zeitschlag, line 507:
```suggestion
            guard let transitInformation else { return false }
```

deltachat-ios/Chat/Views/MessageLabel.swift
```diff
@@ -452,49 +452,65 @@ open class MessageLabel: UILabel {
...
-            guard let phoneNumber = phoneNumber else { return }
+            guard let phoneNumber = phoneNumber else { return false }
             handlePhoneNumber(phoneNumber)
         case let .date(date):
-            guard let date = date else { return }
+            guard let date = date else { return false }
             handleDate(date)
         case let .link(url):
-            guard let url = url else { return }
+            guard let url = url else { return false }
```
zeitschlag, line 503:
```suggestion
            guard let url else { return false }
```

deltachat-ios/Chat/Views/MessageLabel.swift
```diff
@@ -452,49 +452,65 @@ open class MessageLabel: UILabel {
...
+        case let .phoneNumber(phoneNumber):
+            return phoneNumber
+        case let .custom(_, match):
+            return match
+        case .addressComponents, .date, .transitInfoComponents:
+            return nil
+        }
+    }
+
+    internal func handleGesture(_ touchLocation: CGPoint) -> Bool {
```
zeitschlag, line 485: What's the idea behind the boolean return value (that isn't used, from what I see)?

deltachat-ios/Chat/ChatViewController.swift
r10s: that comes from switching 5+ languages all the time :)

deltachat-ios/Chat/Views/MessageLabel.swift
r10s: the return value indicates if the tap was consumed; this is also used that way

zeitschlag: Ah, nevermind, I didn't see it was there before, but in another method 🙈

zeitschlag (approved):

r10s:
CHANGLOG was missing,  added a commit

PR #2448, opened 2024-12-19
feat: different visual representation for pkpass file attachments and added the "add to apple wallet" button

Simon-Laux:
This year I again run into the issue that iOS does not allow you to import tickets/pkpass files form the native File app.
There are many reports on apples support sites, but nothing happened so far, so I assume it is a feature to them in some twisted way (that you can only add to wallet from API, Mail, Safari and iMessage).

This pr uses the PassKit Library to add an "Add to Apple Wallet" button to pkpass file attachments in messages.

image_2024-12-19_07-08-10

on click it opens the standard ticket/pass ui:
image_2024-12-19_07-08-08

The UI I came up with looks acceptable IMO, but I have to admit that my perfectionism is not statisfied.
Maybe you have easy ideas on how to improve it, like adding a small padding.
If so feel free to commit to this pr.

Also when you click on the message it opens the document preview which is nearly useless except for sharing/exporting the file.

### DIFF
--- a/deltachat-ios/Chat/Views/FileView.swift
+++ b/deltachat-ios/Chat/Views/FileView.swift
@@ -1,5 +1,7 @@
 import UIKit
 import DcCore
+import PassKit
+
 public class FileView: UIView {
 
     private var imageWidthConstraint: NSLayoutConstraint?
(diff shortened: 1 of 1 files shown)


r10s:
after little internal discussion among ios maintainer: 

closing, as this is no priority atm and it is not clear if maintainance vs impact is positive.

in general, it is not about getting as many features in as possible (that would be easy), it is about the right features we want to maintain, worth the few resources (this is hard)

Simon-Laux:
> closing, as this is no priority atm and it is not clear if maintainance vs impact is positive.

well the impact is probably negative, because I just hacked it in without doing it properly. (in other words my code here is not maintainable)

But the feature generally is useful - because there seems no easy way to import .pkpass files on iOS: neither the share sheet, quick view/preview, nor the Files app work. And there also seems to be no way to import files from the wallet app.

Afaik the only options are:
- send yourself an email to your iPhones apple mail
- download pkpass file from safari
- send it in a supported app like iMessage. (seems like all apps need to implement it by themselves)

-> in my case none of the 3 above worked (I don't have the normal Mail app set-up, I didn't had the link to website on my phone and iMessage shows a preview and add to wallet button, but you can not press it in the draft and there are no "saved messages" in iMessage unfortunately.)

So seems like apple is suggesting here that the messaging apps should implement it -> though I personally think apple should implement it in both the quick view (that we use), the Files app and the share sheet.

> it is about the right features we want to maintain,

Sure it is somewhat rare edge-case, but on the other hand we want to integrate with the apple system and make it feel native.
But I agree not right time and not mergeable in this hacky state.

Amzd:
> there are no "saved messages" in iMessage unfortunately

Off-topic but you can text your own number in iMessage, which acts kinda like saved messages.

PR #2451, opened 2024-12-19
Access chats from widget

zeitschlag:
This PR lets users access chats from widgets. Users can add and remove their most favorite chats to the widget so that they can access them more quickly.

For the time being the add/remove-option is hidden in the world-famous-3-dots-menu on chatlist and only shows up if one chat in Chatlist is selected (we could think about adding another place to add/remove chats, maybe chat details? LMK what you think, I have a stash flying around).

As discussed, the widsget gets prepopulated with Self-talk and Device-talk. (we should test this thoroughly.)

Chat-shortcuts in widget look like the icons in Chatlist, also to differ them from the App-shortcut (which are squares with rounded corners). Please enjoy some screenshots:

This PR also contains some refactorings (split widget-code in several files) and renames the widget to ✨ `DcWidget` ✨


r10s:
i know it's a draft, but i could not way to try that out ... ... very nice already! ❤️

r10s:
in the emulator, on a fresh install, there is only one icon prepopulated, a white one, tapping the icon gets you to a ghost chat (probably id 0)

https://github.com/user-attachments/assets/a6b1a6ea-6d64-43ea-a31b-ce22c367155c


might be related to the emulator, however, i have a hard time testing currently, as on the real devices, i was running the widget before, resulting in a weird state.

can we exchange all related keys somehow? at least for the UserDefaults, it should be easy to add a `v2` or so. has the widget also a key?

r10s (approved):
was testing that quite a bit, thanks for helping to reset my UserDefaults!

code also lgtm, nothing that hit my eyes :)

very nice!

PR #2452, opened 2024-12-21
Add/remove chats to widgets from chat-details

zeitschlag:
Simulator Screenshot - iPhone 16 Pro - 2024-12-21 at 20 06 35

### DIFF
--- a/deltachat-ios/ViewModel/ContactDetailViewModel.swift
+++ b/deltachat-ios/ViewModel/ContactDetailViewModel.swift
@@ -88,11 +88,9 @@ class ContactDetailViewModel {
             chatOptions.append(.locations)
         }
 
-        if #available(iOS 15, *) {
-            chatActions = [.addToHomescreen]
-        }
-
-        if chatId != 0 {
+        var chatActions: [ChatAction]
+        let chatExists = chatId != 0
+        if chatExists {
             if !isDeviceTalk {
                 chatOptions.append(.ephemeralMessages)
                 chatOptions.append(.startChat)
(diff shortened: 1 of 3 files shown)


deltachat-ios/ViewModel/ContactDetailViewModel.swift
```diff
@@ -87,6 +88,10 @@ class ContactDetailViewModel {
             chatOptions.append(.locations)
         }
 
+        if #available(iOS 15, *) {
+            chatActions = [.addToHomescreen]
+        }
```
r10s, line 93: this needs to go into the `if chatId != 0` branch below - as without a chatId, we cannot add/remove things from the widget

i would also move that _below_ "Archive", which is very probably by far the most often used action, and should be first

btw, this brings me to the question what happens when a chat is deleted - is the widget updated at least after some time? otherwise, we should file and issue for that

r10s (commented):
works for **contact profiles** having a one-to-one chat, but in case there is no chat, the option must not be added, see comment below

for **group profiles**, the option is still missing

(we have that code duplicated for $reasons on iOS, probably came out some "c'mon, it is just a handful lines" :)

deltachat-ios/ViewModel/ContactDetailViewModel.swift
zeitschlag: I didn't check or looked at the code how it should behave, so please feel free to tackle #2457 :) Should be fairly straight forward

zeitschlag: > `if chatId != 0` 

and 

> as without a chatId

means that `0` counts as "no chatId"?
I'll move the code. And search for the group-related stuff.

r10s (approved):

PR #2459, opened 2024-12-22
[WIP] ios 12 byyyyyyeeeeee

### LINKED ISSUE #2453: increase deployment target to iOS 13
iOS 12 is badly supported by xcode, also, we did not manage to get testing devices, so, this is sort of a blind flight :)

but even if we had testing devices, iOS 12 user base is pretty small, <1% - and becoming smaller. these users can stay with the existing, quite stable 1.50.3  - some hinted that it is even possible to get that versions for new installation on iOS 12, but idk.

changing in supported devices:

- having iOS 13 as minimal deployment target will drop support for iPhone 5s (2013), iPhone 6 (2014). iPhone 6s (2015) will be the oldest supported iPhone then
- no changes in "iPad Classic" support, oldest supported iPad will stay at iPad 5 as that got iOS 16
- for iPad Air, minimal is now "Air 2"
- for iPad Mini, minimal is now "Mini 4"

in theory, we could increase directly to iOS 15, as all devices that got iOS 13 also got iOS 15, however, we decided not to go for that now as the additional code simplification is comparable small -- and we do not really know about situations in countries that do not install from apple, there it may be harder to upgrade

**once this issue is closed, deltachat-pages needs to be updated to point to the changed minmal requirements**

zeitschlag:
Closes #2453


r10s:
getting the following error on building:

Screenshot 2025-01-03 at 17 09 42

zeitschlag:
> getting the following error on building:

As both me and the CI can build that thing without getting that error I'd assume that there's something off with your project. Have you tried `CMD+Shift+K`ing?

r10s (approved):

Issue #2461, opened 2024-12-26
widget: avatars not shown in "adaptive icon mode"


r10s:
when setting the system to use "adaptive icon mode", the icons inside the new widget are not shown:

<img width=320 src=https://github.com/user-attachments/assets/63f9d9b5-f22c-48f2-a1ea-18ca40e7ec11>

light/dark mode are fine, and the icons also work as expected - they're just not displayed

PR #2465, opened 2024-12-26
Add two chats to placeholder-widget

### LINKED ISSUE #2462: widget preview should also show "chats" as examples
this esp. makes sense as we prefix with "device messages" and "saved messages", so the first two could be these chats, the other ones the apps as now

zeitschlag:
Adds two round icons to the preview when adding a widget. (Feel free replace the assets, should be pretty straight forward)

Simulator Screenshot - iPhone 16 Pro - 2024-12-26 at 17 08 48

Closes #2462


DcWidget/Assets.xcassets/AccentColor.colorset/Contents.json
```diff
@@ -0,0 +1,20 @@
...
+        }
+      ],
+      "idiom" : "universal"
+    }
+  ],
+  "info" : {
+    "author" : "xcode",
+    "version" : 1
+  }
+}
```
r10s, line 20: how are these changes related? does that already target #2461 ?

DcWidget/WidgetProvider.swift
```diff
@@ -16,13 +16,12 @@ struct Provider: TimelineProvider {
...
-            .app(AppShortcut(accountId: 0, chatId: 0, messageId: 4, image: UIImage(named: "checklist"), title: "checklist")),
-            .app(AppShortcut(accountId: 0, chatId: 0, messageId: 5, image: UIImage(named: "hello"), title: "hello")),
-            .app(AppShortcut(accountId: 0, chatId: 0, messageId: 7, image: UIImage(named: "webxdc"), title: "webxdc")),
+            .app(AppShortcut(accountId: 0, chatId: 1, messageId: 1, image: UIImage(named: "hello"), title: "hello")),
+            .app(AppShortcut(accountId: 0, chatId: 2, messageId: 2, image: UIImage(named: "packabunchas"), title: "packabunchas")),
+            .chat(ChatShortcut(accountId: 0, chatId: 3, title: "Messages to self", image: UIImage(named: "saved-messages"))),
+            .app(AppShortcut(accountId: 0, chatId: 4, messageId: 4, image: UIImage(named: "pixel"), title: "pixel")),
+            .app(AppShortcut(accountId: 0, chatId: 5, messageId: 5, image: UIImage(named: "webxdc"), title: "webxdc")),
+            .chat(ChatShortcut(accountId: 0, chatId: 6, title: "Broadcast", image: UIImage(named: "broadcast"))),
```
r10s, line 24: i would have used  Device Messages" instead as (a) broadcasts are experimental and the user cannot create them easily and (b) this is also very close to the widget that gets created initially

r10s (approved):
very nice! one question and one suggestion, but these are no blockers :)

PR #2466, opened 2025-01-02
Show chat-list for invalid deeplinks (#2464)

### LINKED ISSUE #2464: never open invalid chat/app object from widget
i am not sure what i did, but i got an icon in the widget that does not refer to an existing chat/app. tapping that, shows sort of a "ghost chat":

https://github.com/user-attachments/assets/670a3ad6-22b5-49d8-a718-4cc5e399a860

however, assuming that can happen anyways under some circumstances, races, sync, whatever, we should _not_ open invalid chat objects or invalid app objects.

instead, just open the chatlist (or the last state of the app), that is probably also good enough in case an app is deleted. an error or sth like that seems not to be needed, at least not for the first iteration -- it is already clear by the widget that "sth is wrong" - and also otherwise that chatlist (resp. the last state of the app) is opened

(technically, for historical reasons, always a chat object is returnd - but only if isValid() returns true, it can be displayed)

zeitschlag:
In case of an invalid chat/message, we show the chatlist. Closes #2464.

### DIFF
--- a/deltachat-ios/Coordinator/AppCoordinator.swift
+++ b/deltachat-ios/Coordinator/AppCoordinator.swift
@@ -217,7 +217,12 @@ class AppCoordinator: NSObject {
             }
         }
 
-        showChat(chatId: chatId, animated: false, clearViewControllerStack: true)
+        let chat = dcAccounts.getSelected().getChat(chatId: chatId)
+        if chat.isValid {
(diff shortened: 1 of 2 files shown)


r10s (commented):
ah, here we go :) lgtm, thanks a lot!

PR #2470, opened 2025-01-02
Remove shortcuts when removing chats/apps

### LINKED ISSUE #2457: When deleting a chat, remove that from the Widget
Also: Message, see: `dcContext.deleteMessage`

zeitschlag:
Closes #2457


r10s (approved):
lgtm, these are all cases that come into mind :)

still, as a fallback, i think, we should make sure we do not open a "ghost" chat and check if the chat-id we received from the widget is actually valid. but that can be another PR (or maybe there is sth already, i go through the PR sequencially :)

PR #2477, opened 2025-01-04
Fix no notification and 30 sec stale app when NSE crashed in background from running out of memory

Amzd:
When receiving a notification for a message over ~5mb, the background fetch crashes because the Notification Service Extension is not allowed over 24mb of memory by the system. (and 12mb on older iOS)

This crash means no notification and on launch the app stays stale for 30 seconds because nseFetching user default was not set to false. 

This PR fixes both by just sending the best attempt notification when memory is critical which is just "You received messages" and refactoring nseFetching to never return true for more than 30 seconds since it was set. Now the app is only stale for 26 seconds after the crashed NSE launched, instead of 30 sec after app launch.

To test I send an 8mb GIF (attached) after force terminating the app to make sure it's not still running in background.
 
livekindly-vexquisitstudiohq

### DIFF
--- a/DcCore/DcCore/Extensions/UserDefaults+Extensions.swift
+++ b/DcCore/DcCore/Extensions/UserDefaults+Extensions.swift
@@ -27,19 +27,20 @@ public extension UserDefaults {
 
         if until > Date().addingTimeInterval(30) {
             // user changed their system clock so we reset nse fetching date
-            setNseFetching(until: Date().addingTimeInterval(30))
+            setNseFetching(for: 30)
         }
 
         return until > Date()
     }
 
-    /// Set the time until which the Notification Service Extension could be fetching. Call with nil or Date() when NSE fetching is done.
-    ///
-    /// Note: Do not call with date further than 30 seconds into the future. NSE fetching should not take more than 30 seconds.
-    static func setNseFetching(until: Date?) {
-        precondition(until == nil || until! <= Date().addingTimeInterval(30),
-                     "Don't set NSE fetching for more than 30 seconds")
-        shared?.set(until?.timeIntervalSince1970, forKey: nseFetchingKey)
+    /// Set the amount of seconds for which the Notification Service Extension could be fetching. Call setNseFetchingDone when done.
+    static func setNseFetching(for seconds: Double) {
+        assert(0...30 ~= seconds, "Don't set NSE fetching for more than 30 seconds")
+        shared?.set(Date().timeIntervalSince1970 + seconds, forKey: nseFetchingKey)
+    }
+
+    static func setNseFetchingDone() {
+        shared?.set(nil, forKey: nseFetchingKey)
     }
 
     static func pushToDebugArray(_ value: String) {
--- a/DcCore/DcCore/Extensions/UserDefaults+Extensions.swift
+++ b/DcCore/DcCore/Extensions/UserDefaults+Extensions.swift
@@ -20,12 +20,26 @@ public extension UserDefaults {
         shared?.setValue(value, forKey: mainIoRunningKey)
     }
 
+    /// Check if we are currently fetching using the Notification Service Extension. Never returns true for more than 30 seconds.
     static var nseFetching: Bool {
-        return shared?.bool(forKey: nseFetchingKey) ?? false
+        guard let shared else { return false }
+        let until = Date(timeIntervalSince1970: shared.double(forKey: nseFetchingKey))
+
+        if until > Date().addingTimeInterval(30) {
+            // user changed their system clock so we reset nse fetching date
+            setNseFetching(until: Date().addingTimeInterval(30))
+        }
+
+        return until > Date()
     }
 
-    static func setNseFetching(_ value: Bool = true) {
-        shared?.setValue(value, forKey: nseFetchingKey)
+    /// Set the time until which the Notification Service Extension could be fetching. Call with nil or Date() when NSE fetching is done.
+    ///
+    /// Note: Do not call with date further than 30 seconds into the future. NSE fetching should not take more than 30 seconds.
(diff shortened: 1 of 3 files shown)


Amzd:
I found that even though the NSE can't decrypt the message initially, the background service later on can decrypt so you get a "You have new messages" when the message is received first and sometimes like 10 mins later you can get a "GIF - message" notification. Which I think is OKAY.

DcNotificationService/NotificationService.swift
```diff
@@ -14,21 +14,36 @@ class NotificationService: UNNotificationServiceExtension {
...
-        if !dcAccounts.backgroundFetch(timeout: 25) {
-            UserDefaults.pushToDebugArray("ERR3")
-            UserDefaults.setNseFetching(false)
+        // Send the bestAttempt notification when memory is critical because the process will be killed
+        // by the system soon and any notification is better than nothing.
+        var exitedDueToCriticalMemory = false
+        let memoryPressureSource = DispatchSource.makeMemoryPressureSource(eventMask: .critical)
+        memoryPressureSource.setEventHandler {
+            // Order of importance because we might crash very soon
+            // TODO: Could notify with "received message too big to decrypt in background"
```
r10s, line 31: usually, we avoid "TODO" in the source code, this usually stays forever. in this case, i also think "New Messages" is totally fine, there are also some other situations where this may happen

```suggestion
```

DcNotificationService/NotificationService.swift
```diff
@@ -14,21 +14,36 @@ class NotificationService: UNNotificationServiceExtension {
...
             contentHandler(bestAttemptContent)
+            exitedDueToCriticalMemory = true
+            UserDefaults.pushToDebugArray("ERR5")
+        }
+        memoryPressureSource.activate()
+
+        guard dcAccounts.backgroundFetch(timeout: 25) && !exitedDueToCriticalMemory else {
```
r10s, line 38: i find that hard to read, also not sure if it is correct (if backgroundFetch succeeds, exitedDueToCriticalMemory is ignored)

i would expect the following:

```suggestion
        if !dcAccounts.backgroundFetch(timeout: 25) || exitedDueToCriticalMemory {
```

DcNotificationService/NotificationService.swift
```diff
@@ -14,21 +14,36 @@ class NotificationService: UNNotificationServiceExtension {
...
+            UserDefaults.pushToDebugArray("ERR3")
+            UserDefaults.setNseFetching(until: nil)
+            if !exitedDueToCriticalMemory {
+                contentHandler(bestAttemptContent)
+            }
```
r10s, line 43: there are two more calls to contentHandler() below that we may want to protect the same way

(if we want to do so, i would just repeat 

```
            if !exitedDueToCriticalMemory {
                contentHandler(bestAttemptContent)
            }
``` 

there, factoring out to a function or so may lower understandability of the code, kiss :)

no blocker, tho

r10s (approved):
thanks a lot for pushing that forward and to dive into that complicated issue 🙏

the approach is great all in all, some comments only, esp. the "guard" thingie

DcCore/DcCore/Extensions/UserDefaults+Extensions.swift
```diff
@@ -20,12 +20,26 @@ public extension UserDefaults {
         shared?.setValue(value, forKey: mainIoRunningKey)
     }
 
+    /// Check if we are currently fetching using the Notification Service Extension. Never returns true for more than 30 seconds.
     static var nseFetching: Bool {
-        return shared?.bool(forKey: nseFetchingKey) ?? false
+        guard let shared else { return false }
+        let until = Date(timeIntervalSince1970: shared.double(forKey: nseFetchingKey))
```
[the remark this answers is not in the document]
Amzd: Yep, it will either be 0 or 1, which will basically be 1970. Making a new key just leaks the old key:value unless we also set that to nil.

DcCore/DcCore/Extensions/UserDefaults+Extensions.swift
```diff
@@ -20,12 +20,26 @@ public extension UserDefaults {
...
 
-    static func setNseFetching(_ value: Bool = true) {
-        shared?.setValue(value, forKey: nseFetchingKey)
+    /// Set the time until which the Notification Service Extension could be fetching. Call with nil or Date() when NSE fetching is done.
+    ///
+    /// Note: Do not call with date further than 30 seconds into the future. NSE fetching should not take more than 30 seconds.
+    static func setNseFetching(until: Date?) {
+        precondition(until == nil || until! <= Date().addingTimeInterval(30),
+                     "Don't set NSE fetching for more than 30 seconds")
+        shared?.set(until?.timeIntervalSince1970, forKey: nseFetchingKey)
```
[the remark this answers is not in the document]
Amzd: You are right, precondition should be assert.

I will split this into two functions.

DcNotificationService/NotificationService.swift
Amzd: guard implies it does not continue unless the values are true, which I think makes it more readable. Previously there was an if statement with a return, which is A hard to read and B is not compile safe (the return can be deleted in the future)

the guard basically means "if bgfetch returns true and not exitedDueToCriticalMemory, continue, else return in this block"

so exitedDueToCriticalMemory is not ignored.

DcNotificationService/NotificationService.swift
Amzd: it will never get there if it exits due to critical memory (assuming this exit happens after the fetch background, which I think is safe because that should clean up after itself)

r10s (approved):
very nice! thanks again, also for the recent changes 🙏

PR #2480, opened 2025-01-05
Prevent automatic limited access alert

Amzd:
When sending an image from your gallery the first time each run the system can ask if you want to give the app access to more photos (see image). Delta Chat doesn't need access to photos so this popup is useless. This PR disables the automatic popup as per the docs: 

> Add the PHPhotoLibraryPreventAutomaticLimitedAccessAlert key with a Boolean value of true to your app’s Info.plist file to prevent the system from automatically presenting the limited library selection prompt.
https://developer.apple.com/documentation/photos/phauthorizationstatus/limited

We can still ask permission manually in the future with `PHPhotoLibrary.presentLimitedLibraryPicker(from:)` even with this plist entry.

IMG_3213

How to test:
before: Force Quit > Send image from Gallery > receive popup
after: Force Quit > Send image from Gallery > no popup

### DIFF
--- a/deltachat-ios/Info.plist
+++ b/deltachat-ios/Info.plist
@@ -74,6 +74,8 @@
 	<string>Delta Chat wants to save images to your photo library.</string>
 	<key>NSPhotoLibraryUsageDescription</key>
 	<string>Delta Chat will let you choose which photos from your library to send.</string>
+	<key>PHPhotoLibraryPreventAutomaticLimitedAccessAlert</key>
+	<true/>
 	<key>NSUserActivityTypes</key>
 	<array>
 		<string>INSendMessageIntent</string>


r10s:
> How to test:
> before: Force Quit > Send image from Gallery > receive popup
> after: Force Quit > Send image from Gallery > no popup

this does not work for me on iOS 18 - instead, it seems, not showing a question is kind of default behaviour there, "Global Settings -> Apps -> Delta Chat -> Access -> Photos" shows the following for me:

IMG_7779

and a similar warning was atop of the image picker as well (i discarded that, so i cannot screenshot it again)

r10s (approved):
even though iOS 18 may not need it, the PR seems simple enough and is doing what is recommended by apple, so let's get that in :)

Amzd:
> it seems, not showing a question is kind of default behaviour [on iOS 18]

That is good to hear, it is very annoying on iOS 16

PR #2481, opened 2025-01-05
Reduced motion on chat messages

Amzd:
Reduce the amount of motion the chat messages receive.

| Before | After |
|---|---|
| <video src="https://github.com/user-attachments/assets/6479febf-a87b-43a0-9c75-d01fcaab32c7"> | <video src="https://github.com/user-attachments/assets/191deacd-24b0-43a7-a3f8-12df9d6d27fc"> |
| <video src="https://github.com/user-attachments/assets/cc63c97f-84b8-4bdd-bc0d-586f3564c403"> | <video src="https://github.com/user-attachments/assets/7e81fa9b-7c73-47c6-8c93-3786602c3f0d"> |

This is possible because willShow is always called, even if keyboard hides, because the system will show the accessory.

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -276,11 +276,6 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
                 }
             }
         }
-        keyboardManager?.on(event: .willHide) { [tableView = tableView!, inputAccessoryView] notification in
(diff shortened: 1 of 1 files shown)


r10s:
very nice -- improvements by removing lines of code * mjam * 😋

r10s (approved):
very nice! tested on iOS 18 - and works as expected.

this is now so good, i would rething the idea of re-doing the UITableViewController - additional reason was the upcoming pinned messages, but that probably could also be solved by a static subview atop.

let's see - for now: sooooooo gooood! ❤️

PR #2482, opened 2025-01-05
QuickLook swipe down and less message motion when coming back from GiveBackMyFirstResponder

Amzd:
This PR makes the image preview (quicklook) able to be dismissed by swiping down, instead of the default popGestureRecognizer. This is an improvement because A; the popGesture is only enabled when the navigation bars are shown, B; it interferes with swiping to previous image, and C; the animation feels good.

For this to look good I had to reduce the motion on messages when presenting/dismissing view controllers some more. Now after a view is dismissed the scroll updating is delayed by 0.5s which means that first; the ChatViewController can become firstResponder (bringing up the accessory) and then; the text field can become firstResponder (bringing up the keyboard) without the chat scrolling.

### Quick Look Dismiss (This change is only pre-iOS 18)
| Before | After |
|---|---|
| <video src="https://github.com/user-attachments/assets/9b17775b-093d-4084-aee8-727470d1e5d7"> | <video src="https://github.com/user-attachments/assets/5c697a64-c92d-4774-84b7-0e0679e9554e"> |

### Reduced motion (This change is everywhere)
| Before | After |
|---|---|
| <video src="https://github.com/user-attachments/assets/dfbad789-8e7a-45db-afb9-8d5d7c621445"> | <video src="https://github.com/user-attachments/assets/8ef75b86-d39e-45a4-82f2-4b88b4172fa8"> |

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -202,8 +202,8 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
         set { _bag = newValue }
     }
 
-    private var previewControllerTargetSnapshots: [UIView] = []
-    private var previewControllerTargetHiddenOriginals: [UIView] = []
+    private var previewControllerTargetSnapshot: UIView?
+    private var previewControllerTargetHiddenOriginal: UIView?
 
     init(dcContext: DcContext, chatId: Int, highlightedMsg: Int? = nil) {
         self.dcContext = dcContext
@@ -2625,11 +2625,13 @@ extension ChatViewController: QLPreviewControllerDelegate {
             // pop back in from, when firstResponder is returned.
             snapshot.frame.origin.y = (tableView.superview ?? tableView).frame.maxY
             tableView.superview?.addSubview(snapshot)
-            previewControllerTargetSnapshots.append(snapshot)
+            previewControllerTargetSnapshot?.removeFromSuperview()
+            previewControllerTargetSnapshot = snapshot
             return snapshot
         } else if let msgId = item.messageId, let row = messageIds.firstIndex(of: msgId) {
             if let cell = tableView.cellForRow(at: IndexPath(row: row, section: 0)) as? BaseMessageCell,
                let snapshot = cell.messageBackgroundContainer.snapshotView(afterScreenUpdates: false) {
+                previewControllerTargetHiddenOriginal?.layer.opacity = 1
                 if cell is ImageTextCell { // hide cell while transitioning
                     cell.layer.opacity = 0
                 }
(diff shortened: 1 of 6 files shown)


r10s:
hm, i am not sure i got the gist of this PR completely.

- swipe-down-to-exit-gallery worked also before, and at least on iOS 18, even smoother:
    - really only the image is swiped down, not the whole screen including black borders - this looks esp weird for landscape images
    - no 1-second-delay before the message view with the image is shown again

- opening the gallery without any controls to share etc. may be unexpected: none of whatsapp or telegram do that. the controls only disappear on a second tap

- the message view scrolling in-sync with the gallery left/right: not sure if that is expected by the user or not - whatsapp does it, telegram does not

- the reduced motion is nice

however, always good to have options, so thanks for diving into that. and as said, sorry, if i missed sth.

Amzd:
TODO: on iOS 18 the keyboard is hidden after viewing a photo, can be fixed by a GiveBackMyFirstResponder extension for `UINavigationController.pushViewController(_: QLPreviewController)` that does the same as `UIViewController.present(_: QLPreviewController)`

Amzd:
Scratch that, that opened a massive can of worms. On iOS 18 the keyboard is just hidden when you click on an image in chatted it does not come back automatically, this is a fine UX I think.

Amzd:
Can take another look at iOS 18 having hidden keyboard after QLQuickLook if we want to merge #2536 because swizzling will probably have no negative impacts when pushing either.

r10s:
this needs a rebase - is this PR still needed or conflicting with https://github.com/deltachat/deltachat-ios/pull/2536 ?

Amzd:
I will take another look at this after merge of the other PR

Amzd:
Rebased and checked again for iOS 18 but couldn't get the "return keyboard after preview dismiss" to work with `pushViewController` (which we use because since iOS 18 when you `present(_::)` a QLPreviewController it doesn't show navigation and toolbars by default).

r10s:
swipe-to-close is now indeed smooth again on iOS 18 (it was also smooth before), however, opening is worse:

for a ~second, a weird title bar is shown, with whitespace atop:

Screenshot 2025-01-29 at 15 55 19

<details>
<summary>video</summary>

https://github.com/user-attachments/assets/de0f7905-cf23-4bde-86d2-b492dbc307b8

</details>

it is not super-bad, but it looks as if something is wrong.

and, without this PR, the opening does not have the weird effect.

did not test iOS 15 yet

r10s:
things got weirder: force-restarting the app, opening is fine :)

so maybe related to some system hickup.

so, in general, things look good on iOS 18

@zeitschlag as more into iOS, can you have a look at the code and test on your phones as well?

r10s:
closing stale PR and put it to resurrection, as this makes things only different, creating different bugs.

i think, at some point we need to iterate over this, but just now the focus is differently on multi-transport and related things

Issue #2478, opened 2025-01-04
The New Chat > Search looks weird if you don't have contacts.


Amzd:
Is this a result of #2474?

- iOS version: iOS 16
- Device: iPhone X
- Delta Chat version: master
- Expected behavior:
- Actual behavior:
- Steps to reproduce the problem: New Chat > Search
- Screenshots:
![IMG_3204](https://github.com/user-attachments/assets/899fe1ca-cdd8-427e-9656-2e7a9964dd4e)
- Logs:

r10s:
> Is this a result of https://github.com/deltachat/deltachat-ios/pull/2474?

no, i do not think so. the `NewChatViewController` was not touched.

but maybe it should also get the `.insetGroup` style so that the buttons are more outstanding. but that would not fix the indeed unfitting layout of the error message (i assume this is what the issue is about?)

Amzd:
Yea it seems weird the "no results found" is in a rounded box (when everything else there is squares) and that it has no padding

r10s:
> In my opinion, the design of the app would be more consistent if the New Chat screen also used the new style of the settings screen.

makes sense! i created a PR for that at https://github.com/deltachat/deltachat-ios/pull/2483

this at least adds adds unifies the rounded corners, so that maybe just some padding for the "nothing found" warning is needed to close this issue

PR #2483, opened 2025-01-05
even more modern settings pages 

r10s:
use .insetGrouped also for 'new chat',  'add members…' and 'add contact manually'; each of them may have buttons, which look better. also it is much more consistent now. the only .grouped style left is the main chatlist

successor of https://github.com/deltachat/deltachat-ios/pull/2474

targets https://github.com/deltachat/deltachat-ios/issues/2478


Amzd (approved):

r10s:
> Does this PR also affect the profile information screen of the contacts?

no, this was already handled at #2474

PR #2486, opened 2025-01-07
widget wordings

### LINKED ISSUE #2439: Localize widget-strings
- Move widget-related strings from `untranslated.xml` to `strings.xml` in android-repo

r10s:
this PR changes the wording for the widgets from "Add to Home screen" to "Add to Widget", same for "remove".

moreover, the description is changed to be more actionable, and, as a side effect, does not need to mention apps:

<br><br><br>

this is how the buttons look like now:

<br><br><br>

and finally, if all chats and apps are removed, the widget shows the following hint - looks a but handwavy, but usually, the user does not see the hint here at all as the widget comes prefilled with the chats "Saved Messages" and "Device Messages":

<br><br><br>

targets #2439 , once this is merged, we can add the new strings to android to make them translatable and finally close #2439

### DIFF
--- a/DcWidget/DcWidget.swift
+++ b/DcWidget/DcWidget.swift
@@ -7,7 +7,7 @@ struct DcShortcutWidgetView: View {
 
     var body: some View {
         if entry.shortcuts.isEmpty {
-            Text(String.localized("ios_widget_apps_description"))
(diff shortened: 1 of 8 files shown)


zeitschlag (approved):
I have some doubts about the description, but I won't fight that. So I'll approve.

PR #2487, opened 2025-01-07
end chatlist's editing mode on widget action

r10s:
similar to mute, pin, etc.,
it seems expected that the editing mode is ended once 'Add to Widget' or 'Remove from Widget' is tapped in the chatlist.

this also gives visual feedback,
that is missing without.


zeitschlag (approved):
Good catch, thank you.

PR #2504, opened 2025-01-10
Fix for scrolling issue and keyboard resign when starting an interactivePopGesture but canceling it.

Amzd:
Previously when canceling interactive pop gesture it would scroll back to bottom (or to new messages indicator) and dismiss the keyboard. There is still some lag that happens when you cancel but that is due to recalculation of the navbar title (which might not be needed but I don't know the edge cases).

| Before | After | 
| --- | --- | 
| <video src="https://github.com/user-attachments/assets/167d2f05-bae7-4d52-847e-102aa034193f"> | <video src="https://github.com/user-attachments/assets/f3ad2f00-6586-4729-b8de-537ddcd77862"> |


r10s (approved):
the iOS keyboard never gets boring :)

PR #2505, opened 2025-01-10
webp support for pasting images

Amzd:
Previously you could not paste webp images which could be really confusing:

- Saved an image from Safari to Photos
- In Photos long press it and "Copy Photo"
- Try to paste in Delta Chat
- No "Paste" shows up

At no point will the user have seen it is a webp image and pasting it was not supported, and they might have thought DeltaChat did not support pasting images at all (I did for a while).

I added it in a way that we can add more image types that are not directly supported when we find them.


r10s (approved):
thanks for diving into that issue, and thanks for the clean fix!

did not test, however, the code looks easy enough (and i already received a .webp pasted from you)

Amzd:
🎉 

I should add that it does not send as web, but as jpg

PR #2506, opened 2025-01-10
force a name on editing profile

### LINKED ISSUE #2243: force non-empty profile name also on editing
during onboarding, we enforce a profile name to be set.

for consistency, and [as same reasons apply](https://github.com/deltachat/deltachat-android/issues/3160#issuecomment-2226887687), we should not allow to clear the name later on editing.

there are old profiles where we did not have enforced a name on creation. we would enforce them now during editing, if eg. the user only wants to edit the "bio". however, that seems to be a minor, and also android does it that way since some time, without much complains or issues.

cmp https://github.com/deltachat/deltachat-android/issues/3160, https://github.com/deltachat/deltachat-desktop/issues/4042

r10s:
this PR adds "Cancel" and "Done" buttons to the edit profile screen, where "Done" is only tappable when actually a name is  entered.

if no name is entered the placeholder shows a hint to enter a name - this should be sufficient as the user was actually deleting the name and sees that "Done" is no longer possible. on entering the screen,  usually, the name is set as we also force a name elsewhere, see issue for reasoning

as a side effect, whole dialog is no easier accessible, one can tap around and try out things, and can always "Cancel" things.

this is also what is known by WhatsApp/Signal/Others

 &nbsp; &nbsp; 

closes #2243

### DIFF
--- a/deltachat-ios/Controller/Settings/SelfProfileViewController.swift
+++ b/deltachat-ios/Controller/Settings/SelfProfileViewController.swift
@@ -11,6 +11,14 @@ class SelfProfileViewController: UITableViewController, MediaPickerDelegate {
 
     private let dcContext: DcContext
 
+    lazy var doneButton: UIBarButtonItem = {
+        return UIBarButtonItem(barButtonSystemItem: .done, target: self, action: #selector(doneButtonPressed))
+    }()
+
+    lazy var cancelButton: UIBarButtonItem = {
+        return UIBarButtonItem(barButtonSystemItem: .cancel, target: self, action: #selector(cancelButtonPressed))
+    }()
+
     private lazy var mediaPicker: MediaPicker? = {
         let mediaPicker = MediaPicker(dcContext: dcContext, navigationController: navigationController)
         mediaPicker.delegate = self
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/Settings/SelfProfileViewController.swift
```diff
@@ -47,10 +56,16 @@ class SelfProfileViewController: UITableViewController, MediaPickerDelegate {
...
 
     init(dcAccounts: DcAccounts) {
-        self.dcAccounts = dcAccounts
         self.dcContext = dcAccounts.getSelected()
         super.init(style: .insetGrouped)
         hidesBottomBarWhenPushed = true
+
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(SelfProfileViewController.textDidChangeNotification(notification:)),
```
Amzd, line 65: I don't mind it but ftr you don't have to be this verbose. This is also possible I think.
```suggestion
            selector: #selector(textDidChangeNotification),
```

Amzd (approved):

deltachat-ios/Controller/Settings/SelfProfileViewController.swift
r10s: cool! did not knew that (just did copy+paste :)

PR #2509, opened 2025-01-10
Drag image

Amzd:
You can now drag images from the chat, anywhere (eg different chat, or different app).

https://github.com/user-attachments/assets/0b7fdf56-d4a0-4295-b815-caca97f6621e


r10s (approved):
wow! that's impressive!

and a straight-forward thing, so there seems not to be many reasons to not do that! even if not everyone knows about that, it's easy enough :) ftr, i was first sceptical with collusions with the context menu, but that is perfectly fine.

r10s:
Delta Chat becomes on of the best-integrated apps, step by step :)

PR #2510, opened 2025-01-10
avoid scrolling context menu

r10s:
this PR avoids scrolling context menu as made more probable by adding "Copy Image" at https://github.com/deltachat/deltachat-ios/pull/2508

avoid scrolling is realised by moving 'Info' to the 'More Options / ...' menu - 'info' is currently anyways more a debug view, and it is good to have that a bit more away

this is how the "largest" menu looks on one of the smaller phones (iphone 13 mini):

 &nbsp; &nbsp; 

still, not sure if "Copy Image" is that important, esp. as also available via "Share", however, with this PR menus are not larger than before #2508 -- and we can always resort and iterate


Amzd (approved):
LGTM

Issue #2335, opened 2024-10-24
allow setting a "private tag" in the profile switcher


r10s:
see https://github.com/deltachat/deltachat-android/pull/3373 and https://github.com/deltachat/deltachat-desktop/issues/4249

PR #2511, opened 2025-01-11
set "tags" on profiles, mute and sort profiles

r10s:
this PR replaces the homebrew "Edit" option (that was always a bit weird, eg. deletion was started by tapping the whole cell, not only the icon) by a standard context menu, allowing to delete profiles - as well as editing the new "private tag".

this long-tap is also what we're using elsewhere in the app, so it should be easy enough for users to figure out - also elsewhere, eg. for deleting messages, we do not have an "Edit" button.

~~this also paves the way for adding a "mute" directly there (as on android) and maybe other things. when it comes to reordering, we may want to have an "Edit" back, but we're not there :)~~
it was only a few lines, i added "mute profile" as well as "sort to top" (from #2519) directly to this PR (despite lots of added functionality, still more lines deleted than added :) 

moreover, the "Add Profile" button becomes a standard button, including standard paddings and layout

closes https://github.com/deltachat/deltachat-ios/issues/2335

editing tags:

 &nbsp; 

this is how the tags look in the profile switcher:

### DIFF
--- a/deltachat-ios/DC/DcAccount.swift
+++ b/deltachat-ios/DC/DcAccount.swift
@@ -55,10 +55,9 @@ public class DcAccounts {
     }
 
     public func moveToTop(id: Int) {
-        var maxOrder = 0
-        getAll().forEach {
-            maxOrder = max(maxOrder, get(id: $0).getConfigInt("ui.ios.account_order"))
-        }
+        let maxOrder = getAll()
+            .compactMap { get(id: $0).getConfigInt("ui.ios.account_order") }
+            .max() ?? 0
         get(id: id).setConfigInt("ui.ios.account_order", maxOrder + 1)
     }
 
--- a/deltachat-ios/DC/DcAccount.swift
+++ b/deltachat-ios/DC/DcAccount.swift
@@ -45,6 +45,23 @@ public class DcAccounts {
         return DcUtils.copyAndFreeArray(inputArray: cAccounts)
     }
 
+    public func getAllSorted() -> [Int] {
+        return getAll().sorted { a, b in
+            // no need to check for equality as sorted() is guaranteed to be stable;
+            // meaning it preserves the relative order of elements that compare as equal here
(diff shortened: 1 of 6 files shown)


deltachat-ios/DC/DcAccount.swift
```diff
@@ -45,6 +45,23 @@ public class DcAccounts {
...
+            let orderA = get(id: a).getConfigInt("ui.ios.account_order"), orderB = get(id: b).getConfigInt("ui.ios.account_order")
+            return orderA > orderB
+        }
+    }
+
+    public func moveToTop(id: Int) {
+        var maxOrder = 0
+        getAll().forEach {
+            maxOrder = max(maxOrder, get(id: $0).getConfigInt("ui.ios.account_order"))
+        }
```
zeitschlag, line 61: what about something like this:

```swift
let highestIndex = getAll()
    .compactMap { get(id: $0).getConfigInt("ui.ios.account_order") }
    .max() ?? 0
get(id: id).setConfigInt("ui.ios.account_order", highestIndex + 1)
```

instead of `forEach`?

zeitschlag (approved):
The second line is cut off (as you can see for `irene test` in your third screenshot). I could imagine the the height of the cell is set to a constant somewhere? You could let Autolayout take care of the cell height or increase it manually?

Also we could think about refactoring the `AccountCell` into its own file.

deltachat-ios/DC/DcAccount.swift
r10s: yeah, maybe more modern - and one line less :)

r10s:
> The second line is cut off (as you can see for irene test in your third screenshot). I could imagine the the height of the cell is set to a constant somewhere? You could let Autolayout take care of the cell height or increase it manually?

yes, the height is probably set somewhere to something, however, i could fix it by removing some spacing at https://github.com/deltachat/deltachat-ios/pull/2511/commits/f8419462e2ca080982a4d395c232f6d9df42259e - i think, things even look better without

PR #2517, opened 2025-01-12
improve layout of 'no contact found' message

### LINKED ISSUE #2478: The New Chat > Search looks weird if you don't have contacts.

r10s:
this PR improves the bit weird layout of the message "No results found for FOO" when searching for contacts and nothing is found

closes #2478

this is how it looks now, nothing special (see #2478 for "before"):

 &nbsp; &nbsp;

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - "Message Info" is moved to messages' context menu at "More Options / ..." (#2510)
 - Allow cancelling profile edits, force a name when editing profile (#2506)
 - Fix scrolling issue when cancelling a screen (#2504)
+- Fix layout of 'No contacts found' message (#2517)
 
 
 ## v1.50.4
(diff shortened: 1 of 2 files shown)


Amzd:
Hmm, the scrolling is pretty bad, I don't think this was introduced by this PR but the padding does make it more obvious. Maybe a followup PR is fine. 

Side note: I really hate the scrollToTop extension because it doesn't actually scroll to the top.

https://github.com/user-attachments/assets/8722a825-80fe-4270-94cf-9de7c9f6e531

r10s:
> Maybe a followup PR is fine.

yeah, these small things should better be done in a PR directly. i would not file an issue for it, it is maybe not perfect layout, but it is also not a bug

note, that at some point this controller anyways gets an iteration as the search is not using the core but a ui-filter, making adaptions harder, see https://github.com/deltachat/deltachat-core-rust/pull/2411

PR #2519, opened 2025-01-13
change order of profiles

r10s:
_seems iOS is the first UI that allows profile reordering :)_

this PR adds a "Move to Top" to the new profiles context menu, picking up the idea we had with @adbenitez for android but that never was implemented :)

most ppl only have few accounts, so that 10-sth-line approach should be perfectly fine then - and if you have more accounts, you will figure out how to use it :)

a drag-to-order would be nice as well, but that is out of scope for me currently :)

for storage, a `ui.*` is used, i was first thinking about adding sth to core, however, it seems even better if that part is independent. otoh, it seems good if order is preserved within UI on export/import or so

merges into #2511

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - "Message Info" is moved to messages' context menu at "More Options / ..." (#2510)
 - To easier differ between multiple profiles, set a "Private Tag" (long tap profile switcher) (#2511)
 - In profile switcher's context menu, you have quick access to "Mute Notifications" for a profile (#2511)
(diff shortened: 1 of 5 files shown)


Amzd:
I agree drag to reorder would be a lot nicer. But this is better than nothing

Amzd (approved):

r10s:
if there is sth in core, nice, we could switch. but this was really few effort and few lines of code, for raw sorting max 10 lines 
- most of the 20 loc are ui things ans strings :)

PR #2521, opened 2025-01-14
show context menu on long-tapping proxies

r10s:
this PR shows a context menu with the options 'Share' and 'Delete' when a proxy item is long-tapped.

historicaly, swiping a row left/right
are the default way where ppl expect actions,
esp. as context menus only were added on iOS 13.

however, it happend to me regulary, that i try to long tap - and i have seen that also by others that are using the app, seems long-tap is quite arrived on iOS,
so, let's just support both ways
(as we stop supporting iOS 12, these menus are much more accessible for devs :)

moreover, this PR streamlines the behaviour of the switch, and avoids accidental long tapping the row and avoids wrong expectations when using switches elsewhere (in contrast to Android, on iOS it is more usual to let the user use the concrete switch, not the whole row)

### DIFF
--- a/deltachat-ios/Controller/Settings/Proxy/ProxySettingsViewController.swift
+++ b/deltachat-ios/Controller/Settings/Proxy/ProxySettingsViewController.swift
@@ -297,7 +297,6 @@ extension ProxySettingsViewController {
             }
         }
         shareAction.backgroundColor = .systemGreen
-        shareAction.accessibilityLabel = String.localized("proxy_share_link")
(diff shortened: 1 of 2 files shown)


deltachat-ios/Controller/Settings/Proxy/ProxySettingsViewController.swift
```diff
@@ -297,7 +297,6 @@ extension ProxySettingsViewController {
             }
         }
         shareAction.backgroundColor = .systemGreen
-        shareAction.accessibilityLabel = String.localized("proxy_share_link")
```
r10s, line 300: that was a little bit wrong before, see second commit for details

zeitschlag (approved):

PR #2499, opened 2025-01-09
App Picker

zeitschlag:
Adds an App Picker to download apps from webxdc.org/apps. It works like this:

1. User taps on the new "Apps"-menu-entry after opening the attachmenet-menu in a chat (the previous webxdc-selector got basically replaced)
2. The Picker opens, lets the user browse the website and pick an app
3. That app gets downloaded and staged as a draft in that very chat

<details>
<summary>It looks like this</summary>

Simulator Screenshot - iPad Pro 11-inch (M4) - 2025-01-17 at 10 46 17

</details>

-------

- [x] Only one entry "apps" in menu -> Remove the existing app-picker for now
- [ ] ~Consider keyboard~ It's a mess due to webview (or I'm incompetent), so I'll leave this for now as it is.
- [x] Loading Indicator

Closes #2450 ~eventually~

PR #2522, opened 2025-01-16
Clipper uimenu

Amzd:
The screenshot is with all experimental options enabled which general users will not have probably. 

| Before | After |
| --- | --- |
| IMG_3266 | IMG_3265 |

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -1213,15 +1213,16 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
         }
         var actions = [UIMenuElement]()
         actions.append(action(localized: "camera", systemImage: "camera", handler: showCameraViewController))
-        actions.append(action(localized: "gallery", systemImage: "photo.stack", handler: showPhotoVideoLibrary))
+        let galleryImage = if #available(iOS 16, *) { "photo.stack" } else { "photo" }
+        actions.append(action(localized: "gallery", systemImage: galleryImage, handler: showPhotoVideoLibrary))
         actions.append(action(localized: "files", systemImage: "folder", handler: showDocumentLibrary))
         if dcContext.hasWebxdc() {
             actions.append(action(localized: "webxdc_apps", systemImage: "app", handler: showWebxdcSelector))
         }
         actions.append(action(localized: "voice_message", systemImage: "waveform", handler: showVoiceMessageRecorder))
         if let config = dcContext.getConfig("webrtc_instance"), !config.isEmpty {
-            // person.crop.square.badge.video
-            actions.append(action(localized: "videochat", systemImage: "video.bubble", handler: videoChatButtonPressed))
+            let videoChatImage = if #available(iOS 17, *) { "video.bubble" } else { "video" }
+            actions.append(action(localized: "videochat", systemImage: videoChatImage, handler: videoChatButtonPressed))
         }
         if UserDefaults.standard.bool(forKey: "location_streaming") {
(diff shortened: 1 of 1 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1202,31 +1206,72 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
...
+        }
+        var actions = [UIMenuElement]()
+        actions.append(action(localized: "camera", systemImage: "camera", handler: showCameraViewController))
+        let galleryImage = if #available(iOS 16, *) { "photo.stack" } else { "photo" }
+        actions.append(action(localized: "gallery", systemImage: galleryImage, handler: showPhotoVideoLibrary))
+        actions.append(action(localized: "files", systemImage: "folder", handler: showDocumentLibrary))
+        if dcContext.hasWebxdc() {
+            actions.append(action(localized: "webxdc_apps", systemImage: "app", handler: showWebxdcSelector))
+        }
+        actions.append(action(localized: "voice_message", systemImage: "waveform", handler: showVoiceMessageRecorder))
```
r10s, line 1222:
```suggestion
        actions.append(action(localized: "voice_message", systemImage: "mic", handler: showVoiceMessageRecorder))
```

this is the icon we're using on Android, i think, it makes sense to stay in sync, esp. when it is so easy :)

for the name: "SF Symbol" says, if one needs to support platforms older than iOS 18, one shall use "mic" instead of "microphone":

Screenshot 2025-01-16 at 14 35 02

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1202,31 +1206,72 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
...
+        func action(localized: String, systemImage: String, attributes: UIMenuElement.Attributes = [], handler: @escaping () -> Void) -> UIAction {
+            UIAction(title: String.localized(localized), image: UIImage(systemName: systemImage), attributes: attributes, handler: { _ in handler() })
+        }
+        var actions = [UIMenuElement]()
+        actions.append(action(localized: "camera", systemImage: "camera", handler: showCameraViewController))
+        let galleryImage = if #available(iOS 16, *) { "photo.stack" } else { "photo" }
+        actions.append(action(localized: "gallery", systemImage: galleryImage, handler: showPhotoVideoLibrary))
+        actions.append(action(localized: "files", systemImage: "folder", handler: showDocumentLibrary))
+        if dcContext.hasWebxdc() {
+            actions.append(action(localized: "webxdc_apps", systemImage: "app", handler: showWebxdcSelector))
```
r10s, line 1220:
```suggestion
            actions.append(action(localized: "webxdc_apps", systemImage: "square.grid.2x2", handler: showWebxdcSelector))
```

this icon fits better to what desktop and android are using - and even without, i think this looks more like "icons" than just the single square

android:
Screenshot 2025-01-16 at 14 42 57

desktop:



iOS with this suggestion:

r10s (approved):
veeeeery nice!

also, one animation less in the message list (when opening old action sheet, input field slides out)

zeitschlag:
Thanks, looks good!

I spoke to @r10s earlier, we agreed upon merging this PR rather sooner than later as there will be conflicts with #2499 (and it's easier for us to deal with them in #2499 than here). I'll squash this PR (although I strongly dislike squashing :D) but this way we can ship a Testflight sooner :)

Again: Thanks for your work!

Amzd:
Yep, thanks @zeitschlag 

Only reason it wasn't merged yet is I hadn't had time to commit and test the proposed changes. Thank you for doing that.

Issue #2512, opened 2025-01-12
append dropped URLs to text field (instead of attaching as file)


Amzd:
Should just become text OR give option to download source to send as file?

r10s:
downloading seems over the top, and also unexpected

if easy to do,  appending to the text field instead of attaching text.txt would be nice, however.

r10s:
tested iMessage, Signal, WhatsApp, Telegram, Mastodon with the URL https://delta.chat/en/2024-11-20-webxdc-realtime:

- only iMessage get it right
- Signal, Telegram, Mastodon do nothing on dropping
- WhatsApp does weird download attempts, resulting in weirder attachment than delta



still, if we can improve easily, we should do. but if it turns out to get complicated, well, ppl seem not tobe used to it :)

Amzd:
Discord does it correctly as text tooOn 12 Jan 2025, at 11:44, bjoern ***@***.***> wrote:﻿
tested iMessage, Signal, WhatsApp, Telegram, Mastodon with the URL https://delta.chat/en/2024-11-20-webxdc-realtime:

only iMessage get it right
Signal, Telegram, Mastodon do nothing on dropping
WhatsApp does weird download attempts, resulting in weirder attachment than delta

still, if we can improve easily, we should do. but if it turns out to get complicated, well, ppl seem not tobe used to it :)

Amzd:
> downloading seems over the top

If you try to drag an image from safari sometimes it is only a URL, in which case delta chat doesn't do anything with it because we don't render links to images as an image, unlike eg Discord. In this case it would be useful to get a download option to send the image as attached image instead of URL?

Amzd:
Yep agree.

PR #2523, opened 2025-01-16
append dropped URLs to text field (instead of attaching as file) #2512

Amzd:


r10s (approved):
nice, straight forward enhancement! thanks for that!

PR #2525, opened 2025-01-16
disable blank issue template as it gets much more into focus by recent github changes

r10s:
creating "blank" issues should be a real exception, we really want the track for bugs are agreed enhancements only.

before, the main options were BUG REPORT, FEATURE REQUEST, SUPPORT QUESTION - with a very much tuned down "blank" issue at the bottom.

with the upcoming change, the main options will be BUG REPORT, BLANK ISSUE, FEATURE REQUEST, SUPPORT QUESTION - with "blank" issue even more exposed than FEATURE REQUEST or SUPPORT QUESTION.

this PR disables the blank issue therefore,
if really needed one can create a BUG report and remove the corresponding tag (in fact, that was always what i was doing,
i did not notice "blank" before :)

the following images show the change _without_ this PR:

Screenshot 2025-01-16 at 23 31 50
Screenshot 2025-01-16 at 23 29 07

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-blank_issues_enabled: true
+blank_issues_enabled: false
 contact_links:
   - name: Feature requests
     url: https://support.delta.chat/c/features/6


r10s:
merging, to check if things work out ...

PR #2529, opened 2025-01-17
Always show apps-tab

### LINKED ISSUE #2526: stop hiding "Apps"
the next update will come with an "App Picker", so it does no longer make sense to hide other "Apps" options as tabs

zeitschlag:
Closes #2526


r10s (commented):
`DcContext.hasWebxdc()` should be removed as well, and the conditions that were using it evaluated to `true`. 

note sure if there is even more

r10s (approved):

PR #2534, opened 2025-01-18
Send separate notifications when multiple accounts have new messages when app receives push notification.

Amzd:
This also fixes notifications for classic email always bunching up with notifications of a chatmail account.

Setup of the screenshot:
The device has two accounts in the same group to simulate receiving multiple notifications at once.
| Before | After | 
| --- | --- | 
| IMG_0024 | IMG_0025 |

This PR also adds threadIdentifier which means iOS can decide to show each group in its own notification thread, like in the following screenshot.

IMG_0027

This makes it harder to miss notifications for private messages when you are also in a busy group. 

Tested on iOS 16 and 17, both work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Unreleased
 
 - Support pasted GIFs
+- Group notifications in Notification Center by chat
+- Fix: separate multiple notifications coming in at the same time
 
 ## v1.58.1
 2025-04
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Support pasted GIFs
 - Group notifications in Notification Center by chat
+- More characters in notification before truncating
 - Fix: separate multiple notifications coming in at the same time
 
 ## v1.58.1
--- a/DcCore/DcCore.xcodeproj/project.pbxproj
+++ b/DcCore/DcCore.xcodeproj/project.pbxproj
@@ -25,6 +25,7 @@
 		30E8F2212447357500CE2C90 /* DatabaseHelper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 30E8F2202447357500CE2C90 /* DatabaseHelper.swift */; };
 		30E8F2482449C98600CE2C90 /* UIView+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 30E8F2472449C98600CE2C90 /* UIView+Extensions.swift */; };
 		30E8F24D2449D30200CE2C90 /* DcColors.swift in Sources */ = {isa = PBXBuildFile; fileRef = 30E8F24C2449D30200CE2C90 /* DcColors.swift */; };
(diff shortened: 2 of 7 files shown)


r10s:
as discussed one-to-one, this is a nice experiment, but it is unclear if that works on all versions. it is also unclear if this is a way that apple "wants" or is more a hack. (otherwise the callback would not be needed at all). at least at some point we were not able to get a UNUserNotificationCenter from the NSE, cmp. https://github.com/deltachat/deltachat-ios/blob/eca9dae049aa8c8c158e1d982c5501e8f3ceaf02/DcNotificationService/NotificationService.swift#L19

so, for now, we aim for stability and do not want to risk to get new bugs in again, in the somehow working notifications.

but interesting, we should keep that in mind, and might pick that up or iterate when we interate over that part again

closing this PR and putting it to resurrection. discussion and pushing can continue, of course

r10s:
maybe, things have changed with the "notification center entitlement"  (see https://github.com/deltachat/deltachat-ios/pull/2124 ) 

the "notification center entitlement" allows us to _not_ add a notification. maybe that, as an undocumented side effect, that allows to schedule notifications via UNUserNotificationCenter from NSE (i am pretty sure i read that time posts saying that this is explicitly not possible, but i cannot find a link atm). but even if that's true, i would be hesitant of making use of that hack

Amzd:
> pushing can continue

hmm I pushed more and they don't show up in this PR, so we gotta keep that in mind when picking this up

Amzd:
I found [this](https://stackoverflow.com/questions/55052376/notification-service-extension-creating-local-notifications) which states it was not possible to queue notifications for the future in NSE in 2019. I can not reproduce this, as I can queue notifications in the future too in DC now, but it has been 6 years since then so maybe it was updated.

r10s:
> hmm I pushed more commits and they don't show up in this PR, so we gotta keep that in mind when picking this up

oh, did not knew that, we can also reopen for now (maybe it is related to the the PR not being part of deltachat-org? or os that a general thing, that closed PR are frozen and do not follow the branch any longer?)

r10s:
here is another pointer: https://developer.apple.com/forums/thread/108340

Amzd:
> here is another pointer: https://developer.apple.com/forums/thread/108340 , but also a bit older

> only call the Notification Extension's contentHandler once the UNUserNotificationCenter.current().add() method had fully completed.

I cannot reproduce this, even if I call the contentHandler before `add` it still works (on iOS 16). Might be a bug that was fixed later.

Amzd:
Signal seems to do it this way too (just under many layers of abstraction).

They even just only send a silent notification to update the badge number and all other notifications are through UNUNC

r10s (approved):
very nice!

as signal is using the same approach and as we need to do some additional changes to NSE, i am no longer concerned :)

code lgtm, needs to be tested in real, so, also, from that point of view, it is good timing as we just released 1.58

PR #2536, opened 2025-01-19
Allow interactive dismissal of VCs that need custom first responder handling

Amzd:
Some View Controllers don't give back the first responder correctly (eg when you use the search bar in UIImagePickerController). To combat this we previously wrapped those in a custom view controller that just gave back the responders manually. That sadly had the side effect of disabling interactive dismissal (swipe down to dismiss).

This PR fixes the interactive dismiss by doing away with the custom view controller wrapper and instead swizzling the methods. 

Swizzling is a bit controversial but it is used by Firebase which is in many apps, and the associated objects usage is already in our app because SDWebImage uses it too.

### DIFF
--- a/deltachat-ios/Helper/GiveBackMyFirstResponder.swift
+++ b/deltachat-ios/Helper/GiveBackMyFirstResponder.swift
@@ -1,4 +1,5 @@
 import UIKit
+import Combine
 
 /// https://gist.github.com/Amzd/223979ef5a06d98ef17d2d78dbd96e22
 extension UIViewController {
@@ -24,8 +25,8 @@ extension UIViewController {
         // In iOS 17 and iOS 16 the UIDocumentPickerViewController does not
         // give back the first responder when search was used
         if #available(iOS 16, *) {
-            let vc = GiveBackMyFirstResponder(culprit: documentPicker)
-            present(vc, animated: animated, completion: completion)
+            documentPicker.returnFirstRespondersOnDismiss()
+            present(documentPicker as UIViewController, animated: animated, completion: completion)
         } else {
             present(documentPicker as UIViewController, animated: animated, completion: completion)
         }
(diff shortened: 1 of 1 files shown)


deltachat-ios/Helper/GiveBackMyFirstResponder.swift
```diff
@@ -24,80 +25,75 @@ extension UIViewController {
...
-
-    override var isBeingDismissed: Bool {
-        parent?.isBeingDismissed ?? super.isBeingDismissed || super.isBeingDismissed
-    }
-
-    override func viewWillDisappear(_ animated: Bool) {
-        super.viewWillDisappear(animated)
-        if isBeingDismissed, !lastResponders.isEmpty, let newFirstResponder = UIResponder.currentFirstResponder {
+    @objc fileprivate func _viewWillDisappear(_ animated: Bool) {
+        _viewWillDisappear(animated)
```
r10s, line 66: i do not really understand what is happening here. at a first glance, this looks to me as a recursive call, but it is not and is working as expected.

but what is then called here?

r10s (commented):
in general, i think, the added functionality, of being able to swipe-down-to-close the file-and gallery-view-controller as expected  is worth the complexity. thanks a lot for diving into that and checking out options!

it is incredible, that this complexity is needed, but well, if there is no standard way, then it's that.

issues:

- for the app-selector,  swipe-down-to-close is still not working  - is that expected?
- the camera _lost_ the swipe down functionality - is that expected?

also, i would like to hear from @zeitschlag if he would be fine with getting sth. like that in and if it is maybe better understandable by him :)

deltachat-ios/Helper/GiveBackMyFirstResponder.swift
Amzd: That is how swizzled methods call the method they replaced (so like super. in override methods), I agree it looks like a recursive call.

Amzd:
> the camera lost the swipe down functionality - is that expected?

Ah good catch. That was how it was before I added the wrapper and usually expected behavior. I checked Discord and iMessage and they also cant interactively dismiss the camera.

deltachat-ios/Helper/GiveBackMyFirstResponder.swift
r10s: ahh, now i got it!

if any line of code in the project would deserve a comment, this is the line :)

r10s:
> But maybe we want to support that?

no, it is fine as it is, i even like it more: opening the camera full screen allows to focus on that

deltachat-ios/Helper/GiveBackMyFirstResponder.swift
Amzd: It's because if I would try to call viewWillAppear it would end up here again because this method replaces viewWillAppear through swizzling, so they needed a different mechanism for calling the original method.

zeitschlag:
> for the app-selector, swipe-down-to-close is still not working - is that expected?

This is actually intended, yes, as it was easier to include a one-liner to not take care of managing and canceling downloads than ... managing and canceling downloads.

Amzd:
Yeah I get it, managing download state was extra work. 

I think you've done that extra work now in the new PR though and it would be two extra lines in the showLoading and hideLoading functions here: https://github.com/deltachat/deltachat-ios/pull/2548/files#diff-0573590fd4c29c356f071cd7faf09104ba2311e9fba2a66d2fee0a5743568e6fR89-R103
or am I missing context? I can do it here after that PR is merged too.

deltachat-ios/Helper/GiveBackMyFirstResponder.swift
```diff
@@ -1,4 +1,5 @@
 import UIKit
+import Combine
```
zeitschlag, line 2: What do you need Combine for? I'm a little confused as I didn't see any code using it :)

Amzd: Whoops, left over from experiment. Will remove.

zeitschlag:
> it would be two extra lines in the showLoading and hideLoading

Ah, I think I just understand what you suggested: Disabling the dismiss-swipe during a download instead of disabling it in general, right? Neat :)

Sure, I can do that. I'd suggest to move that to an additional PR, once #2536 and #2548 are merged :)

deltachat-ios/Helper/GiveBackMyFirstResponder.swift
zeitschlag, line 2:
```suggestion
```

zeitschlag (approved):
I spoke to @r10s earlier and in order to move forward, we decided to approve and merge this and add the comment in a subsequent PR. Thanks for figuring this out!

One of my next steps will be to add swipe-dismissal to the catalogue/store/app picker

r10s:
thanks @Amzd, @zeitschlag for iOS-expert-level PR, discussion and review 👌

PR #2537, opened 2025-01-19
Fix for location sharing not ending unless you move

Amzd:
Encountered this at CCC, even while moving around ~10 meters, maybe because it was inside a building.

- Start Location streaming for 5 mins
- Wait 5 mins then stop moving
- Previously location sharing would stay on, draining battery.


r10s (approved):
makes sense, thanks for fixing!

i hope, one day, we can get location-streaming out of experimental - where it is for very good reasons since years :)

PR #2539, opened 2025-01-20
increase deployment target to iOS 14

r10s:
this PR increases the deployment target to iOS 14, meaning that iOS 13 is no longer supported.

the reason is that the "modern" menus are supported since iOS 14 only, and using them has quote some postive effect in UI.

in theory one could provide fallbacks,
as in https://github.com/deltachat/deltachat-ios/pull/2522 , however, that is comparable much effort as we have to duplicate the logic or add an abstraction layer.

that seems not to be worth the effort:

i checked some stats, this change will affect 0.16% of users - but most of them should even be able to update their devices (all devices that got iOS 13 also got iOS 15)
(so, we could also discontinue supporting iOS 14 at some point, but that has 4x more users, so, let's give them a little more time, until we really have bigger incentive)

cmp. https://github.com/deltachat/deltachat-ios/pull/2459


deltachat-ios/Helper/MediaPicker.swift
```diff
@@ -77,19 +77,10 @@ class MediaPicker: NSObject, UINavigationControllerDelegate {
     func showDocumentLibrary(selectFolder: Bool = false) {
         let documentPicker: UIDocumentPickerViewController
         if selectFolder {
-            if #available(iOS 15.0, *) {
-                documentPicker = .init(forOpeningContentTypes: [UTType.archive], asCopy: false)
-            } else {
-                documentPicker = .init(documentTypes: [kUTTypeArchive] as [String], in: .open)
-            }
+            documentPicker = .init(forOpeningContentTypes: [UTType.archive], asCopy: false)
```
r10s, line 80: not totally sure, would be cool if someone double checks:

the `if #available(iOS 15.0, *) {` might always be wrong and should have always been `if #available(iOS 14.0, *) {`.

Xcode gives me a warning otherwise that `documentPicker = .init(documentTypes: [kUTTypeArchive] as [String], in: .open)` is deprecated in iOS 14 (the warning was probably not there before as we deployed to iOS 13)

Amzd:
How many users do we have? 0.16% is not nothing.

deltachat-ios/Helper/MediaPicker.swift
zeitschlag: As, according to [The Docs](https://developer.apple.com/documentation/uikit/uidocumentpickerviewcontroller/3566733-init)™, that initializer is available on iOS 14, so it smells like a misunderstanding, not to say wrong.

Maybe it has [something to do with Documents](https://github.com/deltachat/deltachat-ios/commit/fc88a075f569db6a0238f8436c2dae91a742d73a), so that this Deltachat-feature was made available to iOS 15+, only?

zeitschlag (approved):

r10s:
> 0.16% is not nothing.

sure, but we do not have hundreds of millions of users (yet), so the absolute numbers should be comparable small. it is a but hard to tell absolute numbers, as (a) i do not really trust the stats from apple nor google and (b) there are countries known that do not have access to apples stores and still use iphones, over whatever paths

anyways, if we get backlash at scale, we can iterate and see what to do. we did that few on android, where we raised requirements and lowered them again.

but all in all, in this case, i think, it is fine. i am much less worried than for the step to iOS 13 where we left user behind, having no chance than to buy a new device. there, user can update existing devices, which is probably also a good idea wrt security

Amzd:
Sure, makes sense. #2534 can also be simpler without <13.3 support

deltachat-ios/Helper/ChatDropInteraction.swift
```diff
@@ -8,22 +8,13 @@ public class ChatDropInteraction {
...
-                UTType.item.identifier])
-        }
-        return session.items.count == 1 && session.hasItemsConforming(toTypeIdentifiers: [
-            kUTTypeImage as String,
-            kUTTypeText as String,
-            kUTTypeMovie as String,
-            kUTTypeVideo as String,
-            kUTTypeURL as String,
-            kUTTypeItem as String])
+        return  session.items.count == 1 && session.hasItemsConforming(toTypeIdentifiers: [
```
Amzd, line 11: Has two spaces for some reason? But can also be implicit now
```suggestion
        session.items.count == 1 && session.hasItemsConforming(toTypeIdentifiers: [
```

deltachat-ios/Helper/MediaPicker.swift
Amzd: I think this needs to be checked for all the kUTType strings? They are deprecated in iOS 14 and I think we'll get warnings eg in `ChatDropInteraction.dropInteraction(performDrop:)`

Amzd: I think this needs to be checked for all the kUTType strings? They are deprecated in iOS 14 and I think we'll get warnings eg in `ChatDropInteraction.dropInteraction(performDrop:)`

deltachat-ios/Helper/ChatDropInteraction.swift
r10s: maybe just a stuck space key :)

deltachat-ios/Helper/MediaPicker.swift
Amzd: Sorry I was not very clear. I just meant that everywhere in the app that uses `kUTTypeX` strings instead of `UTType.x` should be updated because the former is deprecated in iOS 14

PR #2543, opened 2025-01-22
defer attach menu creation

r10s:
this PR creates the attach menu only if needed;
this was also the case in the past, when we were using an alert instead of modern menu.

it is only a little bit faster currently (2~3ms vs <0.5ms on iphone7), however, as this also may be used as a pattern and as slower conditions may be added in the future, maybe still worth going for it.
also to exclude that part from being slow.

#skip-changelog

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -1203,34 +1203,35 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
     }
 
     private func clipperButtonMenu() -> UIMenu {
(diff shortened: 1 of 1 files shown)


Amzd:
I don't fully understand the caching that happens. Is it cached until the next time the UIMenu is initialized or longer? Eg if you open the menu then turn off experimental location sharing feature, does it still show the button?

r10s:
imu and tests, it is cached as long as the parent exist:

- when you exit and reenter the view controller, you get a new menu
- if you just close and reopen the menu, without leaving the view controller, the cache is used

to _not_ use the caching at all, there is UIDeferredMenuElement.uncached()

Amzd (approved):

PR #2545, opened 2025-01-22
use modern UIMenu

r10s:
this PR replaces ~~some~~ all of the old "actionSheet" menus with the more modern ones available since iOS 13/14.

it is all pretty much straight-forward. greater refactorings, renamings are not done, to keep the changed lines small.

see commit messages for more details, this is what it looks like:

<br><br>

<br><br>

<br><br>

<br><br>

<br><br>

<br><br>

<br><br>

<br><br>

<br><br>

### DIFF
--- a/deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingViewController.swift
+++ b/deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingViewController.swift
@@ -21,15 +21,12 @@ class InstantOnboardingViewController: UIViewController {
     private var qrCodeData: String?
     private lazy var menuButton: UIBarButtonItem = {
         let image = UIImage(systemName: "ellipsis.circle")
-        let menuButton = UIBarButtonItem(image: image, style: .plain, target: self, action: #selector(InstantOnboardingViewController.showMenu(_:)))
-        menuButton.tintColor = DcColors.primary
-        return menuButton
+        return UIBarButtonItem(image: image, menu: moreButtonMenu())
     }()
 
     private lazy var proxyShieldButton: UIBarButtonItem = {
-        let button = UIBarButtonItem(image: UIImage(systemName: "checkmark.shield"), style: .plain, target: self, action: #selector(InstantOnboardingViewController.showProxySettings(_:)))
-        button.tintColor = DcColors.primary
-        return button
+        let image = UIImage(systemName: "checkmark.shield")
+        return UIBarButtonItem(image: image, style: .plain, target: self, action: #selector(showProxySettings))
     }()
 
     var progressAlertHandler: ProgressAlertHandler?
@@ -221,22 +218,16 @@ class InstantOnboardingViewController: UIViewController {
         present(alertController, animated: true)
     }
 
-    @objc private func showMenu(_ sender: Any) {
-        let sheet = UIAlertController(title: nil, message: nil, preferredStyle: .safeActionSheet)
-
-        let showProxySettings = UIAlertAction(title: String.localized("proxy_settings"), style: .default) { [weak self] _ in
-            self?.showProxySettings(sender)
-        }
-
-        let cancelAction = UIAlertAction(title: String.localized("cancel"), style: .cancel)
-
-        sheet.addAction(showProxySettings)
-        sheet.addAction(cancelAction)
-
-        present(sheet, animated: true)
+    private func moreButtonMenu() -> UIMenu {
+        let actions = [
+            UIAction(title: String.localized("proxy_use_proxy"), image: UIImage(systemName: "shield")) { [weak self] _ in
+                self?.showProxySettings()
+            },
+        ]
+        return UIMenu(children: actions)
     }
 
-    @objc private func showProxySettings(_ sender: Any) {
+    @objc private func showProxySettings() {
         let proxySettingsController = ProxySettingsViewController(dcContext: dcContext, dcAccounts: dcAccounts)
         navigationController?.pushViewController(proxySettingsController, animated: true)
     }
--- a/deltachat-ios/Controller/BackupTransferViewController.swift
+++ b/deltachat-ios/Controller/BackupTransferViewController.swift
@@ -23,11 +23,6 @@ class BackupTransferViewController: UIViewController {
         return UIBarButtonItem(title: String.localized("cancel"), style: .plain, target: self, action: #selector(cancelButtonPressed))
     }
 
-    private lazy var moreButton: UIBarButtonItem = {
-        let image = UIImage(systemName: "ellipsis.circle")
-        return UIBarButtonItem(image: image, style: .plain, target: self, action: #selector(moreButtonPressed))
-    }()
-
     private let statusLine: UILabel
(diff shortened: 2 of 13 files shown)


r10s:
puh, was a little bit more than expected - but now we have all menus modernised :)

deltachat-ios/Controller/ChatListViewController.swift
```diff
@@ -1051,33 +1051,34 @@ extension ChatListViewController: ChatListEditingBarDelegate {
...
             guard let self else { return }
             viewModel.pinChatsToggle(indexPaths: tableView.indexPathsForSelectedRows)
             setLongTapEditing(false)
         })
 
         if viewModel.hasAnyUnmutedChatSelected(in: tableView.indexPathsForSelectedRows) {
-            alert.addAction(UIAlertAction(title: String.localized("menu_mute"), style: .default) { [weak self] _ in
+            actions.append(UIAction(title: String.localized("menu_mute"), image: UIImage(systemName: "speaker.slash")) { [weak self] _ in
                 guard let self else { return }
                 MuteDialog.show(viewController: self) { [weak self] duration in
```
Amzd, line 1083: This could return a sub-menu instead maybe?

r10s: i also thouht about that, but the mute dialog is also available at other places, so it seems better to have the same UI everywhere

Amzd: Maybe those other places could use the menu too? 🤔 But can be another PR, this already addresses a lot of them.

r10s: hm, at least whatsapp and signal also use an actionSheet for the concrete setting. i also have rarely seen a UIMenu in response to a swipe-to-action gesture (where mute is also available)

Amzd: Ah yeah I don't think swipe action is a great place for it.

deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingViewController.swift
```diff
@@ -221,22 +218,16 @@ class InstantOnboardingViewController: UIViewController {
...
-
-        sheet.addAction(showProxySettings)
-        sheet.addAction(cancelAction)
-
-        present(sheet, animated: true)
+    private func moreButtonMenu() -> UIMenu {
+        let actions = [
+            UIAction(title: String.localized("proxy_use_proxy"), image: UIImage(systemName: "shield")) { [weak self] _ in
+                self?.showProxySettings()
+            },
```
zeitschlag, line 225: Probably more of a UX-thing, but if we only have one entry a menu, does it make a lot of sense to show that menu -- and not the menu-entry itself? In other words: Why does the BarButtonItem show a menu and not the Proxy-button?

It feels like an extra step for the user to access the proxy-settings.

deltachat-ios/Controller/BackupTransferViewController.swift
```diff
@@ -264,20 +261,26 @@ class BackupTransferViewController: UIViewController {
...
+
+    private func moreButtonMenu() -> UIMenu {
+        var actions = [UIMenuElement]()
+        if !qrContentView.isHidden {
+            actions.append(UIAction(title: String.localized("menu_copy_to_clipboard"), image: UIImage(systemName: "document.on.document")) { [weak self] _ in
                 guard let self else { return }
-                self.warnAboutCopiedQrCodeOnAbort = true
-                UIPasteboard.general.string = self.dcBackupProvider?.getQr()
-            }))
+                warnAboutCopiedQrCodeOnAbort = true
```
zeitschlag, line 275:
```suggestion
                self.warnAboutCopiedQrCodeOnAbort = true
```

deltachat-ios/Controller/BackupTransferViewController.swift
```diff
@@ -264,20 +261,26 @@ class BackupTransferViewController: UIViewController {
...
+                UIPasteboard.general.string = dcBackupProvider?.getQr()
+            })
         }
-        alert.addAction(UIAlertAction(title: String.localized("cancel"), style: .cancel, handler: nil))
-        self.present(alert, animated: true, completion: nil)
+        actions.append(UIAction(title: String.localized("troubleshooting"), image: UIImage(systemName: "questionmark.circle")) { [weak self] _ in
+            guard let self else { return }
+            navigationController?.pushViewController(HelpViewController(dcContext: dcContext, fragment: "#multiclient"), animated: true)
```
zeitschlag, line 281:
```suggestion
            self.navigationController?.pushViewController(HelpViewController(dcContext: self.dcContext, fragment: "#multiclient"), animated: true)
```

deltachat-ios/Controller/ChatListViewController.swift
```diff
@@ -1086,14 +1087,13 @@ extension ChatListViewController: ChatListEditingBarDelegate {
                 }
             })
         } else {
-            alert.addAction(UIAlertAction(title: String.localized("menu_unmute"), style: .default) { [weak self] _ in
+            actions.append(UIAction(title: String.localized("menu_unmute"), image: UIImage(systemName: "speaker.wave.2")) { [weak self] _ in
```
zeitschlag, line 1090: We do the `guard let self else { return }` dance on several places, but we don't use that unwrapped `self`, it seems to me. What am I missing?

-----

From what I understood, the dance is to break a retain cycle, but this way we still use the old, strong references?

zeitschlag (approved):
LGTM, just some UX and `self`-related questions/suggestions :)

deltachat-ios/Controller/ChatListViewController.swift
Amzd: You can use implicit self after guard let self since swift 5.8: https://github.com/swiftlang/swift-evolution/blob/main/proposals/0365-implicit-self-weak-capture.md

zeitschlag: til

deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingViewController.swift
r10s: > but if we only have one entry a menu, does it make a lot of sense to show that menu 

yes - as the symbol is not self-explaining for most users. therefore, an additional descriptive text is better.

when proxy is then enabled, the icon is shown directly  - this is also what is done in the main app.

deltachat-ios/Controller/ChatListViewController.swift
r10s: k, this was also my impression as someone not that deep in swift: that `self` can now be omitted in more cases, and that this seems to be best practise

zeitschlag: We're using Swift 5.5 (it seems)

Amzd: Hmm, really? The CI is giving Swift 6 errors

PR #2546, opened 2025-01-22
Rewrote sticker detection

Amzd:
Left plenty of comments in and the source, so we can hopefully understand it when we want to iterate this next. The previous logic was quite hard to figure out.

#skip-changelog


zeitschlag (approved):
Checked the code, LGTM #famouslastwords

PR #2548, opened 2025-01-23
2530 recent apps

zeitschlag:


r10s:
very nice!

some high level comment:

- we avoid the wording "Store" somehow, not only because of apple :) so, let's call that tab "Browse", and also rename the source code accordingly

- the other page i would call "Recents" (for translators, "Recently Used" could serve as a maybe better pattern, but that could go to the translator hints)

but let's sleep over that, so we do not rename twice :)

r10s:
and another high-level comment from testing only:

the layout seems not to be adaptive when orientation is changed, see the gallery to get the idea. i always get 4 icons, which are then far too large in landscape.

it'd probably be better to give the icons a reasonable size, and let the "flowing grid collection layout" (however this is called) decide how many it can display in a row. that's also be better when supporting looking at iphoneSE1   (would probably have only 3 apps per row) vs iphone15Max (can have 5 or even 6 apps per row) vs ipad (even more :)

zeitschlag:
> the layout seems not to be adaptive when orientation is changed

the layout considers the orientation now @r10s

r10s:
ftr, i force-pushed to get unrelated translations out of the PR. this part is documented badly: as a rule of thumb, it is fine to edit untranslated and english source in PR, but translations should be pulled in separate PR, i will add some lines to mane that clearer at the appropriate places

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@
 - Fix: Experimental location sharing now ends at the specified interval even if you don't move (#2537)
 - Fix crash on account deletion (#2554)
 - minimum system version is iOS 14 now (all iOS 13 devices can upgrade to iOS 14) (#2459)
-
+- Add screen to share your apps (#2530)
```
r10s, line 11:
```suggestion
- Access recently used apps from app-picker ("Attach / Apps") (#2530)
```

r10s (approved):
code lgtm and also things work as expected.

for the "store" wording, well, we can leave it, also the bot was called "xstore" that time. the strings can be adapted in another PR (i will do that)

for the other tap, however, that is a bit mixed between "my apps" and "recent".  i suggest to use the same wording in the code, maybe "recently used" or "recent" (so only rename "myApps". but that is no blocker.

PR #2553, opened 2025-01-25
show connectivity problems also account switcher

r10s:
unless the state is "connected", this PR adds the connectivity string to the new subtitle in the account  switcher.

this results in having the strings "Not Connected", "Updating" etc. there - just the same strings as in the app title when unconnected.

i was thinking in adding some colored dots as in arcane chat, however, we are (a) not using them in the main title (b) that would need additional explanations where a "Not Connected" is self explaining and (c) the current one is easier :)

main point was to have sth, and also get the changed event in; we can always iterate :)

this is how it looks like if everyrhing is disconnected:

when connection is fine, the switcher looks exactly as always

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Show 'unconnected' and 'updating' states in account switcher (#2553)
 - Detect Stickers when dropped, pasted or picked from Gallery (#2535)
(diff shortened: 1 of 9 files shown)


Amzd (approved):
Adding a new Event could've prevented all the guard statements.

Eg maybe  `Event.connectivityChangedSelf` and `Event.connectivityChangedOther`

But LGTM other than that.

PR #2554, opened 2025-01-25
fix crash on account deletion

r10s:
core might return a NULL pointer for dc_get_fresh_msgs(), which was not handled by the old DcArray constructor and resulted in a runtime crash.

Screenshot 2025-01-25 at 15 33 09

when looking more closely,
it turned out, the while DcArray class is meanwhile only being used to get the count, therefore, it is better to just get rid of it
(in fact, might be the dc_get_fresh_msgs()
will be replaced by a dc_get_fresh_msgs_count() at some point, there was a discussion at some point :)

ftr, the crash was mostly visible on account deletion, but did not happen always there. in theory could also happen for other races on calls to getFreshMessages().


zeitschlag (approved):
LGTM (and plausible)

PR #2555, opened 2025-01-25
cleanup assets

r10s:
this PR cleans up the asset folder, by preferring system images - which are since the move to iOS 13 now always available.

moreover, some icons are streamlined by this PR, see commit message.

the PR is not exhaustive, some assets are left, when it did not work out on the first try (wrong sized, colors etc.) or the layout was changed too much, i moved forward - these things should better be done when we iterate over these areas (eg. audio player)

finally, this PR improves the screen when voice recording permissons are missing - and shows some text instead of a weird, non-telling icon

---

one before/after example showcasing the pin and the mute icons - not that much of a change, but still clearer - _and_ saving some kb _and_ having the same icon as elsewhere:

 &nbsp; &nbsp;


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -110,8 +110,8 @@ class ChatViewController: UITableViewController, UITableViewDropDelegate {
 
     private lazy var muteItem: UIBarButtonItem = {
         let imageView = UIImageView()
-        imageView.tintColor = DcColors.defaultTextColor
-        imageView.image =  UIImage(named: "volume_off")?.withRenderingMode(.alwaysTemplate)
+        imageView.tintColor = DcColors.middleGray
```
zeitschlag, line 113: The idea of using a `customView` is probably due to some margins and constraints, right?

zeitschlag (approved):
LGTM, literally

deltachat-ios/Chat/ChatViewController.swift
r10s: good question, i just don't know, i changed only the image.

maybe it is as the items should not be buttons but more state images

PR #2557, opened 2025-01-27
add a warning when pulling translations to feature branches

r10s:
pulling translations to feature branches clutter diff and make review harder and easily result in merge conflicts.

it should usually not be done

### DIFF
--- a/scripts/tx-pull-translations.sh
+++ b/scripts/tx-pull-translations.sh
@@ -41,3 +41,15 @@ for (( i=0; i<${#IOS_TRANSLATIONS[@]}; i++ )) {
 rm -rf $TMP_ANDROID_TRANSLATIONS
 
 cd ..
+
+
+
+branch=$(git rev-parse --abbrev-ref HEAD)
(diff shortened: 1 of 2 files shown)


scripts/tx-pull-translations.sh
```diff
@@ -48,3 +41,15 @@ for (( i=0; i<${#IOS_TRANSLATIONS[@]}; i++ )) {
 rm -rf $TMP_ANDROID_TRANSLATIONS
 
 cd ..
+
+
+
+branch=$(git rev-parse --abbrev-ref HEAD)
+if [[ ! "$branch" == *"update-transl"* ]]; then
```
zeitschlag, line 48: From what I understand, this will only be a little more text in the console (in case you're on a branch which doesn't have `update-transl` somewhere in its name).

Is there any documentation for the current state of translations, i.e. how to do what?

r10s: yes, but i removed that again, as the branch name detection would already fail for the name recommended in RELEASE.md

instead, i added some lines to CONTRIBUTING.md

PR #2558, opened 2025-01-28
group photo-based actions together

r10s:
as a rule of thumb, 6+ items in a menu is too much for the eye to catch. so, group 'camera' and 'gallery',
which are also the most often used actions.

successor of #2543


r10s:
moving forward to not distract from more important reviews :)

PR #2565, opened 2025-01-30
[WIP] Select multiple assets

### LINKED ISSUE #2484: allow to attach multiple images/files
it would be great, if the "Attach Button" in a chat allows to select multiple images/files. to keep things simple and avoid larger refactoring for a comparable rarely used enhancement, we skip the "staging  area" and ask the user directly if they want to send "N images". this is how it is done on android and desktop:

https://github.com/user-attachments/assets/899a86c4-e725-4726-bf1e-48c6de6431c9

images will be sent as "images", that means being recoded by core.

technically, it seems the `UIImagePickerController` in iOS does not natively support selecting multiple images. as we want to stay with what the system offers and avoid using 3rd party stuff whereever possible, we probably need to switch to `PHPickerConfiguration`, ~~requiring iOS 14 (our current min-deployment is iOS 13) - if in doubt, that means to leave the old code in place [^1]~~ EDIT: meanwhile the min-deployment is iOS 14

ftr, attaching multiple images and files is already today possible using "Share To Delta Chat"

[^1]: ~~i would like to avoid to discuss the question of raising from iOS 13 to iOS 15 here, we do not even have feedback from the removed iOS 12 :) even though all devices that got iOS 13 get iOS 15 in theory, there are reasons of ppl to not upgrade (no access to apple servers eg)~~

zeitschlag:
TODO:

- Handle videos
- Localize

Closes #2484.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Select and send up to four images or videos
 - Support pasted GIFs
 - Group notifications in Notification Center by chat
 - More characters in notification before truncating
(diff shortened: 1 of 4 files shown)


r10s:
wondering if that at the end still require the "swizzling" from https://github.com/deltachat/deltachat-ios/pull/2536

r10s:
continued in https://github.com/deltachat/deltachat-ios/pull/2678

PR #2569, opened 2025-01-31
Block stickers from appending to text in text view

Amzd:


r10s:
great, that does the trick, i can confirm, sticker works again on iOS 18 as supposed to do :)

moreover, i also tested iOS 15 _without_ this PR, there it was always working

r10s (approved):

PR #2527, opened 2025-01-16
improve saved messages

r10s:
> this PR requires core with https://github.com/deltachat/deltachat-core-rust/pull/5606 and improves "Saved Messages" handling as described there. EDIT: core is there, let's wait until it is becoming stable a bit wrt group-consistency

this is how it looks like:

Screenshot 2025-01-17 at 00 02 51 &nbsp; &nbsp; Screenshot 2025-01-17 at 00 05 33

PR #2571, opened 2025-01-31
do not save or private-reply info-messages

### LINKED ISSUE #2567: do not allow to save info-messages
info-messages are more a state of the group, pulling them out of context by saving them, results in weird views of "Saved Messages", weakening the info-messages as such, as now they can be passed around and generated other wise.

we should just not allow to save them, this is also what whatsapp is doing - if in doubt, ppl can do a screenshot or forward them. if this is not sufficient, we can also iterate and think over displaying them differently, but that seems over the top for now

### LINKED ISSUE #2568: "reply" and "reply privately" on info-messages may lead to ghost-chat
some info-messages do not have a sender, one cannot reply to them, and we should not allow that

https://github.com/user-attachments/assets/dbe74cbd-adc7-49e8-9b14-807696a1051e

r10s:
see commit messages for details

successor of #2527

closes #2567
closes #2568

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -1934,7 +1934,7 @@ extension ChatViewController {
                     UIAction.menuAction(localizationKey: "forward", image: image, indexPath: indexPath, action: forward)
                 )
 
-                if !message.isInfo { // info-messages out of context results in confusion, see https://github.com/deltachat/deltachat-ios/issues/2567
+                if !isMarkerOrInfo(message) { // info-messages out of context results in confusion, see https://github.com/deltachat/deltachat-ios/issues/2567
                     if dcChat.isSelfTalk || message.savedMessageId != 0 {
                         children.append(
                             UIAction.menuAction(localizationKey: "unsave", systemImageName: "star.slash", indexPath: indexPath, action: toggleSave)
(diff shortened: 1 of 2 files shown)


zeitschlag (approved):
LGTM!

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1930,14 +1934,16 @@ extension ChatViewController {
...
 
-                if dcChat.isSelfTalk || message.savedMessageId != 0 {
-                    children.append(
-                        UIAction.menuAction(localizationKey: "unsave", systemImageName: "star.slash", indexPath: indexPath, action: toggleSave)
-                    )
-                } else {
-                    children.append(
-                        UIAction.menuAction(localizationKey: "save_desktop", systemImageName: "star", indexPath: indexPath, action: toggleSave)
-                    )
+                if !message.isInfo { // info-messages out of context results in confusion, see https://github.com/deltachat/deltachat-ios/issues/2567
```
zeitschlag, line 1937: Does this mean that you can save marker messages like video invitations? I guess, this is intended?

r10s: did not thought about that. but better use the same isMarkerOrInfo() condition as above for now. 

might otherwise be more confusion than helpful -  they all look like info-messages, so it seems fine if they work similiar. also, these messages maybe need additional cases for displaying the "star" etc. on other platforms, so, for now, just don't allow to save them

r10s: i updated the condition

PR #2574, opened 2025-01-31
prepare 1.52.2

r10s:
this version comes with group-consistency PR and the following device message:

... pointing to the blog, to not get too boring. as the blog was just updated, so that's a good timing


zeitschlag (approved):
"Bug Fixes and Improvements" 👌

adbenitez (approved):

PR #2584, opened 2025-02-05
modern buttons & button icons

r10s:
this PR modernises the buttons used in profiles, settings etc. this is another step much easier possible after dropping ios12

- menus meanwhile also have icons: for better recognition, it makes sense that same functions show up similar (affects eg. "share" or "add to widget" but also "delete")

- esp. the "QR code" icon at "New Contact" is a win; this was figured out on android to make quite some difference in ux

- left-align buttons as meanwhile usual in virtually all comparable apps and screens

for the other icons, we stay conservative and do not add too many things, eg. "add" is a simply plus sign at most places - for adding profiles, groups, members, proxies

some impressions of affected screens:

<br>

<br>

<br>

<br>

<br>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Improve "All media empty" and "Block contact" wordings (#2580)
+- Unify buttons and icons (#2584)
 - Fix: Allow to share contacts that do not have a chat (#2583)
 
 
--- a/deltachat-ios/Controller/AccountSwitchViewController.swift
+++ b/deltachat-ios/Controller/AccountSwitchViewController.swift
@@ -19,6 +19,7 @@ class AccountSwitchViewController: UITableViewController {
     private lazy var addAccountCell: ActionCell = {
         let cell = ActionCell()
         cell.actionTitle = String.localized("add_account")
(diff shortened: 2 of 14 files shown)


Amzd:
I feel like the profile switcher would be nicer if the mute icon was on the right and instead of the checkmark we'd show which was selected using the background color.

r10s:
the checkmark is a standard symbol, and its place is the default place. so, i would not change that. _only_ a background colour would make things less clear, also wrt accessibility. don't make me think :)

for the mute icon: i am not too happy with its position and that a "column" is reserved for it, even if unused. in the other parts of the app, we do not do that. so maybe having the icon inlined, left or right of the title might be better. but it is also good enough as it is now :)

anyways, this is a bit out of scope of this PR

adbenitez:
We put the mute icon in front in Android but the space is not reserved even if the mute icon is not shown, we also do it like that in the chat list,

But in Telegram it is shown at the end of the title instead of the beginning, I think at the end actually makes the UI better because then all titles are perfectly aligned at the start and just put the mute icon appended at the end of the title/name, but at this point with so many incoherences in that account selector (some items with subtitle others centered without subtitle, etc) I would say this is just a small drop in the bucket

adbenitez (approved):

PR #2587, opened 2025-02-07
use 'bookmark' symbol for 'saved messages', tweak forwarding

r10s:
after some discussion with @adbenitez , the bookmark symbol is better for 'save' than a star:

- we're using that already for the save messages (even though there is a star on the bookmark ...)
- by that, it underlines that "save" is really only a shortcut for "forward to -> saved messages"
- 'bookmark' does not mix with an reaction so easily
- it also seems more often used for 'save' in comparable apps (eg. here on github for notifications, also on combination with "save"

moreover, this PR tweaks forwarding, so that forwarding to 'saved messages', if possible, saves the message instead if of forwarding it

<br><br>

the lock icon could also be sharper, but that is another PR then :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Improve "message menu": many options are now faster accessible (#2586)
+- When forwarding to "Saved Messages", if possible, save the message including context (#2587)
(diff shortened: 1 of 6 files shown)


zeitschlag (approved):
lgtm!

PR #2588, opened 2025-02-07
Fixed share extension taking twice the memory per image it needed to

Amzd:
Avoid `NSItemProvider.loadItem` for images because that does some conversion that causes the image data to be twice the size of images retrieved through `NSItemProvider.loadDataRepresentation`. 

`NSItemProvider.loadDataRepresentation` creates a UIImage of the same footprint that the info in Photos app indicates the image should be.

I changed the share extension to use the same code as drag and drop because they both get NSItemProviders.

### DIFF
--- a/DcShare/Helper/ShareAttachment.swift
+++ b/DcShare/Helper/ShareAttachment.swift
@@ -101,23 +101,8 @@ class ShareAttachment {
     }
 
     private func createImageMsg(_ item: NSItemProvider) {
-        item.loadItem(forTypeIdentifier: kUTTypeImage as String, options: nil) { data, error in
-            var result: UIImage?
-            switch data {
-            case let image as UIImage:
-                result = image
-            case let data as Data:
-                result = ImageFormat.loadImageFrom(data: data)
-            case let url as URL:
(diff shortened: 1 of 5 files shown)


r10s (approved):
nice! sounds reasonable, and code lgtm!

for reference, https://github.com/deltachat/deltachat-ios/pull/2020 scaled down the image, but did not change loading

Amzd:
This probably also helps with deduplication because I think the previous solution was not deterministic

Amzd:
Hmm ok so I found another image where it happens that UIImage init doubles the data. This image is 9.1MB but when you init a UIImage with it the size of the UIImage becomes 18MB. 

IMG_2767

We could just load the data and save data straight to file instead of converting to UIImage. (just check file format with the magic byte from ImageFormat.get and don't do anything with the data)

r10s:
> We could just load the data and save data straight to file instead of converting to UIImage

as long as we keep the scaling down. scaling down seems to be needed as core otherwise consumes too much memory - doing the same using iOS seems to be counted differently, see #2020

PR #2589, opened 2025-02-07
contact profile: hide unavailable media/location

r10s:
if there is no chat with a contact,
in the contact's profile,
hide the superfluous media and location options -
similar to eg. shared chats and buttons as clear/delete chat.

this unclutters the profile,
puting things as "send message" more into focus.
and it looks better,
there are otherwise a bit too many grey tones in the upper area :)

cmp https://github.com/deltachat/deltachat-ios/pull/708

before / after:

 &nbsp; &nbsp;


adbenitez (approved):

r10s:
ah, i just noticed, also the mute-button is superfluous ...

PR #2595, opened 2025-02-09
hack around broken 'Add Second Device' QR code

r10s:
core 1.155.2 accidentally broke compatibility
to add-second-device QR codes of previous versions;

this is esp. bad as ppl will have different versions for some days at least, weakening overall UX, esp. of first-time-users
that may come to delta because of praised, seamless multidevice ... :)

this PR is a little experiment, if it is really only the minor JSON change, or sth deeer.

seems, these two repacements fix the issue,
allowing us to release much mess stressful,
while users being happy.

if we think, this is fine,
we should do that better in core, however.

the hack can be removed 3 releases later or so.

see https://github.com/deltachat/deltachat-core-rust/issues/6518 for more details and for snippets how the QR code looks and what this hack is doing

cc @link2xt

### DIFF
--- a/deltachat-ios/DC/DcContext.swift
+++ b/deltachat-ios/DC/DcContext.swift
@@ -301,12 +301,21 @@ public class DcContext {
         return nil
     }
 
+    private func fixAddSecondDeviceQR(_ qrCode: String) -> String {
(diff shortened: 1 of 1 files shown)


r10s:
replaced by https://github.com/deltachat/deltachat-core-rust/pull/6528

PR #2606, opened 2025-02-12
share files generated by webxdc

r10s:
before, one could only send files to chats

TODO: better file name, peek at viewlog


deltachat-ios/Controller/WebxdcViewController.swift
```diff
@@ -528,6 +528,19 @@ extension WebxdcViewController: WKScriptMessageHandler {
                         rootController.popToRootViewController(animated: false)
                     }
                 }))
+
+                if let name = dict["name"] as? String {
+                    alert.addAction(UIAlertAction(title: String.localized("menu_share"), style: .default, handler: { [weak self] _ in
+                        guard let self else { return }
+                        if let base64 = dict["base64"] as? String, let data = Data(base64Encoded: base64), let sourceItem = navigationItem.rightBarButtonItem {
+                            let fileURL = FileManager.default.temporaryDirectory.appendingPathComponent(name)
+                            try? FileManager.default.removeItem(at: fileURL)
```
zeitschlag, line 537: I assume this is to clean up in case there’s already such a file, to make sure that we’re sharing the right data, right?
I’m wondering that no cleanup seems to happen after sharing.

zeitschlag (approved):

Amzd (approved):

deltachat-ios/Controller/WebxdcViewController.swift
r10s: no, this is to make sure, the copying does not fail next line (in case you share files with same names)

for cleanup, we rely on the system, it is a cache, this is also what we're doing elsewhere

maybe thar part could be optimized for all cache hsages (in case there js a callback-or-so when sharing is really done) - but not sure if it is worth the effort

PR #2608, opened 2025-02-14
read reactions in message summary

r10s:


deltachat-ios/Chat/Views/Cells/AudioMessageCell.swift
```diff
@@ -50,9 +50,9 @@ public class AudioMessageCell: BaseMessageCell, ReusableCell {
             mainContentView.spacing = 0
         }
         if msg.type == DC_MSG_VOICE {
-            accessibilityLabel = String.localized("voice_message")
+            a11yDcType = String.localized("voice_message")
```
r10s, line 53: i really wanted a name that made clear, it is _not_ a system var

adbenitez (approved):

PR #2625, opened 2025-03-03
update README screenshots 

r10s:
this PR updates the screenshots in the README.

screenshots taken from https://github.com/deltachat/interface/tree/main/screenshots/2025/ios


Amzd:
The quote reply in that screenshot is missing the name of the person being replied to 🤔 

image

r10s:
hm, a bug? is that reproducible?

PR #2664, opened 2025-04-08
remove email addresses from primary settings screen

r10s:
this is one of the last primary UI elements where the email address was shown. 

~~the email address of contacts is still shown (and will probably stay there), in "Encryption Info" where you can also check the fingerprint.~~
EDIT: this PR only changes the settings screen, the contact profiles are left as is for now

 your _own_ email address is still available at "Advanced / Password & Account".

to avoid confusion, and things sometimes working in a way and sometimes not, we hide the address _unconditionally_ of chatmail: telling _self_ addresses are anyway known and
other addresses will make troubles, eg. when you tell them to a classic mail user, e2ee probably fails, when telling them to a delta/chatmail user, they cannot enter it somewhere

as a side effect, this further improves screenshot privacy and over-the-shoulder attacks, cmp https://github.com/deltachat/deltachat-ios/pull/2329

iOS is spearheading here, once this is merged, we can also do the change on android/desktop, but there is no urgency

review is not so much about the code, but about the fact that remove the addresses there

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 - Protect metadata: encrypt message's sent date
 - Do not fail to send messages in groups if some encryption keys are missing
 - Synchronize contact name changes across devices
-- Hide address in primary profile UI (for a better UX as contacts via addresses do not get encrypted)
+- Improve UX by hiding address in primary settings screen; passing it around without e2ee information won't work
 - Fix: make vcards compatible to Proton Mail again
 - Fix: enncrypt broadcast list and make them work again with chatmail
 - Fix changing group names that was not working in some situations
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Protect metadata: encrypt message's sent date
 - Do not fail to send messages in groups if some encryption keys are missing
 - Synchronize contact name changes across devices
+- Hide address in primary profile UI (for a better UX as contacts via addresses do not get encrypted)
 - Fix: make vcards compatible to Proton Mail again
 - Fix: enncrypt broadcast list and make them work again with chatmail
 - Fix changing group names that was not working in some situations
--- a/deltachat-ios/Controller/ContactDetailViewController.swift
+++ b/deltachat-ios/Controller/ContactDetailViewController.swift
@@ -322,7 +322,7 @@ class ContactDetailViewController: UITableViewController {
             headerCell.showMuteButton(show: false)
         } else {
             headerCell.updateDetails(title: viewModel.contact.displayName,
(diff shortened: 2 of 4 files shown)


Amzd (approved):
I understand that this is probably needed for multitransport but I don't like how much harder this makes it to identify who you are messaging (or mailing, since I think this applies to non-encrypted mail too?). 

Code looks fine.

Hocuri:
I also think that it would be better to leave the email address visible in the profile without an extra tap:
1. We do still have users who use DC to chat with regular email users. As long as we don't want to completely remove this possibility, we should make sure that it's at least somewhat usable. In "Encryption info", the email address is hidden so well that even experienced users will take some time to find it. And for regular users, the email address is what people use to identify contacts, so it's relevant even for existing contacts.
2. The chatmail address is a bit comparable to a WhatsApp phone number (which can also be changed, just as chatmail addresses - only that you can't have multiple numbers in WA). WhatsApp also shows the number in the profile, but not in the chat. After we have multi-transport, we can think about replacing the chatmail addrs by the key fingerprint for multi-transport contacts.
3. I don't think it disturbs from important information in the profile view. You need to have a bit of tech literacy to even open the profile view.

r10s:
it is not about disturbing, we could fine a place there. reasons are that it shall not be used.

but yes, maybe it is too radical :)

what about: 

- leaving address in profile as is, so, you have another data point when looking at the profile.
we anyway aim to improve the profile as prototyped on desktop, when avatar+name is larger, the address can get a little more down

- but: in the settings (leftmost screenshot), remove the self address. i think, this is the more urgent issue, we've several times heard of ppl using this very exposed address, and get issues that they cannot be reached when someone writes to them from email.

i think, _something_ we have to do to nudge ppl to other ways to get in contact - and directly visible email adresses are bad.

Hocuri:
Leaving the profile as-is for now, but removing the email address from the settings main screen, sounds good to me. Good points about the whatsapp phone number and about preventing people from using the email address.

For non-chatmail accounts, the email address is also shown in the account switcher (at least on Android); I think it should stay there, because for non-chatmail accounts, the email address is the main identifier to tell apart your own email accounts.

r10s:
k, i changed the PR, it now only hides the self-address unconditionally in the settings main screen, where it was super-present before. reasons for removal are explained above

thanks a lot for constructive review, this is very welcome in that bit hard to navigate part, in times somewhere between classic and chatmail and different expectations, and things not yet full changed etc.

Amzd (approved):

Hocuri (approved):

Hocuri:
Nice, thanks for driving this effort!

PR #2672, opened 2025-04-15
Tapping a notification when app is terminated now opens context

Amzd:
Previously we did not set the UNUserNotificationCenter delegate synchronously in `application(_:didFinishLaunchingWithOptions:)` which meant that if the app was not already running, a tapped notification did not call `AppDelegate.userNotificationCenter(_:didReceive:withCompletionHandler:)` (because AppDelegate was not yet the ununc delegate) which meant it did not navigate to the chat.

This is specifically mentioned in a comment on `UNUserNotificationCenterDelegate.userNotificationCenter(_:didReceive:withCompletionHandler:)
`
> The delegate must be set before the application returns from application:didFinishLaunchingWithOptions:

image

This was **not** a regression from #2534 therefore I added changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Group notifications in Notification Center by chat
 - More characters in notification before truncating
 - Fix: separate multiple notifications coming in at the same time
+- Fix: tapping a notification when app was terminated now opens the notifications context
 
 ## v1.58.1
 2025-04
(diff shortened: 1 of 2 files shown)


r10s (approved):
okay, so it was probably working, but only not in the test scenario i did, using force quit and then try a notification. thanks for fixing anyways, it's good for testing or first impression of users testing :)

PR #2678, opened 2025-04-25
send multiple images

r10s:
continuing closed https://github.com/deltachat/deltachat-ios/pull/2565 here

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Select and send up to four images or videos
 - Support pasted GIFs
 - Group notifications in Notification Center by chat
 - More characters in notification before truncating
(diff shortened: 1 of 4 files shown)


r10s:
testing a bit:
- iOS 15: sending screencasts is not working, one gets the error "cannot be completed", probably sth. in convertToMp4(), that seems to need more love (change here is that the source comes from the newer picker)
- on iOS 18 that is fine 
- if in doubt, we could allow multiple selection only from some version on @Amzd would be interesting if sending screencasts from iOS 17 is okay

r10s:
so, all in all, i am suspecting `timeRange` somehow

r10s:
error code is -17508 , there is sth about that at https://developer.apple.com/forums/thread/121902

r10s:
k, i give up for now, only iOS 18+ will have the option to select multiple images/videos in one go

r10s:
one difference i encountered is that convertToMp4 for UIImagePickerController is called from main thread,
while for new PHPickerViewController it is called from a worker thread

r10s:
finally: progress 🎉 using `loadInPlaceFileRepresentation` instead of `loadFileRepresentation` makes videos with the new picker work on iOS 15 - according to https://developer.apple.com/documentation/foundation/nsitemprovider/loadinplacefilerepresentation(fortypeidentifier:completionhandler:) , the file is still copied as needed. this is also better performance-wise; we create a copy in convertToMp4

r10s:
finally: progress 🎉 using `loadInPlaceFileRepresentation` instead of `loadFileRepresentation` makes videos with the new picker work on iOS 15 - according to https://developer.apple.com/documentation/foundation/nsitemprovider/loadinplacefilerepresentation(fortypeidentifier:completionhandler:) , the file is still copied as needed. this is also better performance-wise; we create a copy in convertToMp4

still a bit unclear to me what is going on there, the only difference was that we got called from worker thread and not from main thread - but the idea that the file was not long-living enough was there before and we tried copying - without success

r10s:
merging in to move forward

PR #2681, opened 2025-05-05
use 'ppl' icon for 'invite friends' - 

r10s:
use 'ppl' icon for 'invite friends' - the 'heart' will be used for 'donate' in the future,
see https://github.com/deltachat/deltachat-android/pull/3757

#skip-changelog


r10s:
hm, looking at it after a week, i do not really like that. let's close for now and reconsider when we actually do the donate stuff. maybe it is better to just have no icon for "invite" then

adbenitez:
ftr, I think it was not bad,but maybe there is other icons that would fit as well

Issue #1840, opened 2023-04-15
more convenient option to attach image from gallery as original


Simon-Laux:
One argument for always compressing images was that the send as file option is still available.
I recently needed it and couldn't find it.
When sending files it does not allow me to select images from the gallery, I think that is different from android as it might be possible there to send images from the gallery via the file explorer.

The only work around I could think of it to go into the gallery export the image from there to files and then select it in delta chat with attach file. Though that is hardly convenient and you need to know that this possibility exists to use it as it is not really intuitive for normal users.

r10s:
indeed, this is not that simple. not sure, but i think, this was different in the past, where you could just select gallery images from the file explorer; maybe sth. has changed in the os.

we should check that and maybe add sth. more comfortable, until then, exporting from gallery is inconvenient, but not the end of the world

r10s:
asked some more experienced whatsapp and ios user:

the issue is exactly the same for whatsapp. however, it is be pretty known to advanced ios users that one has to export images as files before they can be send as original with whatsapp. so, going the same path here, is good enough for now.

same for signal, seems at least. telegram has other options.

so, we're as good or bad as whatsapp/signal here, i would not consider this as a bug any longer.  there are more important things to fix and to put our limited resources on :)

r10s:
nb: if someone wants to do sth. about this, i suggest an _unobtrusive_ three-dot-menu top-right of the staging area with the option "Send as original file" - once tapped, the staging area converts to the view usually shown for files, so, showing bytesize etc. the option must not be available for already edited files.

this way, we would not clutter the ui for normal users.

same idea can apply to android/desktop.

but again, i do not see that as important in anyway - even less by the information wrt whatsapp/signal.

Simon-Laux:
Just forgot about the compression, send myself some documents as images and they were unreadable. If there were an option for it in the staging area or while attaching I wouldn't have missed that. really annoying issue when you run into it.

r10s:
moving discussion to the forum https://support.delta.chat/t/more-convenient-option-to-attach-image-from-gallery-as-original/2761/2

PR #2683, opened 2025-05-06
attach original images and files directly

r10s:
this PR shows a little dialog before the file selector, allowing to choose from gallery as well. this is also what WhatsApp and Telegram are doing meanwhile - as we were working anyway in that area, that seemed to be an easy improvement (before, you had to go to system gallery, and save a file to the "Files App" before you could attach it)

closes https://github.com/deltachat/deltachat-ios/issues/1840

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -2401,7 +2401,16 @@ extension ChatViewController: MediaPickerDelegate {
                 if let typeIdentifier = itemProvider.registeredTypeIdentifiers.first {
                     itemProvider.loadFileRepresentation(forTypeIdentifier: typeIdentifier, completionHandler: { [weak self] url, error in
                         if let url {
-                            self?.stageDocument(url: url as NSURL)
+                            do {
+                                let copyURL = FileManager.default.temporaryDirectory.appendingPathComponent(url.lastPathComponent)
+                                if FileManager.default.fileExists(atPath: copyURL.path) {
+                                    try FileManager.default.removeItem(at: copyURL)
+                                }
+                                try FileManager.default.copyItem(at: url, to: copyURL) // create a copy as the file is too short-living otherwise
+                                self?.stageDocument(url: copyURL as NSURL)
+                            } catch {
+                                self?.logAndAlert(error: error.localizedDescription)
+                            }
                         } else if let error {
                             self?.logAndAlert(error: error.localizedDescription)
                         }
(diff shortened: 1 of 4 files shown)


deltachat-ios/Helper/MediaPicker.swift
```diff
@@ -75,28 +76,42 @@ class MediaPicker: NSObject, UINavigationControllerDelegate {
...
         documentPicker.modalPresentationStyle = .formSheet
         navigationController?.present(documentPicker, animated: true)
     }
 
-    func showGallery(allowCropping: Bool = false) {
+    func showGallery(allowCropping: Bool = false, sendAsFile: Bool = false) {
         // we have to use older UIImagePickerController as well as newer PHPickerViewController:
         // - only the older allows cropping and only the newer allows mutiple selection
         // - the newer results in weird errors on older OS, see discussion at https://github.com/deltachat/deltachat-ios/pull/2678
+        self.sendAsFile = sendAsFile
```
Amzd, line 108: hmmm mediapicker is stored on chatviewcontroller so setting this and never clearing it makes it set for the next attachment you add too? so:
- send image as file 
  - `onMediaSelected` called with sendAsFile true, as expected
- then send an image from the camera
  - now `onMediaSelected` is called with sendAsFile true (I think?) because `ChatViewController.mediaPicker.sendAsFile` is still true

Amzd (changes requested):

deltachat-ios/Helper/MediaPicker.swift
r10s: i added a commit

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -2396,7 +2397,28 @@ extension ChatViewController: MediaPickerDelegate {
         } else if let itemProvider = itemProviders.first {
 
             // stage a single selected item
-            if itemProvider.hasItemConformingToTypeIdentifier(UTType.movie.identifier) {
+            if sendAsFile {
+                if let typeIdentifier = itemProvider.registeredTypeIdentifiers.first {
```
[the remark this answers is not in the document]
r10s: well, in all my tests, it was fine. 

i found that approach on some stackoverflow, but sure, they can be wrong as well.

i'd say, let's see. the approach is reasonable (send the first type we got) - if it fails, it would not be a worsening as you still can "send from Files" as before, the change is just an additional offering

deltachat-ios/Helper/MediaPicker.swift
Amzd: > paths not starting with `showGallery()` do not end up in `onMediaSelected()`

Ah I did not realize this, feel free to ignore.

PR #2690, opened 2025-05-08
[wip] re-focus profile

r10s:
this PR aims to re-focus the profile so that it can be used easier and nicer to recognize ppl, which is more important by the tune-down of email adresses

im particular:

- make avatar and name bigger
- bring avatar/name/bio together
- more focus on avatar/name/bio and shared-chats
- other, more rarely used options and options that are not used for identification are moved to the menu
- introducing a menu also makes the options easier accessible in case there are many members or shared chats (no need to scroll down)

### DIFF
--- a/deltachat-ios/Controller/ProfileViewController.swift
+++ b/deltachat-ios/Controller/ProfileViewController.swift
@@ -257,12 +257,9 @@ class ProfileViewController: UITableViewController {
             title = String.localized("profile")
         }
 
-        if !isSavedMessages && !isDeviceChat && (contact != nil || isMailinglist || (isGroup && chat?.canSend ?? false)) {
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/AllMediaViewController.swift
```diff
@@ -48,6 +48,12 @@ class AllMediaViewController: UIPageViewController {
         self.dcContext = dcContext
         self.chatId = chatId
         super.init(transitionStyle: .scroll, navigationOrientation: .horizontal, options: [:])
+
+        // avoid accidental flashing, let it be two taps to get to the image gallery, also for global media
+        if chatId == 0, let first = pages.first {
+            pages.removeFirst()
```
Amzd, line 54:
```suggestion
        if chatId == 0 {
            let first = pages.removeFirst()
```

Amzd (approved):
Screenshots looking great!

Issue #2659, opened 2025-04-02
improve share proxy view


r10s:
- the subtitle and the button "share link" is too small, it should use the standard font size
- no left/right padding for the subtitle (see eg. longer translations)
- the subtitle does not wrap (if that's tricky, two static lines should do)

r10s:
closed by https://github.com/deltachat/deltachat-ios/pull/2709

PR #2709, opened 2025-05-17
tweak share proxy layout

r10s:
this works on screens i tested, however, i still do not have an easy way to say "i have three views, let the first consume all additional space" - so on this case, let the qr code consume everything while staying square, cc @Amzd maybe you have a better idea :)

for the other points of https://github.com/deltachat/deltachat-ios/issues/2659  - they are good enough now:
- seems the font size is default, might appear a little smaller as the system adapts. that's fine
- the left/right is also fine, it is now wrapped, so usually the border is no longer touched on long strings, only small or even no padding is good enough

### DIFF
--- a/deltachat-ios/Controller/Settings/Proxy/ShareProxyViewController.swift
+++ b/deltachat-ios/Controller/Settings/Proxy/ShareProxyViewController.swift
@@ -34,6 +34,7 @@ class ShareProxyViewController: UIViewController {
         explanationLabel.font = UIFont.preferredFont(forTextStyle: .caption1)
         explanationLabel.text = String.localized("proxy_share_explain")
         explanationLabel.numberOfLines = 0
+        explanationLabel.textAlignment = .center
 
         contentStackView = UIStackView(arrangedSubviews: [qrContentView, explanationLabel, shareLinkButton, UIView()])
         contentStackView.translatesAutoresizingMaskIntoConstraints = false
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/Settings/Proxy/ShareProxyViewController.swift
```diff
@@ -64,22 +67,14 @@ class ShareProxyViewController: UIViewController {
...
-            qrImageRatio = 1
-        }
-
         verticalCenterConstraint = contentStackView.centerYAnchor.constraint(equalTo: contentScrollView.centerYAnchor)
         contentTopAnchor = contentStackView.topAnchor.constraint(equalTo: contentScrollView.topAnchor, constant: 16)
         contentBottomAnchor = contentScrollView.bottomAnchor.constraint(equalTo: contentStackView.bottomAnchor, constant: 16)
 
         let constraints = [
-            qrContentView.widthAnchor.constraint(lessThanOrEqualTo: view.safeAreaLayoutGuide.widthAnchor, multiplier: 0.75),
+            qrContentView.widthAnchor.constraint(lessThanOrEqualTo: view.safeAreaLayoutGuide.widthAnchor, multiplier: 0.65),
```
r10s, line 75: making it a little smaller, this is fine on all devices i tested.

however, having it more adaptive would be great!

PR #2713, opened 2025-05-20
use default notification sound

r10s:


r10s:
also tested, now i have sound again :)

PR #2722, opened 2025-05-23
align avatar to message bubble

r10s:
this PR aligns the avatar in groups to the bottom  of the bubbles, not to reactions that may or may not be there.

moreover, the avatar is made slightly bigger

before / after:

 &nbsp; 

/me now getting some rum


Amzd (approved):
I love all the small qol changes lately ❤️❤️

Issue #2488, opened 2025-01-08
global search does not find emoji in messages


Raiden-GH:
The global search in DC-iOS only recognizes emojis in the name of the chats but not in the chats themselves. Since this is not the case in DC-Android and DC-Desktop, it is probably not a core bug.

iOS 18.2
DC-iOS 1.50.3

r10s:
i could reproduce that, i created a chat, a contact and a message each containing the emoji `😇`:

<img width=300 src=https://github.com/user-attachments/assets/fef54a39-4750-424e-8917-5de296b63a80><br>

doing a global search for `😇` finds it in chat name and contact name - but not in the message. moreover, the highlighting breaks the display:

<img width=300 src=https://github.com/user-attachments/assets/e5f251f2-7b77-4b85-932f-314b25f05a27>

doing a in-chat-search for the message, however is fine - the message is found (2 hits here) and also the highlighting is fine (see the background color around `😇` in the last message):

<img width=300 src=https://github.com/user-attachments/assets/c2874b57-729e-4f68-873d-3642c2e221b2>

Amzd:
This has to do with how the string is encoded. I tried a couple options in `.cString(using:)`

- .iso2022JP works but idk the side effects 
- .symbol works for an emoji but not for text

Jsonrpc call works so maybe easiest to just convert to that. Have to figure out if we still need DcChatlist.

r10s:
hm, but strings are all returned from core the same way. also, strings are given to core the same way. so, why is the hiliting in the in-chat-search working? (the hilighting is done in UI, not in core, afaik)

Amzd:
I don't know how core works nor can I fully comprehend rust but it says lossy conversion to string here: https://github.com/deltachat/deltachat-core-rust/blob/4ec50d19907f9b64bf144d37a004dcc62fc3ed07/deltachat-ffi/src/lib.rs#L915

And when I do `"❤️".cString(using: .lossy)` in Swift I get nil, so maybe it's something to do with that?

r10s:
maybe switching to jsonrpc for the search api solves the issue, on android it is fine. weird thing is that some searches (for chats, contacts) work

PR #2725, opened 2025-05-24
Fix: No weird results when searching for space

r10s:
treat a search for a single space as any other space.

bug before was that there were two methods to check for "empty" string, just use a single one now, `isEmpty`

this is how the search for a single space looked liked - and tapping the result led to the scary ghost chat 👻


Amzd:
Hmm does this maybe also fix searching for a single emoji?

r10s:
you mean https://github.com/deltachat/deltachat-ios/issues/2488 ? just tried, this is not fixed by  this PR

PR #2730, opened 2025-05-25
add shortcut to start in-chat search

r10s:
in-chat search is started usually from the profile menu, which is straight-forward,
but requires some taps (tap title, tap menu, tap search)

this adds an shortcut to open the search by long-tap title. this is known from telegram, it is advanced, but that is fine, as it should not disturb things if unused/unknown.

this also replaces the extra search icon for the saved messages - for consistency, but also as we have with the 'app & media' another button there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - "All Apps & Media" are available from the settings
 - Clearer app lists by removing redundant "App" subtitle
 - Speed up opening profiles
+- Long-tap chat title as a shortcut to start in-chat search
 - Improve hint for app drafts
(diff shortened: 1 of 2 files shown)


Amzd:
Is having 2 tab bar buttons not easier?

r10s:
we have two buttons for "saved messages", where we do not have the avatar. otherwise, things become a bit dense.

as said, the in-chat-search is also not that inportant, justifying using that space unconditionally

we might get more space, eg "green checkmark" might go away - but still, i would not use that for a search button ... there might be an icon needed by calls as well

PR #2733, opened 2025-05-28
switch to transport API

### LINKED ISSUE #2650: Switch from configure to transport API
In preparation to multi-transport, there are new APIs that directly take the configuration as an argument.
`configure()` is deprecated.
In v1.157.3, the API is buggy; it is fixed in v1.158.0.

This is not urgent, but if you have some time, would be nice to start working on it already so that the UIs are ready when multi-transport is introduced in core.

You need to use these APIs now:

- New API `add_transport_from_qr()` replaces calling `set_config_from_qr()` and then `configure()`.
- New API `add_transport()` replaces repeated calls to `set_config()` and then `configure()`.

What to do:
- Read or skim `drafts/2025-02-10-multi-transport.md` in the bg repo (all the important information is in the beginning, the later sections are less interesting for you)
- Read or skim the new API documentation at https://js.jsonrpc.delta.chat/classes/RawClient.html#addTransport and https://js.jsonrpc.delta.chat/classes/RawClient.html#addTransportFromQr
- Use the new APIs instead of `configure()`. If it helps, here is how I did it for Android: https://github.com/deltachat/deltachat-android/pull/3676/files
  - If anything is unclear, let me know!
- Note that it's not necessary to call `stop_io()` and `start_io()` manually before and after calling `addTransport*()`. (this was necessary for the C `dc_configure()` API).
- Remove the dialog about AEAP (which shows the `aeap_explanation` translated string), because AEAP will be removed. I did it for DC Android here: https://github.com/deltachat/deltachat-android/pull/3698

Core PRs introducing the API: https://github.com/chatmail/core/pull/6589, https://github.com/chatmail/core/pull/6680

r10s:
closes #2650

#skip-changelog as this refactoring is no user-visible change (they will come with [PGP-contacts](https://github.com/chatmail/core/pull/6796) branch later)


deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingViewController.swift
```diff
@@ -282,14 +283,8 @@ class InstantOnboardingViewController: UIViewController {
             guard let self else { return }
 
             let qrCodeData = self.qrCodeData ?? "dcaccount:https://nine.testrun.org/new"
-            let success = self.dcContext.setConfigFromQR(qrCode: qrCodeData)
             DispatchQueue.main.async {
-                if success {
-                    self.dcAccounts.stopIo()
-                    self.dcContext.configure()
-                } else {
-                    progressAlertHandler.updateProgressAlert(error: self.dcContext.lastErrorString, completion: nil)
```
Hocuri, line 291: Have you tested that showing an error still works?

r10s: `addTransportFromQr` does not return an error state, or? how shall i check for errors? do we get the same events as by `configure()`

r10s: k, i will have a closer look. what is a good way to force an error, so i know what to test for? passing "foo bar" or bad accountId to addTransportFromQr?

deltachat-ios/DC/DcEnteredLoginParam.swift
```diff
@@ -0,0 +1,38 @@
...
+    public var password: String?
+    public var smtpPassword: String?
+    public var smtpPort: Int?
+    public var smtpSecurity: String?
+    public var smtpServer: String?
+    public var smtpUser: String?
+
+    public init(addr: String) {
+        self.addr = addr
+    }
```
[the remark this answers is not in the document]
r10s: you're right, i mixed that up with `smtpPassword`

deltachat-ios/Controller/AccountSetup/Instand Onboarding/InstantOnboardingViewController.swift
Hocuri: Enter email address alice@example.net and password 1234. Or disable Wi-Fi on your phone.

PR #2737, opened 2025-06-02
show recode errors

r10s:


r10s:
> Maybe set isCanceled to true in the error function and check for isCanceled before changing message in the other functions?

i added a commit that keeps showing the error, let's see if that is sufficient in practise

r10s:
would be cool to get that in as https://github.com/deltachat/deltachat-ios/pull/2738 is already a nice test for that :)

Amzd (approved):

r10s:
merging that in, #2737 and #2738 will use it and it makes sense to test it there and to avoid conflicts.

we can improve in a follow-up as needed

PR #2738, opened 2025-06-03
fix for not able to compress videos on iOS 16

Amzd:
This PR addresses missing features some of which forgotten when switching from UIImagePicker to PHPicker:

- Attach videos (on iOS 15 and iOS 16)
- Attach webP 
- Attach GIF
- Drop videos (on iOS 15)

I did some refactoring;
- Moved the logic for loading videos and images to a NSItemProvider extension so it can be used from drag/drop and the attach menu (in the future maybe the share sheet too but putting this extension in DcCore requires it to depend on SDWebImage). This means we no longer have slightly different logic for these areas (at least for image and video).
- Made the custom loading functions for videos and images return on main thread which saves a lot of swapping around everywhere. 
- That allowed a simpler `onMediaSelected`

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -2371,14 +2371,16 @@ extension ChatViewController: MediaPickerDelegate {
                         if !progressAlertHandler.cancelled {
                             if itemProvider.hasItemConformingToTypeIdentifier(UTType.movie.identifier) {
                                 itemProvider.loadInPlaceFileRepresentation(forTypeIdentifier: UTType.movie.identifier) { [weak self] url, _, error in
-                                    if !progressAlertHandler.cancelled {
-                                        url?.convertToMp4 { [weak self] url, error in
-                                            if let url, !progressAlertHandler.cancelled {
-                                                self?.sendVideo(url: url)
-                                            } else if let error {
-                                                progressAlertHandler.updateProgressAlert(error: error.localizedDescription)
+                                    if !progressAlertHandler.cancelled, let url {
+                                        NSFileCoordinator().coordinate(readingItemAt: url, error: nil) { url in
+                                            url.convertToMp4 { [weak self] url, error in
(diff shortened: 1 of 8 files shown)


r10s:
this fails on iOS 15, however, after double-checking with current main: 
- iOS 15 was also failing before when  select-from-gallery
- iOS 15 was fine when recording a video directly with delta chat (this is what i was recently testing)
- iOS 18 is fine with this PR for both, record-in-delta and select-from-gallery

will investigate what is happeing on iOS 15 ... why cannot one of the biggest tech company has more friction less api ... they are spending too much money on bullshit

Amzd:
In my working tree while pushing this I had also changed that back to `loadFileRepresentation`

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -2356,35 +2356,27 @@ extension ChatViewController: MediaPickerDelegate {
...
-                    DispatchQueue.main.async {
-                        progressAlertHandler.updateProgressAlertValue(value: (processed*1000) / itemProviders.count)
-                    }
+                    progressAlertHandler.updateProgressAlertValue(value: (processed*1000) / itemProviders.count)
                     if processed >= itemProviders.count {
-                        DispatchQueue.main.async {
-                            progressAlertHandler.updateProgressAlertSuccess()
-                        }
+                        progressAlertHandler.updateProgressAlertSuccess()
                     }
```
r10s, line 2362: so, it is guaranteed that `loadImage` and `loadCompressedVideo` call the provided backup on main now?

probably makes sense

r10s (approved):
thanks a lot again for diving into that issue!

code lgtm, did not test drag'n'drop nor gif nor webp, however

deltachat-ios/Chat/ChatViewController.swift
Amzd: Yeah since all the work already happened inside those helper functions there's no reason to call the completion on bg thread

PR #2740, opened 2025-06-07
Fix message bubble ghosting

Amzd:
Before this PR context menu preparations created a ghosting message cell when canceling the context menu by scrolling.


r10s (approved):
thanks a lot for fixing!

using this since 12 hours without issues, and code lgtm

PR #2742, opened 2025-06-08
Attempt to prevent corruption when forwarding attachment

Amzd:
I have had a few times that the original video is removed when I forward a video from QLPreviewController's share button into DC share extension. I cannot reliably reproduce this but it happened twice so out of caution I remove this deletion of temporary file as temp files will be deleted by the system anyway.


r10s (approved):

Amzd:
Was not released so I skip changelog

Issue #2739, opened 2025-06-06
Reaction applies to wrong message when you receive a message while selecting reaction


Amzd:
- iOS version:
- Device:
- Delta Chat version:
- Expected behavior:
- Actual behavior:
- Steps to reproduce the problem:

Long press message 
Receive message
Select reaction
Reaction is applied to wrong message
- Screenshots:
- Logs:

PR #2743, opened 2025-06-08
Fix contextmenu actions acting on incorrect message

Amzd:
issue #2739 

This PR doesn't fix it for selectMore as that's more involved but should probably also be fixed some day.


r10s (approved):
lgtm, code also running and testing :)

PR #2745, opened 2025-06-14
remove all data but msgId from notification logging

### LINKED ISSUE #2744: Do not log notifications contents
This line logs notification contents:
https://github.com/deltachat/deltachat-ios/blob/67ebb908f5e038be353cb9ee7519f396ca171840/deltachat-ios/Helper/NotificationManager.swift#L107

Users sometimes send logs to developers for debugging and the logs on iOS contain message texts because of this. It's fine to log database msgID, Message-ID or something like this, but not display names or message contents.

r10s:
closes #2744 

thanks for filing the isssue, @link2xt


Amzd (approved):
Great catch

PR #2746, opened 2025-06-18
remove lock signs and use email-symbol instead

r10s:
this PR removes the locks (aka "handbags") shown for encrypted messages and shows instead an email-symbol (an "envelope") if the message is not encrypted.

reason is to have less cluttering in the UI. also, in general, it is better to show a flaw than a feature icon, as there is fewer incentive to fake it.

finally, the email-symbol will be [shown for unencrypted chats as well soon](https://github.com/chatmail/core/pull/6916), so you have that all over the place for classic emails. 

TODO: add a PR in core to remove the email-symbol from the device chat's messages

### DIFF
--- a/deltachat-ios/Chat/Views/StatusView.swift
+++ b/deltachat-ios/Chat/Views/StatusView.swift
@@ -6,7 +6,7 @@ public class StatusView: UIView {
     private let contentStackView: UIStackView
     let dateLabel: UILabel
     private let editedLabel: UILabel
-    private let padlockView: UIImageView
+    private let envelopeView: UIImageView
     private let locationView: UIImageView
     private let stateView: UIImageView
     private let savedView: UIImageView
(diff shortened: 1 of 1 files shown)


Hocuri:
> also, in general, it is better to show a flaw than a feature icon, as there is fewer incentive to fake it.

FWIW, I don't think we're succeeding in this, because the letter-icon will not be perceived as a flaw. I'm still in favor of this change, because we have other measures to enforce encrypted messages, like not accepting any unencrypted messages on chatmail servers.

Hocuri (approved):
I can't test or say anything about the code, but conceptually, this LGTM.

Did you test that the letter-icon looks fine on outgoing messages sent by an MUA that is used on the same account?

Edit: BTW, see also https://github.com/chatmail/core/pull/6925

r10s:
well, for "flaw", i meant more the fact, that there is no incentive to fake it (as for he green checkmark, which may be confused by some emoji), pretending eg. smth is encrypted. i agree that an email-symbol is not regarded as smth bad in general :)

PR #2751, opened 2025-06-24
Properly close webxdc

Amzd:
Fixes memory leak in WebxdcViewController which caused realtime to continue in the background.

Previously the local scope function `actions` captured self strongly which caused the leak in iOS 15+ because the self capturing function was stored on the menu button (on self) through `UIDeferredMenuElement.uncached` closure. I have changed the function to a closure which makes it hopefully more clear that we need to weak ref there.

Regression from #2545


adbenitez (approved):
cant tell much about the code but looks reasonable

Issue #2673, opened 2025-04-15
even more reliable notifications


r10s:
notifications have improved hugely the last year, but there are still some **open ends,** getting more of an issue when it comes to [calls](https://github.com/deltachat/deltachat-ios/pull/2638):

- [x] 1. NSE and App must not use I/O at the same time. currently, this is handled by a boolean that is set/cleared on start/stop. if for whatever reason, clearing is skipped, this at best results in  "You have new messages"
targeted by https://github.com/deltachat/deltachat-ios/pull/2759

- [x] 2. if the NSE uses too much memory, it is terminated by iOS, resulting in a generic "You have new messages".  
in testflight releases, there is an additional, visible crash
targeted by https://github.com/chatmail/core/pull/7431, which avoids using too much memory by not downloading large messages in `dc_accounts_background_fetch()`

- [x] 3. if NSE times out as a downloading a large file takes too much time, or crashes for other reasons, this again results at best in "You have new messages" - at worse, this might lead to double messages (when crash is done after adding but before last-seen is written, not sure on that point, we just observerd that one time). NSE gets ~ 30 seconds.  
in testflight releases, there is an additional, visible crash
targeted by https://github.com/chatmail/core/pull/7431, which downloading large messages in `dc_accounts_background_fetch()`

- [x] 4. even if we can avoid crashing by quitting elegantly, large files may clog up the queue so a call is never ringing. since messages are fetched chronologically only, we will never get past a large attachment in the NSE process because there is not enough memory, until the main app is opened which can then decrypt the large messages.
targeted by https://github.com/chatmail/core/pull/7431

while this "You have new messages" is okayish for normal messages, it is bad for let calls ringing (we get the call "ring" message e2ee through the same NSE, and need to call `reportNewIncomingVoIPPushPayload()` from there - this should be done in time and as reliable as possible.

but even without calls, it is better to have as few "You have new messages" as possible.

there are the following ideas around to **improve the situation:**

- #2759 - avoid running at the same time by IPC - which is really not standard on iOS, therefore the current hack. anyways, thanks @Amzd for figuring out that "Darwin Notification Center" can be used for that. This solved 1.  

- [use less memory by streaming decryption](https://github.com/rpgp/rpgp/pull/453), that alone would improve things hugely, as files of very few mb already cause issues (iirc, depending on iOS version NSE may consule 12~25 mb, and core needs 4x file size for decompression/decryption).  this would target 2. and partly 3.  
[...]

PR #2759, opened 2025-06-27
NSE check if main app is running 

Amzd:
This changes the mechanism to check if the app is running from a UserDefaults boolean to pinging back and forth between app and NSE process using Darwin notification center. This was needed because the UserDefaults boolean would sometimes not be updated when the app was terminated by the system. This meant that notification service extension thought the app was running and did not fetch notifications, resulting in no notifications until the user opened the app again.

This is point 1 on #2673

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,9 +15,10 @@
 - Improve logging and error handling
 - Replace "Broadcast Lists" experiment by "Channels"
 - Indicate which message was scrolled to (eg by tapping a quote or a notification)
+- Fix: No longer missing notifications when app was terminated by the system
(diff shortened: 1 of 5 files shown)


r10s:
ftr, this targets the first issue of https://github.com/deltachat/deltachat-ios/issues/2673 - i am running this PR since some weeks - so far, things work fine for me :)

r10s (approved):
lgtm,

imu, mainIoRunning and nseFetching seems still be needed to show UI eg. as "fetching..." - but it is no longer critical for _not_ starting NSW

i tried a rebase using github, however, there was some conflict in the core-submodule i did not understand, so i removed the last commit ...

PR #2761, opened 2025-06-29
Extend webviews (mainly webxdc) into safe area

Amzd:
Extend webviews into safe area for better UX in webxdc.

| Before | After |
| --- | --- |
|  |  |

webxdc apps can handle safe area themselves, see File Sync webxdc here alligning a UI element to the bottom safe area:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Delta Chat iOS Changelog
 
+## Unrealeased
+- Extend webxdc view into safe area
+
 ## v1.58.6
 2025-06
 


r10s:
how "bad" us that for webxdc that do not care? this will be the majority, webxdc should still be easy to develop. already today, most webxdx are poorly tested on ios and we should not add additional burden.

eg IronCalc is a good example to test, also editor

maybe it is just fine and still workable

r10s (approved):
great! CHANGELOG is missing, otherwise lgtm

thanks a lot for improving here!

PR #2764, opened 2025-07-05
tweak profiles

r10s:
this brings iOS profile to the state of freshly discussed android profile, see https://github.com/deltachat/deltachat-android/pull/3792

see commit messages for details

### DIFF
--- a/deltachat-ios/Controller/AllMediaViewController.swift
+++ b/deltachat-ios/Controller/AllMediaViewController.swift
@@ -12,7 +12,7 @@ class AllMediaViewController: UIPageViewController {
 
     private let dcContext: DcContext
     private let chatId: Int
-    private var prevIndex: Int = -1
+    private var prevIndex: Int = 0
 
     private var pages: [Page] = [
         Page(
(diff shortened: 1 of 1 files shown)


deltachat-ios/Controller/AllMediaViewController.swift
```diff
@@ -64,26 +64,14 @@ class AllMediaViewController: UIPageViewController {
         navigationItem.rightBarButtonItem = UserDefaults.standard.bool(forKey: "location_streaming") ? mapButton : nil
         navigationController?.navigationBar.scrollEdgeAppearance = navigationController?.navigationBar.standardAppearance
 
-        // select the first page that actually contains some media
-        DispatchQueue.global(qos: .userInitiated).async { [weak self] in
+        DispatchQueue.main.async { [weak self] in
```
Amzd, line 67: viewDidLoad is already only called on main thread so the dispatch seems unnecessary here?

deltachat-ios/Controller/ProfileViewController.swift
```diff
@@ -294,8 +294,11 @@ class ProfileViewController: UITableViewController {
             var moreOptions = [UIMenuElement]()
             var primaryOptions = [UIMenuElement]() // max. 3 due to .medium element size
 
+
```
Amzd, line 297: Assuming you didn't mean to add this line?

deltachat-ios/Controller/AllMediaViewController.swift
r10s: can still make a difference wrt responsiveness, or? but probably not needed, i agree

Amzd: Will probably make it worse because it is not deterministic when it will be executed, maybe during an animation which might show as a hiccup

PR #2765, opened 2025-07-09
Share extension refactor

Amzd:
Instead of using the SLCompose sheet we now open the main app to do sharing which prevents issues like running out of memory while compressing/encrypting messages.

This is a big change and will need more testing. 

TODO: 
- [ ] localization (there is TODO comments)
- [ ] check first responder handling in the chat view controller on all iOS versions. The forward flow has a resign and becomeFirstResponder which was not needed in my testing on iOS 16, maybe this is an issue on other versions/platform?
- [ ] audio file support? this is also not supported in drag and drop so maybe a separate PR for this is better
- [ ] wording of changelog? 
- [x] Sharing from screenshot screen doesn't work


Amzd:
moved to #2824

PR #2769, opened 2025-07-11
Adapt to new Broadcast Channel chat types

### LINKED ISSUE #2762: Adapt to new Broadcast Channel chat types
Core release 2.0.0 includes https://github.com/chatmail/core/pull/6901, which includes a new API for broadcasts, which are now called "Channel" in the user-facing UI.

These steps are necessary now (also see https://github.com/deltachat/deltachat-android/pull/3783/):
- [x] Rename "broadcast list" to "channel" in the UI. There are new stock strings in which `broadcast_list` is replaced with `channel`
- [x] Adapt to the new chat types `InBroadcast` and `OutBroadcast` for incoming / outgoing broadcast channels, where the former is similar to a `Mailinglist` and the latter is similar to a `Broadcast` (which was removed).
- [x] Use the new `create_broadcast()` API, which replaced the deprecated `create_broadcast_list()` API
- [x] Make it possible to leave channels
    - In a chat, if the chat is an InBroadcast, and it's not a contact request, then the `Leave` menu option is shown with the translated stock string `menu_leave_channel` as its label.
    - If the user clicks on it, the confirmation dialog has `menu_leave_channel` (rather than `menu_leave_group`) as its positive option.
- [x] Test whether everything works fine using the "UI Testing" list at https://github.com/chatmail/core/issues/6884
- [x] Wait for a core release with https://github.com/chatmail/core/pull/6901

Broadcast lists are experimental, so, it would be fine if you postpone some of these things (e.g. the renaming of strings) and the UX isn't super great. But you should at least make sure that the app doesn't crash when core returns a `{In|Out}Broadcast` chat type.

Note: In the code (except for strings), we use the name "Broadcast" rather than "Channel", because the word "channel" already appears a lot in the code already, which would make it hard to grep for it.

r10s:
closes #2762

counterpart of https://github.com/deltachat/deltachat-android/pull/3783

see links above for reasoning an details, some impressions:

# OUT_BROADCAST

IMG_0116 IMG_0117  IMG_0126 IMG_0120 IMG_0121

# IN_BROADCAST

IMG_0122 IMG_0123 IMG_0124

### DIFF
--- a/deltachat-ios/Controller/ProfileViewController.swift
+++ b/deltachat-ios/Controller/ProfileViewController.swift
@@ -121,7 +121,7 @@ class ProfileViewController: UITableViewController {
         sharedChats = if contactId != 0, !isSavedMessages, !isDeviceChat { dcContext.getChatlist(flags: 0, queryString: nil, queryId: contactId) } else { nil }
 
         sections.append(.options)
-        if isOutBroadcast || isMultiUser {
+        if isMultiUser && !isMailinglist && !isInBroadcast {
             sections.append(.members)
         }
         if let sharedChats, sharedChats.length > 0 {
--- a/deltachat-ios/Controller/ProfileViewController.swift
+++ b/deltachat-ios/Controller/ProfileViewController.swift
@@ -38,12 +38,12 @@ class ProfileViewController: UITableViewController {
     private var contact: DcContact?
     private var memberIds: [Int] = []
     private var sharedChats: DcChatlist?
-    private let isGroup, isMailinglist, isOutBroadcast, isInBroadcast, isSavedMessages, isDeviceChat, isBot: Bool
+    private let isMultiUser, isMailinglist, isOutBroadcast, isInBroadcast, isSavedMessages, isDeviceChat, isBot: Bool
 
     // MARK: - subviews
 
     private lazy var headerCell: ProfileHeader = {
-        let header = ProfileHeader(hasSubtitle: isGroup || isOutBroadcast)
+        let header = ProfileHeader(hasSubtitle: isMultiUser || isOutBroadcast)
         header.onAvatarTap = showEnlargedAvatar
         header.setRecentlySeen(contact?.wasSeenRecently ?? false)
         return header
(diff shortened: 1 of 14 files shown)


deltachat-ios/Controller/ProfileViewController.swift
```diff
@@ -334,24 +336,37 @@ class ProfileViewController: UITableViewController {
...
+                let clearImage = if #available(iOS 16.0, *) { "eraser" } else { "rectangle.portrait" }
                 if isGroup && chat.canSend && chat.isEncrypted {
-                    let image = if #available(iOS 15.0, *) { "rectangle.portrait.and.arrow.right" } else { "arrow.right.square" }
-                    moreOptions.append(action("menu_leave_group", image, attributes: [.destructive], showLeaveGroupConfirmationAlert))
+                    moreOptions.append(action("menu_leave_group", leaveImage, attributes: [.destructive], { [weak self] in
+                        self?.showLeaveAlert("menu_leave_group")
+                    }))
+                    moreOptions.append(action("clear_chat", clearImage, attributes: [.destructive], showClearConfirmationAlert))
+                } else if isInBroadcast {
+                    actions.append(action("menu_leave_channel", leaveImage, attributes: [.destructive], { [weak self] in
```
Hocuri, line 352: If the chat is a contact request, then this action should not be shown, because it will fail, because it's not possible to send the leave message into the contact request chat. (although I'm not sure about this; it would only be ~4 lines of code to allow leaving contact request channels, so, I could just allow it in core)

Hocuri (approved):
I can't test it, and don't know the code of course, but from reading it, looks good!

deltachat-ios/Controller/ProfileViewController.swift
r10s: does one leave the channel implicitly when denying a contact request? 

however, if possible, i think, it make sense to allow leaving channels even if not yet accepted. if it turns out to be too complicated in core, however, i can also easily hide the option

Hocuri: > does one leave the channel implicitly when denying a contact request?

No, messages will just be ignored when you click "Block".

It's the same as we're doing for groups: Until you accept the contact request, you can't send a 'Leave' message, because maybe this comes from a spammer, and sending a 'Leave' message would let the spammer know that this account exists and is read by a human. (although I just noticed that on Desktop, there is a 'Leave' button also for group contact requests, and it just doesn't do anything there).

Meanwhile, we're mostly using chatmail, where this argument carries less weight, because the spam protection works via only allowing encrypted messages (and if a spammer has your addr+key, you will likely need to change profiles, anyways).

> however, if possible, i think, it make sense to allow leaving channels even if not yet accepted. if it turns out to be too complicated in core, however, i can also easily hide the option

I'll create a PR soon, then we can discuss there.

r10s: k, i merge this PR in then, depending on the outcome we can hide or leave things as is

PR #2770, opened 2025-07-12
new email

r10s:
UI implementation of https://github.com/chatmail/core/pull/6932

### DIFF
--- a/deltachat-ios/Controller/NewGroupController.swift
+++ b/deltachat-ios/Controller/NewGroupController.swift
@@ -125,8 +125,7 @@ class NewGroupController: UITableViewController, MediaPickerDelegate {
         guard let groupName = groupNameCell.textField.text else { return }
         let groupChatId: Int
         if createBroadcast {
-            groupChatId = dcContext.createBroadcastList()
-            _ = dcContext.setChatName(chatId: groupChatId, name: groupName)
+            groupChatId = dcContext.createBroadcast(name: groupName)
         } else {
             groupChatId = dcContext.createGroupChat(verified: allMembersVerified(), name: groupName)
         }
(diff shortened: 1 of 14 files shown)


r10s:
> it is not really an email but just creating an email thread

sure, this is what it is internally.

so you suggest to reword to "New E-Mail-Thread" in the UI? not sure if that is more or less helpful :)

adbenitez:
> sure, this is what it is internally.

not internally but visible, you go through that steps as an unsuspecting user and didn't manage to create an email but a thread/group

r10s:
> and they probably know what an email thread is

i doubt, typical email users know. at least ppl around me, in german language, do not know :)

this is the typical thing of ppl answering "Re: Merry Christmas" months later without context :)

but i would not die on that hill, i see the point, and maybe _our_ email users are different :)  let's see if more voices come up, the string anyways needs to be changed in android repo

r10s:
> The option should be hidden for chatmail profiles. Don't see the code checking it now.

it is hidden for chatmail.

see `NewChatViewController.swift`, line 76:

```
        if self.dcContext.isChatmail == false {
            newOptions.append(.newEmail)
        }
```

deltachat-ios/Controller/NewGroupController.swift
```diff
@@ -12,6 +12,7 @@ class NewGroupController: UITableViewController, MediaPickerDelegate {
     private var deleteGroupImage: Bool = false
 
     let createBroadcast: Bool
+    let createEmail: Bool
```
Amzd, line 15: This createEmail and createBroadcast could probably be consolidated into an enum? Or am I misunderstanding and is it possible to have both creatBroadcast: true and createEmail: true?

Amzd (approved):

PR #2785, opened 2025-07-22
Update core to 2.6

r10s:
~~this makes problems, see https://github.com/chatmail/core/issues/7030~~

EDIT: updated PR to core 2.6


r10s:
this makes problems, see https://github.com/chatmail/core/issues/7030

PR #2805, opened 2025-08-08
allow to clone email chats

r10s:
this was disallowed at a time where we had no api
to create ad-hoc "new email" chats

counterpart of https://github.com/deltachat/deltachat-android/pull/3872

came over that via https://github.com/deltachat/deltachat-desktop/issues/5387#issuecomment-3166969826

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Delta Chat iOS Changelog
 
+## Unreleased
+
+- Allow to clone email chats
+
+
 ## v2.10.0
 2025-08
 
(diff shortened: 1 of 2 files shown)


Amzd (commented):
Doesn't the old canSend check also mean you couldn't clone chats that you got removed from?

Amzd:
I mean this PR also changes the behavior that someone kicked from an encrypted group will now be able to make a clone of it while before that was not possible. Just making sure that change is intended/fine.

adbenitez:
> I mean this PR also changes the behavior that someone kicked from an encrypted group will now be able to make a clone of it while before that was not possible. Just making sure that change is intended/fine.

it could also be the other way around, some troll joins your group that you shared publicly (ex. DC community group) and kicked you from the group or destroyed the whole group and with this people can clone and recover reducing the damage made by troll, groups created by blocked contacts are ignored (that area needs some improvements still but it is the way to go)

Amzd (approved):

PR #2847, opened 2025-10-21
Experimental tgs (lottie) support

Amzd:
Questions that still need to be answered:

Do we want to support .tgs? Or just .lottie. Because .tgs requires gzip dependency.

If we want .tgs, can we add a core function for decompression gzip so we don't need to have the dependency on client?


r10s:
similar to vlc kit it is questionable to pull in large c dependencies for less-used areas. maybe okay in this case, however, stickers are no priority atm, so closing this for now. we can always reopen if that comes to focus again

PR #2856, opened 2025-10-29
Some calls improvements

Amzd:
This contains some improvements, but the whole web app will be dropped and remade in native so you can decide if this is worth merging @r10s


r10s (approved):
thanks a lot, even if calls are done natively at the end, it makes sense to get these improvements in

PR #2862, opened 2025-10-30
unify font in fallback avatars

r10s:
this PR unifies the appearance of the "fallback avatars", by always calculating the font size by circle size in init() - no idea why that was ever changed afterwards. 

left: before -  avatar in title and in profile switcher have different font sizes.
right: after - avatar in title and in profile switcher have same font sizes.


Amzd (approved):
I love PRs that only remove code and fix issues

Issue #2865, opened 2025-10-31
cleanup widget orphans, delete widgets when  chat/app is deleted


r10s:
sometimes, chats or apps shown in the widget are left as orphans, showing only as a black item.

tapping the icon opens the app, not the original chat/app (which was deleted)

this should be implemented as a fallback; in the first place, we should check if on chat- or app-deletion, the widgets are removed (if possible)

PR #2868, opened 2025-11-02
don't show invalid items in widget 

Simon-Laux:
solves part of #2865, just the display part, not the cleanup (that's still to do).

The usage of `compactMap`, which removes `nil`, suggests that the check (if message is empty) was just forgotten.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Delta Chat iOS Changelog
 
+## Unreleased
+
+- Fix: Don't show removed chats in widget
+
+
 ## v2.22.2 Testflight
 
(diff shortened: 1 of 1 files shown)


DcWidget/WidgetProvider.swift
```diff
@@ -57,6 +57,8 @@ struct Provider: TimelineProvider {
                 switch entry.type {
                 case .app(let messageId):
                     let msg = dcContext.getMessage(id: messageId)
+                    if !msg.isValid { return nil }
```
Amzd:
```suggestion
                    guard msg.isValid else { return nil }
```

PR #2869, opened 2025-11-03
feat: ignore taps on widget background

Simon-Laux:
Because it is annoying that it opens the app when you miss the chat/app
that you wanted to press.

This uses a large invisible button in the background with a no-op
AppIntent.

This is activated conditionally only on iOS 18+.
And it was tested by me on iOS 26.0.1


Amzd (approved):

r10s:
closing stale PR, the issue is questionable, and potentially also there elsewhere. it is also unwanted by apple that _nothing_ happens when a widget is tapped.

way forward is to improve buttons as such, eg. add titles to them and maybe add a title (as whatsapp) - that makes it more clearer, that also "only the app" may be opened. additional complexity is better spent there. but that is no prio the next weeks.

PR #2889, opened 2025-11-09
mark buttons and external links as such

r10s:
this PR mark buttons that open external links as such; same for offline help

cmp https://support.delta.chat/t/do-not-suddenly-open-links-in-the-browser-without-warning/4327/2

example:

### DIFF
--- a/DcCore/DcCore/Extensions/String+Extensions.swift
+++ b/DcCore/DcCore/Extensions/String+Extensions.swift
@@ -2,8 +2,12 @@ import Foundation
 import UIKit
 
 public extension String {
-    
-	static func localized(_ stringID: String) -> String {
+
+    static func markAsExternal(_ string: String) -> String {
+        return string + " ↗"
+    }
+
+    static func localized(_ stringID: String) -> String {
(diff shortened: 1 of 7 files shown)


DcCore/DcCore/Extensions/String+Extensions.swift
```diff
@@ -2,8 +2,12 @@ import Foundation
 import UIKit
 
 public extension String {
-    
-	static func localized(_ stringID: String) -> String {
+
+    static func markAsExternal(_ string: String) -> String {
+        return string + " ↗"
```
r10s: tested rtl, that looks good as well - it was anyway unclear if the arrow should become first or last, and even more with mixed directions, both seems fine - until we get some different hints

Amzd: isn't a non static func easier to use?
```suggestion
    func markAsExternal() -> String {
        return self + " ↗"
```
then you can use it like:
```diff
- String.markAsExternal(String.localized("example"))
+ String.localized("example").markAsExternal()
```

PR #2920, opened 2025-11-27
Fix sharing a file to delta chat

### LINKED ISSUE #2910: sharing files to Delta Chat is not working from inside Delta Chat
came over that at "settings / advanced / view log" -> share -> delta chat, this results in a file.

after invesigation of @Amzd , this seems to be a general issue with sharing files, however, since https://github.com/deltachat/deltachat-ios/pull/2879

Amzd:
Files that only provide UTType.fileURL provider are now loaded by copying the file at that location.

This adds one more `Capture of 'provider' with non-Sendable type 'NSItemProvider' in a '@Sendable' closure` warning but this warning already exists in this file a couple of times so I think we can pick that up whenever we need to address those warnings.

closes #2910

### DIFF
--- a/DcCore/DcCore/Helper/CodableNSItemProvider.swift
+++ b/DcCore/DcCore/Helper/CodableNSItemProvider.swift
@@ -44,7 +44,7 @@ public enum CodableNSItemProvider: Codable {
             default: continuation.resume(throwing: Error.unknownType)
             }
             func loadFileURL() {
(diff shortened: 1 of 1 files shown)


r10s:
> This doesn't work for files from Files app as they just return their url in the data part.

can you say if that was working before, with the old share extension? 

would still need a fix, but would not be a worsening

Issue #2914, opened 2025-11-23
use new getStorageUsageReportString API in log


r10s:
the log should include the output of the new getStorageUsageReportString api to get storage usage stats

counterpart of https://github.com/deltachat/deltachat-android/issues/4058

PR #2928, opened 2025-12-02
add storage report to log

r10s:
this PR adds the new "storage usage report" to the file created at "view log".

this requires core 2.28.0 or newer

closes https://github.com/deltachat/deltachat-ios/issues/2914

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - stop scanning system address book as it does no longer help on getting in contact
 - Add members to channels by QR codes
 - Channel QR codes can be withdrawn and revived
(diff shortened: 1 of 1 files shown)


deltachat-ios/DC/DcContext.swift
```diff
@@ -810,4 +810,15 @@ public class DcContext {
         }
         return "[]"
     }
+
+    public func getStorageUsageReportString() -> String {
+        do {
+            if let data = try DcAccounts.shared.blockingCall(method: "get_storage_usage_report_string", params: [id as AnyObject]) {
```
Amzd: What does the `as AnyObject` do?

PR #2937, opened 2025-12-05
Chat bubble customization

socksprox:
image

image


link2xt:
For this kind of change better if iOS developers (@r10s) agree first that it can be accepted. I don't think we generally want options that allows users to customize border radius and customize colors one-by-one. If this cannot be accepted anyway, no need to split this into separate PRs.

r10s:
yes, tho timing wise tricky, i can say already now, that we do not offer nor want such levels of configuration details in Delta Chat

PR #2950, opened 2025-12-10
prefer weak pointers in UIAlertAction callbacks

r10s:
this is non-exhaustive, also callbacks easily hold strong references - which, however, are not always evil.

but these were easy to fix and the pattern we want is clearly "weak self", also to avoid creating retain cycles by copy+paste, where the original might have been fine

### DIFF
--- a/deltachat-ios/Controller/Settings/AdvancedViewController.swift
+++ b/deltachat-ios/Controller/Settings/AdvancedViewController.swift
@@ -61,10 +61,10 @@ internal final class AdvancedViewController: UITableViewController {
                     let alert = UIAlertController(title: String.localized("pref_multidevice"),
                         message: String.localized("pref_multidevice_change_warn"),
                         preferredStyle: .alert)
-                    alert.addAction(UIAlertAction(title: String.localized("perm_continue"), style: .destructive, handler: { _ in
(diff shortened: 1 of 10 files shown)


deltachat-ios/Controller/Settings/AdvancedViewController.swift
```diff
@@ -61,10 +61,10 @@ internal final class AdvancedViewController: UITableViewController {
                     let alert = UIAlertController(title: String.localized("pref_multidevice"),
                         message: String.localized("pref_multidevice_change_warn"),
                         preferredStyle: .alert)
-                    alert.addAction(UIAlertAction(title: String.localized("perm_continue"), style: .destructive, handler: { _ in
-                        self.dcContext.setConfigBool("bcc_self", false)
+                    alert.addAction(UIAlertAction(title: String.localized("perm_continue"), style: .destructive, handler: { [weak self] _ in
+                        self?.dcContext.setConfigBool("bcc_self", false)
                     }))
-                    alert.addAction(UIAlertAction(title: String.localized("cancel"), style: .cancel, handler: { _ in
+                    alert.addAction(UIAlertAction(title: String.localized("cancel"), style: .cancel, handler: { [weak self] _ in
```
Amzd: self is not used in this closure

PR #2967, opened 2025-12-21
QuickLook swipe down (pre-iOS 18) and less message motion on every version

### LINKED ISSUE #2918: input-bar-over-msg-bug is back
the input-bar-over-msg-bug with one of the most recent commits.

it is reproducible on iOS 18 eg by enlarging and leaving an image:

https://github.com/user-attachments/assets/a6b1ac1c-f711-4048-94ea-810052279d3d

Amzd:
This is a continuation of #2482, fixing the issue that were discussed in the original PR.

This also fixes #2918


r10s:
closing stale PR, the linked issue is even fixed by other means

Amzd:
I will still want to get some part of this in because QLPreviewController is still buggy on iOS 26 when pushed instead of presented. 

Main issues are share button moving after presentation and swipe down being super laggy and sometimes not even working on the first try.

PR #2978, opened 2026-01-08
truncate file names in the middle

r10s:
that way, the end is always visible,
which contains important information more often than the middle.

this is common knowledge since the 90's or so
however, as quite some previously stable ux patterns, lost on the way. (cmp. Scott Jensen, https://www.youtube.com/watch?v=1fZTOjd_bOQ )

i also checked the file names in the bubbles,
here they're wrapped, so that's fine
(wrapping in the list, otoh, would clutter the list, complicate ux and layout)

sidenote:  i'll aim to do the same for android, this issue might have been part of some ticket confusion, cc @hpk42

before / after:

 &nbsp; &nbsp; 

macOS for reference:

Screenshot 2026-01-08 at 15 14 34

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,10 @@
 
 ## Unreleased
 
+- Truncate file names in the middle, not at the end; important information are more often at the end
 - Fix: Online indicator in the chat navigation bar now updates correctly when returning from background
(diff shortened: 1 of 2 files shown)


Amzd (approved):
Great change

PR #3015, opened 2026-02-19
more strict callkit checks

r10s:
if callkit shall not be used, do not instantiate any of the `CX*` classes.

~~moreover, this PR adds a missing check on incoming call, to not use `CXProvider.reportNewIncomingVoIPPushPayload`~~ EDIT: that was actually okay

#skip-changelog as this is no user-visible change

### DIFF
--- a/deltachat-ios/Calls/CallManager.swift
+++ b/deltachat-ios/Calls/CallManager.swift
@@ -33,25 +33,32 @@ class CallManager: NSObject {
     static let shared = CallManager()
 
     private let voIPPushManager: VoIPPushManager
-    private let provider: CXProvider
-    private let callController: CXCallController
-    private let callObserver: CXCallObserver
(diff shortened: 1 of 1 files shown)


DcNotificationService/NotificationService.swift
```diff
@@ -101,7 +101,7 @@ class NotificationService: UNNotificationServiceExtension {
                         let request = UNNotificationRequest(identifier: "incoming-call", content: content, trigger: nil)
                         try? await UNUserNotificationCenter.current().add(request)
                     }
-                } else if #available(iOSApplicationExtension 14.5, *) {
+                } else if #available(iOSApplicationExtension 14.5, *), canUseCallKit {
```
Amzd: this is implied because it is the "else" of !canUseCallkit

Amzd (approved):

PR #3017, opened 2026-02-20
feat: Allow to switch profile when sharing or forwarding

### LINKED ISSUE #2912: in "Share to Delta", allow selecting target profile
currently, when sharing information to Delta Chat, only changes from the last selected profile are selectable.

this issue is about adding a profile switcher to the "Share To" dialog.

it probably makes sense to use already known same layout and ux elements.

tapping the avatar allows to switch ones account (when using the same account switcher, creating/deleting/editing a new account can be disabled, these things do not really make sense at this point)

when there are no multiple accounts, the account switcher can be hidden.

ftr, the same layout can be used for forwarding messages, however, that requires [some work in core first](https://github.com/chatmail/core/issues/7491)

counterpart of https://github.com/deltachat/deltachat-android/issues/4055

r10s:
this PR removes `RelayHelper`s reference to the selected context, making it possible to forward/share messages to other profiles.

to make this work also in the UI, the  profile switcher is shown also on forwarding/sharing, where it was hidden kind of artificial before.

so, this is quite a usability improvement - with lines _removed_ :)

closes #2912

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Allow to switch profile when sharing or forwarding
 - Add tab bar icon bounce animation on tap
 
 


Amzd (approved):
Nice!

Issue #2973, opened 2025-12-27
iOS 26 liquid glass


Amzd:
- [x] Chat view controller tableview might need to be unflipped to fix the shadow that gets cast over the entire tableview in iOS 26 #2839
- [x] Input bar needs to be redone for iOS 26
- [x] Remove the UIDesignRequiresCompatibility toggle
- [x] New chat icon is blue instead of black
- [ ] "CHATS" is shown in the back button if there is enough space, just a "<" would be fine if we can change this
- [x] Switch profile button has some spacing between the avatar and the background shadow
- [ ] Switch profile detail view avatars are spaced a bit too far to the right

PR #3019, opened 2026-02-21
Implemented chat message lifecycle animations

mikovskii:
Added animation for new messages appearing in chat.
Added animation for message deletions.
Kept incoming behavior lightweight (fade-based) and avoided excessive motion.
Added guarded diff-based row updates with fallback conditions to prevent unstable animations (editing/search/context-menu/reduce-motion/move-heavy updates).
Ensured refresh correctness during rapid/coalesced updates by updating visible rows and preserving seen/scroll button state after deferred refreshes.

Demo:

https://github.com/user-attachments/assets/2fc0e6f9-8f19-4588-8f76-6eddd3a10c6e

https://github.com/user-attachments/assets/f28cdedf-3567-4083-9fd0-d27d4f912e1d

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Allow to switch profile when sharing or forwarding
 - Add tab bar icon bounce animation on tap
+- Add animated insert/delete transitions for chat messages
 
 
 ## v2.43.0 Testflight
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -13,9 +13,14 @@ class ChatViewController: UIViewController, UITableViewDelegate, UITableViewData
 
     private var dcContext: DcContext
     private var messages: [(id: Int, timestamp: TimeInterval)] = []
+    private var messageStableKeys: [String] = []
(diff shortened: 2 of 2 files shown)


r10s:
very cool, thanks a lot!

timing wise, however, we should solve #2918 and #2973 before, to have a more solid ground for this and other possible animations, and also to not make fixing the bugs potentially harder

maybe you can help on that as well? i think this will start these days, cc @Amzd

mslvx:
I'm happy to help, but I'm not that familiar with the app's codebase yet :)  I only started looking into the code base this week. 
I can slowly make PRs with new iOS 26 UI elements.

r10s:
sure, totally understandable. if you like, we can do a little call next week, ans i can tell a bit about background, prios, what is planned next

just now, i am afk, weekend :) have a nice time as well

r10s:
just now, we should not add complexity for such "nice to have" things. once we are really stable in the bubble view, we can reconsider. there are still some open ends after the switch to liquid glass, where 150+ lines of code and their complexity do not help.

there is also still the idea to stop the internal mirroring of the bubble view, might even be, there are means by the system helping on such animations when we follow standards.

closing this PR for now

Issue #2867, opened 2025-11-01
quotes without names are displayed unstable


r10s:
when having a quote without a name, this is displayed as an empty line - where it should compact the display.

only after some scrolling, that is finally compacted.

<img width=320 src=https://github.com/user-attachments/assets/b5c31e4c-cb36-4972-9646-011f1bd153d7>

r10s:
it is hard to  fix without introducing other, subtile layout bugs.

the issue that all of sender, text, image can be optional makes constraints hard.

i think, the following stack view approach in `QuoteView` would be much more stable:

- have a horizontal stack view with `citeBar`, `texts` and `imagePreview`

- the horizontal stack view must not expand to full width, it is better if the image is beside the quote

- `texts` is a vertical stack view with `senderTitle` and `quote`

- that way, things can be hidden and others just "fall" in the gaps

it is all a minor, i am no longer sure if it is worth fixing. maybe we should also just close the issue ...

PR #3031, opened 2026-02-23
fix: clear quote constraints

r10s:
remove imagePreview top clamp
that caused unstable layout for short/nameless quotes.

the clamp was introduced at
https://github.com/deltachat/deltachat-ios/commit/4012fb4417b51f9ab545e8b67bec2c7ac76a783f but seems not to be needed,
even if the quote is only one line without title
(eg. if the original is removed)

closes https://github.com/deltachat/deltachat-ios/issues/2867

### DIFF
--- a/DcCore/DcCore/Extensions/UIView+Extensions.swift
+++ b/DcCore/DcCore/Extensions/UIView+Extensions.swift
@@ -95,23 +95,6 @@ public extension UIView {
         return constraint
     }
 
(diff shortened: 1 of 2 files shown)


r10s:
closing, as this results in partly too small quotes, eg. when drafting the image may overlap the draft area.

maybe a stackview is the easiest option, issue is that all of title, text, image are optional

PR #3040, opened 2026-03-04
Improve QR code visibility

### LINKED ISSUE #2989: better scannable "Add Second Device" QR code
there are several reports, that the QR code of the "Add Second Device" screen is hard hard to scan, esp. by some poor scanners.

the situation got worse by some iroh update, adding more data to the QR code (eg. add-second-device Qr code is more dense than most invite-links)

this issue is about **improving the showing side**, esp. when when the QR code is shown on a small, maybe dirty and dark screen

the currently used `dc_backup_provider_get_qr_svg` is anyway deprecated, so we should use `dc_create_qr_svg(dc_backup_provider_get_qr())`, which will shown more QR code when using the same svg width - also contrast to background is maximized. the borders are removed, but they're anyway no longer state-of-the-art nor good looking

the subtitle can probably be removed alltogether, it does not add much, context is pretty clear

counterpart of https://github.com/deltachat/deltachat-android/issues/4052

request in the forum: https://support.delta.chat/t/enlarge-add-second-device-qr-code/3867

mikovskii:
QR codes are now easier to scan because the app temporarily turns the screen brightness up while they are shown, then restores the previous brightness when you leave the screen or the app goes to the background.

This also updates backup QR rendering ( #2989 )

## demo ( for new qr backup): 

<details>
<summary>Details</summary>

Screenshot 2026-03-04 at 8 22 46 PM

</details>

close: #2989

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Improve QR code visibility and use the new backup QR generation path
 - Allow to switch profile when sharing or forwarding
 - Add tab bar icon bounce animation on tap
 - Fix: remove meaningless address from vCard
--- a/deltachat-ios/Controller/BackupTransferViewController.swift
+++ b/deltachat-ios/Controller/BackupTransferViewController.swift
@@ -3,7 +3,7 @@ import UIKit
 import DcCore
 import SDWebImageSVGKitPlugin
 
-class BackupTransferViewController: UIViewController {
+class BackupTransferViewController: UIViewController, ScreenBrightnessOverrideSupporting {
 
     public enum TranferState {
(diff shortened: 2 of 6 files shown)


Amzd:
I am a bit hesitant to merge this right now because I feel like it would be much nicer if we could use EDR just on the QR view, instead of needing to keep track of viewWillAppear/viewWillDisappear. I haven't tested if that even produces a similar brightness but seems like it does: https://github.com/tsuzukihashi/EDR_Swift

mikovskii:
I'll test it out and try to implement it using an ERD. I looked at how it's implemented in Signal and Telegram. They use an approach similar to mine. But I'll still give the ERD a try.

PR #3052, opened 2026-03-13
Increase reply swipe sensitivity

### LINKED ISSUE #2864: Multi select shows up while showing leading swipe action (reply button) in chat view controller
Apparently `tableView.isEditing` returns true while showing swipe actions so we need a different way to check if we are in multi selection mode.

This affects the navigation bar changing to the selection mode one and when you send a message it does not show up (because sending/receiving messages is intentionally paused during selection).

  - Swipe message cell to show the reply button
  - Send or receive a message
  - Now the navigation bar is the selection one and the sent/received message does not show up

Amzd:
This PR makes it:
- Easier to swipe reply as you no longer need to swipe the full width of the screen
- Impossible to end up in the editing state where a bubble is swiped to the right with the reply action showing.

image

This fixes #2864 and general UX complaints that I don't think we had an issue for.

| Old (half swipe) | Old (full swipe) | New (both) |
| --- | --- | --- |
| <video src="https://github.com/user-attachments/assets/f0e6a1c1-6ca1-428f-a3f0-57083f258db8"> | <video src="https://github.com/user-attachments/assets/435ed51f-4641-4503-b534-098d23b3a43e"> | <video src="https://github.com/user-attachments/assets/9050e7fa-3081-4f28-8df3-2c984c7d581e"> |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Increase reply swipe sensitivity
 - Allow to switch profile when sharing or forwarding
 - Add tab bar icon bounce animation on tap
 - Explain at "Settings / Chats / Outgoing Media Quality" how to send original quality
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -144,7 +144,7 @@ class ChatViewController: UIViewController, UITableViewDelegate, UITableViewData
     /// it if the cell was considered "swiped" by the system (aka when the leading actions stay open).
     /// This means the required swipe distance to activate the reply is just the distance to open the leading swipe action
     /// instead of the much further "performsFirstActionWithFullSwipe".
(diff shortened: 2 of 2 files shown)


r10s (approved):
k, it seems unrelated to liquid glass, i just wanted to raise that concern!

so, lgtm, also tested a bit.

what's missing now is the little "vibration" to indicate that releasing the finger will replay. we had that before, and this is also what whatsapp/signal/telegram are doing. would be cool to get that back. if complex, then please file an issue and do another PR, i am already approving this one :)

Amzd:
Sure, we can do that but it means that instead of relying on the systems definition of what enough swiping is (by checking configurationState.isSwiped after swipe has been released) we then need to use our own distance of how far a swipe needs to be to be considered far enough (because we can't check what the system will do before the swipe ends so we wouldn't be able to send haptic feedback during the swipe). 

Am I understanding correctly you want both:

- A selection change haptic feedback during swipe
- A success haptic feedback on swipe release if reply is triggered

r10s (approved):

PR #3055, opened 2026-03-15
feat: mark messages as "unread"

r10s:
this PR adds an "mark unread" option at the place where otherwise "mark read" is shown on swiping.

this is kind of standard of standard on most messengers on iOS and one of the most often requested missing features in Delta Chat.

https://github.com/user-attachments/assets/a1d23c84-2d8a-430b-ac7a-2ed01fe291bb

~~the PR requires a core with https://github.com/chatmail/core/pull/7990 merged.~~ EDIT: core 2.46 on main comes with the required api

### DIFF
--- a/deltachat-ios/Controller/ChatListViewController.swift
+++ b/deltachat-ios/Controller/ChatListViewController.swift
@@ -388,7 +388,7 @@ class ChatListViewController: UITableViewController {
 
     @objc func markReadPressed() {
         if isEditing {
-            viewModel?.markUnreadSelectedChats(in: tableView.indexPathsForSelectedRows)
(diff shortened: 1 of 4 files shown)


deltachat-ios/Controller/ChatListViewController.swift
```diff
@@ -388,7 +388,7 @@ class ChatListViewController: UITableViewController {
 
     @objc func markReadPressed() {
         if isEditing {
-            viewModel?.markUnreadSelectedChats(in: tableView.indexPathsForSelectedRows)
+            viewModel?.markReadSelectedChats(in: tableView.indexPathsForSelectedRows)
```
r10s, line 391: this was just a typo in the old code

Amzd (approved):

PR #3059, opened 2026-03-20
hide relays from contacts (unpublish relays)

r10s:
this PR adds the option "hide from contacts" to the "remove relay" alert and refine strings and wordings, saying "used for sending" instead of "default" and explaining what the other relays are for.

counterpart of https://github.com/deltachat/deltachat-android/pull/4289 and https://github.com/deltachat/deltachat-desktop/pull/6137

### DIFF
--- a/deltachat-ios/DC/DcContext.swift
+++ b/deltachat-ios/DC/DcContext.swift
@@ -595,6 +595,17 @@ public class DcContext {
         return []
     }
 
+    public func listTransportsEx() -> [DcTransportListEntry] {
+        do {
+            if let data = try DcAccounts.shared.blockingCall(method: "list_transports_ex", params: [id as AnyObject]) {
+                return try JSONDecoder().decode(DcTransportListEntryResult.self, from: data).result
+            }
+        } catch {
+            logger.error(error.localizedDescription)
+        }
+        return []
+    }
+
     public func addOrUpdateTransport(param: DcEnteredLoginParam) throws -> Bool {
         let res = try DcAccounts.shared.blockingCall(method: "add_or_update_transport", accountId: id, codable: param)
         return res != nil
@@ -603,6 +614,10 @@ public class DcContext {
     public func addTransportFromQr(qrCode: String) throws -> Bool {
         let res = try DcAccounts.shared.blockingCall(method: "add_transport_from_qr", params: [id as AnyObject, qrCode as AnyObject])
         return res != nil
+    }
+
+    public func setTransportUnpublished(addr: String, unpublished: Bool) throws {
+        try DcAccounts.shared.blockingCall(method: "set_transport_unpublished", params: [id as AnyObject, addr as AnyObject, unpublished as AnyObject])
     }
 
     public func deleteTransport(addr: String) throws {
--- a/deltachat-ios/DC/DcContext.swift
+++ b/deltachat-ios/DC/DcContext.swift
@@ -584,17 +584,6 @@ public class DcContext {
         dc_send_text_msg(contextPointer, UInt32(id), message)
     }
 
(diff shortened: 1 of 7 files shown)


deltachat-ios/DC/DcContext.swift
```diff
@@ -605,6 +605,10 @@ public class DcContext {
         return res != nil
     }
 
+    public func setTransportUnpublished(addr: String, unpublished: Bool) throws {
+        try DcAccounts.shared.blockingCall(method: "set_transport_unpublished", params: [id as AnyObject, addr as AnyObject, unpublished as AnyObject])
```
Amzd, line 609: Casting all of these value types to AnyObject still makes zero sense to me because value types are not reference types so why does the cast even work and what does it do?.. I know you are just copying from code around it but all of them make no sense. 

AnyObject is usually used to make sure something is a reference type:
```swift
protocol Example: AnyObject {}
extension String: Example {}
- error: non-class type 'String' cannot conform to class protocol 'Example'
```

I think we should change the blockingCall method to accept [Any] instead of [AnyObject] so we dont need to do all these nonsense casts.
https://github.com/deltachat/deltachat-ios/blob/49b834afa2034d3113b11e7bf6b99052b964cb9a/deltachat-ios/DC/DcAccount.swift#L205

Even better would be to delete the AnyObject and use Codable everywhere to make encoding type safe.

/rant

deltachat-ios/DC/DcTransportListEntry.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+public struct DcTransportListEntry: Decodable {
+    public let isUnpublished: Bool
+    public let param: DcEnteredLoginParam
+}
+
+struct DcTransportListEntryResult: Decodable {
+    let result: [DcTransportListEntry]
+}
```
Amzd:
```suggestion
typealias DcTransportListEntryResult = JsonrpcResult<[DcTransportListEntry]>
```

Amzd (approved):
I can't really tell from the UI what the "Hide from contacts" is really supposed to do, is it like "Tell contacts to forget this relay"? or "Don't tell contacts about this relay ever"? 

Code looks fine other than some nitpicking which you can ignore

Hocuri:
> I can't really tell from the UI what the "Hide from contacts" is really supposed to do, is it like "Tell contacts to forget this relay"? or "Don't tell contacts about this relay ever"?

It is both. In every message, you tell your contacts about your relays. "Hide from contacts" means "don't include this relay when telling your contacts how to reach you; they won't send to this relay anymore then".

Hocuri:
In the first screenshot above, I'm wondering whether it is possible to make the question text bigger and less gray? Right now, it looks like you don't really need to read it, but actually, it's important to understand the options.

Apart from that, LGTM!

r10s:
> In the first screenshot above, I'm wondering whether it is possible to make the question text bigger and less gray? Right now, it looks like you don't really need to read it, but actually, it's important to understand the options.

it is the default alert on ios, it looks always like that, and we go with the system wherever possibe (will be different with liquid glass btw). here it is quite okay - it feels different if you have an iphone in your hand, as user attention is exactly on that that area.

r10s:
for "hide from contacts": in normal ui flow, things should be kuch clearer, as user was just about to delete a relay

deltachat-ios/DC/DcTransportListEntry.swift
r10s: cool!

r10s:
@Hocuri turns out we can have the text bold at least, merging the following directly in: https://github.com/deltachat/deltachat-ios/pull/3061

PR #3064, opened 2026-03-24
fix: update in-chat unread counter on messagesChanged

### LINKED ISSUE #3063: in-chat-unread-counter is not always updated
the in-chat unread counter (so the counter on the "back" button) is not updated when chat is marked as unread.

see the video below: when entering "Saved Messages", the counter stays at "1" even tho another chat is marked unread and it should be "2".

the red counter in the tabbar as well as the red counter on the homescreen are correct. once the app is eg. restarted or a chat is marked as read, the in-chat-unread-counter is okay again.

https://github.com/user-attachments/assets/0b16c904-3746-4264-918f-0297becff37d

mikovskii:
https://github.com/user-attachments/assets/d0040c56-087b-4460-8621-b4a3d4404792

close: #3063


Amzd:
The original issue also mentions the app badge counter (on the homescreen) this does not address that right?

mikovskii:
I'm not quite sure what the error on the home screen badge? In the original issue, only the bug related to “in chat count” is mentioned.

Amzd:
Ah you are right, I misready; it mentions that that the app badge works, but I read it as that also doesn't work. Anyway, solution looks simple enough. LGTM.

Amzd (approved):

r10s:
merged at https://github.com/deltachat/deltachat-ios/pull/3075

PR #3072, opened 2026-04-01
fix circular reference between ChatListViewController and ChatListViewModel

Amzd:
This fixes ChatListViewController leaking every time you switch profiles which is can cause lag on slower phones even with less than 10 of them.


r10s (approved):
great catch!

PR #3074, opened 2026-04-08
Implement missed call notification based on core event

Amzd:
This requires https://github.com/chatmail/core/pull/7955


Amzd:
This is no longer necessary as we just notify for any call ended event now, see #3264

PR #3078, opened 2026-04-16
feat: clear all storage

r10s:
add a "Clear All Storage" button to the three dot menu.

it is on purpose, that the option is not directly in the main UI; if ppl are in trouble because of storage, they will find it. maybe we also want to add it as actionable item to the storage warnings, cmp https://github.com/chatmail/core/issues/7730

counterpart of https://github.com/deltachat/deltachat-desktop/pull/6207

requires core with https://github.com/chatmail/core/pull/8105

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Delta Chat iOS Changelog
 
+## Unreleased
+
+- Add option to clear all storage to connectivity view
+
+
 ## v2.49.0 Testflight
 2026-04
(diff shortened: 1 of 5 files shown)


Amzd (approved):
I initially thought this was for local storage, does it make sense to say "Clear Relay Storage"?

Code looks good!

r10s:
with https://github.com/chatmail/relay/pull/929 merged, the main reason for that function is removed

PR #3080, opened 2026-04-20
Implement App Clip to forward invite links to main app

Amzd:
The problem:

- Send someone your invite link or let them scan your QR
- They go to the website (since they don't have delta chat yet)
- They click "Download Delta Chat"
- They click "Download on the App Store"
- They click "Get"
- Open the downloaded app
- Create an account by filling in a name
- ❌ Now they need to scan your QR code again because the invite link was not preserved through this flow

New flow:

- Send someone your invite link or let them scan your QR
- They go to the website and get prompted (or get prompted right away) to use the App Clip
- The App Clip stores the invite link
- They click "Get Delta Chat" in the app clip
- They click "Get"
- Open the downloaded app
- Create an account by filling in a name
- ✅ Get alert to start chat with whomever sent them the invite link. The invite link gets preserved through the download flow.

The App Clip looks like: 

IMG_0011
IMG_0012

This should not happen but in case the app clip stays installed when the user installs the full app, the app clip redirects to the main app right away, before quickly showing that the user may delete this app clip:

IMG_0013


Amzd:
This is tested with invocation from Xcode but invocation from url requires this to be in the appstore for some reason. 

"Testing is done in production" - Apple apparently

r10s (approved):
thanks a lot for figuring that out!

let's give it a try! if smth fails, it fails out there :)

PR #3082, opened 2026-04-20
feat: configurable app picker url

r10s:
this PR makes the app store url configurable, same as on android.

for testing, an alternative store url is `https://webxdc.github.io/website/apps/` (the default is `https://webxdc.org/apps/`)

first, i forgot to adapt the decisionHandler(), while checking for possible reasons that the page was not loaded, i realised that other webview were actually loaded in viewWillAppear, not in init. i adapted the loading accordingly, this seems more correct.

### DIFF
--- a/DcCore/DcCore/Extensions/UserDefaults+Extensions.swift
+++ b/DcCore/DcCore/Extensions/UserDefaults+Extensions.swift
@@ -61,7 +61,11 @@ public extension UserDefaults {
         UserDefaults.standard.string(forKey: appPickerUrl) ?? defaultAppPickerUrlString
     }
 
-    static func setAppPickerUrlString(_ url: String) {
-        UserDefaults.standard.set(url, forKey: appPickerUrl)
+    static func setAppPickerUrlString(_ url: String?) {
+        if let url {
+            UserDefaults.standard.set(url, forKey: appPickerUrl)
+        } else {
+            UserDefaults.standard.removeObject(forKey: appPickerUrl)
+        }
     }
 }
--- a/DcCore/DcCore/Extensions/UserDefaults+Extensions.swift
+++ b/DcCore/DcCore/Extensions/UserDefaults+Extensions.swift
@@ -58,11 +58,7 @@ public extension UserDefaults {
     }
 
     static func getAppPickerUrlString() -> String {
-        if let url = UserDefaults.standard.string(forKey: appPickerUrl) {
-            return url
-        } else {
-            return defaultAppPickerUrlString
-        }
+        UserDefaults.standard.string(forKey: appPickerUrl) ?? defaultAppPickerUrlString
     }
 
     static func setAppPickerUrlString(_ url: String) {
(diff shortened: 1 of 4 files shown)


DcCore/DcCore/Extensions/UserDefaults+Extensions.swift
```diff
@@ -54,4 +56,16 @@ public extension UserDefaults {
         slidingValues.append(DateUtils.getExtendedAbsTimeSpanString(timeStamp: Date().timeIntervalSince1970) + "|" + value)
         shared.set(slidingValues, forKey: debugArrayKey)
     }
+
+    static func getAppPickerUrlString() -> String {
+        if let url = UserDefaults.standard.string(forKey: appPickerUrl) {
+            return url
+        } else {
+            return defaultAppPickerUrlString
+        }
```
Amzd:
```suggestion
        UserDefaults.standard.string(forKey: appPickerUrl) ?? defaultAppPickerUrlString
```

deltachat-ios/Chat/Apps/WebxdcStoreViewController.swift
```diff
@@ -36,6 +37,14 @@ class WebxdcStoreViewController: UIViewController {
 
         NSLayoutConstraint.activate(constraints)
     }
+
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        DispatchQueue.main.async { [weak self] in
```
Amzd: UIKit functions like viewWillAppear are always called on main thread so no switching needed but we also discussed moveing this back to the init, where we could add an `assert(Thread.isMainThread)` if webview.load needs to be called from main(? not sure it does)

deltachat-ios/Chat/Apps/WebxdcStoreViewController.swift
```diff
@@ -62,7 +71,7 @@ extension WebxdcStoreViewController: WKNavigationDelegate {
                 delegate?.pickedAnDownloadedApp(self, fileURL: fileURL as URL)
                 decisionHandler(.cancel)
             }
-        } else if url.host == "webxdc.org" {
+        } else if url.host == appPickerUrl?.host {
```
Amzd, line 70: We discussed removing this as a store might 301 to a new domain which would probably be fine to allow?

Amzd (changes requested):

deltachat-ios/Chat/Apps/WebxdcStoreViewController.swift
r10s: android does the same, so a different host opens in external browser, so not so sure to remove that check

deltachat-ios/Chat/Apps/WebxdcStoreViewController.swift
r10s: moved back to init()

Amzd (approved):

PR #3088, opened 2026-04-24
Disable interactiveContentPopGestureRecognizer on webxdc vc

Amzd:
The interactivePopGestureRecognizer was already disabled, also do this for interactiveContentPopGestureRecognizer


r10s (approved):
in my understanding, this is mainly about the "swipe to go back" gesture?

Amzd:
Yes there is a new gesture introduced in iOS 26 that we need to disable separately from the old gesture. Back gestures are disabled in webxdc so games don't accidentally close when you are swiping.

PR #3090, opened 2026-04-25
feat: play voice/audio messages in background

mikovskii:
Moves chat audio/voice playback from chat-screen lifetime to a global background-capable playback controller. The active player survives chat view disappearance and profile switches, while still updating visible cells if the user returns to the playing message.

 Playback now integrates with `MPNowPlayingInfoCenter` and `MPRemoteCommandCenter` for system controls:
     * toggle play/pause
     * seek

Playback is stopped for audio-conflicting flows:
     * voice recording start / continue
     * outgoing call start
     * incoming call report
     * account deletion for the playing account
     * eadphone unplug pauses playback instead of resuming, to avoid blasting audio through the speaker.

demo (with sound):

<details>
<summary>Details</summary>

https://github.com/user-attachments/assets/998da0b7-304c-43cd-b2de-2eba202e6d84

</details>

part of: #2987

### DIFF
--- a/deltachat-ios/Chat/AudioController.swift
+++ b/deltachat-ios/Chat/AudioController.swift
@@ -1,5 +1,6 @@
 import UIKit
 import AVFoundation
+import MediaPlayer
 import DcCore
 
 /// The `PlayerState` indicates the current audio controller state
@@ -23,11 +24,23 @@ public protocol AudioControllerDelegate: AnyObject {
 /// and also creates and manage an `AVAudioPlayer` states, play, pause and stop.
 open class AudioController: NSObject, AVAudioPlayerDelegate, AudioMessageCellDelegate {
 
+    private enum AudioPlaybackError: Error {
+        case missingFileURL
+        case playFailed
+    }
+
+    private static var backgroundPlaybackController: AudioController?
+    private static var remoteCommandsConfigured = false
+
     open weak var delegate: AudioControllerDelegate?
 
     lazy var audioSession: AVAudioSession = {
         let audioSession = AVAudioSession.sharedInstance()
-        _ = try? audioSession.setCategory(AVAudioSession.Category.playback, options: [.defaultToSpeaker])
+        do {
+            try audioSession.setCategory(AVAudioSession.Category.playback, options: [.defaultToSpeaker])
+        } catch {
+            logger.warning("setting audio session category failed: \(error.localizedDescription)")
+        }
         return audioSession
     }()
 
@@ -50,6 +63,9 @@ open class AudioController: NSObject, AVAudioPlayerDelegate, AudioMessageCellDel
     /// The `Timer` that update playing progress
     internal var progressTimer: Timer?
 
+    private var lastNowPlayingInfoUpdate = Date.distantPast
+    private var playingArtwork: MPMediaItemArtwork?
+
     // MARK: - Init Methods
 
     public init(dcContext: DcContext, chatId: Int, delegate: AudioControllerDelegate? = nil) {
@@ -62,6 +78,10 @@ open class AudioController: NSObject, AVAudioPlayerDelegate, AudioMessageCellDel
                                                selector: #selector(audioRouteChanged),
                                                name: AVAudioSession.routeChangeNotification,
                                                object: AVAudioSession.sharedInstance())
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(appWillTerminate),
+                                               name: UIApplication.willTerminateNotification,
+                                               object: nil)
     }
 
     deinit {
@@ -69,6 +89,19 @@ open class AudioController: NSObject, AVAudioPlayerDelegate, AudioMessageCellDel
     }
 
     // MARK: - Methods
+
+    static func stopBackgroundPlayback() {
+        performOnMainAndWait {
+            backgroundPlaybackController?.stopAnyOngoingPlaying()
+        }
+    }
+
+    static func stopBackgroundPlayback(forContextId contextId: Int) {
+        performOnMainAndWait {
+            guard let controller = backgroundPlaybackController, controller.dcContext.id == contextId else { return }
+            controller.stopAnyOngoingPlaying()
+        }
+    }
 
     /// - Parameters:
     ///   - cell: The `NewAudioMessageCell` that needs to be configure.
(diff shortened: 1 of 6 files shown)


deltachat-ios/Chat/AudioController.swift
```diff
@@ -122,24 +157,36 @@ open class AudioController: NSObject, AVAudioPlayerDelegate, AudioMessageCellDel
...
+            if state == .playing {
+                pauseSound(in: cell)
             } else {
-                // tap occur in a difference cell that the one is currently playing sound. First stop currently playing and start the sound for given message
-                stopAnyOngoingPlaying()
-                playSound(for: message, in: cell)
+                resumeSound()
             }
+        } else {
+            // tap occur in a difference cell that the one is currently playing sound. First stop currently playing and start the sound for given message
```
Amzd:
```suggestion
            // tap occurred in a different cell than the one that is currently playing sound. First stop currently playing and start the sound for given message
```

deltachat-ios/Chat/AudioController.swift
```diff
@@ -122,24 +157,36 @@ open class AudioController: NSObject, AVAudioPlayerDelegate, AudioMessageCellDel
...
+            }
+            AudioController.stopBackgroundPlayback()
+        }
+        guard state != .stopped else {
+            // There is no audio sound playing - prepare to start playing for given audio message
+            playSound(for: message, in: cell)
+            return
+        }
+        if isPlayingMessage(messageId: message.id, contextId: dcContext.id) {
+            // tap occur in the current cell that is playing audio sound
```
Amzd:
```suggestion
            // tap occurred in the cell that is currently playing audio
```

deltachat-ios/Chat/AudioController.swift
```diff
@@ -171,63 +231,184 @@ open class AudioController: NSObject, AVAudioPlayerDelegate, AudioMessageCellDel
     /// - Parameters:
     ///   - message: The `MessageType` that contain the audio item to be pause.
     ///   - audioCell: The `AudioMessageCell` that needs to be updated by the pause action.
-    open func pauseSound(in audioCell: AudioMessageCell) {
```
Amzd, line 174: Please update the doc comment so it describes what the nil case does.

deltachat-ios/Chat/AudioController.swift
```diff
@@ -171,63 +231,184 @@ open class AudioController: NSObject, AVAudioPlayerDelegate, AudioMessageCellDel
...
-        audioPlayer?.pause()
+    open func pauseSound(in audioCell: AudioMessageCell? = nil) {
+        guard let player = audioPlayer else { return }
+        player.pause()
         state = .pause
-        audioCell.audioPlayerView.showPlayLayout(false) // show play button on audio cell
+        (audioCell ?? playingCell)?.audioPlayerView.showPlayLayout(false) // show play button on audio cell
```
Amzd: I don't like the `(x ?? y)?.` syntax
```suggestion
        let cell = audioCell ?? playingCell
        cell?.audioPlayerView.showPlayLayout(false) // show play button on audio cell
```

deltachat-ios/Chat/AudioController.swift
```diff
@@ -171,63 +231,184 @@ open class AudioController: NSObject, AVAudioPlayerDelegate, AudioMessageCellDel
...
+        return .success
+    }
+
+    private static func performOnMainAndWait(_ action: () -> Void) {
+        if Thread.isMainThread {
+            action()
+        } else {
+            DispatchQueue.main.sync(execute: action)
+        }
+    }
```
Amzd: I don't understand why you need to do this synchronously. Either change the else to DispatchQueue.main.async or explain in comments why this needs to be sync.

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -2800,6 +2801,19 @@ extension ChatViewController: AudioControllerDelegate {
     }
 }
 
+// MARK: - Audio Playback
+private extension ChatViewController {
+    func stopMessageAudioBeforeSoundPreview(for viewType: Int32?) {
```
Amzd, line 2710: Could you document this function? Its intention is a bit vague to me. From context I am gathering it should be called before you show any UI with play buttons?

Amzd (changes requested):

deltachat-ios/Chat/AudioController.swift
mikovskii: This is intentionally synchronous because callers may immediately start another audio playback session after `stopBackgroundPlayback()` returns. We need to make sure
  the previous controller has fully stopped and cleared `backgroundPlaybackController` before that happens. I added a comment explaining this.

deltachat-ios/Chat/ChatViewController.swift
mikovskii: Yes, the intention is narrower than "before showing UI with play buttons". It should be called before opening a media preview that may play audio itself, so an already playing voice/audio message does not continue over the preview playback. I added a doc comment to clarify that.

deltachat-ios/Chat/AudioController.swift
Amzd: Would it not be safer to just assert that Thread is main in the functions that use this one?

mikovskii: I think an assert would make this less safe here. `performOnMainAndWait` is used by `stopBackgroundPlayback()`, and that can be called from callbacks that are not guaranteed to run on the main thread, for example the CallKit request completion in `CallManager`. The helper is intentional: it lets those callers stop playback safely on main, while preserving the synchronous contract that playback has already stopped when `stopBackgroundPlayback()` returns. An assert would only reject those callers instead of handling them.

Amzd:
When you delete a message the audio playback should stop. Currently on master that doesn't happen but you can stop it by going back to chatlist. Since this PR changes it to keep playing, that becomes a bit more annoying. Can you cover that case?

mikovskii:
> When you delete a message the audio playback should stop. Currently on master that doesn't happen but you can stop it by going back to chatlist. Since this PR changes it to keep playing, that becomes a bit more annoying. Can you cover that case?

done

Amzd (approved):

PR #3094, opened 2026-05-03
feat: add outgoing ringing sound and call status

### LINKED ISSUE #3005: No connecting/ringing indication in calls
Calls should have a "Connecting..." and "Ringing..." status message on screen and a ringing sound.

mikovskii:
Adds a generated outgoing ringback tone for calls, so users hear call progress while waiting for the remote side to answer.

*  `OutgoingRingbackPlayer`, backed by `AVAudioEngine`, to generate and loop a ringback tone.

demo (`with sound`):

<details>
<summary>Video</summary>

https://github.com/user-attachments/assets/b6c0e881-0cee-4532-8e5f-44d493b5f081

</details>

close: #3005


deltachat-ios/Calls/OutgoingRingbackPlayer.swift
```diff
@@ -0,0 +1,114 @@
+import AVFoundation
+
+final class OutgoingRingbackPlayer {
+    static let shared = OutgoingRingbackPlayer()
```
Amzd: This doesn't need to be static afaict, storing it as a property on CallViewController would be fine I think. In that case you can also make sure the ringing is stopped when this object deinitializes instead of in the deinit of CallViewController

r10s:
did we check if it is not easier to play a simple, short audio file with the desired tone? even if some kb bigger, i would prefer that to the additional code. what is eg. signal doing here?

mikovskii:
I looked at how it's implemented in Signal. I can create a standard .caf audio file and loop that sound. 

[signal ringback sound](https://github.com/signalapp/Signal-iOS/blob/main/Signal/AudioFiles/ringback_tone_ansi.caf)

mikovskii:
[outgoing-ringback.mp3](https://github.com/user-attachments/files/28473710/outgoing-ringback.mp3)

Based on the code OutgoingRingbackPlayer, I used FFmpeg to generate a .caf file.

deltachat-ios/Calls/OutgoingRingbackPlayer.swift
```diff
@@ -0,0 +1,84 @@
+import AVFoundation
+
+final class OutgoingRingbackPlayer {
+    private var player: AVAudioPlayer?
```
Amzd:
```suggestion
    private lazy var player: AVAudioPlayer? = {
        guard let url = Bundle.main.url(forResource: "outgoing-ringback", withExtension: "caf", subdirectory: "Assets") else {
            assertionFailure("Missing resource")
            return nil
        }
        let player = try? AVAudioPlayer(contentsOf: url)
        assert(player != nil, "Failed to init AVAudioPlayer")
        player?.numberOfLoops = -1
        player?.prepareToPlay()
        return player
    }()
```

This is what we do elsewhere instead of the builder paradigm you do with makePlayer. Feel free to log the errors differently if you think them important enough but RingbackError type is not needed.

Amzd (changes requested):

Amzd (approved):

PR #3099, opened 2026-05-06
add "Mark All as Read" to the profile switcher context menu

### LINKED ISSUE #3098: add "mark all as read" from profile switcher
when long tapping a profile in the profile switcher, we currently have the options "mute", "label", "move to top" and delete".

what is missing is an option to mark all chats as read - similar to a chat, to get rid of the unread counter.

desktop has a corresponding option, and there was even a dedicated api added for that at https://github.com/chatmail/core/pull/7709 . string is  `mark_all_as_read`.

for the position: _if_ ppl use that option, this option is likely the one used most, so it should be the first option (all other options are not used on a daily base, more settings)

cmp https://github.com/deltachat/deltachat-android/issues/4406

mikovskii:
close: #3098 

demo: 

<details>
<summary>Details</summary>

https://github.com/user-attachments/assets/0d597c0a-e90e-4688-a13d-553197150960

</details>

*The issue with the entire profile row being redrawn, as seen in the video, has been fixed in this commit*: `f99be28f47d4bc3155e8dace39c7d3468812374b`

### DIFF
--- a/deltachat-ios/Controller/ProfileSwitchViewController.swift
+++ b/deltachat-ios/Controller/ProfileSwitchViewController.swift
@@ -5,6 +5,7 @@ import DcCore
 class ProfileSwitchViewController: UITableViewController {
 
     private let dcAccounts: DcAccounts
+    var onUnreadStateChanged: (() -> Void)?
     private let accountSection = 0
     private let addSection = 1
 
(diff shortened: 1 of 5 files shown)


deltachat-ios/Controller/ProfileSwitchViewController.swift
```diff
@@ -102,6 +105,7 @@ class ProfileSwitchViewController: UITableViewController {
                     UIAction.menuAction(localizationKey: muteTitle, systemImageName: muteImage, with: indexPath, action: toggleMute),
                     UIAction.menuAction(localizationKey: "profile_tag", systemImageName: "tag", with: indexPath, action: setProfileTag),
                     UIAction.menuAction(localizationKey: "move_to_top", systemImageName: "arrow.up", with: indexPath, action: moveToTop),
+                    UIAction.menuAction(localizationKey: "mark_all_as_read", systemImageName: checkmarkImage, with: indexPath, action: markAllAsRead),
```
Amzd: can we conditionally add this action when there are unread messages?

mikovskii: i fixed that in the last commit

Amzd (approved):

PR #3102, opened 2026-05-08
Remove details from logger

Amzd:
The notification token is still in the log file because it is in the prepended device details so it makes no sense to also write this to the device logs.

#skip-changelog


r10s:
what is the gist of that change? more privacy? cleaner logs?

Amzd:
Better privacy as the notification token is an identifier that doesn't change until the user uninstalls the app.

r10s (approved):

PR #3104, opened 2026-05-08
initial Liquid Glass implementations

### LINKED ISSUE #2973: iOS 26 liquid glass

### LINKED ISSUE #3030: Topmost messages are hidden behind  by the chat header.
When the user scrolls up to view older messages in the chat history, the list doesn't scroll far enough. The oldest messages are clipped behind the fixed profile header at the top of the screen, making them unreadable.

- iOS version: iOS 26.3
- Device: iphone 11
- Delta Chat version: v2.35
- Expected behavior: The user should be able to scroll all the way to the top, and the first message should be fully visible below the header area.
- Actual behavior: Top messages are overlapping with or hidden behind the top navigation bar. (Looks like a missing top padding or scroll offset issue).
- Steps to reproduce the problem: Оpen the chat and scroll to the beginning of the conversation
- Screenshots: 
 ### current state 

Image

### expected behavior

Image

Amzd:
TODO
- [x] pasting images into input bar doesn't work
- [x] focusInputTextView doesn't work so eg after pasting the field is not focused
- [x] input bar styling on pre-liquid glass
- [x] previews for attachments other than images
- [x] profile button styling could be better
- [x] badge counter on profile switcher button is cut off
- [x] move input bar stuff to their own files 
- [x] remove all the input bar code that is no longer used
- [x] search/multi-select mode bars in chat view
- [x] multi-select mode bar in chat list
- [ ] Web views should maybe go behind nav bar
- [ ] Animation on tab bar icons no longer necessary 
- [x] Scroll down button #3154 
- [x] Search bar doesn't show up when searching in saved messages
- [ ] on pre liquid glass replying to a tall message moves the textview up too much
- [ ] profile creation screen should use more modern pill button instead of rounded square
- [x] remove all iOS 14 supporting code #3153 
- [ ] during editing the message being edited should show in the quote view #3152 

closes #3030


r10s:
ftr, i ran `pod install` for testing this PR, hope this is okay :)

the PR needs a rebase tho,  DC_SHOW_EMAILS_OFF and all these legacy options are gone in 2.51.0.

looking forward to test the PR then! what a great amount of removed lines!

for iOS 18 (not sure if it affects this PR): it is fine if things are not perfect there, as longs as usable. iOS 18 share that cannot upgrade is small (only iPhone XS/XR) and will become smaller soon. at least currently resources are spent better elsewher

in general, it probably makes sense to get the PR in sooner than later, and fix issues afterwards in smaller PR - review and development becomes too annoying otherwise.

Amzd:
Running pod install is intended as there is a new pod added. This PR still removes some functionality that I miss like pasting images into the input bar which I will fix before this can be merged.

mikovskii:
While testing this PR recently, I noticed that the button that takes users back to the latest messages in the chat is no longer visible.

mikovskii:
<details>
<summary>Before PR</summary>

image

</details>

mikovskii:
One UI suggestion: it might be worth adding a subtle fade/gradient effect at the bottom of the chat where messages disappear behind the message input bar. Without it, the input bar feels a bit visually unpleasant, at least to me.

This is how it works in Telegram, Apple Messenger, and other apps. I've attached a Telegram screenshot as a reference.


<details>
<summary>Photo</summary>

image


</details>

r10s:
there are quite some major and minor issues with LiquidGlass still.

we aim to fix the really worse ones quickly, so we can get this PR in finally - while staying able to release.

for the remaining issues, layout glitches or minor degradations, we will file dedicated issues. lots of them appear only by real usage, in real life :)

deltachat-ios/Info.plist
```diff
@@ -112,6 +112,6 @@
 	<key>ITSAppUsesNonExemptEncryption</key>
 	<false/>
 	<key>UIDesignRequiresCompatibility</key>
-	<true/>
+	<false/>
```
r10s, line 115: a single bit being the cause of all the required changes :)

deltachat-ios.xcodeproj/project.pbxproj
```diff
@@ -2319,7 +2247,7 @@
 				ENABLE_USER_SCRIPT_SANDBOXING = YES;
 				GCC_C_LANGUAGE_STANDARD = gnu17;
 				GENERATE_INFOPLIST_FILE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 14.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 15.6;
```
r10s, line 2250: created a PR to adapt minimal version on the website: https://github.com/deltachat/deltachat-pages/pull/1377

r10s (approved):
wow - what a PR. thanks for going through all that effort ❤️

for casual readers: we know that LiquidGlass is not loved by everyone, however, at some point there is no option. and it is requested more often than not :)

i read through the code roughly only, things lgtm - but i am also using the PR since some days without large issues - some smaller issues are known, but they should not block merging, but be done in follow-ups - also to let the devs test LiquidGlass further :)

deltachat-ios.xcodeproj/project.pbxproj
Amzd: approved

PR #3108, opened 2026-05-13
Mirror the local video in calls

### LINKED ISSUE #3069: self-video is not mirrored
the video made with the front camera is not mirrored (it should behave like a mirror, otherwise, one gets sea sick :) 

(it was mirrored before, without the native calls)

Amzd:
Fixes #3069


r10s (approved):
thanks a lot!

PR #3129, opened 2026-06-04
Tap on location streaming info message to open map

Amzd:
QoL for location streaming because user feedback was that they could not find where to see the location that was being streamed.

Note: It would be nicer if core exposed a DC_INFO_LOCATIONSTREAMING_DISABLED (or TOGGLED) but that's not the case.

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -850,6 +850,9 @@ class ChatViewController: UIViewController, UITableViewDelegate, UITableViewData
             showInvalidUnencryptedDialog()
         case (_, DC_INFO_GROUP_DESCRIPTION_CHANGED):
             navigationController?.pushViewController(ProfileViewController(dcContext, chatId: chatId), animated: true)
+        case (_, DC_INFO_LOCATIONSTREAMING_ENABLED),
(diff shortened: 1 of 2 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -850,6 +850,9 @@ class ChatViewController: UIViewController, UITableViewDelegate, UITableViewData
             showInvalidUnencryptedDialog()
         case (_, DC_INFO_GROUP_DESCRIPTION_CHANGED):
             navigationController?.pushViewController(ProfileViewController(dcContext, chatId: chatId), animated: true)
+        case (_, DC_INFO_LOCATIONSTREAMING_ENABLED),
+            (_, DC_INFO_UNKNOWN) where message.isInfo && message.text == "Location streaming disabled.":
```
r10s: yeah, that line is no-good, it breaks on all translations.

then better not open the map on "disabled" - which also makes some sense - and is at least consistent

Amzd: There is no translations afaict, just the default string in core. Core doesn't even use the enabled string we set (at least not for the info message).

Amzd: Removed this line for now. I looked at core a little but I am not sure how to implement a disabled type.

r10s (approved):

PR #3130, opened 2026-06-04
Fix reactions going to wrong message

Amzd:
Previously if you received a message while long pressing to open the context menu, a reaction could go to the wrong message.

After this change this desync can only happen to "Copy link" action and "Select more" which are way lower priority.


r10s (approved):

r10s:
a CHANGELOG entry is missing :)

PR #3145, opened 2026-06-13
Fix incorrect contact is me check

### LINKED ISSUE #3133: Tapping on "Me" in reactions overview opens Saved Messages profile
- Expected behavior: 
   Maybe open your own actual profile? But probably fine to do nothing. Would be nice if instead this row could have a "Remove reaction" button, which is more interesting than going to your own profile there.
- Actual behavior:
   Opens Saved Messages profile details
  - Add a reaction to a message
  - Tap reaction
  - Tap "Me: <emoji>" row

Amzd:
Fixes #3133

This only happened on profiles that have an account id that's not the same as DC_CONTACT_ID_SELF

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - Add outgoing ringing sound and call status
-
+- Fix: Tapping "Me" in reactions overview would open the wrong contact on anything other than your first profile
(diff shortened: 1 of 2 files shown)


r10s (approved):
thanks for diving into that! thanks for fixing!

PR #3151, opened 2026-06-17
Improve audio recorder session routing

nimbusline:
## Summary

- Configure the audio recorder session as `playAndRecord` with Bluetooth input support and explicitly activate it before starting `AVAudioRecorder`.
- Preserve and restore the previous audio session category, mode, and options when recording ends or fails.
- Restore the idle timer and remove the matching background notification observer during recorder teardown.
- Add an `Unreleased` changelog entry for the end-user-facing recording fix.

## Why

The voice-message recorder only switched the shared `AVAudioSession` to `.record`. On routes such as Bluetooth hands-free or car audio, iOS can fail to select the expected external microphone unless the session is active and configured for recording plus playback-style routing. Using `.playAndRecord` with `.allowBluetooth` lets iOS use HFP microphone routes while keeping the change scoped to the recorder lifecycle.

## Validation

- `git diff --check`
- `xcodebuild -list -workspace deltachat-ios.xcworkspace`
- `xcodebuild -workspace deltachat-ios.xcworkspace -scheme deltachat-ios -destination "generic/platform=iOS Simulator" build`

### DIFF
--- a/deltachat-ios/Controller/AudioRecorderController.swift
+++ b/deltachat-ios/Controller/AudioRecorderController.swift
@@ -21,7 +21,13 @@ class AudioRecorderController: UIViewController, AVAudioRecorderDelegate {
     private let bitrate: Int
 
     // Private variables
-    var oldSessionCategory: AVAudioSession.Category?
+    private struct AudioSessionConfiguration {
+        let category: AVAudioSession.Category
+        let mode: AVAudioSession.Mode
+        let options: AVAudioSession.CategoryOptions
+    }
+
+    private var previousAudioSessionConfiguration: AudioSessionConfiguration?
     var wasIdleTimerDisabled: Bool = false
 
     var recordingFilePath: String = ""
@@ -141,10 +147,10 @@ class AudioRecorderController: UIViewController, AVAudioRecorderDelegate {
     override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
 
-        NotificationCenter.default.removeObserver(self, name: UIApplication.didBecomeActiveNotification, object: nil)
+        NotificationCenter.default.removeObserver(self, name: UIApplication.didEnterBackgroundNotification, object: nil)
         audioRecorder?.stop()
+        endRecordingSession()
         stopUpdatingMeter()
-        UIApplication.shared.isIdleTimerDisabled = wasIdleTimerDisabled
     }
 
     override func viewDidDisappear(_ animated: Bool) {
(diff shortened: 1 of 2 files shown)


Amzd:
Hi, thank you for your contribution. It sadly conflicted with #3090

nimbusline:
Thanks for the heads-up! I updated the branch on top of current `main` / #3090 and folded the conflict resolution into the existing PR commit.

I kept the background playback changes from #3090 in place, including stopping playback before starting or continuing a recording, and kept this PR's recorder session changes. I also switched the Bluetooth recording option to `.allowBluetoothHFP` to avoid the deprecated `.allowBluetooth` warning.

Validation:
- `git diff --check`
- `xcodebuild -workspace deltachat-ios.xcworkspace -scheme deltachat-ios -destination "generic/platform=iOS Simulator" build`

mikovskii:
I looked at the PR and I noticed one issue: `startRecording()` updates the UI before recording is actually started. If `configureAudioSessionForRecording()`, `prepareToRecord()`, or `record()` fails, the toolbar stays in the recording state and Send remains enabled, even though no valid recording is running. This is probably rare, but audio session failures are exactly the kind of thing this PR is trying to handle, especially with Bluetooth/car audio. 

This looks pre-existing: the old code also switched the toolbar and enabled Send before `setCategory`, `prepareToRecord()`, and `record()` succeeded. The PR improves cleanup, but with the new `setActive(true)` / routing setup there is one more realistic failure point, so it may be worth fixing while touching this code. 

Could you move the toolbar/Send state update until after `audioRecorder.record()` succeeds, or reset the UI in the failure paths?

nimbusline:
Thanks, good catch. I updated `startRecording()` so the toolbar and Send button are only switched to the active recording state after `audioRecorder.record()` succeeds. The permission check now keeps Send disabled until a recording is actually running.

Validation:
- `git diff --check`
- `xcodebuild -workspace deltachat-ios.xcworkspace -scheme deltachat-ios -destination "generic/platform=iOS Simulator" build`

nimbusline:
Would it be possible to merge this and make it available via TestFlight?

r10s:
once merged, it will be available on next testflight. this time, due to large liquidglass changes, a testflight is not so far away (otherwise we often do testflight a week or so before a release)

deltachat-ios/Controller/AudioRecorderController.swift
```diff
@@ -141,10 +147,10 @@ class AudioRecorderController: UIViewController, AVAudioRecorderDelegate {
     override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
 
-        NotificationCenter.default.removeObserver(self, name: UIApplication.didBecomeActiveNotification, object: nil)
+        NotificationCenter.default.removeObserver(self, name: UIApplication.didEnterBackgroundNotification, object: nil)
```
Amzd, line 150: Good catch

deltachat-ios/Controller/AudioRecorderController.swift
```diff
@@ -256,6 +269,35 @@ class AudioRecorderController: UIViewController, AVAudioRecorderDelegate {
...
         logger.error("audio recording failed: \(error?.localizedDescription ?? "unknown")")
+        endRecordingSession()
+    }
+
+    private func configureAudioSessionForRecording() throws {
+        let session = AVAudioSession.sharedInstance()
+        if previousAudioSessionConfiguration == nil {
+            previousAudioSessionConfiguration = AudioSessionConfiguration(category: session.category,
+                                                                         mode: session.mode,
+                                                                         options: session.categoryOptions)
```
Amzd:
```suggestion
            previousAudioSessionConfiguration = AudioSessionConfiguration(
                category: session.category,
                mode: session.mode,
                options: session.categoryOptions
            )
```

Amzd (approved):

Amzd:
Tested and correctly picks up BT device now.

nimbusline:
Hey @Amzd, do you know when I’ll be able to test this on my device via TestFlight? 😄

Amzd:
Generally when we resolve all the issues with blocker tag.

PR #3155, opened 2026-06-18
Smaller liquid glass input bar size

iamDecode:
Thanks for #3104! It really makes the app feel more modern 😄. However, I think the input bar is currently a bit large in the new design. For reference, here are telegram, element, imessage:
IMG_9372IMG_9373IMG_9374

  I believe this may have been chosen due to the native UITextView content size, but this is easily circumvented with a negative vertical padding.   
  
### Preview (Old / New)

IMG_9376IMG_9369

IMG_9375IMG_9371

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Delta Chat iOS Changelog
 
+## Unreleased
+
+- Fix: smaller input bar size for new liquid glass design.
+
 ## v2.53.0
 2026-06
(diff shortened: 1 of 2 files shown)


Amzd:
Thank you for your contribution. It is still not perfect when the user has a custom text size that is larger, but this is an improvement

r10s:
much better already! thanks a lot @Amzd and @iamDecode

PR #3156, opened 2026-06-18
add custom fade effects for chat edges

mikovskii:
add a fade effect at the bottom of the chat behind the message input bar. Without it, the input bar feels a bit visually unpleasant

_before_:

<details>
<summary>Details</summary>

image

</details>

_after_:

<details>
<summary>Details</summary>

image

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Allow voice and audio messages to continue playing in the background
+- Improve message fading behind the input bar on iOS 26
 
 
 ## v2.53.0
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -583,7 +583,7 @@ class ChatViewController: UIViewController, UITableViewDelegate, UITableViewData
         edgeEffectDimmingGradients.top.frame = CGRect(x: 0, y: 0, width: view.bounds.width, height: height)
         edgeEffectDimmingGradients.top.startPoint = CGPoint(x: 0.5, y: 0)
         edgeEffectDimmingGradients.top.endPoint = CGPoint(x: 0.5, y: 1)
-        edgeEffectDimmingGradients.top.colors = [0.65, 0.30, 0].map { UIColor.systemBackground.withAlphaComponent($0).cgColor }
+        edgeEffectDimmingGradients.top.colors = [0.70, 0.55, 0].map { UIColor.systemBackground.withAlphaComponent($0).cgColor }
         edgeEffectDimmingGradients.top.locations = [0, 0.6, 1]
 
         let fadeEnd = min(height / bounds.height, 1)
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -49,6 +49,8 @@ class ChatViewController: UIViewController, UITableViewDelegate, UITableViewData
     }()
 
     private lazy var tableViewContainer: UIView = UIView()
+    private let readabilityFadeExtension: CGFloat = 24
+    private let readabilityDimmingGradients = (top: CAGradientLayer(), bottom: CAGradientLayer())
     private lazy var tableView: UITableView = {
         let tableView = UITableView()
         tableView.delegate = self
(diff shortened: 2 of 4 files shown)


r10s:
when on that, and having an idea how that goes, maybe you can do another pr:

i think atop we should have some bluring of the background going behind the title.

otherwise, the title is hard to read if there is some text below.

see to chatlist to get the idea - there the string "Chats" is nicely visible even if some text is scrolled directly below.

r10s:
i also think, the input bar could be a bit further down, if easily possible and still respecting safe area and non-liquid-glass

mikovskii:
I'll take a look tomorrow and get it done).

mikovskii:
I decided not to create a separate PR for the top edge effect, so I put everything in one PR.

mikovskii:
demo:


<details>
<summary>before</summary>


https://github.com/user-attachments/assets/36b17b38-2574-47e2-a457-cefffaa8df76



</details>


<details>
<summary>after</summary>


https://github.com/user-attachments/assets/664bc7ab-12fc-4361-b6f9-e5847fdc4c16



</details>

Amzd:
I removed the changelog entry commit because liquid glass fixes are not published yet.

Amzd:
I don't think the shade is a good idea because the navigation bar title (and statusbar) changes tint color based on what is behind it but doesn't take the shade into account for some reason

image

r10s (changes requested):
> I don't think the shade is a good idea because the navigation bar title (and statusbar) changes tint color based on what is behind it but doesn't take the shade into account for some reason

that looks wrong, indeed. ~~cannot reproduce that part tho, that should be double checked.~~ EDIT: i could reproduce it in dark mode using a white image

apart from that, to me things look much nicer and more readable with that PR, but there are two things:

- bubbles now are not fully opaque. that is not wanted.

- esp. in the upper area, the upper area, that is more visible. but again, it is not wanted at all, bubbles should be 100% opaque

- the "shade heights" can maybe be a bit smaller

mikovskii:
tried to fix the bug with navigation bar title. I used the native blur effect. I wasn't able to make the bubbles completely opaque, because the Liquid Glass buttons react to what's behind them, and since I have to place a custom effect behind them, they also react to that.  To further improve readability, I suggest make in another PR: remove the redundant chat content overview button from the header, since the same view is already available by tapping the chat title. Use a button-style container for the chat title to improve readability and make the interaction more discoverable.

<details>
<summary>Details</summary>


https://github.com/user-attachments/assets/9706034b-7884-414d-9a4e-631647384210



</details>

mikovskii:
If you like this better than how it was before, I can open another PR with these changes

<details>
<summary>Details</summary>
image


</details>

mikovskii:
@Amzd сould you please take a look?

Amzd:
Yes this will merge today, with some additional changes. Thank you for your work.

I like the glass button in the navigation bar and some other apps are also doing this. Add it to this PR.

r10s (dismissed):
this PR now removes the "apps & media" button from all chats :)

<br><i>this button is removed</i>

turning the avatar&name into a button makes some sense, yes, esp. as this is known by many other apps. form follows function

for the fade effect: in dark mode it seems to get slighty lighter to direction of edgest? that is not what is done elsewhere, eg. in our chatlist. by the new "avatar&name button", however, this is no longer a readability blocker, more smth little weird

mikovskii:
I think removing the “Apps and Media” button from the top chat bar is the right move, since it’s already available in the chat profile. In my opinion, this helps keep the top bar from looking cluttered and allows users to focus on the main chat functions. It only makes sense to keep it in iOS 18.

mikovskii:
As for the fade effect, the chat list uses the native effect that Apple added in iOS 26. Since the chat is flipped, we can’t use the native edge effect, but with the effect I added, I tried to preserve the liquid glass look—because if we make the dark edge effect too intense, the bubbles become not fully opaque.

mikovskii:
I tried to fix the issue with the light edge in the dark theme; I'm using a subtle dark tint 



<details>
<summary>Details</summary>

https://github.com/user-attachments/assets/e3f1086a-b1dc-455a-bdab-f3bc74536c9c

</details>

mikovskii:
I think it looks much cleaner without the “Apps and Media” button. 

<details>
<summary>before</summary>

image


</details>

<details>
<summary>after</summary>

image

</details>

r10s:
> I think removing the “Apps and Media” button from the top chat bar is the right move

nope. the "Apps and Media Button" is about stay there for now. 

it was a decision to have all apps via one tap from there, that must not be changed on the side.
also for consistency with android and desktop. 

there are ideas about doing that differently,  but that should be discussed and done independent of a fade effect change

mikovskii:
OK :) No problem. I added the button back in the last commit.

Amzd (approved):

PR #3158, opened 2026-06-22
feat: add scroll down button

### LINKED ISSUE #3154: Add "scroll down" button
This feature was removed in #3104 but it was never as good as the desktop one.

We want it to show when the user scrolled up more than a screen, and when they clicked on a quote which scrolled them further up.

If they clicked on a quote, pressing the scroll down button should take the user back to the quote they clicked, not to the bottom.

mikovskii:
add scroll down button. 

<details>
<summary>demo</summary>

https://github.com/user-attachments/assets/9724a16d-2346-4397-9844-25f436115155

</details>

close: #3154

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -18,7 +18,7 @@ class ChatViewController: UIViewController, UITableViewDelegate, UITableViewData
     private var reactionMessageId: Int?
     private var contextMenuVisible = false
     private var isDraggingScrollView = false
-    private var quoteReturnMessageId: Int?
+    private var quoteReturnMessageIds: [Int] = []
 
     private lazy var draft: DraftModel = {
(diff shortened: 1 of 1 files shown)


mikovskii:
issue 3154 describes that we want the button to appear only after scrolling up more than one screen. I tested this for a couple of days, and it didn't seem very user-friendly. In practice, users often want a quick way to return to their current position even after a small scroll. Showing the button after the first scrolled-past message feels me more predictable and consistent. It also matches the behavior of many popular messaging apps, making it more familiar and intuitive for users since they already expect this pattern.

Amzd:
I removed the changelog because we never released a version without the button. 
I made it scroll to more than one quote message
I changed the icon to indicate what the button does (scroll to message, or to bottom), the same as desktop

r10s:
thanks all a lot <3

PR #3162, opened 2026-06-23
use system buttons more eagerly

r10s:
for non-liquid-glass, there is no visible change, but for liquid-glass, it makes a differerence as controls familar and consistent to the system are used.
    
as a side effect, this forces to use more semantics, which may also help for accessibility, without adding any code - and be even shorter.

#skip-changelog as initial liquid glass is not released yet, and it otherwise does not make a big difference

### DIFF
--- a/deltachat-ios/Controller/QrCodeReaderController.swift
+++ b/deltachat-ios/Controller/QrCodeReaderController.swift
@@ -159,7 +159,7 @@ class QrCodeReaderController: UIViewController {
             if let appSettings = URL(string: UIApplication.openSettingsURLString) {
                 alert.addAction(UIAlertAction(title: String.localized("open_settings"), style: .default, handler: { _ in
                         UIApplication.shared.open(appSettings, options: [:], completionHandler: nil)}))
-                alert.addAction(UIAlertAction(title: String.localized("cancel"), style: .destructive, handler: nil))
+                alert.addAction(UIAlertAction(title: String.localized("cancel"), style: .cancel))
             }
             self?.present(alert, animated: true, completion: nil)
         }
--- a/deltachat-ios/Controller/Settings/Transports/EditTransportViewController.swift
+++ b/deltachat-ios/Controller/Settings/Transports/EditTransportViewController.swift
@@ -236,12 +236,12 @@ class EditTransportViewController: UITableViewController {
     }()
 
     private var cancelButton: UIBarButtonItem {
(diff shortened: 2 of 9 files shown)


deltachat-ios/Controller/Settings/Transports/EditTransportViewController.swift
```diff
@@ -236,12 +236,12 @@ class EditTransportViewController: UITableViewController {
...
 
     private var cancelButton: UIBarButtonItem {
-        let button =  UIBarButtonItem(title: String.localized("cancel"), style: .plain, target: self, action: #selector(cancelButtonPressed))
+        let button =  UIBarButtonItem(barButtonSystemItem: .cancel, target: self, action: #selector(cancelButtonPressed))
         return button
     }
 
     private lazy var loginButton: UIBarButtonItem = {
-        let button = UIBarButtonItem(title: String.localized("login_title"), style: .done, target: self, action: #selector(loginButtonPressed))
+        let button = UIBarButtonItem(barButtonSystemItem: .done, target: self, action: #selector(loginButtonPressed))
```
r10s, line 244: also android uses a simple "check" here, which is totally fine

deltachat-ios/Controller/QrCodeReaderController.swift
```diff
@@ -159,7 +159,7 @@ class QrCodeReaderController: UIViewController {
             if let appSettings = URL(string: UIApplication.openSettingsURLString) {
                 alert.addAction(UIAlertAction(title: String.localized("open_settings"), style: .default, handler: { _ in
                         UIApplication.shared.open(appSettings, options: [:], completionHandler: nil)}))
-                alert.addAction(UIAlertAction(title: String.localized("cancel"), style: .destructive, handler: nil))
```
Amzd, line 162: Hmm this changes it from red to blue doesn't it? Is that intended?

r10s: yes - imu, there is nothing destructive here

r10s: or did i missed smth?

r10s: i had a second look: the "cancel" button should not be "red" here, nor is destructive. it seems to be an old oversight.

the alert is shown when we do not have the permission to access the camera - then it explains how to enable it in settings, offer "open settings" - or "cancel".

at a similar place, for location permission, the "cancel" button is correctly non-destructive already.

Amzd (approved):

PR #3166, opened 2026-06-27
New messages marker now takes daymarkers into account

Amzd:
Previously:
```
- msg
- fresh msg
- new msgs marker
- daymarker
- fresh msg
```

Now:
```
- msg
- new msgs marker
- fresh msg
- daymarker
- fresh msg

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -1143,8 +1143,16 @@ class ChatViewController: UIViewController, UITableViewDelegate, UITableViewData
         var msgs = dcContext.getChatMsgsAndTimestamps(chatId: chatId, flags: DC_GCM_ADDDAYMARKER)
         let freshMsgsCount = self.dcContext.getUnreadMessages(chatId: self.chatId)
         if freshMsgsCount > 0 && msgs.count >= freshMsgsCount {
-            let index = msgs.count - freshMsgsCount
-            msgs.insert((Int(DC_MSG_ID_MARKER1), 0), at: index)
+            /// Fresh count including daymarkers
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: deltachat-ios/Chat/ChatViewController.swift [hunk 1]
```diff
@@ -1142,8 +1143,16 @@ class ChatViewController: UIViewController, UITableViewDelegate, UITableViewData
...
+            /// Fresh count including daymarkers
+            var freshCount = freshMsgsCount
+            var i = 0
+            while i < freshCount {
+                i += 1
+                if msgs.count > freshCount, msgs[msgs.count - freshCount].0 == DC_MSG_ID_DAYMARKER {
+                    freshCount += 1
+                }
+            }
+            msgs.insert((Int(DC_MSG_ID_MARKER1), 0), at: msgs.count - freshCount)
```
r10s, line 1155: what does insert do if the index becomes negative?

currently, it cannot happen at a first glance - but surroundings as core may change or have bugs, would be bad if that crashes.

as a general precaution, if it will crash otherwise, i would check that the index is actually valid

r10s (approved):
thanks for taking care. at least on first check, i did not spot an off-by-one :) but practise will tell :)

deltachat-ios/Chat/ChatViewController.swift
Amzd: Cannot happen because it is checked both before and in the while loop but it would probably crash like other array access methods do.

r10s: thanks for double checking!

r10s (approved):
an CHANGELOG entry is missing tho

PR #3173, opened 2026-06-29
switch back to core 2.53.0

r10s:
i've overseen during review that #3144 contains wrong core version.

this PR switches back to 2.53.0


Amzd:
Ah that keeps happening when rebasing using GitHub Desktop, I should look for a new git client.

r10s:
it will not get unnoticed at least :)

i see that immediately on the command line that shows "submodule is misaligned" in the prompt (using https://github.com/git/git/blob/master/contrib/completion/git-prompt.sh )

PR #3176, opened 2026-06-30
Fix Liquid Glass input bar layout jump

Amzd:
The layout jump happened when you have set a bigger text size (or when using smaller device).

This PR

- removes the jump by doing an extra layout call in onAppear
- limits the height of the input bar to the button size unless there is a draft or it is focused

### DIFF
--- a/deltachat-ios/Chat/InputBarAccessoryView/InputBarTextView.swift
+++ b/deltachat-ios/Chat/InputBarAccessoryView/InputBarTextView.swift
@@ -22,6 +22,8 @@ private struct _InputBarTextView: UIViewRepresentable {
 
     func makeUIView(context: Context) -> ChatInputTextView {
         let textView = ChatInputTextView()
+        textView.text = text
+        textView.imagePasteDelegate = imagePasteDelegate
         textView.keyboardDismissMode = .none
         textView.delegate = context.coordinator
         textView.adjustsFontForContentSizeCategory = true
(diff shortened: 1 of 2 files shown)


r10s:
> the layout jump happened when you have set a bigger text size (or when using smaller device).

for me, on an iPhone 13 mini, the text size does not make a difference - i was always using the default, and tried minimal and maximal text size, all had the "jumps" on entering the chat.

with this PR, the jump is shifted to the time the input field is focused - so it is still there, and smth is probably still wrong. this is true for all font sizes, but visible more on big ones. maybe there is some superfluous padding that scales with the text size but should not?

it is not only the jump: the padding is also wasting valuable space

below two images, first with bigger text, placeholder and typed text have rightfully the same size.  there the issue is very noticable. but also for the default text size on the second image, one notices smth is weird :)

IMG_2060

IMG_2063

Amzd:
> with this PR, the jump is shifted to the time the input field is focused - so it is still there

Yes correct. Since we can't know how huge the user sets their text size that's just how it works, unless we calculate the button size based on the font size.

r10s:
but the text size seems not to be the issue - there is some padding on all font sizes is not needed. one should be able to remove that padding?  the placeholder text is just fine - but once focused, there is the same text size but some padding, that causes the jump.

that said, this  PR makes things already much better as there is no jumping on opening and when the keyboard is opened, things shift anyway. still, it wastes space and looks a bit off, maybe we're doing just too much here? as the input bar on the left is fine already?

r10s:
closed in favour of https://github.com/deltachat/deltachat-ios/pull/3179

PR #3177, opened 2026-06-30
Fix sharing files to a chat in a different profile

### LINKED ISSUE #3101: Share to different profile does not work for files
Introduced in #3017 

When the app is opened from share sheet, the shared items are converted to messages.

https://github.com/deltachat/deltachat-ios/blob/a8af78af613ef22a8038096d066fe6efdb680bd2/deltachat-ios/Coordinator/AppCoordinator.swift#L233-L246

https://github.com/deltachat/deltachat-ios/blob/a8af78af613ef22a8038096d066fe6efdb680bd2/deltachat-ios/Coordinator/AppCoordinator.swift#L256

After switching profile the files attached to the created messages are still in the other profile's db and can't be found from the current profile, so the sent attachments will be empty files.

mvanhorn:
## Summary

Sharing a file from the iOS share sheet into a chat that belongs to a **different profile** than the one selected when the share sheet opened now sends the real attachment instead of an empty (0-byte) file.

## Why this matters

The regression was introduced in #3017. When the app is opened from the share sheet, `AppCoordinator.handleShareDeeplink(url:)` decoded the payload and immediately built `DcMsg` objects against the *currently selected* `dcContext`, calling `setFile(filepath:)` which copies each attachment into that profile's blobdir. The messages were then stashed in `RelayHelper`.

If the user subsequently picked a chat in another profile, `shareAndFinishRelaying(to:)` sent those pre-built messages against the now-selected destination account. The messages and their copied blob files still belonged to the originating account's database, so they could not be resolved from the destination account and the sent attachments arrived empty.

## Changes

- `RelayHelper` now carries the raw `[CodableNSItemProvider]` payload instead of pre-built `[DcMsg]`. The `DcMsg` objects are constructed inside `shareAndFinishRelaying(to:)` against the destination chat's context, so `setFile` copies attachments into the correct profile's blobdir. The same-profile path is unchanged in behavior.
- No change to where the share extension stores files: they already live in the shared app-group container (`shareExtensionDirectory`), reachable by the main app for any account.
- `CodableNSItemProvider` gains a `viewType` accessor so the share-confirmation summary in `ChatViewController` can report item types without a materialized message.

## Testing

Added a regression test (`DcTests.shareAttachmentToDifferentProfileKeepsFileBytes`) that stages a shared file while profile A is selected, switches to profile B, relays to a chat in B, and asserts the resulting message resolves to a non-empty file. The test suite is marked `.serialized` because the new case mutates the globally selected account and shared `RelayHelper` state.

Fixes #3101

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Show notifications when app is in foreground and the respective chat is not on screen
+- Fix: sharing files to a chat in a different profile no longer sends empty files
 
 
 ## v2.53.1 Testflight
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Show in title if an app is still in draft mode
 - Style buttons on welcome screen according to LiquidGlass
 - Fix: Hide 'Show in Chat' and 'Use as widget' from app drafts
+- Fix: sharing files to a chat in a different profile no longer sends empty files
 
 
 ## v2.53.1 Testflight
--- a/DcCore/DcCore/Helper/CodableNSItemProvider.swift
+++ b/DcCore/DcCore/Helper/CodableNSItemProvider.swift
@@ -10,7 +10,14 @@ public var shareExtensionDirectory = FileManager.default
 public enum CodableNSItemProvider: Codable {
     case contentsAt(url: URL, viewType: Int32)
     case text(text: String)
-    
+
+    public var viewType: Int32 {
+        switch self {
+        case .contentsAt(_, let viewType): return viewType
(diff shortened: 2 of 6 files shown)


Amzd (approved):
Original unit test code didn't take into account testing multiple profiles so the way you worked around that is fine, maybe a newOfflineAccount(named: "A") would be nicer but this works.

The rest looks great, thank you.

mvanhorn:
Thanks @Amzd for reworking the test setup for multiple profiles and landing this.

mvanhorn:
Thanks @Amzd! Sharing files into a chat on a different profile working again is a nice quality-of-life fix.

PR #3179, opened 2026-07-01
Reduce padding on the input field

Amzd:
Replaces #3176 

This reduces the padding on the input field and fixes the field being misaligned when text size is smaller.

Below the largest text size (without "Larger Accessibility Sizes" turned on) and the default text size.

Simulator Screenshot - iPhone Air - 2026-07-01 at 14 09 51

Simulator Screenshot - iPhone Air - 2026-07-01 at 14 10 07

Jumping still happens when you use "Larger Accessibility Sizes"

### DIFF
--- a/deltachat-ios/Chat/InputBarAccessoryView/InputBarTextView.swift
+++ b/deltachat-ios/Chat/InputBarAccessoryView/InputBarTextView.swift
@@ -5,13 +5,16 @@ struct InputBarTextView: View {
     @Binding var text: String
     @State private var contentSize: CGSize = .zero
     weak var imagePasteDelegate: ChatInputTextViewPasteDelegate?
+    var textContainerInset: UIEdgeInsets
+    var maxHeight: CGFloat
 
     var body: some View {
         _InputBarTextView(
             text: $text,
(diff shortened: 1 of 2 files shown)


r10s (approved):
this is much better. as you mentioned there is only a small jump visible in really large sizes.

i totally see that it gets more complex to calculate buttonSize more accurate and without flickering.

this PR is good enough to close the issue and call it a day. thanks a lot!

Amzd:
Yea but sadly this PR creates a bug in quote replying that freezes the app.

r10s:
> Yea but sadly this PR creates a bug in quote replying that freezes the app.

can you reproduce that? that would be great news.

i have random freezes on replying _without_ this PR as well, thought it is related to the "copy/paste system text menu", but could not reproduce really.

Amzd:
Ah okay I first caught it on this PR, but might be pre-existing. Quoting messages with text and image can cause an infinite loop due to the .calculated(size: $size) in QuoteViewSwiftUI. I think I should instead use the UIKit view in a UIViewRepresentable, that would also reduce code duplication between SwiftUI and UIKit.

Issue #3183, opened 2026-07-06
cannot close keyboard on "create new profile" screen


r10s:
when opening  "create new profile" screen, the keyboard s opened on purpose, with focus in the name field.

on small screens (iphone 13 mini eg), the "use other server" button is under the keyboard, and one cannot scroll the view.

as one also cannot close the keyboard easily (you can once you've typed smth or so), the button is hard to reach.

idea is to make the view a scroll view in general. 

<img width="320"  src="https://github.com/user-attachments/assets/706f6a26-d019-4295-9e2a-6b3056d8704b" />

Issue #3157, opened 2026-06-22
LiquidGlass follow-ups


r10s:
since #3104 Delta Chat comes with Apples new LiquidGlass UI 🎉🥛 the flag `UIDesignRequiresCompatibility` is no longer required and the Delta Chat is prepared for the upcoming [iOS 27 no longer supporting non-LiquidGlass UI](https://developer.apple.com/documentation/BundleResources/Information-Property-List/UIDesignRequiresCompatibility).

still, there are some things that could be tweaked, this is what this PR is about.

on-point help is very welcome  - at this point thanks to @iamDecode, @mikovskii and others  

- [x] "new chat" hints unreadable/mirrored - issue at #3161 

- [x] add "scroll down" button back - issue at #3154

- [x] editing a message does not show  "edit message in quote view" like before - pr at #3167

- [x] when opening the message view, the input bar changes in height slightly, making all bubbles bump up some pixel #3176, #3179

- [x] animation on tab bar icons no longer necessary (they were introduced at #[3014](https://github.com/deltachat/deltachat-ios/pull/3014)) - PR at #3163

- [ ] Web views should maybe go behind nav bar - this is not totally decided yet, but a PR to test is welcome. this affects "show full message" and "help". for "apps", we probably might still want to have clear separation

- [x] on pre liquid glass replying to a tall message moves the textview up too much

- [x] profile creation screen should use more modern pill button instead of rounded square, #3184, #3203

- [ ] "wallpaper" selection, could have more modern pill buttons, and the wallpaper could also go behind navigation

- [ ] profile switcher transparency could be nicer - should be like "attach contact" or "reaction details"

- [x] messagelist title is worse readable than chatlist. otoh, it is similar to safari. ~~check if blurring background improves things, we can decide if we have that option~~  EDIT: after discussion, it turns out that darkmode is far better so it is probably good enough to tweak the fading - blurring pretty nice, tho, PR at #3156

- [ ] sometimes there is a delay when long tapping a message bubble, the "reactions" are first below the bubble, then above

- [ ] Icon of attached file in draft area doesn't show the "file" icon it did before liquid glass
- [x] multi-select in chatlist should hide lower tabbar ("QR code", "Chats", "Settings") again, PR #3205

- [x] use system buttons more eagerly, pr #3162

- [x] remove all iOS 14 supporting code #3153 

- [X] smaller liquid glass input bar size - fixed by #3155

[...]

PR #3184, opened 2026-07-06
add and use helpers to style UIButton according to liquid glass

r10s:
this PR adds to small helper functions used to style primary and secondary liquid glass style buttons.

that way, `UIButton` look the same as `UIBarButtonItem` buttons used eg. for the "accept" or "block" buttons. i was wondering, why this is needed at all, i thought, that style should just be the default, but after short call with @Amzd , the configuration roundtrip seems really to be needed.

in subsequent PR, these helper function should also be used for "instant onboarding" and "wallpaper" view controller, but they need some additional love (https://github.com/deltachat/deltachat-ios/issues/3183 ,  https://github.com/deltachat/deltachat-ios/issues/3157) before to not make things worse by styling.

targets https://github.com/deltachat/deltachat-ios/issues/3157

before / after:

 

<details>
<summary>using `.large` is too large :)</summary>
 
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Show notifications when app is in foreground and the respective chat is not on screen
 - Show in title if an app is still in draft mode
+- Style buttons on welcome screen according to LiquidGlass
 - Fix: Hide 'Show in Chat' and 'Use as widget' from app drafts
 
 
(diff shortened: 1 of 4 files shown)


Amzd:
To me the text does not look vertically centered in the button

r10s:
> To me the text does not look vertically centered in the button

it is good enough.

i tried to do an explicit `config.contentInsets = NSDirectionalEdgeInsets(top: 15, leading: 15, bottom: 15, trailing: 15)` - that seems to be exactly the top/bottom borders apple is using with the `.large` default.

it looks exactly the same, setting same top/bottom borders, so they're not "wrong".

so, it might be a visual deception, that we should not try to fix by forcefully using _different_  top/bottom borders, implicitly assuming things wrt font and layout and baseline. this will result in other issues when apple changes that. maybe it is even a decision by apple to render fonts that way, if you look at the "Accept" buttons, they're similar, same for these buttons in other apps.

we should not try to tweak what the system provides without very good reasons, and this minimal  visual deception is good enough.

r10s:
however, while playing around, `.medium` button size seems more appropriate, also to match previous layout. the effect is also less visible there (it also comes from the many lower-case chars on the comparable long button texts, shifting center down compared to a button like "OK")

r10s:
this is how it looks on iPhone 7 with **iOS 15 without liquid glass**, btw.

pretty nice and the same as before. this smartness is also a hint to just keep the apple defaults, we could have done that years ago already, the button would be the same would have shifted to liquid glass automatically then :)

Amzd (approved):
You're right we shouldn't do anything custom there. Just feels very odd that it's not centered. I like the medium.

PR #3186, opened 2026-07-08
Fix freeze in input bar when quoting or editing a message with an image

### LINKED ISSUE #3181: quote-reply freezes that app sometimes
when playing around with the input field, copy+paste or so, the app is sometimes freezing.

cannot say concrete steps to reproduce, but @Amzd has seen that happening again.

this issue was taken from the comment at https://github.com/deltachat/deltachat-ios/pull/3179#issuecomment-4864511326

Amzd:
Fixes #3181


r10s (approved):
nice - lots of code removed!

PR #3188, opened 2026-07-09
Remove redundant tap handler

### LINKED ISSUE #3182: editing longer message sometimes scrolls down or freezes
editing a longer message (say, some tens of lines, something quite normal), when moving the cursor up, typing a single character moves cursor down, resulting in lots of confusion.

this does not happen for short text.

while trying out with some text, the whole app freezes reproducible for me:

 
https://github.com/user-attachments/assets/8661259d-4883-453b-bf93-f31fee899408

the [text](https://gist.github.com/r10s/a686030458867b88a7450add790b27f9) has some line breaks and comparable long lines, maybe it is related to that. when forwarding the message to another chat, it does not freeze there - maybe because the chat from the video us really completely empty

Amzd:
This potentially fixes #3182 but I was never able to reproduce that issue reliably.

### DIFF
--- a/deltachat-ios/Chat/InputBarAccessoryView/InputBarView.swift
+++ b/deltachat-ios/Chat/InputBarAccessoryView/InputBarView.swift
@@ -70,9 +70,6 @@ struct InputBarView: View {
                         }
                     }
                     .modifier { glassEffect(view: $0, padding: 0, minHeight: buttonSize, interactive: true) }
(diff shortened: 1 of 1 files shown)


r10s (approved):
looking forward to get that in, for me https://github.com/deltachat/deltachat-ios/issues/3182 happened a lot, we can close the issue with this PR - if it still happens to me, i will re-opren

PR #3189, opened 2026-07-09
Center input bar textfield on pre-liquid glass with regular text size

Amzd:
The HStack aligns items to the bottom so when text field grows, the send and attach button stick to the bottom. However when the Input bar is smaller it should stay centered instead of be at the bottom. The VStack wrapper is required because resizing the InputBarTextView itself with a min-height does not center the cursor inside it.

Before and after:

IMG_4751
IMG_4752

The red box is the text view. The added VStack expands to the full height.


r10s (approved):
very nice!

PR #3194, opened 2026-07-14
Changed the default emoji 👎 to 😂

Amzd:


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -1931,7 +1931,7 @@ extension ChatViewController {
         let myReactions = getMyReactions(messageId: messageId)
         var myReactionChecked = false
 
-        for reaction in [DefaultReactions.thumbsUp, .thumbsDown, .heart] {
+        for reaction in [DefaultReactions.thumbsUp, .faceWithTearsOfJoy, .heart] {
```
r10s: we should keep the order of the defaults used elsewhere:

```suggestion
        for reaction in [DefaultReactions.thumbsUp, .heart, .faceWithTearsOfJoy] {
```

r10s (approved):
very nice! the order should stay consistent, and i think, this is worth a CHANGLOG line.

PR #3195, opened 2026-07-14
Fix: Attaching live photos as file so they aren't compressed did not work

### LINKED ISSUE #3193: Attaching uncompressed image doesn't work
Clipper > File > From Gallery > Pick an image > Nothing happens

```
Failed to set file: Copying new blobfile failed: the source path is neither a regular file nor a symlink to a regular file
```

The `itemProvider.loadFileRepresentation` fails for live photos as their `itemProvider.registeredTypeIdentifiers.first` is `com.apple.live-photo-bundle` which fails.

Amzd:
Fixes #3193 

I switched to using CodableNSItemProvider for extracting file from NSItemProvider

Live photos are still only sent as a jpeg instead of the short video.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Style buttons on welcome screen according to LiquidGlass
 - Fix: Hide 'Show in Chat' and 'Use as widget' from app drafts
 - Fix: sharing files to a chat in a different profile no longer sends empty files
+- Fix: Attaching live photos as file so they aren't compressed did not work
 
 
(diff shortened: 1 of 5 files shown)


r10s:
> Live photos are still only sent as a jpeg instead of the short video

maybe that is even expected and the better tradeoff. "live photos" are not supported on other platforms, and i have more often seen ppl using them accidentally than on purpose. converting them to a video seems poor - player is not made for that and the whole photo will be cluttered by controls that are mostly useless

r10s (approved):

PR #3196, opened 2026-07-14
Allow closing keyboard on onboarding screen

### LINKED ISSUE #3183: cannot close keyboard on "create new profile" screen

Amzd:
Fixes #3183

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Show in title if an app is still in draft mode
 - Style buttons on welcome screen according to LiquidGlass
 - Changed the quick reaction emoji "👎" to "😂" in the message bubble context menu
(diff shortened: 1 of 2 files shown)


Amzd:
Was this an issue pre-liquid glass PR @r10s? (aka: does it need a changelog?)

r10s:
iirc, it was an issue before. i just became more visible by the initially little larger liquidglass buttons

r10s (approved):

PR #3197, opened 2026-07-15
clearer "more" navigation buttons

r10s:
with liquidglass, navigation buttons got a circular or partly circular shape. therefore and additional shape is unneeded clutter, and we use just `ellipsis` instead of `ellipsis.circle`. this is also what many other apps do, it just looks better.

only for the unshaped "more icon" in the bubble's context menu, we stay with `ellipsis.circle`

~~this change also happens for pre-liquid-glass,
however, it is okay also there, even looks fresh, and also before, not surround all ellipsis were surrounded by a circle (eg. reactions). considering that pre-liquid-glass will fade out
(it is less than 20% already), it is not worth to do extra stuff for such side things.~~ EDIT i added a commit to preserve the old icon on pre-liquid-glass

#skip-changelog as liquidglass was not released

### DIFF
--- a/deltachat-ios/Controller/QrPageController.swift
+++ b/deltachat-ios/Controller/QrPageController.swift
@@ -80,7 +80,7 @@ class QrPageController: UIPageViewController {
 
     private func updateMenuItems() {
         let menu = moreButtonMenu()
-        let button =  UIBarButtonItem(image: UIImage(systemName: "ellipsis.circle"), menu: menu)
+        let button =  UIBarButtonItem(image: UIImage(systemName: "ellipsis"), menu: menu)
         navigationItem.rightBarButtonItem = button
     }
 
--- a/deltachat-ios/Controller/QrPageController.swift
+++ b/deltachat-ios/Controller/QrPageController.swift
@@ -80,7 +80,7 @@ class QrPageController: UIPageViewController {
 
     private func updateMenuItems() {
         let menu = moreButtonMenu()
-        let button =  UIBarButtonItem(image: UIImage(systemName: "ellipsis"), menu: menu)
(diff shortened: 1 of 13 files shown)


Amzd:
Hmm, this makes it quite a bit worse pre-liquid glass..

r10s:
> Hmm, this makes it quite a bit worse pre-liquid glass..

yes, but focus is liquid-glass, which is >80% of the user, it is worth making it better for them. i also disagree, that it really is worse for pre-liquid-glass - one can regard it is a fresh look there as well, the ellipsis is larger, and i expect other apps to have a similar change, so there will be a change for the user anyways

however, i get your point as well, before discussing much more, i added a commit that leave things unchanged for pre-liquid-glass

deltachat-ios/Controller/QrPageController.swift
```diff
@@ -80,7 +80,7 @@ class QrPageController: UIPageViewController {
 
     private func updateMenuItems() {
         let menu = moreButtonMenu()
-        let button =  UIBarButtonItem(image: UIImage(systemName: "ellipsis.circle"), menu: menu)
+        let button =  UIBarButtonItem(image: UIImage(systemName: .ellipsisNavigation()), menu: menu)
```
Amzd, line 83: There is two spaces here, but was already there before

Amzd (approved):

deltachat-ios/Controller/QrPageController.swift
r10s: there are quite some more. i do a separate PR to fix that

r10s: did so at https://github.com/deltachat/deltachat-ios/pull/3200

PR #3202, opened 2026-07-17
consistent delete position

r10s:
this PR unified the position of the "delete" button in the chatlist's and chat's toolbars.

the fixedSpace(0) results in the visual separation,on non-liquid-glass, that is a no-op.

still a bit unsure about the position and the added space in general (comments welcome!), but it is an improvement in any case :)

#skip-changelog

## before 

 <br> 

## after

 <br>


Amzd (approved):
Nice simple polish:)

PR #3209, opened 2026-07-20
show all five default reactions

r10s:
this PR uses a "command palette" instead of "preferred element size" to show the default reactions. 

if there is few space, the command palette can be scrolled horizontally.

this is for (a) consistency with other platforms (when reactions were introduced, "command palette" was not or not widely available, or it was too complex, idk :) and (b) to be prepared for different default sets which may come with channel reactions


r10s:
this PR became part of https://github.com/deltachat/deltachat-ios/pull/3225

PR #3215, opened 2026-07-22
add translate text messages using system api

mikovskii:
Add a text message translation using the Apple Translate API

<details>
<summary>Details</summary>

https://github.com/user-attachments/assets/28692ca4-b343-4077-a222-be8a89dd0e1e

</details>

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -2101,7 +2101,7 @@ extension ChatViewController {
                    !message.isMarkerOrInfo,
                    let text = message.text,
                    !text.isEmpty {
-                    children.append(
+                    moreOptions.append(
                         UIAction.menuAction(localizationKey: "translate", systemImageName: "translate", with: text, action: presentTranslation)
                     )
                 }
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -2328,12 +2328,12 @@ extension ChatViewController {
 
     @available(iOS 17.4, *)
     private func presentTranslation(text: String) {
-        let host = UIHostingController(rootView: TranslationPresentationView(text: text) { [weak self] in
+        let host = UIHostingController(rootView: TranslationPresentationView(text: text, onDismiss: { [weak self] in
             self?.translationPresentationHost?.willMove(toParent: nil)
             self?.translationPresentationHost?.view.removeFromSuperview()
             self?.translationPresentationHost?.removeFromParent()
             self?.translationPresentationHost = nil
-        })
+        }))
         translationPresentationHost = host
         addChild(host)
         view.addSubview(host.view)
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -3,6 +3,7 @@ import MCEmojiPicker
 import QuickLook
 import UIKit
 import SwiftUI
+import Translation
 import AVFoundation
 import DcCore
 import SDWebImage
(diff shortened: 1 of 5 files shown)


r10s:
cool function!

however, there is a high chance that this api is not working offline but also online and affects confidentiality therefore.

i did not find any proof that this is not the case, but even if it offline that today, it maybe differently tomorrow, with any iOS update.

otoh, ppl may use worse tools for translation ...

still, as delta chat is not social media, but meant for friends&family, there is a high chance that two ppl speak the same language, so the pr has comparable low effect anyways.

mikovskii:
Thanks for the feedback. I agree that confidentiality is the main concern here.

According to Apple’s documentation, translation is performed on-device after the required language models have been downloaded. I can also verify the behavior in airplane mode and add a clear note that the feature is only available when on-device translation is supported.

Of course, we cannot guarantee that Apple will never change the implementation in a future iOS release. However, the same applies to many system APIs. We can only rely on the documented behavior of the current API and avoid using the feature whenever the system cannot provide an on-device translation.

Regarding the usefulness, I agree that it may not be a core feature for every user. Still, it can be helpful in international groups, communities, and conversations between multilingual friends or family members. It also avoids users copying potentially sensitive messages into third-party translation services.

mikovskii:
Apple documents that TranslationSession uses on-device ML models. I can additionally restrict the feature to language pairs whose LanguageAvailability status is .installed, so message text is passed to the translation session only after the required local models are already present.

This gives us a clear practical offline guarantee for the current documented implementation, and it can be verified in airplane mode and with network monitoring.

A fully enforceable guarantee would require shipping and running our own translation model, which would have substantial app-size and maintenance costs.

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -2315,13 +2326,42 @@ extension ChatViewController {
         UIPasteboard.general.string = stringsToCopy
     }
 
+    @available(iOS 17.4, *)
+    private func presentTranslation(text: String) {
+        let host = UIHostingController(rootView: TranslationPresentationView(text: text) { [weak self] in
```
Amzd: Explicitly add the onDismiss label here so it is more readable what this closure does.

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -2315,13 +2326,42 @@ extension ChatViewController {
         UIPasteboard.general.string = stringsToCopy
     }
 
+    @available(iOS 17.4, *)
+    private func presentTranslation(text: String) {
```
Amzd: Move this and the TranslationPresentationView to their own file eg Translation.swift

Amzd (changes requested):
I would really appreciate this feature, as even friends sometimes don't know how to say something in one language so use another, so translating in app would make this a lot quicker @r10s

scripts/untranslated.xml
```diff
@@ -5,6 +5,7 @@
     <string name="connectivity_low_power_mode">Throttled by system's \"Low Power Mode\".</string>
     <string name="backup_successful">Backup successful</string>
     <string name="backup_successful_explain_ios">You can find the backup in the \"Delta Chat\" folder using the \"Files\" app.\n\nMove the backup out of this folder to keep it when deleting Delta Chat.</string>
+    <string name="Translate">Translate</string>
```
Amzd: Wrong capitalization in the name

Amzd (changes requested):

Amzd (changes requested):
Please move Translate to More Options sub-menu so small screens wont have to scroll to see all options. Yes this is less convenient but we anticipate that Translate is not an often used feature in private 1:1 communications.

Amzd:
Apple says this on their Translation & Privacy page

> The audio and text (including text recognized in images) you choose to have translated may be sent to Apple to process your requests. [...] By default, Apple stores transcripts of your audio interactions and your text inputs for two years and may review a subset of these transcripts. The small subset of requests that have been reviewed may be kept beyond two years, for ongoing improvement of Translate, System-wide translation, and other natural language processing features such as Siri.

https://www.apple.com/legal/privacy/data/en/translation/

So unless users turned on the "On-Device" toggle in Translate settings their messages will be stored for two years (minimum) which is not desired I think. I looked around the APIs a bit and I don't think we can enforce private offline translation, so lets not do this integration at the moment. Sorry.

PR #3219, opened 2026-07-24
Bring back "back" in the back-button on older versions

Amzd:
This was removed for liquid glass but is ugly on older versions


r10s (approved):
so many "back" and "b" in the issue title.

could be a rap or a punk song "BRING BACK BACK IN THE BACK BUTTON"

PR #3220, opened 2026-07-24
Remove restore last chat feature

Amzd:
On liquid glass this feature has a visual bug with the back button and on all versions this feature is questionable: if a user manages to find an attachment to send with a preview that freezes/crashes the app when the chat is opened this feature would permanently lock a user out of the app because it keeps opening that same chat.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Retire provider database
 - Add experimental "Team Profile" mode to profile creation
+- No longer open last chat when app is restarted
 - Fix notification info line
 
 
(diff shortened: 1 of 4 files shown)


r10s:
hm, but isn't it expected that an app restores itself in its last state?

or are we assuming that in the vast majority of cases the system "freezes" the app, so that that anyway happens implicitly? (unless the app is not used for a long time, but then it does not matter)

Amzd:
Yes I am assuming that the system does not restart the app faster than a user forgets the context they were in inside the app.

r10s:
probably reasonable in 2026, let's merge, if it causes issues, we can iterate

r10s (approved):

PR #3225, opened 2026-07-27
allow reactions to channels

r10s:
this PR allows channel subscribers to react to messages.

- **reactions are broadcasted** as other messages from the channel owner, this requires core with https://github.com/chatmail/core/pull/8450 and is done about every ten minutes

- **allow to mute/unmute outgoing channels**

- **shows all full 5 default reactions,** in the future, it is planned to make that configurable, so having a different default than on android/desktop/ubuntutouch becomes worse.
for direct display, iOS 17 is required. otherwise a submenu is used (before, direct display required iOS 16 only, but that allows to display only  3~4 defaults easily directly) (as around 2-3% are still on iOS 16, they will get a submenu to react now, as for iOS 15. this drawback is accepted for easier maintainance and to make it better for 95% of the user)

- **reaction summary** shows counts for incoming channels - channel subscribers do not know the exact details

- **reactions in channels** are not be considered mentions, that's too annoying, cmp android PR

the layout of the 5 default reactions can be tweaked a bit in a subsequent PR, i'd like to have the typical separator below, and maybe also little larger emojis, but these are no blockers, and i prefer to keep diff and review easy.

requires a core with https://github.com/chatmail/core/pull/8450

counterpart of https://github.com/deltachat/deltachat-android/pull/4560

### DIFF
--- a/DcCore/DcCore/Helper/UNMutableNotificationContent+init.swift
+++ b/DcCore/DcCore/Helper/UNMutableNotificationContent+init.swift
@@ -27,7 +27,8 @@ public extension UNMutableNotificationContent {
     /// Initialiser that returns a notification for an incoming reaction. Returns nil if no notification should be sent (eg if chat is muted)
     convenience init?(forReaction reaction: String, from contact: Int, msg: DcMsg, chat: DcChat, context: DcContext) {
         guard !context.isMuted() else { return nil }
-        guard !chat.isMuted || (chat.isMultiUser && context.isMentionsEnabled) else { return nil }
+        guard !chat.isMuted || (chat.isMultiUser && !chat.isOutBroadcast && context.isMentionsEnabled) else { return nil }
+
         let contact = context.getContact(id: contact)
         let summary = msg.summary(chars: Self.pushNotificationCharLimit) ?? ""
         self.init()
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -2019,35 +2019,37 @@ extension ChatViewController {
             })
         }
 
-        let showPicker = myReactions.isEmpty || myReactionChecked
-        let title: String
-        let accessibilityLabel: String?
-        if showPicker {
-            title = "•••"
-            accessibilityLabel = String.localized("pref_other")
-        } else {
-            title = (myReactions.first ?? "?") + "✓"
-            accessibilityLabel = nil
-        }
-        let action = UIAction(title: title) { [weak self] _ in
-            guard let self else { return }
+        if !(dcChat.isInBroadcast || dcChat.isOutBroadcast) {
+            let showPicker = myReactions.isEmpty || myReactionChecked
+            let title: String
+            let accessibilityLabel: String?
             if showPicker {
-                reactionMessageId = messageId
-                let pickerViewController = MCEmojiPickerViewController()
-                pickerViewController.navigationItem.title = String.localized("react")
-                pickerViewController.delegate = self
-
-                let navigationController = UINavigationController(rootViewController: pickerViewController)
-                if let sheet = navigationController.sheetPresentationController {
-                    sheet.detents = [.medium(), .large()]
-                }
-                present(navigationController, animated: true)
+                title = "•••"
+                accessibilityLabel = String.localized("pref_other")
             } else {
-                dcContext.sendReaction(messageId: messageId, reaction: nil)
+                title = (myReactions.first ?? "?") + "✓"
+                accessibilityLabel = nil
             }
+            let action = UIAction(title: title) { [weak self] _ in
+                guard let self else { return }
+                if showPicker {
+                    reactionMessageId = messageId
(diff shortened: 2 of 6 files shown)


DcCore/DcCore/Helper/UNMutableNotificationContent+init.swift
```diff
@@ -27,7 +27,8 @@ public extension UNMutableNotificationContent {
     /// Initialiser that returns a notification for an incoming reaction. Returns nil if no notification should be sent (eg if chat is muted)
     convenience init?(forReaction reaction: String, from contact: Int, msg: DcMsg, chat: DcChat, context: DcContext) {
         guard !context.isMuted() else { return nil }
-        guard !chat.isMuted || (chat.isMultiUser && context.isMentionsEnabled) else { return nil }
+        guard !chat.isMuted || (chat.isMultiUser && !chat.isOutBroadcast && context.isMentionsEnabled) else { return nil }
```
Amzd, line 30:
```suggestion
        guard !chat.isOutBroadcast else { return nil }
        guard !chat.isMuted || (chat.isMultiUser && context.isMentionsEnabled) else { return nil }
```
Would make it much more readable that broadcasts never get reaction notifications

deltachat-ios/Chat/Reactions/ReactionsOverviewViewController.swift
```diff
@@ -8,17 +8,32 @@ protocol ReactionsOverviewViewControllerDelegate: AnyObject {
...
-        self.context = context
+    init(reactions: DcReactions, showFrequencies: Bool, context: DcContext) {
+
+        self.showFrequencies = showFrequencies
+
+        // layout by strings is not great, but good enough for now for a secondary UI
+        if showFrequencies {
+            self.contactIds = []
+            self.texts = reactions.reactions.map { reaction in
+                return "\(reaction.emoji)   " + String.localized(stringID: "n_reactions", parameter: reaction.count)
```
Amzd:
```suggestion
                "\(reaction.emoji)   " + String.localized(stringID: "n_reactions", parameter: reaction.count)
```
We generally use implicit return in map, especially one liners

deltachat-ios/Chat/Reactions/ReactionsOverviewViewController.swift
```diff
@@ -8,17 +8,32 @@ protocol ReactionsOverviewViewControllerDelegate: AnyObject {
...
-    init(reactions: DcReactions, context: DcContext) {
-
-        self.reactions = reactions
-        self.contactIds = Array(self.reactions.reactionsByContact.keys)
-        self.context = context
+    init(reactions: DcReactions, showFrequencies: Bool, context: DcContext) {
+
+        self.showFrequencies = showFrequencies
+
+        // layout by strings is not great, but good enough for now for a secondary UI
```
Amzd, line 21: This will fail for emojis not supported on your device yet, eg shake head emoji will become 🙂 and ↔️ when not supported. Not saying it needs to change, just a case you might've not thought about. Possibly \t would be better than spaces.

deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -2012,35 +2019,37 @@ extension ChatViewController {
...
+                if showPicker {
+                    reactionMessageId = messageId
+                    let pickerViewController = MCEmojiPickerViewController()
+                    pickerViewController.navigationItem.title = String.localized("react")
+                    pickerViewController.delegate = self
+
+                    let navigationController = UINavigationController(rootViewController: pickerViewController)
+                    if let sheet = navigationController.sheetPresentationController {
+                        sheet.detents = [.medium(), .large()]
+                    }
```
Amzd:
```suggestion
                    let navigationController = UINavigationController(rootViewController: pickerViewController)
                    navigationController.sheetPresentationController?.detents = [.medium(), .large()]
```

Unnecessary if statement (I know it's copied)

Amzd (approved):
All my notes are nitpicks. PR looks good, can be merged without the suggestions too.

DcCore/DcCore/Helper/UNMutableNotificationContent+init.swift
r10s: no. we want to show reaction notifications if a channel is not muted - therefore the option to be able to unmute the by default muted channel.

this is what eg. telegram is doing as well

deltachat-ios/Chat/Reactions/ReactionsOverviewViewController.swift
r10s: was failing before then as well. but anyways, the dialog can be improved at some point to the level of deltachat-android resp. whatsapp (see screenshots in the android PR)

DcCore/DcCore/Helper/UNMutableNotificationContent+init.swift
Amzd: Ah that makes more sense now, I misunderstood initially.

PR #3228, opened 2026-07-28
Fix low padding on rounded chat title on iOS 26

Amzd:
Before and after:
IMG_0021 IMG_0020


r10s:
is that dealing with rtl accordingly?

Amzd:
> is that dealing with rtl accordingly?

I think so. The image is always on the left. I don't have any RTL test setup though.

r10s (approved):

PR #3229, opened 2026-07-29
Fix status bar edge effect contrast

### LINKED ISSUE #3214: Automatic status bar color changes in iOS 26 should change the edge effect color
Since #3156 we have custom edge effects. These should change color to reflect the statusBarStyle.

I can't find a way to detect these style changes other than checking `view.window?.windowScene?.statusBarManager?.statusBarStyle` when the scrollview scrolls. Maybe it's okay to then compare it to a `lastStatusBarStyle` and if changed, change the edge effect material from light to dark or back.

It's fine to just do this for the top edge effect for now since that's the one that makes the statusbar unreadable.

domyshev:
Fixes #3214.

On iOS 26, the status bar text color changes while scrolling depending on the content behind it, but the top edge material did not follow. This could leave dark text over dark content.

This updates only the top edge effect when `statusBarStyle` changes:

- Uses dark material for light status bar content
- Uses light material for dark status bar content
- Keeps adaptive material for the default style

The mirrored table view and bottom edge effect are unchanged.

Tested:

- iPhone 17 Pro simulator, iOS 26.5: verified the `darkContent` → `lightContent` transition while scrolling
- iPad mini (A17 Pro) simulator, iOS 18.5: build succeeds

### Before / after

| Before | After |
| --- | --- |
|  deltachat-3214-before | deltachat-3214-after |

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -614,14 +614,13 @@ class ChatViewController: UIViewController, UITableViewDelegate, UITableViewData
               statusBarStyle != lastStatusBarStyle else { return }
 
         lastStatusBarStyle = statusBarStyle
-        let blurStyle: UIBlurEffect.Style
-        switch statusBarStyle {
+        let blurStyle: UIBlurEffect.Style = switch statusBarStyle {
         case .lightContent:
-            blurStyle = .systemUltraThinMaterialDark
+            .systemUltraThinMaterialDark
         case .darkContent:
-            blurStyle = .systemUltraThinMaterialLight
+            .systemUltraThinMaterialLight
         default:
-            blurStyle = .systemUltraThinMaterial
+            .systemUltraThinMaterial
         }
         edgeEffectBlurViews.top.effect = UIBlurEffect(style: blurStyle)
     }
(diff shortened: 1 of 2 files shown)


deltachat-ios/Chat/ChatViewController.swift
```diff
@@ -606,6 +608,24 @@ class ChatViewController: UIViewController, UITableViewDelegate, UITableViewData
...
+        lastStatusBarStyle = statusBarStyle
+        let blurStyle: UIBlurEffect.Style
+        switch statusBarStyle {
+        case .lightContent:
+            blurStyle = .systemUltraThinMaterialDark
+        case .darkContent:
+            blurStyle = .systemUltraThinMaterialLight
+        default:
+            blurStyle = .systemUltraThinMaterial
+        }
```
Amzd: We can avoid late assigning here I think :)
```suggestion
        let blurStyle: UIBlurEffect.Style = switch statusBarStyle {
        case .lightContent:
            .systemUltraThinMaterialDark
        case .darkContent:
            .systemUltraThinMaterialLight
        default:
            .systemUltraThinMaterial
        }
```
Nitpick, feel free to ignore

Amzd (approved):
Exactly what I had in mind, thank you :)

Amzd (changes requested):
Actually, after testing: The problem still kind of persists in dark mode. It does correctly change to the "light" material but the statusbar is still not easy to read. Might need some tweaks in the future

Amzd (approved):
Good enough

domyshev:
Should the dark mode fix be included in this PR, or should that be handled in a follow-up PR?

Amzd:
If you have one you can add it but don't stress about it, this is already way better.

r10s (approved):

deltachat-ios/Chat/ChatViewController.swift
domyshev: agreed

domyshev:
late assigning removed, conflicts resolved

Issue #3211, opened 2026-07-21
unstable input field


r10s:
the input field was noticed as being unstable and hang here and there already, however, i did not manage to create smth reproducible, until now :)

video 1:

- tested iOS 26 on iPhone 13 (but much older ones have issues as well, see #3198 )
- use a fresh profile
- paste more than one line to the input field of "saved messages", leave as draft
- go to "saved messages profile"
- go back
-> the input field is gone, similar to #3198

video2: another one, even worse as it freezes the app:

- again in "saved messages", quote a message with more than one line
- paste more than one line to the input bar, leave as draft
- go back and re-enter the chat: that already shows text outside the input field
- terminate the app
- reopen the app, go to "saved messages"
-> app freezes

EDIT: i cannot reproduce the freeze any longer

# the videos:

https://github.com/user-attachments/assets/9be7e551-73de-4f84-9192-9afde58d4e22

https://github.com/user-attachments/assets/16bb97b3-45c2-418a-a3c4-24fcc18e8257

Amzd:
Likely not the same issue as the now closed iOS 15 one, because this is handled without layout issues on my iOS 16 device: it just closes the keyboard when you navigate away. I will look into why this doesn't happen on iOS 26.

r10s:
k, then let's have a look how iOS 15 is doing once this is fixed

PR #3234, opened 2026-08-04
Use native input bar size calculation instead of manual solution

Amzd:
This aims to address input bar sizing issues noted in #3211 in case they were not all fixed in #3221


r10s (approved):
code, lgtm

PR #3235, opened 2026-08-04
Increase the remove quote/attachment button size

Amzd:
Clearing quote was sometimes impossible due to small button on pre liquid glass.

### DIFF
--- a/deltachat-ios/Chat/InputBarAccessoryView/InputBarView.swift
+++ b/deltachat-ios/Chat/InputBarAccessoryView/InputBarView.swift
@@ -30,7 +30,7 @@ struct InputBarView: View {
                         } else {
                             draft.setQuote(quotedMsg: nil)
                         }
-                    }.layoutPriority(-1).labelStyle(.iconOnly)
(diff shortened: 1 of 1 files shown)


r10s:
hm, it looks a bit bold and clumsy for liquid glass, which is the more important group today.

also other messenger seem to prefer the size we were using without this PR.

> Clearing quote was sometimes impossible due to small button on pre liquid glass.

can you be more precise where it was impossible?

i just tested on iPhoneSE1  - the smallest phone we support - and there it was perfectly fine.

i am not against changing the size as such, but i am against using font sizes and tweaks again without many reasons :)

i am very much for just go with the system defaults, which seem okay here, for liquid glass or not

<br><i>"remove" button on iphoneSe1 with iOS15 - if anything, one could complain about the padding, but also that is good engough, esp. as we do not want to optimize for iOS15 much</i>

Amzd:
The screenshot you sent is not the default size, I think you are using a larger font from accessibility settings. 

But I agree that changing this with font is inconvenient since it changes based on what the user has selected.

PR #3236, opened 2026-08-05
Fixed onboarding page buttons

Amzd:
On pre liquid glass setting background *and* capsule breaks the rounded corners of the capsule. Background changes automatically in the capsule button anyway based on enabled state. 

Removed changing the title color of "Use Other Server" button because it does not make sense to do that when the button stays enabled. If it should be disabled then title color would change automatically anyway.


r10s (approved):

Amzd:
TODO: Disable both

PR #3249, opened 2026-08-06
Don't show new messages indicator in the back button on liquid glass

### LINKED ISSUE #3192: Back button bug on long title and unread chats
This is how it looks normally with a small title:
Image

On a long title, as you can see the unread badge is missing and the back icon is much smaller and visually shifted to the left.
Image

- iOS version: 26.5.2
- Device: iPhone SE gen2
- Delta Chat version: 2.53.1
- Expected behavior: back button looks the same in chat with long and small title
- Actual behavior: bug in chat with long title
- Steps to reproduce the problem: enter a chat with a long title while you have unread messages in another unmuted chat

Amzd:
Fixes #3192

Signal doesn't do this either and it is currently broken on liquid glass, so lets remove this feature for now and create an issue if we really want this still. 

Now we show notifications from other chats even if app is in foreground this feature is not as important anymore (#3174)

### DIFF
--- a/deltachat-ios/Chat/ChatViewController.swift
+++ b/deltachat-ios/Chat/ChatViewController.swift
@@ -2905,6 +2905,9 @@ extension ChatViewController: ReactionsOverviewViewControllerDelegate {
 
 extension ChatViewController: BackButtonUpdateable {
     func shouldUpdateBackButton(_ viewController: UIViewController, chatId: Int, accountId: Int) -> Bool {
+        // On liquid glass we don't show new messages indicator in the back button
+        guard #unavailable(iOS 26.0) else { return false }
+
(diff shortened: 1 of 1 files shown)


r10s (changes requested):
i agree that the option is less important nowadays with changed notification options.

but then, remove the option alltogether, so less code to think about :)

it is a nice-to-have, an issue is not needed, there are more important things to target

Amzd:
This feature can probably be turned on again in iOS 27 through 
```swift
navigationItem.backBarButtonItem?.visibilityPriority = .high
```
In combination with a compression resistance reduction on the title view

r10s (approved):

PR #3254, opened 2026-08-11
do not open connectivity from main title

r10s:
tune down connectivity view;
the connectivity view is kind of advanced debugging view, a bit raw, with multi-relay even more.

it is sufficient if one can open that at "settings", at some point, we probabyl want to combine that with "relay"

cmp https://github.com/deltachat/deltachat-android/pull/4589

### DIFF
--- a/deltachat-ios/Controller/ChatListViewController.swift
+++ b/deltachat-ios/Controller/ChatListViewController.swift
@@ -16,7 +16,6 @@ class ChatListViewController: UITableViewController {
 
     private lazy var titleView: UILabel = {
         let view = UILabel()
-        view.isUserInteractionEnabled = false
         view.font = UIFont.systemFont(ofSize: 17, weight: .semibold)
         view.accessibilityTraits = .header
         return view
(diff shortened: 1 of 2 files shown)


Amzd (approved):
This feature was hidden to me until this PR so removing it is fine imo

deltachat-ios/Controller/ChatListViewController.swift
```diff
@@ -16,9 +16,7 @@ class ChatListViewController: UITableViewController {
 
     private lazy var titleView: UILabel = {
         let view = UILabel()
-        let navTapGesture = UITapGestureRecognizer(target: self, action: #selector(onNavigationTitleTapped))
-        view.addGestureRecognizer(navTapGesture)
-        view.isUserInteractionEnabled = true
+        view.isUserInteractionEnabled = false
```
Amzd: UILabel does not have user interaction enabled by default so this line can be removed
```suggestion
```

Amzd (changes requested):

Amzd (approved):
LGTM

PR #3255, opened 2026-08-11
Wait until the share sheet is dismissed before initiating the app open so we do not get "can't present" issues

### LINKED ISSUE #3185: "share to delta" is sometimes unreliable on first try
it happened several times to me that when using "share to delta" eg. from the gallery, delta chat pops up, but not with the "share to" view, but as the normal app.

using "share to delta" a second time then, it works as expected.

not sure how to reproduce that, it seems happen more often on first try, after the app is inactive for some time

Amzd:
Fixes #3185


r10s (approved):

r10s:
looking forward to check that out!

thanks a lot for fixing!

PR #3262, opened 2026-08-16
Harden deeplinks parameters parsing fixes crashes and missbehaviours

trufae:
* Add tests to verify all possible bad combinations
  - IDs cant be malformed, negative or larger than UInt32.max, or missing if required
* Current deeplink parameter parsing misses many corner cases which resulted in:
  - Crash when chatId is out of range (application closes, DoS on click)
  - Reference chats from the wrong accountId (invalid accountId was accepted as valid), which may confuse users or trick them.
  - WebXDC handler took negative ids as invalid but missed large ints

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Do not open "Connectivity View" from main title; use "Settigns" instead
 - Fix: If one relay is connected, assume overall connectivity
 - Fix marking messages as seen across devices
+- Harden deeplinks parameters parsing to address crash and missbehaviours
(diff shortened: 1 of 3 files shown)


Amzd:
I think this whole PR boils down to

```swift
let int = Int("4294967296") // fine
let int32 = Int32("4294967296") // fine
let int32fromInt = Int32(int ?? 0) // crash
```

and the case where this matters is from deeplinks which can be modified by users.

I have condensed this to an 8 line change to make maintenance easier.

PR #3263, opened 2026-08-16
Disable logging by default and add settings toggle

trufae:
The DcLogger class wraps osLogger forcing all log messages to be .public, this breaks the privacy design of logged information and there are actually quite many logging lines leaking sensitive information in the logs, considering only debug builds will benefit from this information and i assume the purpose of this is to help users determine why something is not working i think the easiest and cleanest solution is to keep that buggy dclogger abstraction (that just adds a heart emoji to each call) and dont do anything unless the user permits it.

Ideally I would like to refactor the whole logging system to solve this problem from the root because its a bad secure coding practice to force .public privacy in oslog, But the PR would be much larger and it will be better to clarify the need for some of those logs before moving forward.

Sumarizing, this PR aims to address the problem without losing debugging capabilities for advanced users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Do not open "Connectivity View" from main title; use "Settigns" instead
 - Fix: If one relay is connected, assume overall connectivity
 - Fix marking messages as seen across devices
+- Disable logging by default and add an option in advanced settings
 - Update to core 2.59.0
 
 
--- a/DcCore/DcCore/DC/Logger.swift
+++ b/DcCore/DcCore/DC/Logger.swift
@@ -13,7 +13,16 @@ public class DcLogger {
     let osLog: Logger
 
     static var enabled: Bool {
-        get { UserDefaults.standard.bool(forKey: UserDefaults.loggingEnabledKey) }
+        get {
+            if UserDefaults.standard.object(forKey: UserDefaults.loggingEnabledKey) == nil {
+                #if DEBUG
+                return true
+                #else
+                return false
+                #endif
+            }
+            return UserDefaults.standard.bool(forKey: UserDefaults.loggingEnabledKey)
+        }
         set { UserDefaults.standard.set(newValue, forKey: UserDefaults.loggingEnabledKey) }
     }
 
--- a/DcCore/DcCore/DC/Logger.swift
+++ b/DcCore/DcCore/DC/Logger.swift
@@ -12,23 +12,30 @@ public class DcLogger {
     static let category = "deltachat"
     let osLog: Logger
 
+    static var enabled: Bool {
+        get { UserDefaults.standard.bool(forKey: UserDefaults.loggingEnabledKey) }
+        set { UserDefaults.standard.set(newValue, forKey: UserDefaults.loggingEnabledKey) }
+    }
+
     public init() {
         osLog = Logger(subsystem: DcLogger.subsystem, category: DcLogger.category)
     }
 
     public func error(_ message: String) {
-        osLog.error("❤️ \(message, privacy: .public)") // "public" is needed to show lines; core takes care of privacy
+        guard DcLogger.enabled else { return }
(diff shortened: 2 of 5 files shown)


r10s:
thanks for having a look! 

hm, we very likely do not want to have an option for logging or log by other means. that makes debugging bugs harder and is also nothing we are doing elsewhere. it'd be a bad tradeoff.

do other apps have such a switch? eg. signal? 

in general, the operating system is not regarded as a threat.

logging to a file has disadvantage wrt disk lifetime and battery (only OS can log to ram and survive crashes), and is generally a larger maintenance effort. note, that the log is not a long-living one and can only be accessed when the phone is unlocked - and at that point you can access many other things as well.

instead, can you name the log lines that you think are problematic and we can see if we can improve them in core. 

and we can also check if we can use other defaults

as a rule of thumb for all changes: we aim for fewer options and fewer code :) it is super easy to add things, but maintainance over years is hard :)

trufae:
Private data shouldn't be logged in system logs or shared for bug reporting, the dclogger abstraction kills the privacy contract of the OSLog api because there's no way outside the os module to handle the interpolated strings filtering. i found several critical .. can't say a number rn but it's quite high and i would preferibly discuss that privately.

But my approach to handle this problem with a general switch was based on simplicity:

* i don't want apps to fill my logs with lots of messages for normal use, even less in release builds
* i dont want private info like ip addresses, file paths, contact names to be shared in bug reports
* these logs can be accessed even if you protect the app with faceid if you have the device unlocked
* dclogger abstraction breaks the .privacy contract because they are all .public (no redacted masks)
* all messages from core debug are logged, those can't be filtered from here

a proper fix will require changes in DcCore and in this app and that won't be a small patch. Imho it's preferiibly to add an option as a quick aid and inform users about private data being exposed when sharing those logs for bug reporting.

I would be happy to help on a deeper fix addressing each of the oslog calls later on.

trufae:
Signal have its custom logger api which identifies and filters all the private information using regexes (SignalServiceKit/Debugging/ScrubbingLogFormatter.swift) which is not perfect and actually i have found some private information leaked in logs for signal for ios too, but they apply some measures for most of the cases.. i may find more time to report this issue to them too

r10s:
k, if you have concrete critical log lines in mind, please just file a [security advisory](https://github.com/deltachat/deltachat-ios/security/advisories). 

i agree in general, that logging is worth a look, it is also considered too verbose here and then. however, note, that  the first screens shown in "View log" (the one without date tags) are _in-app only_. the other log lines are usually overwritten quickly. still, using different means of logging may be worth to think over as well

what we do not want to have is logging disabled by default nor have a switch not make debugging harder, for reasons mentioned above, therefore closing this PR

trufae:
those lines are not in-app only, open the console app in mac and you will see the very same logs streamed over wifi or usb. the security advisory link doesnt allows me to register anything, i can do that in my own fork, but i dont think requesting a CVE from a fork of the official project is the way to go. i can share the complete list with details, but as said, the logs coming from deltachat-core also include sensitive information and those are sent to the logs too.

if you want i can adjust the patch to just keep those logs inside the app and stop using osLog, the sensitive information will still be there, but it will be up to the user to export it and won't be accessible via the system console logs.

I've sent the doc to @Amzd , accept my dc invite and i'll send it to you if you are interested

Amzd:
> open the console app in mac and you will see the very same logs streamed over wifi or usb

Only when the device is unlocked and when that's the case you can also grab the whole database right?

trufae:
> > open the console app in mac and you will see the very same logs streamed over wifi or usb
> 
> Only when the device is unlocked and when that's the case you can also grab the whole database right?

The message database should never be accessible, but i assume is found in encrypted backups like all messaging apps do

only debug logs are streamed live, the rest of log levels are stored in disk and retriveable via sysdiagnose tool. About device lock, if you use a previously paired computer it is possible to retrieve the logs if you havent rebooted the device. But i didnt tried that much, thats what apple specifies

PR #3267, opened 2026-08-18
Play voice messages on ear speaker and lock screen when phone is held to the ear

trufae:
ref https://support.delta.chat/t/blank-screen-when-listening-to-voice-messages-on-ios/5660

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


Amzd:
We want to move to an audio session abstraction class which would handle audio based on priorities (call, voice msg, video attachment, etc) similar to Signal. This feature would then be handled by that class. This adds complexity now which would make it harder to transition which is one of the reasons I am hesitant to approve this atm.

PR #3268, opened 2026-08-18
Log to a file in release builds instead of the system oslog

trufae:
following for https://github.com/deltachat/deltachat-ios/pull/3263

The idea here is that release builds will use a custom logger that just save the logs inside the app home, these files are not accessible from outside the app and are protected by the sandbox.

Also, i have enforced the files to not be backuped or saved in the icloud. i think that we shuold be careful about doing that with other files...because me as a user i wont like to backup/restore anything but the profile credentials.. but that's another topic outside the scope of this pr.

The new code have 2 logger implementations

* in debug builds it works like before
* in release builds we save the logs for the last hour (i tried to make this as optimal as possible to reduce syscalls and disk access, because as said, ideally the app shouldnt be logging anything unless the user really wants that, but considering this path is not an option).

im open to discussion about the strategy to rotate logs too,but i hope this appraoch fits better

PD: users must be warned not to share the logs as verbatim without review/redacted because it can include sensitive information. this message must be appearing somewhere around the "View Logs" button.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Log to a private file instead of the system log, so logs are no longer readable from outside the app (e.g. via USB or Wi-Fi); "View Log" keeps the last two executions
 - Allow to react in channels
 - Offer five default reactions when long-tapping a message
 - Show phased out relays in connectivity view as such
--- a/DcCore/DcCore/DC/Logger.swift
+++ b/DcCore/DcCore/DC/Logger.swift
@@ -79,9 +79,13 @@ public class DcLogger {
     }
 
     private func openCurrentLogFile() {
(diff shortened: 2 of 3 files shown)


r10s:
after internal discussions, we do not want to not use system logging for reasons mentioned at https://github.com/deltachat/deltachat-ios/pull/3263#issuecomment-5312370960 

as also mentioned there, we can change log lines as needed. we removed an sdp and a token at https://github.com/deltachat/deltachat-ios/pull/3289 . as mentioned, please tell us what you find critical as well

in general, i understand that you do not like the system logging, however, we want to stay with it for reasons mentioned at https://github.com/deltachat/deltachat-ios/pull/3263#issuecomment-5312370960

trufae:
Si you dont want to use the system log apis but you want to use them :?? I’m confused

PR #3269, opened 2026-08-18
Delete leftover logfile after sharing accessible via usb

trufae:
* When the user shares the logs, the log file is saved in disk
* file is accessible via libimobiledevice because ifuse mounts the entire sandbox

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Send messages to up to 5 relays
 - Do not open "Connectivity View" from main title; use "Settigns" instead
 - Fix: If one relay is connected, assume overall connectivity
+- Fix: Delete the temporary log file after sharing it, it stayed readable via USB otherwise
 - Fix marking messages as seen across devices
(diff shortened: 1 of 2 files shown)


trufae:
This is another vulnerability i found while digging into the logging support in deltachat for iOS. when the user taps in view logs and then shared the log, right when the toast appears this tmp file is created and it's done with the wrong permissions, but more importantly, this file is never deleted, so , even after that and without opening the app, an authorized computer can mount via libimobiledevice the app tmp directory and download that file.

the patch removes the file when the toast is closed

Amzd:
This is a temp file and will be cleaned up by the system, but sure we can delete it earlier.

trufae:
And why not delete it as soon as it’s no longer needed like the pr do? Even if the system will delete it its accessible via usb and thats a file that contains sensitive information and should never be exposed that way.

Amzd:
The app you share it to is likely also leaking it but you are right, we can delete the temp file, which is what I do in #3276

PR #3264, opened 2026-08-17
Make background fetch in NSE asynchronous

Amzd:
This is a Notification Service Extension rework that fixes the following issues while reducing duplicate code:

- Quicker notifications when one profile has a relay that can't be reached. We no longer wait for background fetch to complete before queuing user notifications. This is required for https://github.com/chatmail/core/pull/8592
- Fix: Calls that come in after 30 seconds of the app being in background now correctly start IO. This fixes #3119 #3107 #3068 #3043
- Notification burst when starting the app after no network should now only include messages that weren't read as we delay sending them for a second.
- Remove deleted notifications. #3042
- Update notifications for edited messages (in foreground this only happens for the selected profile until we fix issue 3275)
- Fix: Notifications for missed calls while offline

Note: We assume that dc_accounts_background_fetch returns 0 immediately if it fails otherwise dc_get_next_event is called which would hang and leak the event emitter because no events will ever be sent there.

Related https://github.com/chatmail/core/pull/8554 https://github.com/chatmail/core/issues/8631

PR #3280, opened 2026-09-01
Use chat/user avatar instead of deltachat picture in notifications

trufae:


Amzd:
This will create a merge conflict with #3264 which has priority.

I think the API can be `UNNotificationContent.forMessage(...)` instead of `UNMutableNotificationContent(forMessage:...).updatingForMessage(...)` so the consumers (NotificationService and NotificationManager) don't need to think of doing this and it's just always attached to every notification. I can fix that too.

trufae:
I can rebase once merged

PR #3281, opened 2026-09-01
Use stock clip icon instead of emoji

trufae:


Amzd (changes requested):
Did you mean to delete the icon asset?

trufae:
Yes. Theres no need to ship an asset that looks uneven with the rest of icons of ios and the app when the system provides it already

Amzd:
The asset is not deleted in this PR

PR #3282, opened 2026-09-01
Drop allow_internet support for webxdc

trufae:
* WebXdc by spec doesn't support internet
* The change also needs to be extended to patch core
* This code was introduced in 2024 ref https://github.com/chatmail/core/pull/5678#issuecomment-5318156827
* Apps can be crafted to use internet without user consent or being notified

^ i'm opening this pr as an excuse to discuss about the half baked feature


r10s:
> WebXdc by spec doesn't support internet

this is and was always true.

the flag is not about webxdc you add to a chat.

the flag is only internal, and is about maps integration shipped with the app, which shares most the code with webxdc for practical reasons. and that integration can be replaced by some means, which is on purpose an obscure hack for devs only (similar to tap 10 times version number on some os). it is not for end user.

this "integration replacement" part is important for some countries with different connectivity, for simple experiments, cross platform. note, that the whole maps is experiemental.

one can argue that  "integrations" should have a different name, i would tend to agree with that, but renaming now is not worth the effort.

we aim to get maps out of experimental in the next year, until then, there is no much actionable

if we would merge this PR, we would break maps alltogether

trufae:
The issue is that you can get apps to bypass the checks and make them use the internets. But anyway

r10s:
> you can get apps to bypass the checks and make them use the internets

no, not "just so".

for casual readers:

one cannot "just so" get an webxdc app that has access to the internet. the receiving user  needs to attach it manually to ones own "saved messages" first. and then, it is visible that this is not a normal webxdc but a map

as said above, this is not _on purpose_ not easy but obscure, even if one knows what to do. if an evil user wants to trick you, it is far easier if they send you a link to  a website. also note, they need to discover you first. we are not aware of any real-life issue with that. of course, as for many things, one can invent theoretical things, or can have a different opinion on that

finally, as said, all that is subject to change anyways

trufae:
This obscure workflow seems like a hack to me for something that should require user aproval at least but yes its not just run an app and get that, the user needs to follow some steps that maybe done by accident too. The user will never know the xdc app they saved is now accessing internet because theres no indicator or prompt asking for this permission.

PR #3288, opened 2026-09-11
Add button to switch call to speaker

### LINKED ISSUE #3002: Video calls should use the speaker by default
Currently calls are always using the very quiet earpiece speaker or the connected bluetooth speaker. Video calls should use the loud speaker instead of the quiet call speaker by default.

### LINKED ISSUE #3054: Audio playback switches to call speaker after phone call interruption
- iOS version: 26.3
- Device: iphone 11
- Delta Chat version: 2.43
- Expected behavior: After a phone call ends, audio messages and voice messages should continue playing through the main (media) speaker as usual.
- Actual behavior: If a phone call interrupts audio/voice message playback, after the call ends the playback switches to the call/earpiece speaker instead of the main speaker.
     1. Start playing an audio or voice message in Delta Chat.
     2. Receive an incoming phone call.
     3. Answer the call (audio playback stops).
     4. End the call.
     5. Play the audio/voice message again.
- Screenshots: n/a
- Logs: n/a

Amzd:
Fixes #3002 and #3054 

Note: I tried to make the call louder which works but the speaker button is better 

```swift
func peerConnection(_ peerConnection: RTCPeerConnection, didStartReceivingOn transceiver: RTCRtpTransceiver) {
        if transceiver.mediaType == .audio, let newTrack = transceiver.receiver.track as? RTCAudioTrack {
            newTrack.source.volume = 2 // 2x gain
        }
```


r10s (approved):

Amzd (commented):
Still need to `try? rtcAudioSession.overrideOutputAudioPort(.none)` when deinit callviewcontroller I think

PR #3293, opened 2026-09-12
adapt to new relay api

r10s:
- use delete_transport() instead of set_transport_unpublished()
- delete_transport() can be used to remove the sending relay
- list_transports() was already used instead of deprecated delete_transport_ex()

the confirmation [is adapted on android](https://github.com/deltachat/deltachat-android/pull/4655/changes), and gets corrected on the next translation pulling; this should not block merging

#skip-changelog as the changes are already mentioned


deltachat-ios/Controller/Settings/Transports/TransportListViewController.swift
```diff
@@ -161,17 +161,15 @@ extension TransportListViewController {
...
         var actions: [UIContextualAction] = []
 
-        if !transport.isDefault(dcContext) {
-            let deleteAction = UIContextualAction(style: .destructive, title: String.localized("remove_desktop")) { [weak self] _, _, completion in
-                DispatchQueue.main.async {
-                    self?.deleteTransport(at: indexPath)
-                    completion(true)
-                }
+        let deleteAction = UIContextualAction(style: .destructive, title: String.localized("remove_desktop")) { [weak self] _, _, completion in
+            DispatchQueue.main.async {
```
Amzd: Why do we swap to main here? Does the UIKit callback come on a different queue?

r10s: idk - i just removed the outer "if", so that also used-for-sending relay can be deleted directly (core picks up a new sending relay in that case)

review without whitespace shows a clearer diff

but indeed, dispatching to main seems superfluous and is not done on most comparable places

r10s: i added a commit to remove the dispatching to main

Amzd (approved):

PR #3301, opened 2026-09-17
onboard using initTransports()

r10s:
see https://github.com/chatmail/core/pull/8718 for the new initTransports() api


deltachat-ios/Handler/ProgressAlertHandler.swift
```diff
@@ -75,12 +75,10 @@ class ProgressAlertHandler {
...
-        // delay so the user has time to read the success message
-        DispatchQueue.main.asyncAfter(deadline: .now() + 1, execute: {
-            progressAlertController.dismiss(animated: true) {
-                onComplete?()
-            }
-        })
+        DispatchQueue.main.async {
+            progressAlertController.dismiss(animated: true)
+        }
+        onComplete?()
```
Amzd, line 80: This flips the execution order of dismiss and onComplete which we shouldnt do because it means we can't present alerts in oncomplete

Amzd (changes requested):

Amzd (approved):

